PR #8, opened 2014-10-20
Update observatories.py

matteobachetti:
The clock corrections were read with a numpy.loadtxt call, and this only works if the clock file has no unexpected entries. Alas, in the Parkes file and maybe in others there are empty clock values and other problems, and so it crashes. 
If we stick to TEMPO-style clock files, we should tell the clock file reader the exact number of digits to read into the float values. This is a quick and dirty solution; there are probably faster ways you can imagine, in which case feel free to discard this pull request and change it with your preferred solution.

### DIFF
--- a/pint/observatories.py
+++ b/pint/observatories.py
@@ -8,6 +8,22 @@ from pint import pintdir
 class observatory(object):
     pass
 
+def read_clock_file(filen):
+    file = open(filen)
+    mjds = []
+    ccorr = []
+    for l in file.readlines():
+        try:
+            m = float(l[0:9])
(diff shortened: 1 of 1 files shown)


paulray:
@scottransom and @matteobachetti - I propose we close this issue. The Parkes file reading is now fixed (at least for MJD > 50844).  Discussion of better/more general solutions should probably be in Issue #88

matteobachetti:
@paulray Oh, I had forgotten about this issue... It's ok for me to close and discuss a better solution elsewhere at this point

PR #16, opened 2015-01-20
Merge issue

luojing1211:
I have the most version of nanograv/PINT version on my local remote branch nano. But when I merge the branch to master, It shows conflict. I tried to fix it, it merges without the change on Nano branch. So I am here to try this way. 

-L


scottransom:
@luojing1211  This needs to be broken up into several pull requests, I think.  And there are a bunch of things that I'm not sure we actually want in PINT itself (like all the TEMPO2 tests).  Perhaps those should be an add-on part of PINT?  Is it useful for any other reason than just testing our development?

luojing1211:
I don't think all the test for the fundamental calculation should be part of the released version. 

For the pull request, I think it was because I did not fully understand how to use pull request back to that time.

scottransom:
I guess I'll close this pull request then.  Feel free to submit others, though!

PR #17, opened 2015-02-26
Photontoas

paulray:
This pull request consists of some comments that I put in as I read through the code, denoted by SUGGESTION(paulr), as well as allowing a TOAs object to be initialized from a list of TOA objects, and a couple small changes including propagating usepickle up to the get_TOAs convenience function.


paulray:
Interestingly, after I made this pull request, I pushed up a couple more commits to this branch and they just got added to the pull request.  The main change there is to move the assignment of astropy units into the TOA class instead of the TOAs class (for TOA.error and TOA.freq).

paulray:
I guess this behavior will be different once the pull request is merged?  Presumably new commits on this branch will require a new pull request after it is merged and accepted?

paulray:
No, it was just me being kind of lazy and letting my setup.py modification get into my commit.  I don't normally "git commit -a" unless I'm sure what will happen from "git status".  But, I'll be more careful with making atomic commits going forward.  

Plus, I didn't really want to delete the whole commit since that would have removed my comments as well.  The pushing of the new commit to fix the things I should not have committed seems to have worked.

paulray:
OK, I think this is ready to merge. I got the nose tests running without an error.
Someone should still go through and improve the tests, however.  They aren't very consistently done.

Issue #30, opened 2015-03-16
No unit test for BT model


paulray:
There should really be a unit test in PINT/tests that verifies that bt.py is working and computing the expected results.

Also, when a BT model is read from a par file, a warning is issued that 'BINARY BT' is not recognized. This should probably be fixed.

paulray:
@matteobachetti I think this one is for you.

paulray:
While I'm at it, I'll add a note about the other issue in bt.py, previously reported by email to the list.

There is an issue with the handling of the OM parameter in both dd.py and bt.py.

The code looks like this:
```
       self.add_param(Parameter(name="OM",
           units="deg",
           description="Longitude of periastron passage",
           parse_value=lambda x: np.double(x) / 180 * np.pi))
```

See how it (correctly) claims that the units on OM are degrees, then it converts it to radians when it parses the value.  It handles it in the delay function ok, because it assumes the value is in radians. However, if you print the model or use model.as_parfile(), you get OM in the *WRONG* units!

matteobachetti:
The best thing would be to use Astropy units everywhere. Right now, Parameter does not understand astropy units and I couldn't make the conversion more consistently than this (without making this even uglier). At least, now the value will appear in the right units (degrees) when you do model.as_parfile()

vhaasteren:
I have started working on this. The BT model was basically a complete copy from the tempo2 code, so unit tests would not really make sense.

I have started from scratch. See btmodel.py in pint/models/. It does the delay calculations, and the period calculations, and derivatives of all. A unit test would then make sense against tempo2/tempo. Currently, there are errors in the d_delay_d_xxx functions, but the following work:

Pobs()
delay()
d_Pobs_d_xxx

demorest:
There is now a test for the BT model (test_B1953.py).

PR #31, opened 2015-03-17
test of BT model

matteobachetti:
This is a quick test of the validity of the delay calculation in the BT model as per Issue #30.
Data are simulated and the comparison is done with the results of libstempo.


matteobachetti:
Yes, for the future I agree that we should use astropy units.

PR #32, opened 2015-03-17
Add get_toa_list() function and double precision feature in get_mjds()

luojing1211:
Hi all, 

I added get_toa_list() wrapper for a list of toa objects to access TOAs.table and calculate TDB and planets position. 

And I also added a feature that enable the toa.get_mjd() function return a double precision mjd value. 

See my last three commits, I don't know how to select only last three commit and not show the old commits. If you know how, please let me know, so I can avoid this happen in the future. 

-L


scottransom:
@luojing1211 some of these commits are still not appropriate for going in to the master branch (for example, a custom change to the default PySPICE directory in the setup.py file.  Please check each of these and push up new versions with those changes removed (it will automatically update your pull request).

Also, I'm not sure that I like or understand why there seems to be nearly duplicated code for the TOA list thing compared to the way we process TOAs now.  Can you explain that?  For example, why is get_TOAs_filelike so similar to what we have already?  Why isn't it just a special case of it?  Duplicated code is not good.

And why is there a ton of the chebyshev polynomial stuff built directly into the d phase/dt function?  Isn't that used only for a special phase-prediction mode?  Do you have an example use-case for that now?

Finally, there are quite a few typos in your doc strings (mostly spelling) and you should not be using things like 86400, but instead SECS_PER_DAY (which have used in other parts of the code).

luojing1211:
@scottransom Thanks for your input. I agree some of the earlier commits is not ready to merge the master branch. 

The duplicated code for TOAs_filelike should not be there. I don't know why the update my master branch from nanograv/PINT did not remove it, perhaps, I merge it with my development branch before I send pull request. But I will fix it. 

About the Chebyshev polynomial stuff in d_phase_d_toa() function I was confused with polyco. By referencing from the discussion, I thought we need a Chebyshev method for calculate the apparent frequency. Do you think a finite difference is enough for the calculation?

Of course I will fix the typos. And something like 86400. 

-L

PR #36, opened 2015-06-23
PINT Polyco generator and evaluator 

luojing1211:
Hi Nanograv/PINT branch manager, 

Please take a look about the new features and leave comments on these. I have not put the binary phase and logrms into the code. Instead, I just put 0.0 as the value. But In the future, it will be there. 

-Luo Jing


pint/models/polycos.py
```diff
@@ -0,0 +1,612 @@
...
+
+        # Make sure the number of nodes is bigger then number of coeffs.
+        if numNodes < ncoeff:
+            numNodes = ncoeff+1   
+    	# generate the ploynomial coefficents
+    	if method == "TEMPO":
+            
+           
+    	# Using tempo1 method to create polycos
+            for i in range(len(entryIntvl)-1):
```
scottransom: This seems to be an indentation issue.  Will you please bring the for loop left one level?

luojing1211: This For loop is under the if  method == "TEMPO" statement. I believe is should be right

PR #46, opened 2015-09-29
PSR binary generic model and DD model 

luojing1211:
Hi pint developers, 
In this pull request, the binary model are not independent for PINT. It is a subclass of timeing_model.

I finish the code for PSR binary generic model and DD model. I still have some question about it. But I would like you to take a look of the code and make comments to improve it. 

Thank you. 

-Luo Jing


scottransom:
@luojing1211  Can you please see what the conflicts are with this?  I would like to merge this and get some good tests with it.  And I also want to modify it to add ELL1 model.

luojing1211:
Hi Scott,

I will take a look. I talked to Rutger about the binary model. He would
like to have a more independent binary model and PINT timing model is just
a wrapper of the independent binary model. So I am writing this part right
now in another branch. Is it better to have a decision before we merge it
together?

-Luo Jing

vhaasteren:
Ok, I finally had a look at the new binary model branch. These are the
comments I have for now:

1) There is an error in your master branch right now:
byte-compiling /Users/vhaaster/env/dev/lib/python2.7/site-packages/pint/models/parameter.py to parameter.pyc
File "/Users/vhaaster/env/dev/lib/python2.7/site-packages/pint/models/parameter.py", line 155
)

Easy fix, though.

2) The function model.as_parfile() does not output the binary model. I expected
to see the line:

BINARY DD

3) After reading in a par file with a DD model, I do not see the delay function
for the binary listed. I expected to see something like m.delay_funcs:

bound method J0613-sim.BT_delay of pint.models.timing_model.J0613-sim object at 0x10ce83fd0,
bound method J0613-sim.solar_system_shapiro_delay of pint.models.timing_model.J0613-sim object at 0x10ce83fd0,
bound method J0613-sim.dispersion_delay of pint.models.timing_model.J0613-sim object at 0x10ce83fd0,
bound method J0613-sim.solar_system_geometric_delay of pint.models.timing_model.J0613-sim object at 0x10ce83fd0

But instead (changing the binary model to DD), I see:

bound method J0613-sim.solar_system_shapiro_delay of pint.models.timing_model.J0613-sim object at 0x1123ed750,
bound method J0613-sim.dispersion_delay of pint.models.timing_model.J0613-sim object at 0x1123ed750,
bound method J0613-sim.solar_system_geometric_delay of pint.models.timing_model.J0613-sim object at 0x1123ed750

The binary delay function is now kept separately. Is that desirable? I think it
was kind of neat that everything was in delay_funcs, and I think it is a mistake
to single out the binary delay like this. My proposal:
a) Keep the binary delay function just in model.delay_funcs. That makes
the 'delay' function of TimingModel also neat.
b) The PSRBinary class adds the 'binary_model' attribute, which is a
reference to the actual binary delay function. We might decide to
have the binary_delay attribute already defined in the TimingModel
class, which initially refers to a lambda function that returns all
zeros. The PSRBinary class can then overwrite that. But I am not
sure whether we should, since from a design point of view, the
TimingModel class should not know about anything in the
sub-components. In other words, line 153 of timing_model.py is
specific to the PSRBinary component, which it should not know
anything about.

4) I do not like the type tests that are at line 186 of PSR_binary.py. This is
the type of behavior that should be handled by the Parameter class. We
should just be able to call a member function of Parameter (or overloaded
by the subclass). This type of stuff is happening in other locations of
PINT as well at the moment. We should fix that.

5) Shouldn't the function 'der' on line 158 of PSR_Binary be a part of
TimingModel? This seems like general functionality we would want there.

6) Suggestion: use numdifftools as a fallback to obtain parameter derivatives.
I have been using those in the past months, and they are (slow but) correct
down to near machine precision.

luojing1211:
Thanks for your comments,

I will take a look and make change on those.

-Luo Jing

luojing1211:
​Thank you Rutger,

I have looked at your comments. I recommend you to take a look on
binarywrapper pull request. It is a branch that more close to the design
you were talking about before. But the comments you have made are still
vilde, since I used same idea for those part.  ​


On Sun, Dec 13, 2015 at 8:17 PM, Rutger van Haasteren <
notifications@github.com> wrote:

> Ok, I finally had a look at the new binary model branch. These are the
> comments I have for now:
>
> 1) There is an error in your master branch right now:
> byte-compiling
> /Users/vhaaster/env/dev/lib/python2.7/site-packages/pint/models/parameter.py
> to parameter.pyc
> File
> "/Users/vhaaster/env/dev/lib/python2.7/site-packages/pint/models/parameter.py",
> line 155
> )
>
> Easy fix, though.
>
​This should be in the prefix pull request. I think I was not in the right
branch when I pushed commits. ​


> 2) The function model.as_parfile() does not output the binary model. I
> expected
> to see the line:
>
> BINARY DD
>
​I did not look at this function. Good catch. ​

> 3) After reading in a par file with a DD model, I do not see the delay
> function
> for the binary listed. I expected to see something like m.delay_funcs:
>
> bound method J0613-sim.BT_delay of pint.models.timing_model.J0613-sim
> object at 0x10ce83fd0,
> bound method J0613-sim.solar_system_shapiro_delay of
> pint.models.timing_model.J0613-sim object at 0x10ce83fd0,
> bound method J0613-sim.dispersion_delay of
> pint.models.timing_model.J0613-sim object at 0x10ce83fd0,
> bound method J0613-sim.solar_system_geometric_delay of
> pint.models.timing_model.J0613-sim object at 0x10ce83fd0
>
> But instead (changing the binary model to DD), I see:
>
> bound method J0613-sim.solar_system_shapiro_delay of
> pint.models.timing_model.J0613-sim object at 0x1123ed750,
> bound method J0613-sim.dispersion_delay of
> pint.models.timing_model.J0613-sim object at 0x1123ed750,
> bound method J0613-sim.solar_system_geometric_delay of
> pint.models.timing_model.J0613-sim object at 0x1123ed750
>
> The binary delay function is now kept separately. Is that desirable? I
> think it
> was kind of neat that everything was in delay_funcs, and I think it is a
> mistake
> to single out the binary delay like this. My proposal:
> a) Keep the binary delay function just in model.delay_funcs. That makes
> the 'delay' function of TimingModel also neat.
> b) The PSRBinary class adds the 'binary_model' attribute, which is a
> reference to the actual binary delay function. We might decide to
> have the binary_delay attribute already defined in the TimingModel
> class, which initially refers to a lambda function that returns all
> zeros. The PSRBinary class can then overwrite that. But I am not
> sure whether we should, since from a design point of view, the
> TimingModel class should not know about anything in the
> sub-components. In other words, line 153 of timing_model.py is
> specific to the PSRBinary component, which it should not know
> anything about.
>
​For this part, the main concern is that, the binary delay has to take the
barycentric TOAs, which means we have to remove solar system ​delays first
and then apply the binary delay. If TimingModel class don't know which
delay should calculate first, it will not return the right delay. It would
be nice to make TimingModel class no nothing about sub-components, could
you explain a little bit more about how do your proposal to fix this?


    4) I do not like the type tests that are at line 186 of PSR_binary.py.
This is

> the type of behavior that should be handled by the Parameter class. We
> should just be able to call a member function of Parameter (or overloaded
> by the subclass). This type of stuff is happening in other locations of
> PINT as well at the moment. We should fix that.
>
> 5) Shouldn't the function 'der' on line 158 of PSR_Binary be a part of
> TimingModel? This seems like general functionality we would want there.
>
> 6) Suggestion: use numdifftools as a fallback to obtain parameter
> derivatives.
> I have been using those in the past months, and they are (slow but) correct
> down to near machine precision.
>
​This function was designed for call the analytical derivatives from binary
model. It would be nice we can both do numdifftool and call analytical
function at same place, maybe der() would just be a wrapper function.

PR #54, opened 2015-11-13
Binary wrapper

luojing1211:
A branch using independent binary model. PINT binary model is a wapper


luojing1211:
I push the newest commit to a new branch. So now it is in the new pull request.

PR #66, opened 2015-11-30
Binarywrapper

luojing1211:
This branch's binary model is independent class. 
Somehow I push the new commit to a new branch. I will just use this new one.


luojing1211:
If this branch are chosen to be merged. The name of the file, DDwrapper for example, need to be change to the standard name. Please let me know which pull request is chosen.

vhaasteren:
Jing, is the other PR then not valid anymore (PSR binary generic model and DD model)? If so, can you close that PR?

luojing1211:
It is valid. It is just not independent from PINT platform.

vhaasteren:
Ok, I finally had a look at the new binary model branch. These are the
comments I have for now:

1)  There is an error in your master branch right now:
    byte-compiling /Users/vhaaster/env/dev/lib/python2.7/site-packages/pint/models/parameter.py to parameter.pyc
      File "/Users/vhaaster/env/dev/lib/python2.7/site-packages/pint/models/parameter.py", line 155
        )

    Easy fix, though.

2)  The function model.as_parfile() does not output the binary model. I expected
    to see the line:

BINARY         DD

3)  After reading in a par file with a DD model, I do not see the delay function
    for the binary listed. I expected to see something like m.delay_funcs:

[<bound method J0613-sim.BT_delay of <pint.models.timing_model.J0613-sim object at 0x10ce83fd0>>,
 <bound method J0613-sim.solar_system_shapiro_delay of <pint.models.timing_model.J0613-sim object at 0x10ce83fd0>>,
 <bound method J0613-sim.dispersion_delay of <pint.models.timing_model.J0613-sim object at 0x10ce83fd0>>,
 <bound method J0613-sim.solar_system_geometric_delay of <pint.models.timing_model.J0613-sim object at 0x10ce83fd0>>]

    But instead (changing the binary model to DD), I see:

[<bound method J0613-sim.solar_system_shapiro_delay of <pint.models.timing_model.J0613-sim object at 0x1123ed750>>,
 <bound method J0613-sim.dispersion_delay of <pint.models.timing_model.J0613-sim object at 0x1123ed750>>,
 <bound method J0613-sim.solar_system_geometric_delay of <pint.models.timing_model.J0613-sim object at 0x1123ed750>>]

    The binary delay function is now kept separately. Is that desirable? I think it
    was kind of neat that everything was in delay_funcs, and I think it is a mistake
    to single out the binary delay like this. My proposal:

        a)  Keep the binary delay function just in model.delay_funcs. That makes
            the 'delay' function of TimingModel also neat.
        b)  The PSRBinary class adds the 'binary_model' attribute, which is a
            reference to the actual binary delay function. We might decide to
            have the binary_delay attribute already defined in the TimingModel
            class, which initially refers to a lambda function that returns all
            zeros. The PSRBinary class can then overwrite that. But I am not
            sure whether we should, since from a design point of view, the
            TimingModel class should not know about anything in the
            sub-components. In other words, line 153 of timing_model.py is
            specific to the PSRBinary component, which it should not know
            anything about.


4)  I do not like the type tests that are at line 186 of PSR_binary.py. This is
    the type of behavior that should be handled by the Parameter class. We
    should just be able to call a member function of Parameter (or overloaded
    by the subclass). This type of stuff is happening in other locations of
    PINT as well at the moment. We should fix that.

5)  Shouldn't the function 'der' on line 158 of PSR_Binary be a part of
    TimingModel? This seems like general functionality we would want there.

6)  Suggestion: use numdifftools as a fallback to obtain parameter derivatives.
    I have been using those in the past months, and they are (slow but) correct
    down to near machine precision.

vhaasteren:
Oops, I posted this in the wrong pull request. Apologies

PR #70, opened 2015-12-04
Dmx

luojing1211:
Hi All folks,

This is PINT DMX branch. 
New Features:
1. Parameter module has prefixParameter 
2. Timing model can handle prefixParameter
3. Model builder can add the right prefixParameters from parfile
4. Dispersion module is able to calculate dispersion delay with DMX_ DMXR1_ DMXR2_  , not DMXF_, DMXEP_
5. Test is in PINT/tests/test_dmx.py

Please take a look and leave comments. I will make improvement as soon as possible when I see the comment.


luojing1211:
Wrong branch. Close it

Issue #58, opened 2015-11-23
What paths should be used for tests?


paulray:
The tests in the tests subdirectory make different, inconsistent, assumptions about the current working directory.  For example test_bt.py tries to read '../tests/J1955.par', while test_model.py tries to read 'tests/J1744-1134.basic.par'.  This means that test_bt.py runs correctly if you are in the main PINT directory (which is how it is run if you use the run-tests.sh script), while test_model.py runs correctly if your CWD is PINT/tests.

So, what is the preferred way to specify the path to files that need to be loaded by the scripts in the test suite?

paulray:
The same issue comes up for the examples/fit_NGC6440E.py script.  It runs if you do:
`python examples/fit_NGC6440E.py`
from the PINT directory, but not if you do
`python fit_NGC6440E.py`
when your current working directory in PINT/examples.

demorest:
This is fixed in PR #132

PR #71, opened 2015-12-04
Prefix

luojing1211:
Hi All folks,

This is PINT Prefix branch. 
New Features:
1. Parameter module has prefixParameter 
2. Timing model can handle prefixParameter
3. Model builder can add the right prefixParameters from parfile
4. Dispersion module is able to calculate dispersion delay with DMX_ DMXR1_ DMXR2_  , not DMXF_, DMXEP_
5. Test is in PINT/tests/test_dmx.py

Please take a look and leave comments. I will make improvement as soon as possible when I see the comment.


paulray:
I figured out how to add @luojing1211 repo as a remote and checked out this branch to test. I found a couple of issues:
* I instantiated a model using `m = pint.models.get_model(parfile)`. This parfile had 3 glitches, but no DMX parameters. But, somehow the model got DMX parameters added and it fails when trying to convert to a parfile.
```
In [16]: m.params
Out[16]: 
['PSR',
 'GLPH_1',
 'GLEP_1',
 'GLF0_1',
 'GLF1_1',
 'GLF0D_1',
 'GLTD_1',
 'GLPH_1',
 'GLEP_1',
 'GLF0_1',
 'GLF1_1',
 'GLF0D_1',
 'GLTD_1',
 'PLANET_SHAPIRO',
 'DM',
 'DMX',
 'DMX_0001',
 'DMXR1_0001',
 'DMXR2_0001',
 'F0',
 'F1',
 'F2',
 'TZRMJD',
 'PEPOCH',
 'RAJ',
 'DECJ',
 'POSEPOCH',
 'PMRA',
 'PMDEC',
 'PX',
 'GLPH_2',
 'GLPH_3',
 'GLPH_2',
 'GLPH_3',
 'GLEP_3',
 'GLEP_2',
 'GLEP_3',
 'GLEP_2',
 'GLF1_3',
 'GLF1_2',
 'GLF1_3',
 'GLF1_2',
 'GLF0_2',
 'GLF0_3',
 'GLF0_2',
 'GLF0_3']

In [17]: m.DMXR1
m.DMXR1_0001     m.DMXR1_mapping  

In [17]: m.DMXR1_0001.value
Out[17]: 0.0

In [18]: m.DMXR1_0001.as_parfile_line()
<ipython-input-18-e9600acf2ea0> in <module>()
----> 1 m.DMXR1_0001.as_parfile_line()

/Users/psray/src/PINT/pint/models/parameter.pyc in as_parfile_line(self)
    204         if self.value is None:
    205             return ""
--> 206         line = "%-15s %25s" % (self.name, self.print_value(self.value))
    207         if self.uncertainty is not None:
    208             line += " %d %s" % (0 if self.frozen else 1, str(self.uncertainty))

/Users/psray/src/PINT/pint/utils.pyc in time_to_mjd_string(t, prec)
    134     astropy does not seem to provide this capability (yet?).
    135     """
--> 136     jd1 = t.jd1 - erfa.DJM0
    137     imjd = int(jd1)
    138     fjd1 = jd1 - imjd

AttributeError: 'float' object has no attribute 'jd1'
```
Question: I see that in test_dmx.py, the model is build with model_builder.get_model(). What is the difference between that and pint.toa.get_model()?

* My test with 3 glitches fails to compute the model correctly.  I don't know how to commit to this branch, so I'll just push the files to my repo under a prefixtest branch

* Finally, I notice that test_dmx.py assumes it is running in the tests subdirectory.  I think all the other tests assume they are running in the main pint directory and prepend 'tests/' to filenames. I'm not sure which is best, but we should be consistent (see issue #58 )

luojing1211:
Very good catch Paul Ray. 
1. Why do we still have DMX parameters in the model even the par file does not have them in it?
    The reason for that is because in our dispersion model, one set of DMX parameters are added in the class __init__ function. That is why there is only one set of DMX parameters. The way prefixParameter how works is 

1. Add one prefix parameters in the model components. 
2. Model builder goes to the parfile, select components and build the model.
3. Model builder detects more prefixed parameters depend on the selected components. For example, if dispersion has DMX_0001 in it, model builder will try to find more DMX_ prefix. Then it will add correct amount of DMX_ model. 

I think that is why in your test, DMX_ parameters are still in the model. If it is not there, our dispersion component will break, since it has the DMX calculation in it. 

Of course to fix the print par file issue, one straightforward way is just add a is statement " if par.value is None or 0.0: continue." But if a parameter's value is 0.0, it will not print. I have seen this case in your testprefix.par file.  One parameter's value is 0.0. 

Answer for your question: 
toa module has no get_model() method. It has get_toa() method. 


2. I think when you update has merged to the master, I will update my branch, so I can do a quick fix. 

3. I am the responsible  for the issue 3. The test I made is majorly run under the tests directory. If we have a decision, I will change it.

scottransom:
Note that parameters can and should often be zero, so we don't want that to be a condition on whether to use them or print them or not!

luojing1211:
I took a look about the print_as_parfile issue. The reason is the when I initializing the DMR1 parameter which is a MJD, the value should be a time object. So I changed it to time object. It works right now. But the output value will be 0.00000, which is not a valid MJD. So do we want print the parameters we never use? 

One way could be that we build a simple dispersion model for example. No DMX calculation included. This way, it solves the problem and saves calculation power. On the other hand, to have this simple model, it requires a more robust model builder which can select the right model components with similar parameters.

luojing1211:
Do not merge this pull request. !!!!!!
Due to some conflict with new update. I create a new pull request #92. We can switch the discussion there.

paulray:
@luojing1211 Regarding my question about get_model(), I had meant to type pint.model.get_model(), not pint.ta.get_model(), but anyway now I see that pint.model.get_model() is the same as pint.model.model_builder.get_model(), so you can ignore my question.

Other discussions, I will move to issue #92 as you requested.

Should we "Close" this pull request?

luojing1211:
Sure, we can close it.

paulray:
Closing this pull request in favor of #92

PR #80, opened 2016-01-01
Binarywrapper

luojing1211:
This new version should fix the problem. Let me know if there are more issues .


vhaasteren:
I am taking over this PR, due to conflicts in the current git commit history (that was my fault). I'll fix it, and merge it through a PR from my own account.

PR #83, opened 2016-01-13
add bare value to parameter class unit

luojing1211:
Hi, 
This is the branch I add bare_value to parameter class. Please try it out and break it. 

-Luo Jing


vhaasteren:
This seems to work for me. A couple of remarks regarding the behavior at the moment.

Currently only 'value' is a property with a setter. The attributes bare_value and base_unit are not properties. It is therefore possible to set bare_value and base_unit without actually changing the value of a parameter.

I think they all need to be properties. Moreover, it should be impossible to change the value of 'base_unit' to an inconsistent unit. So you'd be able to change base_unit from seconds to hours, but not to Hz.

If I change the numerical value of bare_value, the actual value of the parameter should also change. They should be coupled through the use of properties.

I do like the use of _value. I think all that the properties should do is to re-set _value. When they return a value through the property 'getter', they are returning a processed form of _value. Also, when we change the base_unit, we will also change _value.

Does that make sense?

luojing1211:
I was think about that. But I was designing as the only properties can be changed by user is parameter.value. Instead of being changed, the base_unit is the standard unit, which will not be reassigned in the process and bare_value will be the value associate with base unit. In other words, 
par.base_value =((par. value).to(par.base_unit)).value. 

From your comment, I guess that fitter will use the bare_value as well. I can set it up as a property, so when it gets changed par.value will be updated. But it could be a little bit tricker, since par.value is not always a astropy quantity type.

vhaasteren:
Yes, I think it would be best if the fitters can use bare_value. That way, the fitters do not have to worry about units at all, and always use the values of bare_values.

The fact that implementing this is trickier (but not _that_ tricky) is _exactly_ why we need to implement it. Otherwise we need to deal with that in places like the fitter. This needs to be done at the Parameter class level.

paulray:
It seems to me that bare_unit should be a constant property of any parameter, and that it should not be changeable.  That way, things like fitters wouldn't have to worry about it changing and could make assumptions safely, like a PB is always in days.  Does that sound reasonable?

vhaasteren:
At the moment that is not how it works. For instance:

    m.RAJ.value
    >> 6h13m43.9777s
    m.RAJ.value = 0.15
    m.RAJ.value
    >> 0.15

However, using base_unit also does not work, because it is allowed to be 'None'. If we do:

    m.RAJ.value = 0.15 * m.RAJ.base_unit

this can lead to an error.

I do not have a strong opinion on whether bare_value should be changeable or not. If we do, it should internally modify m._value. If we make it unchangeable, then we need to make sure that we can set m.PAR.value in a way like value * base_unit. Or even better, detect whether or not the unit is ok. So:

    m.RAJ.value = 0.15

Results in m.RAJ.value = 0.15 * m.RAJ.base_unit (hours). And

    m.RAJ.value = 6h13m43.9777s

Works as well. But

    m.RAJ.value = 0.15 * u.sec

Will throw an exception, because the units are inconsistent.

scottransom:
It seems that if value is a property (which I think it should be) then its setter can (and should) verify that it has consistent units with bare_unit.  If they are not the same unit, it can and should convert so that __value is in the correct units.  bare_unit should also be a property where its setter only allows it to be set to the "correct" unit!

luojing1211:
This is how I do it right now. In value setter, it will detect value unit and assign to base_unit, if base_unit is None. If the value assigned as an astropy quantity with a different unit from base_unit but convertable , it will convert this quantity to the base unit and get the bare_value after converting. So it will keep the base_unit and get the bare_value associate with the base_unit
Piece of code
```
@value.setter
    def value(self,val):
        self._value = val
        if self._value is None:
            self._bare_value = None
            return

        if self.base_unit is None:
            if hasattr(self._value,'unit'):
                self.base_unit = self._value.unit

        if hasattr(self._value,'unit'):
            value_base_unit = self._value.to(u.Unit(self.base_unit))
            self._bare_value = value_base_unit.value
        else:
            self._bare_value = self.get_bare_value(self._value)
```

paulray:
Regardig @vhaasteren comment above, couldn't the problem with base_unit being None be solved by having it default to [u.dimensionless_unscaled](http://docs.astropy.org/en/stable/units/standard_units.html#the-dimensionless-unit)?

Also, this seems like a good place to employ test-driven development.  We could write a short script that does a bunch of manipulations on parameter values and bare_values that should either work or cause exceptions, then we just massage the code  until it does the right thing.  This forces the outward facing interface to be defined first, then the implementation specifics to be worried about second.  And, anyone can look at the test code and either agree or disagree that it is how things "should" work.

scottransom:
Agree with @paulray on both points

luojing1211:
Another reason I allow base_unit as None is some parameter does not really match a quantity type. For example, 'PSRJ' has a value as string represent a pulsar name. It really don't have unit.

luojing1211:
Hi all, 
I ready update the code.  Par.value and par.bare_value are property right now. The tests/test_parameter.py is a short testing file. Please go ahead check it out. As Paul R. said, if you can write a small test program as the way you want to use it, it will be very helpful. 

Thank you.

pint/models/model_builder.py
```diff
@@ -57,6 +57,8 @@ def __init__(self, name, parfile = None):
         self.name = name
         self.model_instance = None
         self.param_inparF = None
+        self.param_unrcgnz = {}
```
scottransom: self.param_unrcgnz?   That is a terrible abbreviation...  Since this is internal code, I *highly* recommend a full and understandable name lik eself.param_unrecognized

pint/models/parameter.py
```diff
@@ -44,15 +45,31 @@ class Parameter(object):
...
     """
 
     def __init__(self, name=None, value=None, units=None, description=None,
             uncertainty=None, frozen=True, aliases=None, continuous=True,
-            parse_value=fortran_float, print_value=str):
-        self.value = value
+            parse_value=fortran_float, print_value=str,get_value = None,
+            get_bare_value=lambda x: x):
         self.name = name
         self.units = units
```
paulray: This could use a comment describing what is meant by each attribute. In particular, it should clearly specify whether self.units is a string or an instance of an astropy.unit

luojing1211:
Comment and docstring will be updated very soon.

PR #86, opened 2016-01-27
Base_value add valueError

luojing1211:
Add a valueError for base_value. Make sure base_value is a number. 

-Luo Jing

### DIFF
--- a/pint/models/parameter.py
+++ b/pint/models/parameter.py
@@ -144,7 +144,7 @@ class Parameter(object):
     def base_value(self,val):
         self._base_value = val
         if not isinstance(val, numbers.Number):
-            return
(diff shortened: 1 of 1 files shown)


pint/models/parameter.py
```diff
@@ -144,7 +144,7 @@ def base_value(self):
     def base_value(self,val):
         self._base_value = val
         if not isinstance(val, numbers.Number):
-            return
```
scottransom: Might it not be better to compare vs numbers.Real rather than numbers.Number?  What are the advantages of numbers.Number?

luojing1211: If we are not deal with complex numbers, I believe numbers.Real works as
well.
-Luo Jing

PR #89, opened 2016-01-29
Binary opt2

luojing1211:
This is new update for binary model based on Rutger's comment 1. There will be more updates, so don't merge yet.


vhaasteren:
@luojing1211 can you fix the conflict? Just make sure that test_derivates ends with:

    parfile = os.path.join(datapath, 'B1855+09_NANOGrav_dfg+12_modified.par')
    timfile = os.path.join(datapath, 'B1855+09_NANOGrav_dfg+12.tim')
    ddm = mb.get_model(parfile)
    t = toa.get_TOAs(timfile,planets = True)
    ddob = ddm.get_dd_object(t.table)

pint/models/pint_pulsar_binary.py
```diff
@@ -120,3 +144,10 @@ def binary_delay(self,toas):
         for bdf in self.binary_delay_funcs:
             bdelay+= bdf(toas)
         return bdelay
+
+    @Cache.use_cache
+    def d_delay_d_xxxx(self,param,toas):
+        """Return the DD timing model delay derivtives"""
```
vhaasteren: The comment to return the 'DD' binary model seems slightly wrong. DD is only the 'default' now, right?

pint/models/pint_dd_model.py
```diff
@@ -19,48 +20,33 @@ class DDwrapper(PSRbinaryWapper):
...
-             units="second",
-             description="Binary Einsten delay GAMMA term",
-             parse_value=np.double),binary_param = True)
-
-        self.add_param(Parameter(name="DR",
-             units="",
-             description="Relativistic deformation of the orbit",
-             parse_value=np.double),binary_param = True)
+        self.binary_model_name = 'DD'
+        self.binary_model_class = DDmodel
```
vhaasteren: So we are using the 'DD' model by default?

I know I have not updated the BT model, so definitely leave it for now, but is this going to be standard?

luojing1211: This is only happened in pint_dd_model.py. So it is DD model specific. I believe DD model should not be the standard model.

pint/models/pint_pulsar_binary.py
luojing1211: I have not modified derivative part now. Yes, I am going to change it .

PR #92, opened 2016-01-31
Prefix New

luojing1211:
Due to some conflict. I have to create a new prefix pull request.


paulray:
I merged my two recent pull requests, since they fix the Parkes clock bug and add a prefix_test that I wrote that uses glitches.  This should help.

luojing1211:
I tried the test. Is this what you get?
figure_1

paulray:
Yep, that shows the problem.  If you try it with Tempo2, the residuals are flat.

luojing1211:
I see, I will diagnose it right now.

-Luo Jing ​

luojing1211:
Fixed. Test passed.  PINT detecting wrong number of glitches causes the trouble. I also fixed the setup part. 

I have a quick question, do the prefixed parameter has to be continues? For example, can I have a GLPH_2, and GLPH_4 instead of  GLPH_1 and GLPH_2. I know it does not make any sense. But for some parameters, GLF0D_  for example, technically It can start with GLF0D_2. Another example is can we have F0, F2, given in par file but not F1?

paulray:
For a polynomial expansion like F0, F1, F2, I think PINT should require that all parameters from _1 to _N should be specified.

For gliltches, one could imagine that you might want to drop glitch 3 from the model and have glitches 1, 2, and 4 still work.

luojing1211:
For the set up right now. Spin down term should have F1 to FN, but glitch does not have to.

luojing1211:

pint/models/dispersion.py
```diff
@@ -34,7 +71,27 @@ def dispersion_delay(self, toas):
...
             warn("Using topocentric frequency for dedispersion!")
             bfreq = toas['freq']
-        return self.DM.value * DMconst / bfreq**2
+
+        # Constant dm delay
+        dmdelay = self.DM.value * DMconst / bfreq**2
+
+        # Get DMX delays
+        epoch_ind = 1
+        while epoch_ind in self.DMX_mapping:
```
scottransom: Another way of doing this without looping over them every time that you call disperson_delay() would be to explicitly put TOAs into DMX ranges and then add a flag with the dmxrange number to each TOA.  You could do that once.  Then after that, you could use the astropy table groupby code to apply the appropriate DM delays.  I think that would save a whole bunch of array comparison operations during a fit.

pint/models/glitch.py
```diff
@@ -8,13 +8,13 @@
     from astropy.erfa import DAYSEC as SECS_PER_DAY
 except ImportError:
     from astropy._erfa import DAYSEC as SECS_PER_DAY
-from .parameter import Parameter, MJDParameter
+from .parameter import Parameter, MJDParameter,prefixParameter
```
scottransom: I often see a lot of python style issues in your code (like the missing space after the comma here, or typos in documentation / comments.  I highly recommend that you run your code through "pep8" before you push it.  In fact, it would be great if you could do that now for this pull request!

luojing1211:
I will do it right now.

pint/utils.py
```diff
@@ -288,6 +289,33 @@ def str2longdouble(str):
     return str2ldarr1(str)[0]
 
 
+def split_prefixed_name(name):
```
scottransom: Isn't this just a one-liner doing something like the following with a better regular expression?
int(re.match(r'([a-z]+_*)(\d+)', name, re.I).groups()[-1])
If you wanted the prefix string and the value, you could do:

def split_prefixed_name(name):
    s, v = re.match(r'([a-z]+_*)(\d+)', name, re.I).groups()
    return s, int(v)

luojing1211:
It does simplify the process. I was aiming to get the index format, such as "0001" or "1",as well.

scottransom:
If you really need that, then that is a tiny change as well:

```
def split_prefixed_name(name):
    s, v = re.match(r'([a-z]+_*)(\d+)', name, re.I).groups()
    return s, v, int(v)
```

scottransom:
Note that you could compile the regular expression once, and use it repeatedly:  x = re.compile(r'([a-z]+_*)(\d+)')  then x.match(name, re.I).  And you would do the compile out of the function, possibly at the top of the module.

luojing1211:
I changed some program style issue. It should be looking good now. I also update the name split function and DMX delay calculation.

luojing1211:
I think I fix the pep8 issue, except the very long last line in the utils.py. 
Prefix pattern, 
typo fixed. Pattern example 
pp1: For the prefix like DMX_3, 
pp2: For the prefix like DMXR1_3, 
pp3: For the prefix like F12

It is good to know the d_delay_d_GLF0_1() method is not working. I fixed it by setting ii to be 1 and fixed some other bug in other derivative functions. When I made the modification about this module I was focused on the delay function.

scottransom:
I don't think that we need all three pp1, pp2, and pp3.  For your pp2 example, DMXR1, DMXR2, DMXF1, and DMXF2 are parameters themselves.  There will never be a DMXF4, for instance.  So I think pp1 can work for all the cases (i think there is a way of getting the regexp to not return the _ if you don't want it).

scottransom:
Worst case you could just return v.strip('_') if needed...

luojing1211:
Here is a little test for the pattern pp2. It fails on the pattern 'DMX_001'
```
>>>pp2 = re.compile(r'([a-zA-Z]+\d+_*)(\d+)') 
>>>b = pp2.match('DMX_0001')
>>>b.groups()
<ipython-input-17-e756149bd244> in <module>()
      1 pp2 = re.compile(r'([a-zA-Z]+\d+_*)(\d+)')
      2 b = pp2.match('DMX_0001')
----> 3 b.groups()

AttributeError: 'NoneType' object has no attribute 'groups'
```
Or we can do a non-fancy way. 
```
def split(name):
       if '_' in name:
           namefield = name.split('_')
           prefix = namefile[0]+'_'
           idxstr = namefile[1]
           idx = int(idxstr)
       else:
           Figure out a way to split a number for letter. for example F12. 
        return prefix, idxstr, idx
```
The problem is how to split a paramtere like F12, I can not think of a way that can avoid the re.match method.

luojing1211:
I just found out. This way works fine. 

```
>>>pp2 = re.compile(r'([a-zA-Z0-9]+_*)(\d+)') 
>>>b = pp2.match('DMX_0001')
>>>b.groups()
('DMX_', '0001')
>>>b = pp2.match('DMXR3_0001')
>>>b.groups()
('DMXR3_', '0001')
```

scottransom:
So I'm about ready to merge this, but I just followed to command-line instructions above and tried to run test_dmx.py and test_parameters.py and both fail.  Can you please make sure that those work?

scottransom:
Also note that test_dms.py is *not* in standard testing format.  You should be able to run those via nosetests....

luojing1211:
Yes, I am going to work on it, after I fix the MJDparameter bug.

scottransom:
Just to clarify, nosttests should be runable from the top-level directory.  You shouldn't have to go into tests to run that.

paulray:
The test directory question is issue #58 
We should write the proper answer in there and close it.
I guess the answer is that tests should run from the main directory as python tests/test_xxx.py and all the files needed should be found.  Should this always be done with relative paths, or should test scripts use $PINT/tests/datafile.  It seems to me that would be bad since you wouldn't be sure  you were testing the local directory version of everything.

luojing1211:
I fixed test file and move all the data to datafile director under tests file. It will run under the PINT top-level directory. The test_dmx.py will raise an AssterationError saying "PINT and tempo residual difference is too big. " That is due to I am using a 9 years data set, which we can not process totally yet.

luojing1211:
But I recommend to merge the fix MJDparameter Pull request first. This Pull Request has directory change, which can cause trouble if merged first.

scottransom:
@luojing1211 So I'm trying to merge MJDparameter first and then this one, but there is a fairly big conflict in parameter.py when I try that....  We've got to start doing a better job keeping our branches up-to-date for these pull requests.  And I think more smaller pull-requests is the way to go.  As these PRs get bigger and bigger, they diverge more and more from changes that others are making.

luojing1211:
I totally understand that. It was a small PR, but when with our discussion going on, I made update on it. That could be the reason. But let us try to keep it small. I am getting the special parameters class setting up in the MJDfix branch. I should have it in another branch. I will let you know when to merge. After you merge the MJD one, let me know, I will fix conflicts on my side. Sorry for the trouble.

scottransom:
So one problem that I noticed is that you removed all those test data files, but you didn't commit them or the directory to this pull request!  So I can't test anything!  If you fix that, I can probably merge these two pull requests.  And then we can work on stuff from that point.

luojing1211:
Ok I will fix it very soon.

paulray:
What's the right way to keep conflicts to a minimum when you have a feature branch you are doing pull request for?  I think that you can keep doing "git rebase master" or "git merge master" in that branch to make sure that it is up to date and resolve any conflicts so it is easy for the release manager to merge.  Then only the deltas from the current master will be in the pull request.  Is that the right way?

luojing1211:
I completely agree with your @paulray. And this is what I am always doing currently. The problem is sometimes, like right now, there are two PR underdevelopment at the same time. Even those two star at same point and keep update from master, but it can not totally avoid conflict, since they are developed independently and may conflict at some point.

luojing1211:
This branch is ready.

paulray:
Ah, good point @luojing1211 - having two pull requests going is considerably trickier!

luojing1211:
Maybe setup a PR priority function.
def pull_request_priority(Pull_Request):
      result = tell_scott_merge_this_Pull_Request_first(Pull_Request) 
      if not result:  # If no response 
          for i in range(100):
                result = tell_scott_merge_this_Pull_Request_first(Pull_Request)

scottransom:
Nice algorithm... @luojing1211 
@paulray   In general, I think that most people prefer rebase as you don't get a lot of merge commits in your log, and everything gets merged into master via one nice clean set of patches at the end.  But both merges and rebasing work.

paulray:
Thanks @scottransom - I heard that you should NOT do rebase if you've already pushed to a server, but it is fine if you just have a local branch.  Do you agree?

scottransom:
Yes, absolutely.  There are a lot of examples of that online that explain how it can mess people up.  (Note that if you are the only one working on your branch of Fork on github, that could be considered a "local" branch as well.  But it is more dangerous, just in case someone cloned your branch and you don't know it.)

paulray:
My problem is that I always have 2 clones of paulray/PINT (one on laptop and one on desktop), so I think they'd get messed up if I rewrite commit histories

scottransom:
yeah...  I'm not sure how you'd handle that.  I think the issues are really not a problem when it is you working on 2 private machines, though.  The issues are when others are pulling/pushing/cloning from the same repository, because then you can end up with inconsistent histories.  I think with yourself you can always sync up between the two before you rebase.  This is probably discussed online in some tutorial somewhere...

luojing1211:
I should be good, now.

scottransom:
I just merged it, but there are still lots of issues with params.  Loading in a model most of the params still don't have proper astropy units....

scottransom:
Another issue seems to be that DMX params are added even if there are none defined in the parfile.  That shouldn't happen...

luojing1211:
For the first issue is strange. I am me looking at it. 
For second part, I kept at least on prefixed parameter for running the delay calculation. If there is no DMX parameter, it will fail when dispersion delay calls it. But we can add a quick patch.

scottransom:
I'm looking at the test_paramaters.py file.  Many of the newly added tests are not correct.  And we should have tests in there for *tons* of parameter stuff....

luojing1211:
Thank you Scott. I do need these feedbacks. since this is a big pull request, I probably miss a lot stuffs. I will fix it as soon as possible.

-Luo Jing

PR #95, opened 2016-02-04
Fix mjd parameter

luojing1211:
This a a quick fix to MJDparamter. I added a time_scale optional argument, which has not been used in the real model yet. But it will help us simplify our MJDparameter setup. 

For now, it seems the fitter.set_params does nothing right now. Not changing parameter value.


pint/fitter.py
```diff
@@ -50,16 +50,13 @@ def set_params(self, fitp):
         for p, val in zip(fitp.keys(), fitp.values()):
             # If value is unitless but model parameter is not, interpret the
             # value as being in the same units as the model
-            modval = getattr(self.model, p).value
+            modval = getattr(self.model, p).num_value
             # Right now while the code below preserves the unit, Angle types
```
scottransom: You should probably remove the comment(s) that are not applicable anymore since the Angle stuff has been removed.

luojing1211:
How about this get_MJD_parameter function. If we like it, I will remove the MJDparameter class.

luojing1211:
I think this is ready to merge. After this merged I will fix prefix branch.

scottransom:
OK.  Now you should be able to fix up the prefix one.

PR #96, opened 2016-02-05
Implementing priors on parameters

paulray:
Hi,

I felt like doing a little coding, so I started an implementation of priors on parameters.  This pull request is probably not ready to be merged, but I figured it would be a good place to start the discussion of how this should be done.

What is here initially is a way to compute prior probabilities on individual parameters, so a fitter could easily use them.  What is missing is a way to have priors on combinations of parameters (e.g. total proper motion or sky position).  This probably needs to be implemented at a higher level (e.g. in the TimingModel class, instead of just in the Parameter class).

Currently,  you can evaluate the prior probability of any parameter at its current value doing something like:
`m.F0.prior_probability()`

If you want to check what the prior would be given a value without having to set the model value in advance:
`m.F0.prior_probability(value=10.0)`

Note that priors all operation on the `num_value` values. I did this rather than the Quantity values because I think most fitters operate that way.

The default for every parameter is to return 1.0 for any value.  I've made subclasses to represent Gaussian and Bounded Uniform priors as well.  All you have to do is:
`m.F0.prior = pint.priors.GaussianPrior(10.0,1.0)`

Thoughts? @scottransom @vhaasteren @demorest @luojing1211


pint/models/parameter.py
```diff
@@ -134,6 +138,17 @@ def value(self,val):
...
+        """Return the prior probability, evaluated at the current value of
+        the parameter, or at a proposed value.
+        
+        Probabilities are evaluated using the num_value attribute
+        """
+        if value is None:
+            return self.prior.prior_probability(self.num_value)
+        else:
+            return self.prior.prior_probability(value)
+
```
scottransom: A slightly different (and perhaps a bit more pythonic) way to do something like this would be:

``
val = self.num_value if value is None else value
return self.prior.prior_probability(val)
``
With a shorter method name, it could easily be a one-liner:
``
return self.pp(self.num_value) if value is None else self.pp(value)
``
Not a big deal, though...

scottransom:
I like it.  We need priors...

pint/models/priors.py
```diff
@@ -0,0 +1,72 @@
...
+
+class GaussianPrior(Prior):
+    def __init__(self, mean, sigma):
+        super(GaussianPrior,self).__init__()
+        self.mean = mean
+        self.sigma = sigma
+        self.var = sigma**2
+
+    def prior_probability(self,value):
+        return np.exp(-(value-self.mean)**2/(2*self.var))/np.sqrt(2.0*np.pi*self.var)
```
paulray: This computation could have been done, perhaps more clearly and with less chance for error, by using scyipy.stats.norm(). Would that be encouraged or discouraged? Is it bad to have another dependency on scipy (obviously scipy is already used elsewhere in PINT, but I didn't know if we were trying to minimize dependencies like that).  What do you say @scottransom ?

pint/models/parameter.py
paulray: The 'pythonic' way reads too much like passive voice to me.

paulray:
I don't see any harm in merging this since it should break anything, but I want to hear from @vhaasteren first, in case he has a different idea on how this should be implemented.

pint/models/parameter.py
scottransom: I used to not like it, but it has grown on me.  And unlike passive voice, it is almost always shorter.  And in this particular case, I like how there is only 1 return call.

paulray: Maybe it's not so much like passive voice as it is like "You're pretty.... ugly!"  Or, "I like you.... NOT!" 
:-)

scottransom: My feelings for you are strong if you like me, otherwise, not so much....

paulray:
The most recent commit I added fixes the fitter to use num_value instead of value. I was failing when it was trying to fit for T0.  I probably should have made this a separate pull request, but I just added it on to this one...

pint/models/priors.py
vhaasteren: I think that we should also allow for log_probability to be returned. In border cases, 'incorrect' priors, or high-dimensionality, logarithms become essential.

vhaasteren:
I do like this implementation of priors. Regarding the modeling of these things, you can make it as complex as you'd like. Especially when resorting to hierarchical models. But we want to keep this aspect of PINT simple at this level. It would be very easy to override the behavior of these priors when doing more complicated stuff.

I think it is great to have the priors implemented as a part of the Parameter class. Every parameter should have a prior placed on it, so it is the natural place. And the prior is a distribution, which it is now.

The only thing I wonder is whether the class 'Prior' is not reinventing the wheel. Couldn't we use classes derived from (for example) the scipy continuous random variable class rv_contious? Those are pretty standard, and there are TONS of functions readily implemented...

paulray:
Thanks for the input @vhaasteren. The idea of deriving from the scipy class is interesting.  I have to think a bit about whether this is best done by inheritance or composition.

paulray:
@scottransom - I think this is ready to merge. You should take a look.  To make sure the priors were useful, I revised event_optimize to use them, and added a more complete command-line interface and a set of test datafiles.  So, I think the Prior() class and event_optimize are ready to be used and built upon.  There is definitely more to do, like making other types of priors (most especially a gaussian that has a limited range, so that you can have E = 0.95 +/- 0.02 and also not allow E>1.0)

paulray:
@vhaasteren - This updated implementation uses your suggestion of rv_continuous, and composition, not inheritance (the Prior class has a member that is an rv_continuous, but Prior does not inherit from that).

scottransom:
Looks good to me!  I'll play with it and maybe push up some small tweaks....

paulray:
That would be great. Any changes/improvements are welcome!

scottransom:
Hmmm...  Very first thing I tried has a problem:

lhotse:/data1/git/PINT$ python tests/test_priors.py
WARNING: Unrecognized parfile line 'SOLARN0               0.00' [pint.models.timing_model]
WARNING: Unrecognized parfile line 'T2CMETHOD           TEMPO' [pint.models.timing_model]
F0 is 61.485476554
  File "tests/test_priors.py", line 11, in <module>
    print("Uniform Prior on F0 {0}".format(m.F0.prior_probability()))
AttributeError: 'Parameter' object has no attribute 'prior_probability'

Also note that test_priors.py is not in a format that nosetests can run it....

paulray:
Oops, I forgot about that test and hadn't updated it.  Will fix.

paulray:
OK, made a new PR with the fix. Gotta run.

PR #103, opened 2016-02-09
tempo1 clock file parsing

demorest:
This function parses tempo1 clock files in a way more consistent with how tempo actually does it.  The main differences are:  to parse values based on fixed character offsets within the line (fortran-style) rather than expecting a certain number of columns; and use the info provided in both clock offset columns.  I don't know the historical reasons for having two columns, but it doesn't really matter.  If we are going to read these files we should use them the same way tempo does.

Features in tempo that are not currently implemented here are the ability to have INCLUDE statements in time files, and to disable interpolation for  certain data points (via the 'f' flag).

### DIFF
--- a/pint/observatories.py
+++ b/pint/observatories.py
@@ -8,6 +8,63 @@ from pint import pintdir
 class observatory(object):
     pass
 
+def load_tempo1_clock_file(filename,site=None):
+    """
+    Given the specified full path to the tempo1-format clock file, 
+    will return two numpy arrays containing the MJDs and the clock
+    corrections.  All computations here are done as in tempo, with
+    the exception of the 'F' flag (to disable interpolation), which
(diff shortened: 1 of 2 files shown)


demorest:
Also if we really want to be more consistent with tempo, we just should read $TEMPO/clock/time.dat and process the INCLUDE statements therein.  This would allow removal of the kludgey/hard-coded dictionary inside the get_clock_corr_vals() function that is currently needed to match the telescope name to the right time_XXX.dat filename.

paulray:
Fantastic @demorest !   I just did a spot check on one set of TOAs and the values look right for GBT.

One request:  Could you add a test to the tests directory that has TOAs from a few different observatories, and particularly Parkes TOAs from before and after the format change in the clock correction file?

Issue #104, opened 2016-02-10
Can not import Cache


luojing1211:
Hi folks, 
I was trying to use cache in parameter.py. But if failed to import. Do you have any idea what is happening? Thank you. 

ERROR: Failure: ImportError (cannot import name Cache)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/jingluo/anaconda/lib/python2.7/site-packages/nose/loader.py", line 413, in loadTestsFromName
    addr.filename, addr.module)
  File "/Users/jingluo/anaconda/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Users/jingluo/anaconda/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/jingluo/Research_codes/PINT/tests/test_parameters.py", line 1, in <module>
    from pint.models import parameter as p
  File "/Users/jingluo/Research_codes/PINT/pint/models/__init__.py", line 5, in <module>
    from .timing_model import TimingModel, generate_timing_model
  File "/Users/jingluo/Research_codes/PINT/pint/models/timing_model.py", line 4, in <module>
    from .parameter import strParameter
  File "/Users/jingluo/Research_codes/PINT/pint/models/parameter.py", line 16, in <module>
    from .timing_model import Cache
ImportError: cannot import name Cache

luojing1211:
Does any have the same problem? If this issue get resolved, the parameter opt2 branch will be ready for merging.

paulray:
Can  you give a specific code example that causes this issue?

luojing1211:
In the parameter.py file I have
```
from .timing_model import Cache
```
But when I ran test in the tests directory,  it returns the error. 
But in the pint/models directory same directory as parameter.py, we have astrometry.py it has the line:
```
from .timing_model import TimingModel, MissingParameter, Cache
```
And it works. Is there anything I forget to setup?

luojing1211:
I think I understand why. I was doing Circular import dependency. Timing_model imports parameter at the same time parameter is importing timeing_model.

paulray:
That is fine with me, but @demorest wrote Cache, so he should probably weigh in

demorest:
[...]

PR #105, opened 2016-02-10
Parameter opt2

luojing1211:
Hi developers, 

I updated parameters. Now you can use (floatParameter, strParameter, boolParameter, MJDParameter,
AngleParameter, and prefixParamters).

New Features:
* Add floatParameter strParameter boolParameter  MJDParameter AngleParameter class. Each class has a set of specific value process functions. 
* Optimise prefixParameter it uses new Parameter classes mentioned above. 

Issue:
* Timing_model cache seems not working, see issue #104 
* Some basic tests are provided, but more tests are needed.


demorest:
I think we could have left out a special class for floats, and just left it as default for Parameter.  Since most parameters are floats.

luojing1211:
The floatParameter setup right now is more than a standard Parameter, it has this set_value_float specialized function for value setter and the choice for longdouble or just float64. So having a floatParameter seems make sense to me.

luojing1211:
This branch should be ready for merge. I would be better to double check it by other folks, since there are a lot of changes in API. Please take a look or make a test on this branch. 

Thank you.

vhaasteren:
@luojing1211 At the moment the BT model is not fully ready yet, am I right? When I load a BT model parfile, then I get an error about line 39 of bt.py, where 'x' is a string, not a float

pint/models/parameter.py
```diff
@@ -58,44 +61,49 @@ class Parameter(object):
...
+        self.get_num_value = get_num_value
+        self.get_value = get_value  # Method to update value from num_value
+        self.parse_value = parse_value  # method to read a value from string,
+                                        # user can put the speicified format
+                                        # here
+        self.print_value = print_value  # method to convert value to a string.
+        if parse_uncertainty is None:
+            self.parse_uncertainty = fortran_float
+        else:
+            self.parse_uncertainty = parse_uncertainty
```
vhaasteren: Would it not make more sense to just do self.parse_uncertainty = parse_uncertainty, and set the value of parse_uncertainty to fortran_float by default? Seems shorter to me

pint/models/parameter.py
```diff
@@ -105,16 +113,16 @@ def units(self):
...
-            except:
-                log.warning('The value unit is not compatable with'
-                            ' parameter units,right now.')
+            if self.units is not None:
+                wmsg = 'Parameter '+self.name+' units has been reset to '+unt
+                wmsg += ' from '+self._units
+                log.warning(wmsg)
+                try:
+                    if hasattr(self.value, 'unit'):
+                        temp = self.value.to(self.num_unit)
```
vhaasteren: Instead of 'temp', you could also write '_'. Sometimes that is a clearer indication that the output won't be used.

pint/models/spindown.py
```diff
@@ -25,29 +25,28 @@ def __init__(self):
         # The number of terms in the taylor exapansion of spin freq (F0...FN)
         self.num_spin_terms = maxderivs
```
vhaasteren: Why do we have 'maxderivs' hardcoded in here? It should be easy to check how many of a certain prefix parameter we have in the model_builder class, and then pass that number on. Right now, this is an unnecessary limitation IMO.

Is there a limitation such that we don't do this right now?

pint/models/parameter.py
```diff
@@ -321,21 +479,59 @@ def __init__(self, name=None, value=None, description=None, units='rad',
             'deg': (u.deg, 'deg', '%.15fdeg'),
         }
```
vhaasteren: I like how you are dealing with the various units using this dictionary. But is 'separator' the right name of this dictionary? When I think of a separator, I think of things like the colon in 'h:m:s'.

How about 'unit_operator' or something like that? (I don't have much inspiration, sorry)

pint/models/parameter.py
```diff
@@ -441,22 +649,49 @@ def __init__(self, name=None, prefix=None, indexformat=None, index=1,
             self.unit_template = lambda x: self.units
         if self.description_template is None:
             self.description_template = lambda x: self.descrition
+        # Using other parameter class as a template for value and num_value
+        # setter
+        self.type_separator = {
```
vhaasteren: Same comment as for 'separator' above

pint/models/spindown.py
scottransom: This is probably left over from my initial kludgey implementation (which I did in about an hour when we needed it for BWM analyses of Fermi pulsars).  :-)

vhaasteren: I see. Well, the code is still being run. And the code fails if I add more F's in the parfile than in the maxderivs limit. Perhaps @luojing1211  can have a brief look at this, as it doesn't seem right.

scottransom: It's actually not a big deal, though.  There are similar limits in TEMPO (and probably in T2).  If you have more derivs than that, you should't be using a Taylor expansion...

vhaasteren: True, but I am not trying to be practical here.

scottransom: Fair enough.  (He sounds serious.  You should probably fix this, @luojing1211 ....)

vhaasteren: Arg, did I forget my smiley face again? :-)

pint/models/parameter.py
luojing1211: How about unit_identifier

pint/models/parameter.py
luojing1211: That is right, BT model has not been set use the new style parameters. But I can change it right now.

pint/models/spindown.py
paulray: If there is  a fixed maxderivs, it should probably be based on some numerical reason why going beyond that is likely to cause bad results and users should be prevented from doing so.

vhaasteren: I don't think there is a single identifiable number like that. It depends on the tim file, which we do not have access to here. There are two cases that could cause problems:

1) The number of TOAs is lower than the number of timing model parameters
2) The condition number of the design matrix becomes too large

Number 1 is solved by using appropriate priors. Number two is solved by the fitter, by normalizing the design matrix when creating inversion matrices. I have done both of these before. IMO, setting a number of maxderivs is ennecessarily limiting, even though in practice it would (almost) always make sense.

paulray: Good points, @vhaasteren . Plus, if we set arbitrary limits on array sizes by setting a variable called MAXDERIVS, we are in danger of looking like FORTRAN programmers. All we are missing is the all caps constant name.  :-)

luojing1211: Quick question, does F1 consider as a prefixed parameter?

paulray: I'm not sure I understand the question, but I haven't read over the prefix parameter code much yet. Isn't F* the prefix parameter?  Or, do prefix parameters necessarily count 1, 2, 3, .. instead of 0, 1, 2, 3, ..?  Looking at the code below, it seems that only F2 and above are prefixParameters.  Why is that?

scottransom: I agree.  I think that F* all should be prefix parameters.... (unless F0 has to be treated specially because of precision issues or some other reason).

luojing1211: Right now, all the spin term are initialized with long double, to keep the precision. I will make the change right now.

scottransom: The only one that needs longdouble, though, is F0 (and that only in *extreme* cases).  The others *never* have even close to that precision requirement.

paulray: But do we want to evaluate polynomials with a mix of float and longdouble? Seems simplest to keep all the F* parameters in the same representation.  F1+ would have to get promoted to longdouble anyway during the evaluation, wouldn't they?

vhaasteren: Agreed with all that was said.

If prefix parameters are not long double, then we can just define F0 separately, for the time being. (Yes, now I am being pragmatic)

luojing1211:
Spindown part has been modified. Please take a look.

luojing1211:
I set F0 as a prefixed parameter with longdouble, and F1+ term set as longdouble as well. This makes the code a lot cleaner.

pint/models/parameter.py
vhaasteren: Sure

vhaasteren:
@luojing1211 when is add_prefixed_param being called?

luojing1211:
Do you mean add_more_prefix_params? I don't think it has been called.

vhaasteren:
No, I mean this one, at the end of model_builder. What is its use?

  def add_prefixed_param(modelIns,names):
      """Add a prefixed parameter into the timing model.
         Parameter
         ----------
         modelIns : PINT Timing_model class
             The timing model of parameters adding to
         names : string or list of strings
             The name list of prefixed parameters.s
      """
      if type(names) is str:
          names = [names,]
      for n in names:
          prefix,indexformat,indexV = split_prefixed_name(n)
          if prefix in modelIns.prefix_params:
              units = modelIns.prefix_params_units[prefix]
              des = modelIns.prefix_params_description[prefix]

          modelIns.add_param(prefixParameter(name = n,units = units,value = 0.0,
                             description = des))

luojing1211:
I see. It was the old adding prefix parameter way. But for the set up right now, the equivalent function is  parameter.prefixParameter.new_index_prefix_param().

vhaasteren:
For as far as I am concerned, this is ready to be merged now. Anyone else any comments?

I do have extra comments, actually, but those follow from the new unit test I am writing. So I propose to first merge this one, so that I can make a PR on top of the new master.

vhaasteren:
Hmm, PR on top of PR is very ugly. I re-created the PR of the unit test on top of the current master. Much cleaner.

luojing1211:
I see what you would like to do. You want to set the units and update the num_value to new units. You can not do it right now. But it will be soon.

vhaasteren:
Yes, I put the test in there for the following reason.

Currently, if you set units to, say, 'kHz', the value of F0 does not change. So it is still 326 Hz. But then, if you do m.F0.value = 0.368, suddenly the value is 0.368 kHz. That, I think, does not follow the principle of least surprise.

luojing1211:
Test passed.

vhaasteren:
Great!

One thing. There is now A LOT of output during the tests. Snip:

WARNING: Parameter example units has been reset to Hz/s from Hz/s [pint.models.parameter]
WARNING: Parameter example units has been reset to Hz/s from Hz/s [pint.models.parameter]
WARNING: Unrecognized parfile line 'EPHVER              5' [pint.models.timing_model]
WARNING: Unrecognized parfile line 'MODE                1' [pint.models.timing_model]
.......WARNING: Parameter example units has been reset to Hz from Hz [pint.models.parameter]
.WARNING: Parameter example units has been reset to Hz from Hz [pint.models.parameter]
.WARNING: Parameter example units has been reset to Hz from Hz [pint.models.parameter]
WARNING: Parameter example units has been reset to Hz from Hz [pint.models.parameter]
.WARNING: Parameter example units has been reset to Hz from Hz [pint.models.parameter]
WARNING: Parameter example units has been reset to Hz from Hz [pint.models.parameter]
.WARNING: Parameter example units has been reset to Hz from Hz [pint.models.parameter]
WARNING: Parameter example units has been reset to Hz from Hz [pint.models.parameter]

Do we want that? We could suppress the log level in the unit test code I suppose. What do you think?

vhaasteren:
btw, that should be done in the setUpClass and tearDownClass functions of the unit test, where the tearDownClass would set it back to the level it was at before

luojing1211:
It is in the units setter. If we change unit frequently, I will remove this
part

vhaasteren:
I don't think we change units frequently, but we could consider changing the WARNING to an INFO. Otherwise we can do log.setLevel('ERROR') in the setUpClass.

Question is, do we consider it worthy of a WARNING? Thinking about it, the user probably knows what he/she is doing, so perhaps INFO is more appropriate.

scottransom:
Given that fitters etc expect parameters to have a certain unit, it is dangerous to do change the default unit.  So I think WARNING is appropriate.  I do think we can change the logging to only show errors when running the tests, though....

vhaasteren:
Is that true, @scottransom ? The fitters will expect the parameters to be in the units of the parameter derivatives. So if we modify the parameter derivatives to always return the numerical values of the units in m.par.units, I think there is no warning necessary. Or is there another reason?

vhaasteren:
By the way, I just realized that the m.par.uncertainty value is not in a different unit after we change 'units'. What is the behavior we want there? I'd say we want uncertainty to refer to the numerical value (without units? seems awkward, btw, I'd vote with units) wrt to the unit in m.par.units.

I am somewhat unsure now what we had agreed upon regarding the behavior during the telecon, and we definitely need to have the unit test reflect that behavior.

scottransom:
But not all fitters use derivatives.... But why couldn't the derivatives return the same units?  Although maybe that is a PITA...  As for uncertainties, I *definitely* think they should always be in consistent units with the values.   Personally, I think this is all nice "extras"...  users will likely almost never explicitly change the units.  They can get any unit they want using x.value.to(COMPATIBLE_UNIT).

luojing1211:
Right now this branch has a problem. If we build model from basic model, for example StandardTimingModel, rather than model_builder,  the prefix parameters will not be parsed, since add prefix part is in the model_builder. Do we need to fix this part and keep the functionality, or just build model from model_builder?

vhaasteren:
@scottransom The fitters that do not use derivatives would always work, as far as I can see. Derivatives should indeed return the same values as well. I just don't think they do that at the moment. I will fix that in conjunction with the wls_fitter.

@luojing1211 That is a slight paper cut indeed. But is it a severe problem? I think the recommended use should be the model builder. The separate timing model classes could implement the most use cases separately (like F0 and F1), but stuff like the DMX parameters and JUMP parameters will have to be done with the model builder.

However, at the very least, we should provide the _interface_ so that users can do it by themselves.

luojing1211:
The model_builder class has the functionality to allow the use select their own components. Just not use model_builder() instead.

luojing1211:
Don't merge it yet. I just found a bug. I will let you know when it is done.

vhaasteren:
Before merging, I'd also like to reach consensus on the parameter uncertainties.

I think they *should* have units, and they should respect the units set in model.par.units. Currently, they do not.

scottransom:
Yes.  They should definitely have units and I agree that they should be the same as for .value.

luojing1211:
Ok I will fix this.

luojing1211:
Should the .uncertainties  have the same type as .value?

vhaasteren:
The same units you mean? Yes, the same units and the same type.

luojing1211:
I mean the data type, like time object, angle object.

vhaasteren:
I would say so.

Thinking about it, this means we also need a num_uncertainty property that gives the numerical value of the uncertainty. Don't you agree?

luojing1211:
That make sense. These things should be properties as well, right ?

vhaasteren:
Yes

luojing1211:
I add uncertainty part and fixed bug. @vhaasteren If you want write more test for it, please do so.

luojing1211:
I don't have too much to add on this branch. I would like this branch get merged soon, so that we can get other modules coded, like dispersion and FD parameters. Please make comments on it.

paulray:
I just ran into a problem in my code where I was using p.num_value and p.uncertainty. Now that the uncertainty has units, this causes a problem.  I guess I need p.num_uncertainty.  So, I think what I need is for this pull request to be merged as well.

scottransom:
Let's do this.....

vhaasteren:
Perfect, y'all. Now we can move along with stuff!

PR #106, opened 2016-02-15
Generate Fake TOAs

paulray:
Generating fake TOAs is useful for several things, including making datasets for doing tests.
This PR includes:
* The ability to format TOAs as Tempo2 or Princeton format strings for output to a .tim file
* The ability to init a `TOA()` with an `astropy.Time`
* The ability to apply a vector of time adjustments to a TOA table
* A very simple top level script (currently called `zima.py`, as in fake beer) that generates a set of TOAs for a given frequency, observatory, and cadence, with 0 residual for a specific model, and writes them out as a .tim file.  This could be expanded to do a lot more things, like adding white and red noise and so on.

### DIFF
--- a/examples/zima.py
+++ b/examples/zima.py
@@ -9,13 +9,13 @@ import numpy as np
 import pint.toa as toa
 import pint.models
 import pint.fitter
-import pint.residuals
 import astropy.units as u
 import matplotlib.pyplot as plt
 from astropy.time import Time, TimeDelta
 import argparse
 
 from astropy import log
+log.setLevel('DEBUG')
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description="PINT tool for simulating TOAs")
@@ -29,7 +29,7 @@ if __name__ == '__main__':
 
     ntoa = 100
     duration = 400*u.day
-    start = Time(56000.0,scale='utc',format='mjd')
+    start = Time(56000.0,scale='utc',format='mjd',precision=9)
     error = 1.0e-6*u.s
     freq = 1400.0*u.MHz
     scale = 'utc'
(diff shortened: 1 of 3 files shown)


luojing1211:
To have a working d_phase_d_toa(), we need your fake toa generator. If this branch is merged I can change d_phase_d_toa() function to the new API.

paulray:
@luojing1211 - What function exactly do you need?  The generation of fake TOAs isn't very well encapsulated in my code.  I should probably wrap it up in a function or something.

luojing1211:
Init a TOA with an astropy Time. By the way, since astropy time can take a list of times, does your function has the ability to take one time array with list of time in it? Or just take a list of astropy time. 

-Luo Jing

paulray:
OK, this code shouldn't hurt anything, so I'm going to merge it.
You can init a `TOA()` with an `astropy.Time` as the MJD argument. Since a `TOA()` can only represent a single TOA, it does not work with a list of times.

scottransom:
Note that we need d_phase_d_toa to get an instantaneous topocentric 
pulsar freq as well.  And the new method/function adjust_TOAs() that 
Paul just input could be perfect for that.

Basically you would make a copy (if necessary) of the TOAs in question, 
and then add known offsets plus and minus the TOA so that you could 
determine a good numeric phase derivative via the .phase() method.

Scott

On 02/15/2016 11:20 AM, Jing Luo wrote:
> To have a working d_phase_d_toa(), we need your fake toa generator. If
> this branch is merged I can change d_phase_d_toa() function to the new API.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nanograv/PINT/pull/106#issuecomment-184278840>.Web
> Bug from
> https://github.com/notifications/beacon/AAIHIsn8C_K6ie0igOynmHahbIUj9owXks5pkfJlgaJpZM4Had_T.gif
>

-- 
Scott M. Ransom            Address:  NRAO
Phone:  (434) 296-0320               520 Edgemont Rd.
email:  sransom@nrao.edu             Charlottesville, VA 22903 USA
GPG Fingerprint: 06A9 9553 78BE 16DB 407B  FFCA 9BFA B6FF FFD3 2989

PR #107, opened 2016-02-16
Fix d_phase_d_toa

luojing1211:
This is a new d_phase_d_toa function. 
New features
* Use new fake toas and toas.adjust_toa() API
* Enable to calculate d_phase_d_toa at observator , SSB and geocenter. 
* Test file is included
* Using second order finite difference. phase'(t) = (phase(t+h)-phase(t-h))/2+ 1/6*F2*h^2 + ..

Issues
* Results have to be double checked. 
* Geocenter time scale. What is the right time scale for geocenter toas?
Please leave your comments.


demorest:
VLBI gating needs geocentric polycos.

paulray:
Fermi also frequently uses Geocentric TOAs.  It is in PINT as one of the observatory special cases (the others being Barycenter and Spacecraft.

luojing1211:
I added geocenter into d_phase_d_toa function. 
But my question is what is the time scale for geocenter toas class?
This is what I did, 
geoTOA = toas['tdbld'] + topo_center_delay(toas)
geoTOAsObj =  toa.read_fake_TOAs(geoTOA, obs='Geocenter', scale='utc')

Should it be 'tdb'?

paulray:
I just looked this over a bit and I have some questions. I think we need to go back to thinking about the use cases for this function.  In particular, I don't understand why `d_phase_d_toa()` has an `obs` argument.  I think this was causing some confusion in our discussion at the telecon.

What I had been thinking was that this would compute dphase/dt at a set of TOAs, *in the time system of the TOA*.  So, if you want topocentric frequency, you call this with topocentric TOAs.  If you want barycentric frequencies, you call this with a set of barycentric TOAs (which might be fake TOAs generated on a grid).  When would you want to give a set of topocentric TOAs and get dphase/dt at the barycenter?

My two use cases are (1) knowing the instantaneous local period, perhaps for folding or generating polycos, and (2) knowing the barycentric period at a set of times, perhaps for fitting an orbit to a set of measured barycentric period from PRESTO.

@luojing1211 and others, what are the other use cases that drive the obs argument?

scottransom:
I don't know why there should be an obs argument either.  If you want topocentric d_phase_d_toa, the TOAs already know what observatory they are from.  So specifying an obs is redundant....

luojing1211:
Yeah, I understand since the telecon, and I am fixing it right now.

luojing1211:
Do we have a method to test the error? I am using centered finite difference method, I would like to do the test 
```
| p'(toa)_real - p'(toa)_numerical | <= 1/6 * p'''(toa) * h^2
```
Here h is the sample step. 

How do we get p'(toa)_real part? 

-Luo Jing

luojing1211:
Hi all, 
I have modified my test file. I am using PITN generated Barycenter TOAs from a par file only have F0 and F1. 

But I have noticed that the simulated toas has a 1e-7 s level residual. @paulray , could you tell me more about it. It could cause my test issue. 
Test code:
```
def test_d_phase_d_toa(self):
        # TODO: How do you Calculate real derivative. To analysis the error?
        analog = self.ana_diff()
        emsg = ""
        for fac in [0.1, 1, 10, 100, 1000, 5000]:
            step = 1.0/self.model.F0.num_value*fac
            dpdtoa_model = self.model.d_phase_d_toa(self.toas, step)
            diff = dpdtoa_model - analog
            emsg += str(max(np.abs(diff))) + " "
        assert False, emsg

    def ana_diff(self):
        der = self.model.F0.num_value + \
              self.model.F1.num_value * ((self.toas.table['tdbld'] - \
                                self.model.PEPOCH.num_value)* 86400.0)
        return der
```
Since I want to see the result, I put assert False
Result:
```
  File "/Users/jingluo/Research_codes/PINT/tests/test_d_phase_d_toa.py", line 39, in test_d_phase_d_toa
    assert False, emsg
AssertionError: 8.67044666414e-06 9.80430027206e-07 1.41208955591e-07 1.40903139087e-08 1.47089302543e-09 2.34026117929e-10
```
I am not sure why I increase the step size, it gives me a lower error.  Could it be the residuals?

paulray:
My code that makes fake TOAs does not achieve 0 residuals exactly.  It does two iterations and that does seem to get the residuals down to about 1e-7 seconds, as I recall.  This could be improved with a third iteration, or by using a better estimate of the local period (i.e. d_phase_d_TOA instead of 1/model.F0)

luojing1211:
I see. I want to use this TOAs to test d_phase_d_TOA, so a third iteration
would make more sense.

-Luo Jing

luojing1211:
The test is running. Now pint d_phase_d_toa compare to an analog calculation is around 1e-8 level. Depend on the step size, it can be lower. Compare to tempo polyco, it is around the same level as well. 

Let me know if the precision is enough.

luojing1211:
This is the d_phase_d_toa test result. 
The test uses a simple model:
PSR              1744-1134
RAJ      17:44:29.39171911  0          0.00002312
DECJ     -11:34:54.5671567  0           0.0010157
PMRA               18.6744  0              0.0308
PMDEC             -10.1488  0              0.0932
PX                  2.9362  0              0.1678
POSEPOCH        50434.0000
F0    245.4261237074677808  0  0.0000000000041300
F1     -5.380851374862D-16  0  1.284265997768D-20
F2     1.2083172450962419925e-23 
PEPOCH        50434.000000
START            53216.126
FINISH           55122.747
It compares the analog d_phase_d_toa calculation and numerical calculation which is used in PINT. 
And this is the result. 
screen shot 2016-04-05 at 10 02 49 am

luojing1211:
This pull request is too old. See pull request #185

PR #113, opened 2016-02-22
Changed 'setup' import in setup.py from setuptools to fix python setu…

vhaasteren:
…p.py develop

The setup import in setup.py now comes from setuptools. I had issues doing

    python setup.py develop

This fixes the issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,10 @@
-from setuptools import setup
+try:
+    from setuptools import setup, Extension
+except ImportError:
+    from distutils.core import setup
+    from distutils.extension import Extension
+
 import numpy, os, sys
-from distutils.extension import Extension
 try:
     from Cython.Distutils import build_ext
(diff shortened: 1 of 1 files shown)


vhaasteren:
@demorest , I placed this in a PR, since you were also doing something with setup.py. It should be easy to check whether this works, since it is one line.

demorest:
That seems fine to me.  Should we also then import Extension from setuptools rather than distutils?  It actually seems to work either way...

vhaasteren:
I looked up the difference in a bit more detail. distutils is still the standard. Perhaps we should try to import setuptools, and if that fails we default back to distutils for both.

demorest:
Ok, that sounds like a good plan.  Do you want to add those changes into this PR?

vhaasteren:
Yes, done. I'll merge it.

PR #114, opened 2016-02-23
Convert uses of np.float128 to np.longdoble

paulray:
Simple PR to remove uses of `float128`, which is not portable to all architectures. Using `np.longdouble` is preferred.


paulray:
Note that _many_ tests are currently failing but I don't think that is due to anything in this pull request. Hopefully they will all be fixed when the other PRs on binary models are merged.

PR #115, opened 2016-02-23
Remove need for PINT environment variable.

demorest:
As discussed on the call yesterday, it would be nice and modern for users to not need to set a PINT environment variable.  This was actually pretty easy, please check out this branch and see what you think.  Before merging this in, a few things to discuss:

Note that setup.py will now install the datafiles under the installation target directory.  Before this can happen, 'download_data.sh' still needs to be run manually.  I don't see any reason why setup.py couldn't just download the files itself rather than needing an extra step.  If this sounds OK to you all, I will add this in and remove the download_data script.

Right now, PINT will still first check for a PINT environment variable and if found use that rather than the install dir.  This is why the datafiles/ subdir is now symlinked into the top-level repository dir.  If this is a feature we want, we can keep this, or we could just have PINT ignore the environment variable.  Any preferences?


paulray:
So, we'll have the datafiles in the place where we did the git checkout, and another copy in the install location.  One potential drawback is that people may have some trouble actually locating the install location (in some site-packages somewhere) and may or may not be able to modify files there.
So, what is the process for updating datafiles?  Say I download de430 to my local datafiles directory. Do I run setup.py install again to get it to the place where PINT will find it?

Also, will the auto downloads of things like IERS A still work in this setup?

demorest:
Yes, the files will exist both under the source tree and the install dir.  However, this is not any different than what happens with all the *.py code files.

The permissions thing could be a reason to keep the environment variable as an option.  I think the situation of someone really wanting to have their own copy of ephemeris files, but not their own install of PINT seems pretty unlikely.  But if it does happen they could set the PINT env var to override the system-wide install.

Yes, currently to update data files you would run the download script, and then run setup.py to install them (unless you are running PINT from within the source dir rather than using the install option).  I think this is kind of confusing and is why I was suggesting setup.py should just download the files itself as needed.

The IERS downloads are handled totally separately, using the astropy download utils.  I thought about using the same setup for the ephemeris files but I think the current way is probably better.  Mainly because the ephemeris files are not constantly changing/updating like the IERS stuff.  So a site-wide install rather than a per-user install is OK, and there's not a need to check for changes every time the file is used.

vhaasteren:
I would vote for no environment variable, ever. It is very user un-friendly.

How about the following:
- Install the datafiles like @demorest has already implemented. Those are the reference datafiles, and they are updated with all the *.py files that are installed.
- Those are not the ones that are used in practice though
- When run, pint checks whether the user data files are present. We can use appdirs for that:
https://pypi.python.org/pypi/appdirs/1.4.0
- If the user data dir is not set, we copy the content from the installation datafiles. Otherwise, we use the content set in the user data directory we get from appdirs.

This way, the data files are in a completely standard location, just like any other program on your system. When we update PINT, we should have a check whether or not we want to overwrite the user data, and we can write some scripts for that.

paulray:
I think I follow what you are suggesting, @vhaasteren .
So, I will likely have 3 copies of the datafiles directory (which is currently 81 MB, by the way):

1. In my git checkout of PINT 
2. In  ~/Library/Python/2.7/lib/python/site-packages/pint (assuming I did `python setup.py install --user`)
3. In ~/Library/Application Support/PINT

And the one actually used when I run is 3.
Is that right?

vhaasteren:
That is indeed what I am suggesting, @paulray . Location number (3) is completely standard, and that is where you would modify things if you wanted to. If location number (3) does not exist, it is created on runtime from number (2) at the start of PINT.

Note here that (2) can vary, depending where and who installed PINT, but it is always know to PINT itself where that is. And the user will always have access to (3), since it is unique to each user.

The only tricky thing would be if a sys-admin would want to update clock files system wide. That would only work if the sys-admin updated (2), and then somehow gave the command that nr. (3) needs updating. I am not sure how that would work...

demorest:
I don't think we should automatically copy/update anything in the per-user directories.  I think we could check a per-user location first, so people who want to override the system-wide defaults are able to.  But if they are doing something non-standard it should be up to them to keep their files updated (or not) as they see fit.  For example what if someone wants to do tests using an old clock file version for some reason.  It would be really annoying if PINT keeps overwriting it with a "better" one.

Otherwise I think this plan sounds good, although it would be best to not introduce another package dependency.  I think there is some similar code in astropy for locating appropriate config file directories, maybe we could make use of that.

vhaasteren:
That is a good suggestion @demorest , I like that. If the user wants to take matters in their own hands, they can. But by not doing it, the default in (2) would be used.

And all the better if there is code in astropy to do that. I had only used appdirs once, but it is indeed not installed by default.

paulray:
Yeah, I like the check for user overrides first, then default to the installed versions.

Of course, the full path to all datafiles read by PINT should be printed with log.info(), so it can always be seen by the user.

demorest:
Hi all, these changes are ready to merge in.  I did end up using the appdirs package; it is a single file, so I've just added it into the PINT repository (it seems to be intended for this kind of usage).  @vhaasteren , thanks for the pointer to this!  I added an extern/ subdir to hold this and any other external packages we want to bundle with PINT.

Note that many of the test routines still refer to the PINT environment variable.  I will fix these but thought it might be cleaner (git history wise) to do as a separate PR after this one is merged in.  Thoughts?

vhaasteren:
Thanks @demorest , that looks good. Yes, separately would indeed be cleaner, as you figured now already, but I don't think it would have mattered that much.

Anyways, great that you managed to get this done, and using the appdirs package like a bundled dependency sounds fine. I'm sure we'll encounter more such examples. One question I have is about the licensing. The copyright is still held by ActiveState Software Inc. and Eddy Petrisor. Is it ok to bundle it in a package like this? Do we have permission?

demorest:
Yes, it is released under the MIT license.  We can do anything we want with it as long as their copyright message stays there.

demorest:
Actually a copy of the license is supposed to stay with the code as well.  This was in a separate file in the original distribution, I have just added it into our copy of appdirs.py.

This does raise the question (which I think we've discussed but never answered) of how/if we should license PINT...

vhaasteren:
Ok, great. The MIT license is perfect.

IMO, we want to be explicit about licensing, so we should pick a license. I'd vote for the MIT license, but perhaps we can talk about it next week?

PR #120, opened 2016-02-24
Changed bounded priors to be inclusive of endpoints.  

paulray:
This change allows E==0.0 when you set UniformBoundedPrior(0.0,1.0).  It also allows 1.0, so we might want to consider whether we need to implement [lower,upper], (lower,upper], and [lower,upper).  But for now, I think that just allowing inclusive ranges should be OK.  For something like SINI, it is certainly appropriate.


paulray:
I hope no one minds if I just merge these tiny changes.

PR #121, opened 2016-02-25
Fix eccentricity check in pulsar_binary.py

paulray:
With this change, eccentricity == 0.0 is now allowed


luojing1211:
I tested it, it works.

PR #123, opened 2016-02-27
Generic DM model

luojing1211:
This is generic DM model. 

Changes: 
* Change the file name from dispersion.py to dispersion_models.py
* Add dispersion class as the base class from different dispersion model. 
* Add basic dispersion class 
* Add DMX dispersion model class 
Todo:
* Get model builder work with new dispersion models. 

I think it would be nice to make the model builder smarter to select the right model based on parfile, even they are very similar.


paulray:
Closing for re-factoring, as per telecon

PR #125, opened 2016-02-29
Fd parameter

luojing1211:
This is a pull request for FD model:
* Add frequency_dependent.py file
* Add FD model for FD_delay() and d_phase_d_FD() functions. 
* Add FD parameters in parfile B1855+09_NANOGrav_dfg+12_modified.par

TODO:
* Add test for FD parameters. 
* Figure out what is the expected test results. 

PS: This pull request is comparing with nanograv/PINT dev branch.


paulray:
It is great that you are doing this against the dev branch!

However, before we do that we should make sure that the dev branch is up to date with master. That is why there are a ton of commits on this pull request, which makes it hard to see what is really new.

I just updated the dev branch to be current with master and pushed. I had hoped that would clean up this pull request, but it did not.  Anyone know how to do that?  Or doe we just have to delete and re-submit this PR?  @scottransom?

luojing1211:
I can just submit it again.

-Luo Jing

PR #126, opened 2016-02-29
Add FD model

luojing1211:
This is a pull request for FD model:

Add frequency_dependent.py file
Add FD model for FD_delay() and d_phase_d_FD() functions.
Add FD parameters in parfile B1855+09_NANOGrav_dfg+12_modified.par
TODO:

Add test for FD parameters.(Already implemented)
Figure out what is the expected test results.
PS: This pull request is comparing with nanograv/PINT dev branch.


pint/models/frequency_dependent.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    def __init__(self):
+        super(FD, self).__init__()
+        self.add_param(p.prefixParameter(name='FD1', units="second", value=0.0,
+                       descriptionTplt=lambda x: ("%d term of frequency"
+                                                  " dependent  coefficients" % x),
+                       unitTplt=lambda x: 'second',
+                       type_match='float'))
+
+        self.delay_funcs['L1'] += [self.FD_delay]
```
paulray: Here, it would be good to put in a comment about why these delays were being assigned to 'L1'.
As I understand, 'L1' is for delays local to the Solar System that should be applied to get (what we usually call) barycenter-corrected TOAs.  I'm actually not sure whether FD terms should be L1 or L2.

This also made me realize that L1 and L2 are not very descriptive names.  We should discuss a better naming convention.  Maybe 'SolarSystem' and 'PulsarSystem' or something like that?

Also, should the FD terms really be a phase lag, rather than a delay term?  I guess they are implemented as delays in TEMPO.  What do you think @demorest ?

pint/models/frequency_dependent.py
```diff
@@ -0,0 +1,61 @@
...
+        FD_terms.sort()
+        FD_in_order = range(1,max(FD_terms)+1)
+        if not FD_terms == FD_in_order:
+            diff = list(set(FD_in_order) - set(FD_terms))
+            raise MissingParameter("Spindown", "FD%d"%diff[0])
+        self.num_FD_terms = len(FD_terms)
+
+    def FD_delay(self, toas):
+        """This is a function for calculation of frequency dependent delay.
+        See nanograv 9 years data set (2015) Eq.(2)
```
paulray: Would be good to put the full reference citation here.

pint/models/frequency_dependent.py
luojing1211: I agree with you that 'L1' and 'L2' are not the perfect name. 'SolarSystem'  and 'PulsarSystem' seems hard to apply to dispersion delay.

pint/models/frequency_dependent.py
```diff
@@ -0,0 +1,61 @@
...
+        FDdelay = sum(c_i * (log(obs_freq/1GHz))^i)
+        """
+        FD_delay = np.zeros_like(toas, dtype=np.longdouble) * u.second
+        for ii in range(1,self.num_FD_terms+1):
+            FD_coef = getattr(self, self.FDmapping[ii])
+            log_freq = np.log(toas['freq'] / (1 * u.GHz))
+            FD_delay += FD_coef.value * (log_freq) ** ii
```
paulray: It probably doesn't actually matter for the FD parameters, but it would be good if we were consistent in PINT about how we evaluate polynomials, since there are numerical precision issues you can run into.  In utils.py, there is `taylor_horner()` which is close to what you want, but not the same since it includes the factorials in the denominator appropriate for a Taylor expansion.   Maybe we should make a similar evaluate_polynomial() function where you specify the coefficients c_i and the factorials are not assumed?  This would allow those two functions to be optimized at some later date and all the places that use them would automatically get the benefit of the improved speed or accuracy. Thoughts? @scottransom @demorest @vhaasteren

pint/models/frequency_dependent.py
paulray: Good point @luojing1211 . The names need more thought.

pint/models/frequency_dependent.py
scottransom: Yeah, I completely agree with this.  Should be simple to split the function you are talking about out of taylor_horner() and then call that function within taylor_horner.

paulray: Actually, it turns out numpy already includes this function.  We should just use it:
http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.polyval.html

luojing1211: It is done.

pint/utils.py
```diff
@@ -370,11 +371,9 @@ def taylor_horner(x, coeffs):
     Out[1]: 40.0
 
     """
-    result = 0.0
-    fact = float(len(coeffs))
-    for coeff in coeffs[::-1]:
-        result = result * x / fact + coeff
-        fact -= 1.0
+    fact = factorial(range(len(coeffs)-1, -1, -1))
```
scottransom: Hmmm...  I'm not sure that I like changing the version of taylor_horner that I had to using factorial() and np.polyval()..  I think that this new version us not as stable floating-point-wise.  That's because for high-order expansions, when the factorial values become very large, you end up with large terms in the numerator and denominator  which can cause floating point issues.  With the older implementation, the factorial calculation is accumulated piecewise as x/fact which means that you never get huge numbers.  Also, a minor point is that np.flipud() is kind of opaque for 1-D arrays or lists.  you should use coeffs[::-1] which is Python for "a copy of the array but in reversed order".

luojing1211:
Hi all, I add some test for FD parameter. This is how I did the test. I use tempo2 to generate set of good toas. Then, I just use a little script "simulate_FD_model.py" add FD delay to the tempo2 generated toas. 
I compare how this two software correct the FD delays. There are some odd points in the tempo-PINT difference. I am going to track down what is going on.

luojing1211:
Hi all, 
This is the test result against tempo2:
screen shot 2016-03-11 at 3 26 36 pm
The odd peak comes from the difference between Tempo2 clock correction and PINT clock correction.
Here is the plot:
screen shot 2016-03-11 at 3 27 07 pm

PR #129, opened 2016-03-07
Jump

luojing1211:
# Implement JUMPS

Hi all, 
I implement jump. I would like this pull request get commented. 
New Features:
* Add new parameter type maskParameter
* Model_builder recognise JUMP
* Implement JUMP phase
* Add param_register in timing model

TODO:
* A better test is needed.


paulray:
Needs conflicts resolved before merging

luojing1211:
This should be OK.

PR #136, opened 2016-03-22
Development

luojing1211:
Hi all, 

This is quick fix for JUMP units and part of removing tempo dependency.


paulray:
I don't quite understand this. It is a big pull request against master. Can you explain?  I think it may be trying to simultaneously bring master up to date with dev.  This probably isn't the best way to accomplish this.

paulray:
Note that because we haven't yet fully implemented our process for pull requests only being against dev, the master branch had some commits that were not on dev.  I just fixed this by merging master into dev. Now dev is ahead of master (it has commits that are not on master), and not behind master (there are no commits on master that are not on dev as well). So, you might want to fetch from nanograv and figure out if this PR should be against master or dev.

luojing1211:
I think I merged some of my other branch to my development branch long time ago. I will fix it.

luojing1211:
I made a mistake. But I made a new one, looks a lot better.  This can be closed.

PR #137, opened 2016-03-22
Quick fix for jump unit and part of tempo independent test. 

luojing1211:
Hi all, 

This is quick for JUMP units and removing tempo dependent tests. 

This branch should be ready.
-Luo Jing


pint/models/phase_jump.py
```diff
@@ -9,11 +9,13 @@
 
 
 class PhaseJump(TimingModel):
-    """This is a class to implement phase jumps 
+    """This is a class to implement phase jumps
     """
     def __init__(self):
         super(PhaseJump, self).__init__()
-        self.add_param(p.maskParameter(name = 'JUMP'))
+        # NOTE: it is dangours to put jump unit as second
```
paulray: In a comment like this, please put some more explanation about why it is dangerous.  Someone looking at it should be able to understand what the issue is.  Link to an issue on github, if needed.

pint/models/phase_jump.py
```diff
@@ -9,11 +9,14 @@
...
 
 class PhaseJump(TimingModel):
-    """This is a class to implement phase jumps 
+    """This is a class to implement phase jumps
     """
     def __init__(self):
         super(PhaseJump, self).__init__()
-        self.add_param(p.maskParameter(name = 'JUMP'))
+        # NOTE: it is dangours to put jump unit as second.
+        # Since JUMP is not always in the unit of time.
```
paulray: I'm confused here. When are JUMPs not in time units? I thought all JUMPs were time, or maybe I'm forgetting something.

Also, these are called `PhaseJump` and are implemented in the `phase_funcs` part of the timing model.  Is that the way TEMPO implements jumps?  What the the procedure for converting between the time units of the jump and phase?

demorest: In TEMPO, JUMPs are always in time units, and are included as a time delay.  (They get added into the clock correction;  note, this is _not_ how we should do it here, just explaining how it is done in the tempo code.)

In TEMPO2, JUMPs are implemented the way @luojing1211 has done it here.

I think we should follow the TEMPO version, have JUMP specified in seconds and apply it as a delay (not phase).  I don't think there is anything dangerous about this, this seems totally straightforward to me.  As we've discussed before, we may also want to allow an actual phase jump (with units of pulse phase) using a different parameter name.

luojing1211: Ok, I will change it very soon. Is this a 'L1' (before binary ) delay or 'L2' (after binary ) delay?

demorest: I think a time delay would be L1 and a phase jump would be L2.

luojing1211:
I think this is ready. Please merge and try it out. If there is any problem, let me know.

PR #140, opened 2016-03-25
New dispersion

luojing1211:
Hi All, 

This is a new design for dispersion model
New Features
* DM variances are treated as linear combination for example, dm = constant_dm + dmx
* One delay function are provided. 

Note:
* Please merge the quick fix for jump branch first. 
* DM model selector are not provided.

### DIFF
--- a/pint/models/dispersion_model.py
+++ b/pint/models/dispersion_model.py
@@ -39,8 +39,9 @@ class Dispersion(TimingModel):
     def dispersion_time_delay(self, DM, freq):
         """Return the dispersion time delay for a set of frequency.
         This equation if cited from Duncan Lorimer, Michael Kramer, Handbook of Pulsar
-        # Astronomy, Second edition, Page 86, Equation [4.7]
-        Here we assume the reference frequency is at infinity. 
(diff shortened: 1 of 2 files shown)


pint/models/dispersion_model.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def constant_dm(self, toas):
+        cdm = np.zeros(len(toas))
+        cdm.fill(self.DM.num_value)
+        return cdm * self.DM.num_unit
+
+    def dispersion_time_delay(self, DM, freq):
+        """Return the dispersion time delay for a set of frequency."""
+        # dm delay
+        dmdelay = DM * DMconst / freq**2.0
```
paulray: Both this line and the DMconst definition above should have a reference to the literature for where the equation and value come from, such as Eqns 4.4 and 4.6 from Lorimer and Kramer.  In fact, looking at that, I think the DMconst should be DMconst = 4148.808.  Also, we should include a note that the equation assumes that freq >> f_p (the plasma freq)

luojing1211:
As discussed on the telecon, this branch can be merged.

PR #153, opened 2016-06-08
bugfix and added GLF2 in glitch.py

dcfidalgo:
Hi,
- glitch.py: Fixed a small bug (mix-up with astropy units) and added the parameter GLF2_ since tempo2 has it and i guess it does not hurt to have it as well. I also extended a bit the check on the parameters.
- fermi_toas.py: Added two more arguments to the load_Fermi_TOAs method (minmjd and maxmjd, so it does not create TOA instances for events you are not interested anyways). 

David

### DIFF
--- a/pint/fermi_toas.py
+++ b/pint/fermi_toas.py
@@ -96,7 +96,9 @@ def phaseogram(mjds, phases, weights=None, title=None, bins=100, rotate=0.0, siz
     else:
         plt.show()
 
-def load_Fermi_TOAs(ft1name,ft2name=None,weightcolumn=None,targetcoord=None,logeref=4.1, logesig=0.5,minweight=0.0):
+def load_Fermi_TOAs(ft1name, ft2name=None, weightcolumn=None, targetcoord=None,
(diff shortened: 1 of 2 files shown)


paulray:
Why is this not merged yet?

@dcfidalgo Can you rebase this on the current master to clear up the conflicts?  It would be great to add a small test to test the GLF2 functionality as well.

paulray:
The Fermi part of this has now been added and merged.

@luojing1211 do you want to add GLF2 to the glitch model?  Enough has changed since this PR was filed that we should probably just close this and make that change independently.

luojing1211:
Will do.

luojing1211:
Hi @paulray, Does GLF2 the highest order we are currently modeling the Glitch?

paulray:
I've never actually used GLF2 myself, but I think we can safely assume that GLF2 is the highest order we need to be able to model.

luojing1211:
This problem should be fixed by PR #293. @paulray do you want to take a look at PR #293

paulray:
This has been superseded.  Closing.

PR #156, opened 2016-07-07
Timing model update

luojing1211:
Those are some update from the timing model code review.


paulray:
I think this is fine to merge, but I find the name `get_type_params` confusing. Would it be clearer to use something like `get_params_of_type` or something?

luojing1211:
This branch is ok. Merge to dev branch.

PR #157, opened 2016-07-12
Update parameter

luojing1211:
Hi all, 

Here is the new update on parameters.


luojing1211:
This pull request is ready to merge.

luojing1211:
Can we merge this branch now?

paulray:
Hi @luojing1211 
I looked it over and there are a couple of things I think you should fix before merging, primarily in the docstring for the Parameter class.  In the description, it still mentions 'num_value', which is actually no longer used, so that should be fixed. Then I'd like to see an additional paragraph added with the philosophy for how Parameter is implemented.  That is, make it say clearly to developers that normally you should use p.quantity to access the current value of the parameter and that it will always be a quantity with units. And, for developers who need to access the bare numeric value, they should use p.value and that the value returned will alway correspond to the unit p.units (even if p.quantity has a different unit). This is for consistency so that someone writing a fitter that isn't units-aware will always get the same unit for that parameter.  

That docstring paragraph should also explain how uncertainty is defined.  I don't really love the lack of parallelism between p.quantity/p.value and p.uncertainty/p.num_uncertainty.

Next, I don't quite understand some of the examples in the notebook. For example, why is `p.quantity = 200` allowed?  Shouldn't the quantity always have to be set to a number with units?
And, is there really a need to allow floating point values and quantities to be set as strings, like `par.quantity = '300'`. That seems confusing to me.

In addition, there are a number of other places in the code where num_value still appears. These should probably all be cleaned up.

luojing1211:
Hi @paulray , Thanks for the comment. These are very good point. I will fix the num_value part right away. The docstring will be added to the class. 

For lacking of parallelism, how about p.uncertainty and p.uncertainty_value. For definition of uncertainty, do you mean that in the parameter section, we should put uncertainty definition for each parameter, or a more exact definition like wikipedia. 

Actually, setting p.quantity is more for MJDParameter or AngleParameter. We can set TZRMJD using a string or float not necessarily only use astropy.Time object.  Maybe we should discuss about this.

luojing1211:
Hi, I just updated docstring. For parameter subclass, I updated the part for floatParameter only. Could you please check if anything I can improve.

For the setting .quantity using string or a pure number part, what do you think @scottransom @demorest .

PR #166, opened 2016-08-17
Update prefix parameter

luojing1211:
Hi all, 
First of all, dev branch needs to rebase on master. 

This is a new update for the prefix parameter and masked parameter. It simplifies building the prefix parameter and mask parameter. But some questions raised. 
1. Does maskParameter (The type for JUMP parameter type) only gives an float parameter value? Could it be an MJD?
2. For the changes in the parameter.py line 797. I change the setatter to customized setattr. Does it has potential problems?

Folks, please comment on this pull request.

Thank you.


luojing1211:
@scottransom  I ready corrected the getter and setter. Test on my computer is passed. If you feel it is good. We can merge it.

PR #167, opened 2016-08-17
change dmx time compare to a low presicion one

luojing1211:
A little change. When comparing DMXR1 and DMXR2 using float mjd data type. Avoid comparing two time object directly. 

TODO: Maybe adding a low precision mjd column in the toa table could help a little bit when doing toa selection.


luojing1211:
This update will reduce the time of computing delays. This branch pass the tests. If there is no other problems. This should be merged.

luojing1211:
I already made the change. This should be ready to merge.

PR #170, opened 2016-08-18
Comp order

luojing1211:
Hi all, 

This is the component order code I have developed in the last two weeks. Please check it out, and comment on it.


paulray:
@luojing1211 should we close this for now? It seems to have been overtaken by events.

luojing1211:
I agree.

PR #174, opened 2016-09-02
Binary opt2

luojing1211:
This is an update for binary model. New Features:
* Bug fix
* Changed class names
* Make binary functions more generalized. 
* Optimize the Binary input. 
* Update docstring.

PR #175, opened 2016-09-02
Ell1 model

luojing1211:
Add ELL1 model. Merge Pull request #174 first then, merge this Pull request


demorest:
The test fails because the .tim file for J0613 hasn't been added to the repository.

luojing1211:
It should be ok now.

PR #176, opened 2016-09-20
Update bt model

luojing1211:
Hi all, 
This is the updated version of BT model.


demorest:
Few small problems:

1. binary_generic.py lines 109 and 110 are missing commas inside the parameter dict.

2. pint/models/__init__.py needs the "from .bt import BT" line removed since bt.py no longer exists.

3. tests/test_bt.py either needs to be updated to use the new BT classes, or removed if it is no longer needed.

luojing1211:
It should be fixed. Please test is again.

PR #180, opened 2016-10-05
"pulsar MJD" time format

demorest:
This fixes an incompatibility between the traditional pulsar definition of UTC MJD that is used in TOA lines, versus what is used in astropy/ERFA/SOFA.  The latter define fractional MJD as "seconds_in_day/length_of_day" where length of day is usually 86400 but is 86401 on leap second days.  Pulsar TOA fractional MJDs are _always_ "seconds_in_day/86400" (which means TOAs can never happen during leap seconds).  The approach is to define a new astropy-compatible TimeFormat class that is then used when initializing Time objects from TOA information.

### DIFF
--- a/pint/pulsar_mjd.py
+++ b/pint/pulsar_mjd.py
@@ -30,8 +30,21 @@ class TimePulsarMJD(TimeFormat):
 
     @property
     def value(self):
-        # Note this is not quite right in the UTC case.  I'm not sure 
-        # if this matters, but might want to clean it up.
-        return (self.jd1 - erfa.DJM0) + self.jd2
+        if self._scale == 'utc':
+            # Do the reverse of the above calculation
+            # Note this will return an incorrect value during 
+            # leap seconds, so raise an exception in that 
+            # case.
+            y, mo, d, hmsf = erfa.d2dtf('UTC',9,self.jd1,self.jd2)
+            if 60 in hmsf[...,2]:
+                raise ValueError('UTC times during a leap second cannot be represented in pulsar_mjd format')
+            j1, j2 = erfa.cal2jd(y,mo,d)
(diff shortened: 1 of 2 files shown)


luojing1211:
Hi Paul, 
I did a little test for this pull request. It seems not working for the test. Did I do anything wrong?
This is what I did. 
```
>>>from astropy.time import Time
>>>import numpy as np
>>>import matplotlib.pyplot as plt
>>>from pint import pulsar_mjd
>>># The leap second day is MJD 56109
>>>test_leap = np.linspace(56107,56110,500)
>>>test_time = Time(test_leap, format='pulsar_mjd', scale='utc')
>>>plt.plot(test_time.mjd, (test_time.tai.mjd - test_time.pulsar_mjd)*86400.0)
>>>plt.show()
```
pulsar_mjd
There is still a fractional day problem. 
I did a test on the 9 years data J1741 as well. The odd point is still there. 

Can you do a test on your machine, just make sure if it is my PINT version problem?

-Luo Jing

demorest:
The plot does not look right because when you use `test_time.pulsar_mjd` that calls the `TimePulsarMJD.value` method which was not implemented correctly yet.  I have a fix for this that I will check in shortly.  Note this does not affect the underlying time values.  When I tested last night, these changes definitely did fix the large residuals on leap second day for J1923, J1741.  Are you sure you've deleted any old TOA pickle files that may be hanging around?

luojing1211:
Sorry I forget to delete the pickle file. It runs fine. With the last commit, the simple test looks good. I think it can be merged. 
Thank you Paul. 

-Luo Jing

demorest:
Ok, great.  I'm writing up a separate issue about pickling...

luojing1211:
When this get merged. I will close the leap second issue.

Issue #181, opened 2016-10-05
reconsider TOA pickling


demorest:
I've recently been testing PINT more frequently which reminded me of something I've been meaning to bring up.  This is that the behavior of storing/re-using pickled TOAs by default seems kind of dangerous to me.  If/when PINT gets into more widespread use I think this will be the source of lots of confusion among users.  This is because the contents of the pickle file depend on many things:

1. The original TOA file, obviously.  We currently compare file modification times to see if the pickle needs updating.  However there are very likely situations where this breaks, for example replacing a .tim file with a previous version (ie `mv pulsar.tim.bak pulsar.tim`).  Or changing the contents of INCLUDE'd files.

2. The PINT code that was used to generate the pickle.  If some bug is fixed, or features added, in the PINT TOA-reading code, pickles need to be regenerated.  I'm surprised this hasn't bitten us more during testing (or maybe it has and we just didn't notice).  I am imagining lots of emails in the future like "I updated PINT, why didn't it fix this bug?"

3. The clock correction files.  "I updated my clock files, why didn't my timing results change?"

You probably get the idea.  We might be able to work around some of these by adding more checks, like maybe md5sums instead of file modification times, embedding a PINT version ID of some sort in the pickle, etc.  I think it would be better to just not do the pickling ever, and work on speeding up the TOA-processing code.  Or if any of you want to keep the functionality, I feel pretty strongly that it should _not_ be done by default; it should only be enabled when specifically asked for (presumably by someone who knows what they are doing and understands the risks).

Thoughts?

scottransom:
Yeah, I like pickling.  And, in general, I think it is and will be useful for most people and for most things.  The examples that you gave that can break things are all quite rare for most users (esp once PINT becomes stable).  And we can easily handle that by having it be the first answer in the FAQ for troubleshooting (or some such -- "Have you tried re-running after deleting your .pickle file?").  It is *hugely* helpful when you have tens of thousands of TOAs because the barycentric pos/vel calculations especially are extremely expensive.

I'm not completely against making it not the default behavior, but I think that most people will actually want to use it most of the time.

demorest:
[...]

PR #182, opened 2016-10-10
Change default pickling behavior

demorest:
This changes default TOA-reading behavior to not pickle by default, as discussed in issue #181 .  The main "top-level" usage changes are:

1. get_TOAs() function: usepickle now defaults to False.  A pickle file will only be created if usepickle=True (previously it was always created).  Otherwise this function acts the same as before.

2. TOAs class: no longer accepts usepickle argument in constructor.  To read a pickle, call with toafile argument set to the name of the pickle file (including extension).

I also did some minor reorganizing of the pickling code.  The only really notable thing is adding the _check_pickle() function which is meant to be a single place to consolidate any tests we want to do to see if a pickle needs to be regenerated.  For now it just compares file modification times same as before.

### DIFF
--- a/examples/event_optimize.py
+++ b/examples/event_optimize.py
@@ -325,8 +325,11 @@ if __name__ == '__main__':
         ts.compute_TDBs()
         ts.compute_posvels(ephem="DE421", planets=False)
         ts.pickle()
-    else:  # read the events in as a picke file
-        ts = toa.TOAs(eventfile, usepickle=True)
+    else:  # read the events in as a pickle file
+        picklefile = toa._check_pickle(eventfile)
+        if not picklefile:
+            picklefile = eventfile
+        ts = toa.TOAs(picklefile)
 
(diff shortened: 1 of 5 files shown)


demorest:
OK, I'll merge this later today.

Does anyone care about removal of the ability to read a list of multiple TOA file names (see comments in code changes here)?  I'm guessing not because a) this didn't actually work correctly before and b) I didn't see it used anywhere in tests or examples.

luojing1211:
Hi Paul, 

The reading from list was for Polycos. But the polycos will be updated. So removing it will be ok. 
-Luo Jing

demorest:
Jing, to clarify, I am talking about reading from multiple TOA files.  The ability to create a TOAs object from a list of individual TOA objects is still there.

luojing1211:
I see. Sorry about the misunderstanding.

PR #183, opened 2016-10-11
Astropy ephem

luojing1211:
Hi all, 

This is a branch that switch the spice calculation to astropy solar system calculation. I did not remove the old spice files for the purpose of testing. But in the future , PySpice can be removed from PINT. 

Please check this branch out and make tests. Let me know if there is any problems. 

Thank you.


luojing1211:
Compare to the wrong branch closing.

PR #184, opened 2016-10-11
Astropy ephem

luojing1211:
Hi all,

This is a branch that switch the spice calculation to astropy solar system calculation. I did not remove the old spice files for the purpose of testing. But in the future , PySpice can be removed from PINT.

Please check this branch out and make tests. Let me know if there is any problems.

Thank you.


demorest:
Looks mostly OK to me.  What is the difference between `topo_posvels` and `topo_posvels2`?  If it's just the final output formatting, this should have been implemented as an option rather than copy/pasting the entire function.

luojing1211:
The topo_posvel is for testing. It will disappear. Is there any reason to
keep the old format for topo_posvel?

demorest:
It doesn't appear to be used anywhere in the code so I don't see why we would need to keep it.

luojing1211:
Hi all, 
I remove the most of pySpice code. But there is one function str2ldd we are using in spice_utils.so. We have to figure out a way to make it work and simple.

demorest:
Well, the str2ldarr1() function does not actually use spice.  I think it was just put into the spice_util_py.pyx file out of convenience.  I think the easiest change would be to just delete everything in that file except for that one function, then rename the file to str2ld.pyx (or similar).  It is such a simple function that we could probably just write it in C directly, and remove the dependence on cython as well.  But I don't think that's necessary to do right now as part of this pull request.

luojing1211:
The problem is the setup. I was trying to remove the spice dependency in setup.py until I hit that problem. So this pull request's setup is not working.  If it is fast to do it, maybe just to have a simple ctype function? So we can remove spice dependency completely.

demorest:
You just need to remove spice but still build the (renamed) extension module.  I just tried this and it worked fine, I'll check it in to my repo in a sec.

demorest:
Ok, see my astropy_ephem branch.  I think you should just be able to merge this into yours and update the pull request..

luojing1211:
This should be it.

demorest:
Ok, this is merged.  Note I made one small change (changing location of str2ld module) directly in dev branch after merging.

luojing1211:
Hi Paul,

I don't know if you have the same problem. On my machine, I can not import
str2ldarr1 function. I did reinstall it, but it gives me the error:
>>> import pint
  File "<stdin>", line 1, in <module>
  File "/Users/jingluo/Research_codes/PINT/pint/__init__.py", line 12, in
<module>
    from . import utils
  File "/Users/jingluo/Research_codes/PINT/pint/utils.py", line 12, in
<module>
    from .str2ld import str2ldarr1
ImportError: cannot import name str2ldarr1

What should I do to fix it?

Thank you.
-Luo Jing

demorest:
I don't know.  I did test this before committing it, and it worked fine for me.  How are you installing PINT (ie, what is your setup.py command line)?  Can anyone else reproduce the problem?

luojing1211:
I tried 'python setup.py build_ext --inplace'. I can import str2ld module
but it can not find str2ldarr1 function.
>>> import str2ld
>>> dir(str2ld)
['__builtins__', '__doc__', '__file__', '__name__', '__package__',
'__path__']

-Jing Luo

demorest:
Hm, OK.  I'm not using the inplace option, I'm just doing a standard install (`python setup.py install --prefix=$HOME`).  I'll see if I can reproduce the problem here.

luojing1211:
Hi Paul,

I asked Kevin to install it. He got the same error.

-Luo Jing

demorest:
Ok, grab the latest and try again.  subdir name shouldn't have been the same as the module.  You might need to delete PINT/pint/str2ld/ subdir if git doesn't do it for you.

luojing1211:
Hi Paul,

It works. Python must import the folder other than .so file.
Thank you.

PR #185, opened 2016-10-12
fix d_phase_d_toa

luojing1211:
This is a fix for d_phase_d_toa based on new updates. Same change with pull request #107, but without conflicts.


luojing1211:
Hi all, 

I think we should merge this branch. I did a test, it works. I am not sure if the current code still works.

demorest:
Can you add a test to the tests/ subdir?  I'm assuming there's not one otherwise we would know for sure whether or not the current code is broken.

luojing1211:
This branch has been tested. It should be ready to merge.

PR #187, opened 2016-10-14
Bugs fix and little optimize. 

luojing1211:


demorest:
Those look fine to me.

Regarding the data file download comment, the download uses `wget -N` so the files should only be downloaded if the existing ones are out of date.  You should see messages like "Remote file no newer than local file -- - not retrieving." in this case when you run setup.py.  Is this not working for you?

demorest:
Actually, maybe it would be good to add a --no-download option or something similar.  Could be helpful when doing development work when no network connection is available, etc.

PR #188, opened 2016-10-17
Fix wls fitter

luojing1211:
Dear all, 
This is a branch for fixing wls fitter. It is still in the test phase. But I would like our developers get a chance to take a look and make comments. 

-Luo Jing


luojing1211:
Hi all, 

Now I am finalizing the test for our derivatives. I am test the derivative functions against the numerical derivatives. This is what I am doing. 
```
diff = analog_derivative- num_derivative
der_mean = (analog_derivative + num_derivative)/2.0
relative_diff = np.abs(diff/der_mean)
``` 
I would like to know how small the relative_diff meets our requirement. Generally I set the threshold 1e-3. But in some case, it is quite hard to get to that level. 

-Luo Jing

luojing1211:
Hi All, 

The fitter has been fixed. Please test it out. 

-Luo Jing

scottransom:
So I just did a super quick check of this by running examples/fit_NGC6440E.py with and without wls.

Without wls, I get this:
```
Fitting...
Best fit has reduced chi^2 of 1.04891025955
RMS in phase is 0.00204343921463
RMS in time is 33.2345023433 us

```
And with wls I get this:
```
Fitting...
d_phase_d_F0
d_phase_d_F1
d_delay_astrometry_d_RAJ
d_delay_astrometry_d_DECJ
d_delay_d_DM
Best fit has reduced chi^2 of 1.08841833625
RMS in phase is 0.00222081782284
RMS in time is 36.1193886312 us

```

Note that the chi^2 are different as well as the RMSs, and wls is worse.  Also, there seem to be some debugging statements coming out on stdout.  If I have some time this weekend I'll poke around a bit more.

I'm still concerned that there might be some sign issues in some of the derivatives.  I think that could cause the fitter to bounce around and miss the minimum of the fit.  I think we need a robust (and simple!) way of checking that for all model params.

luojing1211:
Hi Scott,
Thanks for testing this branch. Could you run more iterations on wls
fitter? Right now call_mini() function is only running one iteration.


-Jing Luo

demorest:
I was testing this yesterday and also had problems.  I don't have time to post a detailed report right now, but on the B1855 test data (nanograv 5-year), the reported uncertainties are either huge or nan.  And when I put a few-sigma offset into F0 and run a fit, the parameter value does not change back to the original post-fit.  This should not be a iteration thing, a simple linear param like that should go pretty much directly to the correct value in one fit step.

scottransom:
Running call_minimize() a couple times does improve the fit.

luojing1211:
About the test derivatives, I put those tests in each unit test file. The
analytical derivatives is compared with numerical one. The relative
difference is the quantity to give the tolerance. Due to the property of
some parameters,  the numerical steps and tolerances has to be adjusted.
This test is running slow, since it is testing all the parameters.

I agree it is not the perfect tests, so I would like to know a better way
to do it.

luojing1211:
Hi @demorest, 

I did some tests about the fitter. The miss fitting problem occurs when we fit more than one parameters together. For example, If only F0 get offset, fitting F0 only, gives me the good fit. But If I have fitting F0 and F1, it goes bad. 

Here is the test I did:
```
p = 'F0' # set perturbate parameter. 
perturb_param(p, 1e-12)
plot_res(f)
```
screen shot 2016-11-21 at 11 04 55 am

Fit F0 only
```
f.set_fitparams('F0')
f.call_minimize()
plot_res(f)
print "Fitted",  f.model.F0
print "INIT", f.model_init.F0
```
screen shot 2016-11-21 at 11 07 33 am

But when fit F1 and F0 together. 
```
f.set_fitparams('F0','F1')
f.call_minimize()
plot_res(f)
print "Fitted",  f.model.F0
print "INIT", f.model_init.F0
```
screen shot 2016-11-21 at 11 10 02 am

It is not able to have a good fit. 
Do you have any idea why this happen?

demorest:
Yes, I was just looking into this also.  Part of the problem may be that PEPOCH is very far in the past in this par file, which makes F0 and F1 highly covariant.  So there can be numerical issues.  However, tempo and tempo2 both seem to be able to deal with this parfile OK...

luojing1211:
The problem is when Fitting F0 and F1, Sigma has a very big dynamic range. This give a bad singular value decomposition. How can we reduce the d_phase_d_F1 value in this case?

-Luo Jing

luojing1211:
Hi @demorest, I changed PEPOCH to MJD 55000 in the par file, but it lost phase connection. Did you change anything else?

-Luo Jing

demorest:
Yes, you also need to change F0, ie: 

`F0_new = F0_old + F1 * (PEPOCH_new - PEPOCH_old) * 86400`

But that just works around the problem.  I think some of the matrix computations in the wls fitter should be changed to be less susceptible to numerical issues..

luojing1211:
The "bogus sci notation input" problem is due to the MJDparameter treats
the uncertainty as a MJD. I will change it right now.

-Luo Jing

demorest:
@luojing1211 please see the new_fitter branch I just pushed to my repository.  I made changes to the matrix calculations in wls_fitter.call_minimize() to make it more numerically stable, as we discussed on the phone.  The main changes were to mean-subtract and re-scale the design matrix columns before doing anything else; and compute SVD of the design matrix itself rather than of the "M^T M" product.

This version returns parameter uncertainties that are mostly within ~10% of the tempo2 values using the B1855 5-year test data set.  Except for RAJ and DECJ, where I think there is a units problem (probably arcseconds vs degrees or similar).

luojing1211:
Hi Paul, 

The changes you made do fixed the problem. But are we expecting that it needs more iterations to return the best fit? I did the same test I posted above. It took more iterations both in the signal parameter fitting and multiple parameter fitting.

luojing1211:
About the way model and toas stored in fitter class.  fitter.Model is the deep_copy of model object. But the fitter.toas object is the pointer points to the pint readed in toas object. So changing fitter.toas will change the original toas object.

demorest:
I don't think it should take very many iterations since the starting par file is already close to the right solution.  I tried iterating a few times and it doesn't really seem to be converging, the chi2 jumps around by a few between iterations.  It is also significantly higher than the tempo/tempo2 value (~1428 vs ~1391).  So something might still be not quite right.

demorest:
I found part of the problem:  In the residuals class, the mean subtraction was being done using an unweighted mean of the residuals, this results in wrong chi2 being calculated.  I've pushed a fix for this to my new_fitter branch.

When starting from the tempo/tempo2 best fit params, chi2 is still ~1 unit higher when using PINT.  And when I perturb F0 the convergence is still slow (actually not clear if it will ever reach the minimum), whereas tempo/tempo2 go straight back to the original chi2 in a single iteration.  So still could be some problems with derivatives or something.

luojing1211:
I checkedthe tempo2 general2 d_F0 and d_F1 output. When comparing the mean
value, tempo2 d_F0 and d_F1 value difference is significantly smaller than
PINT value.
Tempo2
d_F0 mean : 2147908.3950013951
d_F1 mean : 10895315.436514735

PINT
d_phase_d_F0
​mean ​
: −7625317.8
​
d
_phase_d_F1
​mean ​
: −1.9795142
​e​
15

​Could this be a problem?

​


 ​

demorest:
I'm not sure it's a meaningful comparison.  

For one, the mean value is not the right thing to compare due to likely differences in how TZRMJD is handled.  

And second, we don't know what units tempo2 is using for those derivatives.  There does not seem to be much consistency here.. for example I just took a look and tempo2's "d_F0" seems to differ from our d_phase_d_F0 by a factor of F0.  "d_F1" seems to differ from our d_phase_d_F1 by a factor of 86400**2.  I don't know that we can make much sense of this without going through the tempo2 code and checking what is done for each parameter.  I don't think we can necessarily even rely on the signs being consistent as Rutger pointed out recently during our design matrix discussion.

luojing1211:
I am trying to find out what could cause the slow converge. The F0, and F1 analytical derivatives is very close to numerical one. Is there any other way to diagnose the derivative problem? Or maybe the problem is not at derivatives.

luojing1211:
Hi Paul, 

I made the fitter converge faster by not subtracting the mean value of  design matrix. Could you make some tests on this change? I would like to know if it is safe to do so.

luojing1211:
Hi all, 

I just change the Fitter API in this branch. It is now working but I am not sure if it is the best way. I would like have it reviewed by the folks.
New updates:
* Add fitting method module : move all the different fitter class to fitting module. 
* Change fitter class to a wrapper class : Fitter class now is just an API class with the ability to select fitting method. 

Please let me know how to improve the code.

luojing1211:
Hi @demorest, 
I think I found a possible reason for slow convergence when subtracting mean value from design matrix. Correct me if I am wrong. 

Subtracting mean value from the design matrix may change the sign of derivative at some data points. This sign alternating could leads to a slower converge. From the math:

(J.T  * W * J )* dB  =  J.T * W  * y, Here J is the design matrix, W is the weight , y the residual and dB the delta parameter. 
if we subtract mean value matrix M from J, we have:

((J-M).T * W * (J-M))*dB  = (J-M).T * W *y, 
If we multiply the ((J-M).T * W * (J-M))^-1 on both side. we still have an -M term which makes dB a smaller value. 

What do you think?

paulray:
@luojing1211 This PR broke several of the command line scripts like pintempo.py and event_optimize.py.  They had lines like this:
`    f = pint.fitter.fitter(t, m)`
that did not get updated.  This is a good reason to include the command line scripts in the unit tests! I'll be trying to do that going forward.

Issue #186, opened 2016-10-13
Better observatory data structures


demorest:
For a while I have been thinking about adding a better framework for dealing with observatories (site locations, clock corrections, etc).  This issue is mainly to let you know that I've started doing some work on this, so we don't end up duplicating effort.  I think I have a pretty clear plan in mind already, but please send ideas if you have them.  The basic goal is to be able to remove all the special-case if statements (ie barycenter vs geocenter vs spacecraft vs standard tempo site) from toa.py and into a set of new data structures that can be extended to deal with any new special cases we come up with in the future.  This can also provide a way to add in the small GPS clock correction that we currently don't do (assuming this won't be done in astropy).

luojing1211:
Thanks Paul. This effort wilt change the empty observator class case.

demorest:
I am mostly done with this, but it still needs a bit more work to remove the old observatories code and clean up (fix/add tests, etc).  When this is done I'll put a pull request in.  If anyone wants to take a look ahead of time, you can see my "observatory" branch, most of the new code is under pint/observatory.

PR #189, opened 2016-10-19
Observatory classes

demorest:
These are the Observatory changes described in issue #186 and discussed recently on telecons.  I think these are ready to be merged in; all the tests pass and the basic functionality is in place.  For now, the TEMPO clock files are still read by default but the machinery for changing this is available.  I have added the GPS clock corrections, for now just by including a copy of the tempo2 `gps2utc.clk` in PINT.  In the future it might be nice to auto-dowload these.  Anyway, it would be great if some of you could check that these changes work for you.


luojing1211:
Hi Paul, 
I tried to run the test. It failed on test_model.py test. Tempo did not compute the utc2gps clock correction for J1744-1134.t1.par. If the gps clock correction is switched off, it passes the test. So this brings an question about API design. Right now we have to switch off the gps correction at the observatory level. Should there be an API to control it at toa class?

-Luo Jing

demorest:
Oops, you're right, I'll fix that test.

I thought it made sense to include GPS correction as part of the observatory-specific code because some "observatories" don't need it applied (just geocenter and barycenter for now but could be others in the future).  We can always add a high-level option to disable it if we think this is useful.

demorest:
Ok, the tests are fixed, I think this is good to go.  Any final comments from anyone else?

PR #200, opened 2016-12-08
Remove TEMPO dependency

paulray:
I converted the default clock directory from TEMPO to PINT, so it searches for time*.dat files in pint/datafiles, and I added the current set of TEMPO clock files to pint/datafiles. This should allow travis-ci to test PINT without any TEMPO dependency.


paulray:
This is pretty close to working. I think it is failing when it tries to connect to an X11 display.

paulray:
Python2 now passes with this PR! @cdeil @demorest @luojing1211 

I think we should merge this and go from there.

cdeil:
Great!

In the commits in this branch:
https://github.com/nanograv/PINT/pull/200/commits

there's this one, which in some form should already be in master:
https://github.com/nanograv/PINT/pull/200/commits/a520365f7757996370af4cee4cf78a6601643ef7

I don't know how much you care about commit history. Just thought I'd mention it, in case you want to clean this up.

paulray:
@cdeil I think I got the commit history somehow mixed up and I don't understand enough about git to fix it.  I don't think it really matters, though --  do you?

luojing1211:
I don't think this is a big deal. If we really want to fix this, does git
rebate master helps?

-Jing Luo

cdeil:
> @cdeil I think I got the commit history somehow mixed up and I don't understand enough about git to fix it. I don't think it really matters, though -- do you?

It does matter for some big projects or projects where auditing or security is a concern.

But no, I don't think it really matters here or for any of the other scientific project I contribute to or maintain.

PR #203, opened 2016-12-19
Bugfix for infinite frequencies, with test

paulray:
This fixes an issue when TOAs had frequency = 0.0, to indicate infinite frequency. When dispersion_delay was evaluated, it came out NaN instead of 0.0. I added a small test to prove that a TOA file with infinite frequency, barycentric TOAs works.

Note that this is not a great solution, since it fixes the problem only at the DM delay calculation stage. A better solution would be to have the code that reads TOAs from files convert a 0.0 frequency to inf, so that all calculations that use frequency would use inf and not the confusing 0.0.  The convention that freq=0.0 means freq=inf would only apply to reading and writing TOAs to TEMPO or Tempo2 file formats. (If PINT makes its own TOA file formats, including pickled TOAs, those should NOT use the 0.0 convention)


paulray:
I think that `TOA` objects should never have a frequency of 0.0. The frequency should always be meaningful.  I tried to implement this but it caused some other error that I couldn't track down in the time I had.

paulray:
Oops, I also realize that I messed up the units on `freq` when replacing it with infinity, so this PR should not be merged until that is fixed.

paulray:
This has been corrected. It should be ready to merge. The changes are:

- Fixed FMAX and EMAX to not use 1e100 as a synonym for infinity
- When reading TOA files, convert 0.0 frequencies to infinity
- When writing TOA files in TEMPO/Princeton or Tempo formats, convert freq=infinity to freq=0.0

Issue #207, opened 2016-12-27
Astropy 1.3 release


luojing1211:
Hi all, 
Astropy has 1.3 version released officially. Its solar system ephemeris module now returns the velocity of solar system body. We should be able to use it and remove the ephemeris download process. Since astropy will do it for us. 

-Jing

paulray:
What exactly does this mean?  With astropy 1.3, we no longer need to call jplephem ourselves, everything is done via astropy?  That is certainly handy, but does it allow PINT users to use custom ephemeris files that may not be available in astropy?  How would that be handled?

luojing1211:
We are not losing the flexibility of using different ephemeris, and astropy will handle the download work.  

Currently, we calculate positions using astropy, which has the flexibility to use different ephemeris. The only difference is the velocity. Astropy 1.2 did not provide an API for velocity, so we use jplephem to do the velocity calculation by ourselves. That requires us to download ephemeris separately. I means we have two copy of ephemeris in our compute, one is downloaded by astropy and on is downloaded by PINT. With the new update, position and velocity are integrated in astropy. So we do not have to deal with ephemeris download.

PR #209, opened 2016-12-27
Change obj velocity calculation method to astropy

luojing1211:
This is an update for solar system body's velocity calculation. Since astropy 1.3 has released, the velocity calculation is supported by astropy now. More detail see issue #207 

Note: This branch only works for astropy 1.3. Please update it.

### DIFF
--- a/pint/observatory/special_locations.py
+++ b/pint/observatory/special_locations.py
@@ -12,7 +12,7 @@ from ..solar_system_ephemerides import objPosVel_wrt_SSB
 
 class SpecialLocation(Observatory):
     """Observatory-derived class for special sites that are not really
-    observatories but sometimes are used as TOA locations (eg, solar 
+    observatories but sometimes are used as TOA locations (eg, solar
     system barycenter).  Currently the only feature of this class is
     that clock corrections are zero."""
     def clock_corrections(self, t):
@@ -22,7 +22,7 @@ class BarycenterObs(SpecialLocation):
     """Observatory-derived class for the solar system barycenter.  Time
     scale is assumed to be tdb."""
     @property
-    def timescale(self): 
+    def timescale(self):
         return 'tdb'
     @property
     def tempo_code(self):
@@ -38,7 +38,7 @@ class BarycenterObs(SpecialLocation):
 class GeocenterObs(SpecialLocation):
     """Observatory-derived class for the Earth geocenter."""
     @property
-    def timescale(self): 
+    def timescale(self):
         return 'utc'
     @property
     def earth_location(self):
--- a/pint/observatory/topo_obs.py
+++ b/pint/observatory/topo_obs.py
@@ -20,9 +20,9 @@ class TopoObs(Observatory):
     computed, observatory coordinates are specified in ITRF XYZ, etc."""
 
     def __init__(self, name, tempo_code=None, itoa_code=None, aliases=None, 
-                 itrf_xyz=None, clock_file='time.dat', clock_dir='PINT', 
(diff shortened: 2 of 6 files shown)


paulray:
I prefer not to update until MacPorts supports astropy 1.3.  As of now it is at 1.2 (https://www.macports.org/ports.php?by=name&substr=astropy).

Also, if this PR removes the need to download data files then that should be corrected at the same time.  Though, I assume it only removes the need for de*.bsp, not other files, right? I'm a bit confused about the datafiles scheme in PINT.  Initially, they are downloaded by `setup.py`, but there is also a `download_data.sh` script in the datafiles directory, however that script does not get copied to the installed directory.  If `download_data.sh` is obsolete, it should be deleted.  What do you think @demorest?   PINT should certainly include a function for the user to update the downloaded data files, but it needs to be able to update both the local copy and the installed files and I don't think `download_data.sh` does that.

luojing1211:
Hi @paulray 

This branch does not remove the download files yet. The file is going to be removed is *.bsp files, definitely not other files.  `download_data.sh` was used at earily PINT development period. Since we are downloading those file in the `setup.py`, it is no longer used.

luojing1211:
Hi @paulray and @all, 

I remove the download and check sum for *.bsp file in the setup.py. Since astropy will download and cache them internally, *.bsp file will not show up in our datafile directory. 

Also I added NASA ftp site for ephemeris link. Please check it out see if it can be improved.

paulray:
Before this can be merged, setup.py should be updated with this line (currently it says 1.2):
```
install_requires = ['astropy>=1.3'],
```

paulray:
MacPorts is now up to date with astropy 1.3, so people installing astropy with either MacPorts or pip will get 1.3.  To me, this means this PR is OK to merge.

luojing1211:
I just realized that astropy download the ephemeris '.bsp' file as a format of cache file saved in the ~/.astropy directory other than a '.bsp' file. This maybe limit us for using the ephemeris that does not come from JPL site. 

Is this matters a lot to you @jellis18 and @all. To use astropy, we do not have to worry about downloading the '.bsp' files from jpl site, but we lost some control as the price.

jellis18:
@luojing1211, ok just to be clear, it seems that `astropy` will always try to download the JPL ephemeris if it is not cached. In other words you can't point it to a `.bsp` file without it being cached or downloaded.

I really think that we want to have a decent amount of flexibility on the ephemerides especially since it appears that this could be limiting factor in GW detection and red noise modeling. There are French (http://www.imcce.fr/fr/presentation/equipes/ASD/inpop/) and Russian (I don't know the link) ephemerides that one may want to use. At least the French versions (I haven't tried the Russian version yet) are in spice format and can be read by `jplephem`. Maybe we could cache them the same way that `astropy` does so that it can be used?

Lastly, if `tempo2` has this functionality (which it does) then `PINT` should definitely have it :)

luojing1211:
Hi @all 
I hecked astropy to allow astropy read data file from a local directory. I have to change the private attributes of solar_system_ephemeris class.  I don't know if it is the right thing to do, but it seems working. 
```
coor.solar_system_ephemeris._kernel = SPK.open(path + "%s.bsp" % ephem)
coor.solar_system_ephemeris._value = datapath("%s.bsp" % ephem)
coor.solar_system_ephemeris._kernel.origin = coor.solar_system_ephemeris._value
```
What do you think?

paulray:
I actually already made the name change of objPosVel2SSB in my PR.  Hopefully they will merge smoothly.

luojing1211:
I think this is ready. We should consider get this merged.

Issue #211, opened 2016-12-30
parfile_from_line


paulray:
The routine `parfile_line_regular(self, line)` in parameter.py fails on a parfile line like:
```
F3 0 0
```
It never sets `ucty` but then tries to use it at line 338

luojing1211:
There is a fix in the wls_fitter pull request. May be I should move it to a separate pull request.  It is kind of confusing for me. Does this mean value = 0.0 and fit is False or value = 0.0 and uncertainty = 0.0

paulray:
It means the value = 0.0 and fit is False. To have an uncertainty, don't you need 3 values (value, fit_flag, uncertainty)?

luojing1211:
So there is no chance to have:( NAME value uncertainty) without fit_flag?  If that is the case, the code can be little bit more clean.

paulray:
I'm not sure if that is allowed in TEMPO or Tempo2.

paulray:
The docstring in our line parsing code should *definitely* specify what formats are allowed.

luojing1211:
I personally never see the usage like that. Why don't we set it like this way. If it happens, changing  this in .par file should be easy enough.

luojing1211:
After this fix, the (Name value uncertainty) is not working. But it is easy fix.

PR #213, opened 2016-12-31
Fix parse parameter line

luojing1211:
This is a fix for the unct did not assigned error when parsing a par file line like: F3 0 0

### DIFF
--- a/pint/models/parameter.py
+++ b/pint/models/parameter.py
@@ -308,6 +308,12 @@ class Parameter(object):
         """
         Parse a parfile line into the current state of the parameter.
         Returns True if line was successfully parsed, False otherwise.
+        Notes
+        -----
+        The accepted format:
+            NAME value
+            NAME value fit_flag
+            NAME value fit_flag uncertainty
         """
         try:
             k = line.split()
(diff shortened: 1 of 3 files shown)


paulray:
Thanks for this fix!

I just looked at this page of the Tempo2 docs: http://www.atnf.csiro.au/research/pulsar/tempo2/index.php?n=Documentation.Parameters

It gives three ways parameters can be specified:
```
In more detail, a pulsar which has a spin period of P0=1.23456 s and no fitting is required then use:
 P0 1.23456 
To request that tempo2 fits for this parameter:
 P0 1.23456 1
or to include an uncertainty on the measurement (which is ignored by the main tempo2 software)
 P0 1.23456 1 0.00003
```
Notice that it does *not* say there is a form where the uncertainty is specified without having fit_flag present.

However, in their example J0437 par file on the same page, they have this line:
```
 OMDOT           0.016                     1.000e-02 
```

How is that distinguished from:
```
OMDOT 0.016 0
```
or
```
OMDOT 0.016 1
```

I fear that it may check to see if that third value is an integer or not.  This is not a great way of doing things.  What should PINT do? @demorest @scottransom

paulray:
Just a note that this addresses Issue #211

luojing1211:
How about to have the ability to read the line like : 
```
 OMDOT           0.016                     1.000e-02 
```
And force user to have an no-integer for uncertainty. Or more specifically no "1" and "0"  in the uncertainty.

luojing1211:
Now the fit_flag is set to only accept 0 and 1, otherwise, it will be treated as uncertainty.

luojing1211:
As the discussed in the meeting. This branch is ready to merge. Does anyone have more question? I will merge it at 5:00pm CST, if no one has questions and no one merge it before that.

Issue #216, opened 2017-01-06
Single-source version in a better way


cdeil:
@paulray - In 10c1057321f8f8195e77730e10064b0003ef8928 you manually implemented a way to get the git version sha by generating a `pint/_commit.py` file from `setup.py`.

I think this didn't work properly for @luojing1211 (see comment at https://github.com/nanograv/PINT/commit/10c1057321f8f8195e77730e10064b0003ef8928#commitcomment-20331200), and it's not working properly for me either.

This is on Python 3:
```
In [2]: import pint
  File "/Users/deil/Library/Python/3.5/lib/python/site-packages/pint/_commit.py", line 1
    __commit__ = 'b'0.5.1-10-g58036db''
                      ^
SyntaxError: invalid syntax

In [3]: !cat /Users/deil/Library/Python/3.5/lib/python/site-packages/pint/_commit.py
__commit__ = 'b'0.5.1-10-g58036db''
```

I think also when you start releasing the package, by running `python setup.py sdist` and then `pip install` on the source tarball, the code you have now will not work. There will be no git repo and an error. (Didn't try this yet.)

---

Now how to change to make this to work more reliably (on Python 2/3 and from git/released package)?

Setting this up is always a pain an seems way more complex than it should be.
Possible solutions are listed here: https://packaging.python.org/single_source_version 

There's a solution for version (including git version) implemented in astropy-helpers. I'm not a huge fan because of the many files and complexity of astropy-helpers, but obviously that's an option.

I've seen other projects (e.g. Sherpa) use https://github.com/warner/python-versioneer
Apparently since recently the "Python packaging authority PyPA" put out a new package to handle this: https://pypi.python.org/pypi/setuptools_scm
I haven't looked at that yet, but possibly that could be bundled as a boilerplate file (to avoid the extra dependency) and used by PINT.

@paulray - Let me know if you want to look into that issue and make a pull request proposing a solution, or if you want to investigate / fix yourself.

PR #217, opened 2017-01-09
Use versioneer.py for versioning

cdeil:
This pull request is a suggestion how to do versioning for `pint`.

It fixes the issue mentioned in #216 and for me this works locally with Python 2 and 3:
```
python setup.py install
cd /tmp
python -c 'import pint; print(pint.__version)'
```

It uses https://github.com/warner/python-versioneer .

As you can see by looking at the python-versioneer source code or by reading [this blog post](https://blog.mozilla.org/warner/2012/01/31/version-string-management-in-python-introducing-python-versioneer/), implementing this by hand and getting it working properly is **very** hard and brittle.
So even if I don't like boilerplate files like `versioneer.py` and `pint/_version.py` added to a repo, for this I think it's better than trying to roll your own solution.

Further work will be needed e.g. on `MANIFEST.in` to make `python setup.py sdist` and pint releases work. But that's for future follow-up PRs.

@paulray and all - Thoughts?


luojing1211:
Is it possible to put pint/_version into pint/external directory? We would like to keep the non-pint core code into that directory.

cdeil:
> Is it possible to put pint/_version into pint/external directory?

Yes. I can make the change tonight.

Any other comments?

cdeil:
>> Is it possible to put pint/_version into pint/external directory?
> Yes. I can make the change tonight.

Although I have to say: all packages using versioneer I've looked at had it where it's now, and Astropy and the affiliated packages also have the `_version.py` file in the top-level package folder.
It's "hidden" or "private" by convention via the underscore.

So my suggestion would be to leave where it's not. But your suggestion also makes sense. Just let me know what you decide.

luojing1211:
@demorest What do you think about this?

demorest:
I think it should go into extern/ since holding code like this is the entire point of that subdir.  In general I think PINT is getting large enough that we should be trying to "un-clutter" the top-level dir as much as possible.

paulray:
I thought that `versioneer.py` was what we wanted to move to extern, not `pint/_version.py`

luojing1211:
@paulray  Maybe I misunderstand this during the meeting. @demorest could you verify this?

cdeil:
I've moved `pint/_version.py` to `pint/extern/_version.py` in e3fe8522cbb56955705981acc814ecbc7f8d489d .

I don't think moving `versioneer.py` is advisable:
https://github.com/warner/python-versioneer#theory-of-operation

> To allow setup.py to compute a version too, a versioneer.py is added to the top level of your source tree, next to setup.py and the setup.cfg that configures it.

It's a file outside the package, that doesn't get installed (like setup.py or setup.cfg).

luojing1211:
Is this PR looking OK? @paulray, @demorest and @scottransom

paulray:
I think the idea of extern was that if we had to include code in PINT that was not written by us we would put it there.  This would both show that we (the PINT authors) weren't the ones who wrote it, and would provide some encapsulation if that code had a different license that PINT (once PINT gets a license, that is).  For this PR, I'm not sure what the right answer is for where the two files should reside.  It kind of seems silly to move _version.py to extern if versioneer.py is in the top level of our distro. That hasn't accomplished either of the above goals.  So, I guess in this case my vote would be to return _version.py to its natural location and leave it at that.  But @demorest may feel differently.

cdeil:
It doesn't matter so much where `_version.py` is located, Python 3 support and working versioning are more important, no?

It's also important to do more testing of this versioning scheme, from Github tarballs, tagged versions and sdist releases, that might reveal other things that you want to change about the versioning setup.

Let me know if there's something else I should do here, like remove the last commit.

paulray:
I'm fine with merging this now and then fixing or re-arranging things later as any issues come up.  With this, is the version accessible as `pint.__version__` ?

cdeil:
> With this, is the version accessible as pint.__version__ ?

Yes, that should always work.

If you find something not working or not what you expect, let me know here or in a new issue.

paulray:
@cdeil I was thinking of updating the PINT version number because of all the recent changes.  I think this Wiki page is now out of date: https://github.com/nanograv/PINT/wiki/Tagging-PINT-Versions

What is the process now?  Do we not touch _version.py and just make a tag?

paulray:
@cdeil Never mind the previous comment, I think I figured it out and updated the wiki page. I tested by tagging 0.5.2 and it seemed to work fine.

Issue #204, opened 2016-12-20
format_toa_line is buggy


paulray:
The `format_toa_line()` function in `toa.py` is buggy.

- It assumes that `obs` is a single character which may not be correct. It should convert the observatory specified to the 1-character alias for TEMPO format.  For Tempo2 format, it may need a different behavior, particularly for observatory special cases. Also, it should raise an exception for observatories that can't be represented in a particular TOA file format. 
- It assumes that `toaerr`, `freq`, and `dm` are all in the correct units. It should do correct unit conversions to put values into the units assumed by the particular TOA file formats.

paulray:
This is mostly fixed, but writing Tempo2 TOA files with TOAs that are from the barycenter or geocenter does not write the correct observatory code.  For observatory barycenter, the code should be '@' or 'bat'.  For the geocenter, it should be 'coe'.

paulray:
OK, fixed that too.

PR #219, opened 2017-01-23
Move scripts to scripts directory

paulray:
Added new script `pintbary.py` that allows simple barycenter time corrections from the command line, like `bary` from PRESTO.

Moved other command line scripts to PINT/scripts and made setup.py install the ones that are meant for end-user use (`pintempo.py`, `zima.py`, `pintbary.py`, `fermiphaseogram.py`)

Fixed a bug in `get_mjds()`, which was accessing a `t.value` when it meant `t.mjd`. To get an MJD float from a Time.

Fixed bugs in `fermiphaseogram.py`

Added 1-letter tempo_code properties to Barycenter and Geocenter special locations

Improved writing out TOA files, including better use of Quantities with units


paulray:
This PR also fixes issue #204

cdeil:
Maybe add tests for the command line scripts on travis-ci?
I just mean adding one line executing `some_pint_tool --help` for now at the end of `.travis.yml`.

Of course the usefulness of this is limited, but I think it's still useful, because it can catch regressions such as when files or functions / classes are changed elsewhere and imports in the scripts need to be updated. Or later when we add a Python 3 build, it will check some extra things where Python 3 is more strict (e.g. no mixed tabs / spaces) or different (e.g. no print statement).

paulray:
@cdeil I tried adding lines to `.travis.yml` to test out the command line scripts, but it can't find them in the path, even though they should be installed.  Any suggestion on how to get them to run?  Should I just run them with `python scripts/pintempo.py`?

paulray:
Nope, running `python scripts/pintempo.py` does let the script start running in travis, but the imports fail.  Not sure how to make this work.  This [article](http://stackoverflow.com/questions/16916880/how-to-test-command-line-scripts-with-nose) says I should re-write them to encapsulate them in functions and test those in python, not from the command line in travis.

cdeil:
I see the fail here:
https://travis-ci.org/nanograv/PINT/jobs/208402860#L236

By default Python will always try to import modules / packages from the current working directory first, i.e. in this case the source folder.

The problem is (I think) that on travis-ci we never execute a "build" or "install" command, so the C extensions aren't built.

It's a bit weird that the import fails with `pint.tao` and not some C extension module. I don't understand that part and would have to investigate why the ImportError isn't something different mentioning a C extension.

To fix you could try this:
```
pip install .
cd scripts
python pintempo.py
# runs the version from site-packages
```
or this:
```
python setup.py build_ext --inplace
python scripts/pintempo.py
# runs the version from the source folder
```

---

There might be better ways to handle scripts.

Here's one option:
* Put the scripts folder inside the pint package
* Add a `pint/scripts/__init__.py` so that it's part of the package
* Use `python -m pint.scripts.pintempo` to execute

Several other Python packages work that way, it's considered a best practice, because it can avoid mismatches between the Python version or package version used and the command line script:
* `python -m pip`
* `python -m http.server` (on Python 3)
* `python -m pytest`

---

Another option could be to add the command line interface into `pint/__main__.py` and to use sub-commands with ArgumentParser. Then the scripts could be in `pint/scripts` and the main would just dispatch.

So users would call:
```
python -m pint pintempo
python -m pint pintbary
...
```

Personally, I like that option best and plan to change Gammapy to work that way (instead of having ~ 10 command line tools installed and on the user PATH like we have now).

---

In Astropy and affiliated packages, and many other packages, setuptools `console_scripts` entry points are used:
http://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point

A small advantage there is that you can test the scripts without "shelling out", as part of your Python test runner. A possible disadvantage is that you always have to install pint before you can use the command line tools, installing generates a wrapper script for you.

---

Personally, I've seen many users confused about which script / package version they are running, and thus like the methods where `python -m` is used, and there never is a separate command line script outside the package. If `python -m pint` is considered too cryptic for users, then installing a single `pint` script that has subcommands would be my personal second favourite.

But in the end: I think all of these options are pretty good, and you can get any of them to work without much effort, so just pick the one you like best.

If you have any questions, I'd be happy to help.

matteobachetti:
@cdeil @paulray I'm using the "encapsulation" à la Astropy, that basically boils down to defining a `main` function in the library source, and using that function as an `entry_point` in `setup.py`. I think it is very convenient. To test the functions, I just call these `main` functions, whether they are installed or not. 

E.g. this main function https://github.com/matteobachetti/MaLTPyNT/blob/master/maltpynt/fspec.py#L1006 gives birth to the MPfspec here https://github.com/matteobachetti/MaLTPyNT/blob/master/setup.cfg#L37 (config file that is loaded by setup.py later) and it is tested here https://github.com/matteobachetti/MaLTPyNT/blob/master/maltpynt/tests/test_fullrun.py#L247

paulray:
Thanks for the helpful discussion!  I'm not sure exactly which solution to adopt, but certainly the first thing to do is to rewrite the scripts to have a main() function, so I'll do that now.  That will actually make @scottransom happy since he never liked that I changed event_optimize.py to have command-line arguments as he prefers to have a script that he edits and runs.  With this method, he can do this (example abbreviated) and edit and run that script to his hearts content, while other users can just call it from the command line.:
```
# run_event_optimize.py
import event_optimize
eventfile = 'xxx.fits'
parfile = 'psr.par'
event_optimize.main([eventfile, parfile])
```

Issue #78, opened 2015-12-30
par files output by model.as_parfile() should work with TEMPO and Tempo2


paulray:
Currently, there are some issues with the par files output by model.as_parfile() that should be dealt with.
The first is there is no "UNITS" line, so Tempo2 assumes it to be TCB, which is wrong.  Since PINT only supports TDB so far, this is an easy fix, and I added "UNITS TDB" to the output produced by as_parfile().

In addition, the binary model should output the "BINARY BT" line.  This has to be implemented differently since it breaks the 1 par file line per parameter paradigm that as_parfile() assumes.

There may be some other control-type parameters that should be included in the output as well (CLK, MODE, etc...).  I haven't thought this all through yet, but I wanted to capture it as an issue.

scottransom:
Should be smartly sorted as well!

paulray:
I just found, thanks to Pablo Saz Parkinson, that when the binary model is written to the par file, it writes the OM line in DD:MM:SS.s format, not decimal degrees. This is not compatible with Tempo2, I believe.  This needs fixing.

luojing1211:
Hi Paul, 

Which branch are you in? I did not find this problem. 
![screen shot 2016-07-29 at 10 44 01 am](https://cloud.githubusercontent.com/assets/5324931/17254260/aa99db7e-5579-11e6-8e54-f4b391b50c03.png)

paulray:
Interesting. I didn't try it myself, I just saw the par file generated by Pablo. Possibly he is just running an older version of PINT and it is now fixed.

paulray:
@kerrm just found that this issue was causing him pain.  Its priority should probably be elevated.

paulray:
PR #236 should fix at least the writing of the binary model name.

A more complete solution seems like it would require more code, perhaps by making the parameter names somehow be grouped by what model they apply to.  This could, for example, be by adding the name of the TimingModel subclass to the Parameter.  Then, `model.F0.model_class` could be `Spindown` or something like that. And, the as_parfile() method could print all the Spindown parameters together, then all the BinaryBT parameters together, etc...

paulray:
@kerrm just pointed out that the error on RAJ and DECJ is supposed to be in arcseconds. This is how it is interpreted by TEMPO, Tempo2, and even PINT. But, when writing with `as_parfile()`,  PINT outputs the errors in degrees. This is a bug.

luojing1211:
[...]

Issue #218, opened 2017-01-09
Make get_componets in model_builder work on Python 3


cdeil:
Currently `import pint.models` doesn't work with Python 3:
```
$ python3 -c 'import pint.models'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/deil/Library/Python/3.5/lib/python/site-packages/pint-0.5.1+11.g5028af7.dirty-py3.5-macosx-10.12-x86_64.egg/pint/models/__init__.py", line 12, in <module>
    from .model_builder import get_model
  File "/Users/deil/Library/Python/3.5/lib/python/site-packages/pint-0.5.1+11.g5028af7.dirty-py3.5-macosx-10.12-x86_64.egg/pint/models/model_builder.py", line 37, in <module>
    ComponentsList = get_componets()
  File "/Users/deil/Library/Python/3.5/lib/python/site-packages/pint-0.5.1+11.g5028af7.dirty-py3.5-macosx-10.12-x86_64.egg/pint/models/model_builder.py", line 26, in get_componets
    exec('import %s as tmp' % mod)
  File "<string>", line 1, in <module>
ImportError: No module named 'astrometry'
```

As the traceback shows, the issue is that the `get_components` function in `pint/models/model_builder.py` doesn't work in Python 3.

I could try to change that code so that it simultaneously works in Python 2 and 3.
But @luojing1211 , since you've written that code (see https://github.com/nanograv/PINT/commit/bf6df0c1c2f041e747557d274a57eb24d407c954), maybe you could do it?

Usually calling `exec` is not a good way.
There are better Python coding patterns to have a model registry or dynamic class generation or ...
I haven't tried to understand what that code is doing yet, and there's no comment about that.

@luojing1211 or @paulray  - Thoughts?

luojing1211:
Hi @cdeil,

This part is to search all the files in the 'pint' folder and find all the existing timing model classes. One way I know of is using meta class. Is it what's in your mind?

-Jing Luo

cdeil:
@luojing1211 - So the purpose of this code is to have a registry?

Would you be OK to have to manually call "register" or add them to a global list of dict?

Or do you want auto-registry of sub-classes?
Then yes, I think using a metaclass is the common pattern.
But I'm not very familar with metaclasses or possible registry patterns, so I've put it as a question here: https://github.com/astropy/regions/issues/18#issuecomment-271410958
[...]

PR #220, opened 2017-01-23
Meta class timing model

luojing1211:
This is a metaclass TimingModel update for solving the problem mentioned at issue #218

### DIFF
--- a/pint/models/timing_model.py
+++ b/pint/models/timing_model.py
@@ -100,15 +100,15 @@ class Cache(object):
 class ModelMeta(abc.ABCMeta):
     """
     This is a Meta class for timing model registeration. In order ot get a
-    timing model registered, a member called 'name' has to be provided in a
+    timing model registered, a member called 'register' has to be set true in the
     TimingModel subclass.
     """
     def __init__(cls, name, bases, dct):
         regname = '_model_list'
         if not hasattr(cls,regname):
             setattr(cls,regname,{})
-        if 'is_register' in dct:
-            if cls.is_register:
+        if 'register' in dct:
+            if cls.register:
                 getattr(cls,regname)[name] = cls
         super(ModelMeta, cls).__init__(name, bases, dct)
 
--- a/pint/models/timing_model.py
+++ b/pint/models/timing_model.py
@@ -8,6 +8,7 @@ import astropy.time as time
 import numpy as np
 import pint.utils as utils
 import astropy.units as u
+from astropy.table import Table
 import copy
 import abc
 
@@ -192,6 +193,8 @@ class TimingModel(object):
 
         if binary_param is True:
             self.binary_params +=[param.name,]
+        if hasattr(self, '_param_table'):
+            param_row = [param.name, ]
 
     def remove_param(self, param):
         delattr(self, param)
@@ -247,6 +250,14 @@ class TimingModel(object):
             if par.is_prefix == True and par.prefix == prefix:
                 mapping[par.index] = parname
         return mapping
+
+    def match_perfix(self, param):
+        par = getattr(self, param)
+        if not hasattr(par, 'prefix'):
+            return ''
+        prefix = par.prefix
+        
+
 
     def match_param_aliases(self, alias):
         p_aliases = {}
@@ -719,6 +730,11 @@ def generate_timing_model(name, components, attributes={}):
         numComp = len(components)
     except:
         components = (components,)
+
+    param = []
+    comps = []
+    param_type = []
+
     for c in components:
         try:
             if not issubclass(c,TimingModel):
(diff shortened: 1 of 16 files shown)


luojing1211:
All TimingModel classes are added to __init__.py 
The class member 'name' is changed to 'is_regisiter'. If 'is_register' true, the class will be registered to the model list. The name will be parsed to the list automatically.

pint/models/timing_model.py
```diff
@@ -96,6 +97,21 @@ def use_cached_results(*args, **kwargs):
         return use_cached_results
 
 
+class ModelMeta(abc.ABCMeta):
+    """
+    This is a Meta class for timing model registeration. In order ot get a
+    timing model registered, a member called 'name' has to be provided in a
```
cdeil: The docstring here still says "name", but it's now "is_register" I think.

"is_register=True" sounds a little off to me, because it's a mix between "is_registered" and "register me". Maybe call it "register"?

luojing1211:
Quick question:
    How do we swith on Python3 test for travis?

cdeil:
> How do we swith on Python3 test for travis?

We have to do a significant extension of `.travis.yml` to specify multiple builds.
There's different ways to specify the build matrix.
Here's a simple example: https://github.com/rootpy/rootpy/blob/master/.travis.yml
Here's a more complicated example: https://github.com/gammapy/gammapy/blob/master/.travis.yml

It usually takes a few hours and several commits to get `.travis.yml` right, especially if you haven't done it before.  So I don't think you should attempt it in this PR. I can make a new PR extending `.travis.yml` in the coming days.

jellis18:
I've started testing this with `python3`. There are lots of little fixes that are needed but as @cdeil says, the whole point was to make `PINT` `python3` compatible so I think this may be the place. although maybe we should update the title of the PR.

luojing1211:
I think we should merge this branch, since this is on optimization on the PINT structure. We should make a new pull request for fixing compatibility of Python 3.x

jellis18:
@luojing1211, ok but what is this PR supposed to fix then if not the Python 3 compatibility?

luojing1211:
This is to change the way of timing model selection. It is better than the
current method.

luojing1211:
Hi @ALL, 

I added some new features for this branch:
* All timing model components has metaclass for registering the components. 
* Add very basic components information class and order structure. 
    PR #170 seems over complicated. 
* Modified par file print out function. It handles basic parameters grouping and sorting. 
   This feature will fix issue #78   
Problem:
* No unittest for modified as_parfile() function . 
If there is no good way to test it yet. I would recommend this branch get merged.

paulray:
As a unit test for as_parfile(), how about writing a moderately complicated model constructed programmatically (not from a par file), writing it out as a par file, then reading it back in, and verify that the phases computed (for a few random TOAs) are the same between the original model and the one read in.

luojing1211:
This is a good idea. Besides, I would like to add a convenient function for getting the information  from the toas class. Which module do you think this function should be in? Maybe model builder is a good place?

paulray:
@jellis18 I see that the coverage/coveralls check is failing.  Is it simply a test to see if the coverage decreases and that constitutes a fail, or is it more subtle?

paulray:
> Besides, I would like to add a convenient function for getting the information from the toas class. Which module do you think this function should be in? Maybe model builder is a good place?

I don't understand what you are asking here. Can you clarify @luojing1211 ?

luojing1211:
Sorry for the confusion. I was talking about the convenient function for
getting the "EPHEM", "CLK", information from a parfile.

jellis18:
@paulray, I think that is it. There are setting to set the threshold for how much coverage can decrease before failure although I can't find what the defaults are (maybe it is 0.5%). I changed it to 10% and am re-running travis to see if it passes.

I guess it is up to us to set this number though. It may be a good idea to make this stringent so that it requires that any new features have tests. This also brings me back to the protected branch idea. Right now we could merge this even though the tests failed. If we make `PINT/master` a protected branch then the tests have to pass before we can merge.

paulray:
@jellis18 thanks for the explanation! For at least a little longer, while PINT has such a small set of developers and is under such active development, I don't think I'd make a protected branch (especially if a 0.5% drop in coverage causes a fail). But, putting in a milestone to do that at some point seems like a good idea. Having travis has been great in keeping us from merging things that break the tests. We definitely need to focus more on building good tests as well and the coverage stuff does help encourage that.

On the other hand, the code quality checker has some insanely strict rules, like 80 column lines and too many local variables. I hope it has some way to tune what it reports. For example, flags me every time I use log.info("message {0}".format(val)), which I use ALL the time.  I don't agree that it is poor style, and I don't really understand how to write it in a different way.

paulray:
Is it OK to merge this? I'd like to get at least the parfile writing changes.

luojing1211:
Not yet. This afternoon should be ok.

luojing1211:
If the check is successful, this branch should be ok to merge.

luojing1211:
Hi @paulray , do you want to double check this branch and merge this one?

Issue #193, opened 2016-12-04
TT(BIPM) 


luojing1211:
Currently we are using astropy time object to handle the TT~UTC converting, which takes TT(TAI) as default. But in our dataset, a more accurate realization TT(BIPM) is used. I did not see any option to change astropy default to TT(BIPM). Does anyone know if it exists in astropy? If not maybe we should get contact with astropy.

paulray:
This is done in PR #223, so this probably can be closed.

I have some issues with the TT(BIPM) correction being done in TopoObs, which feels like the wrong place, as I described in this [wiki page](https://github.com/nanograv/PINT/wiki/Clock-Corrections-and-Timescales-in-PINT).  But, that can probably be opened as an issue later if we decide to move where the correction is done.

Issue #102, opened 2016-02-08
Improve clock correction process and specify whether to use PINT, TEMPO or Tempo2 files


paulray:
For clock corrections, PINT should have the option of reading Tempo2-style clock corrections from $TEMPO2/clock.  In Tempo2, the clock correction chain goes all the way from observatory time to some (specified) realization of TT.  In TEMPO, if I understand correctly, the clock corrections only go from observatory time to GPS (which is assumed to be close enough to UTC for most work, <10 ns).  So, if PINT is given an option to use the Tempo2 clock files, it should probably just use obs2gps.clk and gps2utc.clk, and not go all the way to TT.  In PINT the conversions from UTC to TT and TDB are handled differently (using astropy.time).

paulray:
Telecon 2016 Feb 22:
* Should really make a more structured, sensible framework for clock corrections
* Should handle the case where an observatory has multiple backends, each of which require different clock corrections (e.g. Nancay)
* Should be able to tell PINT if you want to use TEMPO, Tempo2, or PINT-internal clock correction file
* Need to have different ways of updating PINT-internal clock correction files, if we implement them.  Could be kept in sync with TEMPO, or Tempo2, or whatever.

luojing1211:
Hi @paulray,
Does the current new observatory structure fix this problem? PR #189

paulray:
Renamed this since the original issue is solved but the general clock correction issues are still there. @aarchiba was just discussing this on Slack, in fact.

paulray:
Related to #915

aarchiba:
Can we consider this solved now?

abhisrkckl:
Can this be closed?

Issue #37, opened 2015-07-02
TDB vs TCB


scottransom:
We should be able to handle both barycentric timescales.  This is important especially since TEMPO2 used TCB by default.

paulray:
This requires some design decisions.  Will PINT always use TDB internally, and just have timing model files converted upon reading in?  Currently the TOA tables are all written assuming that TDB is the basic time scale.  Or, do we change PINT to use "TB", which can be either TDB or TCB depending on some setting?  One issue is that this choice is kind of a global thing and might want to be known before a model is defined, so perhaps the UNITS keyword in the par file isn't the best place to make this setting. But, of course, any par file should specify UNITS so PINT knows what is being read.

paulray:
Regarding this, as of 2006, TDB is actually defined in terms of TCB, so I guess TCB is really the more fundamental timescale. They encourage use of TCB for solar system ephemerides, though I don't think JPL has taken them up on that so far.  Here is the resolution (2-page PDF):
https://www.iau.org/static/resolutions/IAU2006_Resol3.pdf

abhisrkckl:
Is there any interest at present in supporting TCB in PINT?

scottransom:
@paulray How much work is it worth for that capability?  Especially given that T2 will convert for you in a single-line command?

paulray:
Good question. It depends on how likely JPL is to change to TCB for ephemeris time in the future, I guess.
I'm not sure I trust the conversion to be exact without a re-fit, but maybe it is.

abhisrkckl:
I tried the TCB -> TDB conversion with a few InPTA par files. Without a re-fit, there was a ~10 ns difference in residuals for some par files. I don't know if that is cause for worry or not.

gooddc:
I hesitate to say this out loud, but I should probably volunteer for the TCB to TDB wrapper task.

scottransom:
No good deed goes unpunished, Deborah!  :-)  But thank you for even considering!

gooddc:
I'm pretty sure I can do it; it just probably won't be fast.

abhisrkckl:
Are you planning to develop a TCB -> TDB converter like the transform plugin or a method to enable PINT to read TCB files? (There is probably a lot of overlap between these, I guess.)

scottransom:
Given that I think we decided that we don't want the PINT internals to handle TCB directly, I think both of those two different solutions are effectively almost identical (which you also suggest).

[...]

PR #223, opened 2017-02-15
BIPM clock corrections 

jellis18:
Here is a first go at hacking in the BIPM clock corrections as a way of solving Issue #193. I just added them in exactly the same way that the gps clock corrections were coded. At the moment this does not work as the residuals produced are on the order of 1000s of microseconds. I am using the `tempo2` clock file `tai2tt_bipm2015.clk`, which I suspect is probably not the proper file.

I'm just making this PR so that we can try to work this out.


jellis18:
OK, I just made some corrections to the clock file to subtract off the 32.184 to deal with the TT -> TAI correction part. I'm assuming that is already done in the clock corrections done by `astropy`. 

Below is a figure showing the difference in residuals between `tempo2` and `PINT` when using BIPM clock corrections. The difference is on the order of 10 ns and appears to be pretty white (i.e no obvious structure) so things may be working although I'd like to find a better way of testing to be sure.

tempo2-pint_bipm

luojing1211:
Hi @jellis18  Is it better to subtract 32.184 seconds in the code other than rewrite a new clock correction file?

jellis18:
OK @luojing1211, I added the offset to the code. I think this is probably better since now we are using the exact same file as `tempo2`.

@all, what other tests do we need to do?

luojing1211:
Hi @jellis18 , 
I had the test with this branch with B1855+09 9 years data. It looks working working. It would be nice if we can switch this flag on or off just using par file CLK TT(TAI). Since it is default on now, it could cause the failure on the tests. I know to solve this would need to have a 'configure' class. We have this discussion in the future. 

Another potential problem is the precision of clock correction. This is actually to @demorest or @scottransom.   Our time precision is using float128, but when we pass the time into interpolation function, it is using float64. Do we need to worry about that?

jellis18:
Hi @luojing1211,

That is a good point about the tests and indeed it does fail now; however travis ci did not catch it as you can see [here](https://travis-ci.org/nanograv/PINT/builds/201973618). Maybe we should come up with a better testing scheme...

I'll look in to ways to turn off BIPM if not in the par file.

demorest:
Hi guys, about the precision question I don't think there would be any issues.  Using double-precision times to for interpolating the clock values, as well as for the corrections themselves is not a problem.  I think the only place we could possibly get into trouble is when applying the corrections.  However (in toa.py) it looks like correction values are first converted to TimeDelta objects before being added to the TOAs so I think this is all OK.

jellis18:
OK, we are somewhat confident about the precision. Does anyone have any ideas how we should deal with having a check for CLK TT(TAI) or BIPM to determine whether or not to do the BIPM corrections? Maybe we could have a check in `preprocess_parfile` but I am not very familiar with the overall architecture.

jellis18:
As I discussed this issue with @luojing1211 and @scottransom recently, we think it may be best to allow the user to specify if they want GPS or BIPM corrections turned *off* when making TOAs. The default behavior should be to have them on but the user should have the option nonetheless.

I've started making these changes and I've hit a roadblock (probably because I'm not very familiar with meta classes or registries). It seems like the desired way to initialize and access an `Observatory` instance is through the `get` method via
```    
    @classmethod
    def get(cls,name):
        """Returns the Observatory instance for the specified name/alias.
        If the name has not been defined, an error will be raised.  Aside
        from the initial observatory definitions, this is in general the 
        only way Observatory objects should be accessed.  Name-matching
        is case-insensitive."""
```

However, it seems to me that there is no way for the user to specify any of the keyword arguments in the `TopoObs` class via this method.

This could be an easy fix but I'd appreciate suggestions as to how to set keyword arguments when initializing an `Observatory` instance.

luojing1211:
That is a valid point @jellis18. Those flags are carried on when the observatories class instance get initialized in the pint/observatory/observatories.py. 

May be a function that changes the observatory attributes in the get() function and make get() take **kwarg? 
@demorest what do you think?

jellis18:
Does anyone have any suggestions on this? It looks like it could be a major overhaul since the observatories are registered in `observatories.py` and the options to `TopoObs` are not accessible anywhere but here...

luojing1211:
@jellis18. The possible access points in toa.py are
1.
```
TOA.__init() line 305
 site = Observatory.get(obs)
```
After get the site instance, we can have a line that switches the flag on or off

2. At 
```
TOAs.apply_clock_correction()
```
we can add the flags into this method, change site flag in front of line 578
```
gcorr = site.clock_corrections(time.Time(grp['mjd']))
``` 

May be there a some other place like before read in TOAs, we change all the site settings. Is there any other way to make this more elegant? @scottransom @demorest

scottransom:
I'm not sure how "elegant" this needs to be....  All we want are options to turn on/off certain corrections or ephemerides when we grab TOAs, either using the "helper" function get_TOAs(), which already has options, or via the TOAs class itself.  And those options would simply do exactly what you suggest, just tun on and off the application of various things.  Am I missing something?

luojing1211:
I can not think of any other good way to do this without a deep level code reorganise.

jellis18:
OK, I'll try to put something together a little later today.

paulray:
I think the clock corrections could use an improvement, as mentioned in #102 and PINT still does not support TCB as mentioned in #37. So patching a little thing on to get BIPM working is fine for now, but a more extensive rewrite with a more general/elegant solution is still worth considering.

scottransom:
I think my only point was that this doesn't need nor warrant (I don't think) some kind of specialized beautiful and elegant Pythonic treatment.  All we want are switch options, where the branches in the switch have a normal default value, but the user can override that.  I don't think that anything precludes that at all.

jellis18:
I have added a `get_observatory` convenience function that can turn the gps or bipm clock corrections off. I have replace all instances of `Observatory.get` with `get_observatory` where the default behavior is to use both gps and bipm clock corrections. The functions `get_TOAs` and `get_TOAs_list` now have options to turn off the gps or bipm clock corrections. This functionality is added to the `apply_clock_corrections` method and carried through the `get_observatory` function.

I think I have updated all of the checks to turn off the bipm corrections so that they still work. Let me know what you think. If we agree that this is the way to go then I can write a few tests.

pint/observatory/topo_obs.py
```diff
@@ -122,6 +127,16 @@ def gps_fullpath(self):
         return os.path.join(os.getenv('TEMPO2'),'clock',fname)
 
     @property
+    def bipm_fullpath(self):
+        """Returns full path to the TAI TT(BIPM) clock file.  Will first try PINT
+        data dirs, then fall back on $TEMPO2/clock."""
+        fname = 'tai2tt_bipm2015.clk'
```
luojing1211: @jellis18, It this hard coding file name ok? Is there anyway to tell which BIMP file to use?

jellis18: I'm not sure, there is probably a better way to be doing this. I just copied the way that the gps corrections were handled.

jellis18: If we could have the files updated and cached when run that would be better but I'm not sure how the clock correction files are made and if they are stored anywhere in the correct format.

luojing1211: We get most of our clock_correction from TEMPO or Tempo2 by copying them into the datafile directory. We have not created those files yet. I don't know if we are going to in the future.

jellis18: In that case I would think hardcoding them may be ok for now.

luojing1211:
As discussed in the telecon, this should get merged. So merging.

Issue #224, opened 2017-02-17
Tests fail but travis still passes


jellis18:
As I was working on the BIPM clock corrections in PR #223 I noticed that many tests failed (as expected since we are changing clock corrections) but `travis` still [passes](https://travis-ci.org/nanograv/PINT).

I'm not exactly sure why this happens but perhaps a better way to do tests is to use `nosetests` or `pytest` directly with `setuptools` so that one can just use `python setup.py test` instead of the shell script used now.

I see a similar issue mentioned in Issue #195, maybe it is time to re-visit this...

luojing1211:
Hi @jellis18 , If you go to travis-ci details, a lot of ERROR and FAILs are showed in the run-test section. I think the problem is the run-test script does not report the failure as an exit 1 status.

paulray:
Strange. The shell script used to properly exit with nonzero status [when the tests failed](https://travis-ci.org/nanograv/PINT/builds/187302140).  I don't know why this would change.

However, changing travis to run nosetests directly instead of via the sh script, would be fine.

Eventually, we'd like to switch to pytest, but it seems to be in a chaotic state where no one can even agree on what the command to run tests is called.  Also, it never really worked for me.

PRs that improve the testing framework are most welcome!

jellis18:
OK,  I have a branch with some new tests but now I'm getting [this](https://travis-ci.org/jellis18/PINT/builds/202834111) problem even on the master branch.

Somehow `t.utc.iso` is giving `2006-01-15 21:24:37.834136` instead of the expected `2006-01-15 21:24:37.834078`. I get the same error on my computer and on `travis`. Although if I just run the commands from the test separately then it gives the correct result. 

Has anyone seen this before?

luojing1211:
> Somehow t.ut1.iso is giving 2006-01-15 21:24:37.834136 instead of the expected 2006-01-15 21:24:37.834078. I get the same error on my computer and on travis. Although if I just run the commands from the test separately then it gives the correct result.

Can I see your test? I need to see the way you create the time object.

jellis18:
@luojing1211, its not a new test. It is in `test_astropy_times.py`. 

```
self.lat = 19.48125
self.lon = -155.933222
earthloc = astropy.coordinates.EarthLocation.from_geodetic(self.lon,self.lat,0.0)
self.t = astropy.time.Time('2006-01-15 21:24:37.5', \
                           format='iso', scale='utc', \
                           location=earthloc, precision=6)
```

[...]

PR #225, opened 2017-02-21
Different way of doing tests and new travis ci with python3 tests

jellis18:
Here are some changes that should make the ci tests work more smoothly. It appears that when running the shell script for tests, it would only fail if the tests failed in the setup stage and not in the test part of the unit tests as unit test is not meant to fail there but simply report that the test failed. They are as follows

1. Use `nosetests` directly by adding a few lines to `setup.cfg`. Now to run tests one can do `python setup.py nosetests`. 
2. Change `.travis.yml` to work with python version 2.7 - 3.6 but allow failures for all python 3 versions until problems related to PR #220 are worked out.
3. Change `.travis.yml` to use requirements files and pip instead of `Anaconda`. This is actually useful for developers too.
4. Don't test `test_astropy_times.py` to get around problems mentioned in #224.
5. Use `unittest` wherever possible. 

With this changes `travis` will fail when it is supposed to and we can switch off the `allow_failures` line once we get python 3 working. I think we should try to merge this ASAP to help out with other PRs so that we know if our tests are actually passing or not.


matteobachetti:
Hi, one small and easy fix that allows to at least import toa in Py 3: use 
`from builtins import super`
and substitute all occurrences of `super(class).__new__(something)` with `super().__new__()`

jellis18:
Another thing that we may want to do along the lines of tests is to make PINT/master a [protected branch](https://help.github.com/articles/defining-the-mergeability-of-pull-requests/) and have a requirement that the CI tests pass before being able to merge. 

@nanograv/pint, @nanograv/core, any comments on this PR?

jellis18:
Should we merge this?

paulray:
It looks pretty good to me.  Two questions:
(1) Why do we need to support so many versions of Python 3? It it important to have so many?
(2) Shouldn't Sphinx be in the requirements somewhere (e.g. requirements-dev.txt), since we use it to build the docs?

jellis18:
> (1) Why do we need to support so many versions of Python 3? It it important to have so many?

I guess a better question is why not? I'm really not very accustomed to Python 3 but most projects I have seen have tests for all versions. Although as you see, at the moment most tests will fail with Python 3 as there are a bunch of fixes that need to be made in a separate PR.

> Shouldn't Sphinx be in the requirements somewhere (e.g. requirements-dev.txt), since we use it to build the docs?

Yes that is definitely true. I forgot about docs since I was just making sure we had what we needed to run the tests. I'll go back and add any extra dependencies (including) Sphinx that we may need for the docs.

paulray:
@jellis18 Cool, I'm fine with all the Python 3 versions, I guess.  Someday when we get Python 3 support working, we may find that we have to drop some of those versions if they have some issue we can't deal with.

On Sphinx, the tests used to require sphinx because they ran doctests, but we no longer build the docs as part of the test script because there are many that aren't written in a way that will pass.

jellis18:
OK, I've pushed changes adding Sphinx and astropy helpers (which the docs seem to need). 

One thing related to this PR is that we will also want to change some statements in the wiki relating to testing and installation.

For example in [Installation Notes](https://github.com/nanograv/PINT/wiki/Installation-Notes) we can add that `pip install -r requirements.txt` will install required packages and that `pip install -r requirements_dev.txt` will install development tools. Also, I find that a better way to do development builds is to use `python setup.py develop` which makes symbolic links so that changes are seen without the need to re-install. This has the added benefit that you don't need to set paths and it all works just like a normal install. In the test section we can use `python setup.py nosetests`. These can then be put in the [Git Workflow](https://github.com/nanograv/PINT/wiki/Git-Workflow) section as well.

luojing1211:
@all, do you think this is ok to merge now? We can do the modification on wiki quite easy.

paulray:
I fine with merging this now

luojing1211:
I second. Merging.

PR #228, opened 2017-02-22
move chi2 calculation out of for loop

luojing1211:
This moves chi2 calculation out of the for loop of assign the new parameter values. The old version should be a bug.


luojing1211:
I believe this should be able to merge.

luojing1211:
Anyone wanna to double check this branch?

PR #237, opened 2017-03-09
Fix likelihood in event_optimize

paulray:
This corrects the normalization of the gaussian template used in event_optimize so that log Likelihood values can be compared with others (e.g the E@H people). This bug was reported by @kerrm


scottransom:
Is there no issue, or summary of one somewhere about this?  I don't really understand why that change needs to happen...  In the previous version the profile summed to 1.0.  And in the current version it sums to the number of bins in the profile.

kerrm:
Hi Scott,

The profile is a probability density, so the integral from 0 to 1 should be
1.  Integral ~ sum_i f_i *dx.  dx is 1/N_bin, so you get the resulting
change.

Cheers,
Matthew

scottransom:
That helps.  Thanks.  Fine with me to merge!

PR #238, opened 2017-03-09
Finish support for Fermi and NICER event files

paulray:
With this, PINT can read Fermi, NICER, and RXTE event files as TOAs and compute pulsar phases for each photon.

For NICER and RXTE this works with unmodified spacecraft event times, or barycentered times (as output by` axBary`).  Spacecraft times are handled by the `NICERObs` class, which reads an FPorbit file.

For Fermi, this works with unmodified event times, barycentered times (from `gtbary`), or geocentered times (from `gtbary tcorrect=geo`). Fermi spacecraft times are handled by the `FermiObs` class, which reads an FT2 file.


paulray:
Note: Do not merge this yet. The Fermi spacecraft velocities are not correct.

Issue #221, opened 2017-01-26
Switch from GH wiki to Sphinx docs?


cdeil:
Have you considered using Sphinx docs and Readthedocs instead of Github wiki pages?

IMO it's much better and pleasant to maintain, e.g.
- docs in same repo as code, versioned docs, changes go through pull requests and can be reviewed
- RST much more powerful than MD (e.g. links within your docs, or to Astropy docs, plot directive, ...) even if you don't generate API 
- Get a PDF version of the docs

It's very easy to add and reformat the existing docs content you have on the wiki, overall probably a few hours of work. This is something I'd be happy to do or help with.

Thoughts?

paulray:
Those all sound like good things.  Right now we are supposed to be using Sphinx, but not much has been done with that.  The Github wiki is only meant to document the install process and be a place for developers to use, but maybe that is all better put into Sphinx. I just don't know how to use it for general docs yet. I know it auto-generates docs out of the code, but that's it.

paulray:
How do we get started with this?  How do you write general documentation, not just auto-generated reference stuff?  I'm happy to contribute once it gets set up.

jellis18:
@paulray, I can help with this. I have set this up for our new data analysis code [enterprise](https://github.com/nanograv/enterprise) and its pretty simple. I do have a few tricks to get the code to autogenerate API docs automatically, without having created the html files before hand. This is something that is not default Sphinx behavior.

I can try to set this up in my fork and send a PR once I get the basics setup and then you and others can put in the non-autogenerated bits.

jellis18:
Should we close this now?

PR #244, opened 2017-03-11
Read The Docs integration

jellis18:
I have started to set up `PINT` for auto-generated docs using Read The Docs related to Issue #221. So far most of these docs are boilerplate and can be modified but the basics are there. These docs should build whenever we commit to the master branch (these settings can be changed I think).

This may take a few iterations to get all the integrations set up correctly but I wanted to start this PR to make you aware that I'm working on this.

When it all works we should only have modify the static content (i.e user guides, tutorials READMEs etc and the docs from the docstrings will be generated automatically).

I have also added a `Makefile` that is very useful for building and viewing the docs easily (as well as other things). You should just be able to run `make docs` and it will build the docs and open the html in your browser.


jellis18:
OK, I had to do a few things to get it to work but it seems ok now. I made a temporary Read The Docs page for my fork and you can see what the auto-generated docs would look like [here](http://pint-dev.readthedocs.io/en/latest/index.html).

paulray:
Thanks for getting that going, it looks good!

jellis18:
@paulray, thanks. In terms of the implementation I think it is basically done. You and others just have to add static material that you want. We should also probably figure out a better way of listing the modules. There are various Sphinx options that we could explore...

paulray:
Cool, then should we merge this so I, and others, can start editing?  Will the docs show up at pint.readthedocs.io?

I noticed that the docs mention being able to `pip install pint`.  Sadly this won't work because of a namespace conflict in PyPI (see #33).

jellis18:
> Cool, then should we merge this so I, and others, can start editing? Will the docs show up at pint.readthedocs.io?

They should show up at http://nanograv-pint.readthedocs.io/en/latest/, I had to use this name because pint was already taken. I also have a label in the github README like the travis label. We can put links in the README too.

>I noticed that the docs mention being able to pip install pint. Sadly this won't work because of a namespace conflict in PyPI (see #33).

OK, as I said a lot of the docs are boilerplate and will have to updated to `PINT` specifics. For example the contributing page is probably decent but should involve some more `PINT` specific stuff.

With that said I'm fine with merging, although we should probably do a squash and merge so we don't get the intermediate commits in the history.

paulray:
Hmm. I never squash my PRs, so they are filled with all my ugly intermediate commits! :-)

jellis18:
@paulray, well in some cases I think it is good so we don't have things like "Trying again" as commit messages ;)

jellis18:
Should we merge this?

PR #245, opened 2017-03-12
Coveralls and landscape integrations

jellis18:
I have set up integrations for `PINT` with Coveralls (a CI service to check out code test coverage). This is now part of `.travis.yml` and it will run after the tests pass. With this PR I also add a new badge like the one below (it still says unknown because the changes aren't merged yet).
[Coverage Status](https://coveralls.io/github/nanograv/PINT?branch=master)

I have also set up Landscape integration which shows "code health" which includes things like unnecessary imports, function arguments but also things like coding practice no-nos like long lines, missing spaces, etc. It will have a badge like the one below.
[Code Health](https://landscape.io/github/nanograv/PINT/master)

Let me know if you like these integrations and we can merge. For the most part there are no code changes that go along with this, just the CI stuff and extra badges.


paulray:
This looks very useful! I just glanced at the Landscape output and it seems to be flagging many real errors! I think we should merge this.

jellis18:
@paulray, I think we should merge too but some of the errors are just things that landscape doesn't recognize. We can probably play around with the `.landscape.yml` as we move forward to not pick up on those things.

paulray:
Concur. It is OK if some are false positives. At least it forces a person to look at them and find which ones are real errors.  We will want some way to shut it up about things that are not errors, eventually.

PR #253, opened 2017-03-16
Fix landscape error 1

luojing1211:
This is a fix for landscape errors. 
@paulray, This branch fails at test_photonphase.py with error message 
"IOError: File 'fermitest.fits' already exists." on my computer. Is this the pull request #252 trying to fix?

### DIFF
--- a/pint/models/frequency_dependent.py
+++ b/pint/models/frequency_dependent.py
@@ -1,7 +1,7 @@
 """This module implements a frequency evolution of pulsar profiles model"""
 from warnings import warn
 from . import parameter as p
-from .timing_model import TimingModel, Cache
+from .timing_model import TimingModel, MissingParameter
 import astropy.units as u
 import numpy as np
 import pint.utils as ut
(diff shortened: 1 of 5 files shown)


paulray:
Yes, that won't fail on travis since the file won't exist. That is fixed in #252.   If you want it to work for you, there is a one word change ("False" to "True") in fermiphase.py:
```
@@ -122,6 +121,6 @@ def main(argv=None):
         else:
             # Write to new output file
             log.info('Writing output FITS file '+args.outfile)
-            hdulist.writeto(args.outfile,overwrite=False, checksum=True, output_verify='warn')
+            hdulist.writeto(args.outfile,overwrite=True, checksum=True, output_verify='warn')
```

luojing1211:
Hi @paulray , 
Most fixable error has been fixed. The other errors can be fixed with other PR or needs a bigger modifying. After this, health increase to 81%. I think this should be merged.

paulray:
OK, merging...

PR #262, opened 2017-03-21
Profiling

luojing1211:
Add a little script for profiling. It is a generic profiling script.
Package required: 
cProfile
argparse
subprocess
pstats 
gprof2dot
for the last one see page. 
https://github.com/jrfonseca/gprof2dot

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -9,5 +9,6 @@ omit =
     pint/models/__init__*
     pint/observatory/__init__*
     pint/orbital/__init__*
+    pint/profiling/*
 	pint/str2ld.py
 	pint/tmtestfuncs.py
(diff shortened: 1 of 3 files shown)


paulray:
For directories like this, we should make sure that it is ignored by things like nosetests and coverage. I'm not exactly sure how to do this.

Also, this script should use `from __future__ import print_function` since we want all our code to by Python3 compatible.

jellis18:
You should add `pint/profiling/*` to the `omit` section of `.coveragerc` and `- pint/profiling` to the `ignore-paths` section of `landscape.yml`. That way it won't be picked up by landscape or by coveralls.

luojing1211:
Can we merge this branch?

paulray:
I think you just need to change `print cline` to `print(cline)`

PR #264, opened 2017-03-23
Make Py3 work

matteobachetti:
This is a series of fixes that should make tests work on Py3. 
A large part of them is about using dictionary keys as lists, which is not the case by default in Py3. There are a few minor subtleties, like modules with changed names, a different way to use metaclasses, integer division (Which is generally the most complicated thing to detect).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ matrix:
         - python: "2.7"
         - python: "3.5"
         - python: "3.6"
-    allow_failures:
-        - python: "3.5"
-        - python: "3.6"
  
 sudo: required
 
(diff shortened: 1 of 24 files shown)


matteobachetti:
@paulray Tests are passing. I disallowed failures in Py3 for clarity

paulray:
Thanks for working on this!  What are `.keys` in Py3? It seems strange that you have to wrap them in `list()` in so many places. How would this have been done if the code were written natively in Py3?  Just curious.

matteobachetti:
Dict keys are iterables in Py3, so the functionality is the same if you use them in loops, e.g. 
```
for i in blabla.keys():
    do_stuff_(i)
```
But they do not support item assignment. In Py 2 you could do something like `blabla.keys()[2]`, in Py3 it doesn't work anymore.

matteobachetti:
In Py3, `list(blabla)` would also work, without asking explicitly for the keys

paulray:
OK, thanks for the explanation! So, should we go ahead and merge?

matteobachetti:
Yes, for me it's ready to merge

jellis18:
I know this is merged but it may be cleaner to use something like
```
sorted_list = sorted(DMX_mapping.keys())
```
rather than
```
sorted_list = list(DMX_mapping.keys())
sorted_list.sort()
```

Issue #274, opened 2017-03-24
Avoid code duplication in X-ray data


matteobachetti:
`nicer_toas`, `xmm_toas` and `rxte_toas` are currently practically the same code, duplicated. I would suggest to have a generic (`Xray_FITS_TOA`?) class called by different functions that specify the correct FITS extension and keywords. This will make it easier to extend it to different satellites (e.g. NuSTAR).

paulray:
Concur. Just have to figure out how to do it gracefully.  I did try for code reuse in putting the fits reader in fits_utils.py and the plotters in plot_utils.py. So, the individual mission codes were supposed to be what is unique to each mission, but there is still significant replication.

matteobachetti:
One way might be to have a configuration file with entries like (just an incomplete example)
```
[rxte]
    telescop: XTE
    local_TOAs: True
    message: None

[xmm]
    telescop: XMM
    local_TOAs: False
    message: Currently only handles barycentered TOAs (produced by the SAS tool barycen)
                     Eventually should be extended to handle raw events, but that requires
                     understanding the orbit data in the XMM data products.

```
And call the relevant entries in the config in the reader

matteobachetti:
I can work on this, if it doesn't clash with other planned PRs

paulray:
That would be fine with me.  For photon TOAs, my focus is on getting the times correct (there are currently unresolved differences between PINT and Tempo2 answers #246 ), so I will be messing with `observatory/fermi_obs.py`.  Once I get that working, they may be some more code replication between *_obs.py that we can work on. But at the moment, I want to get it correct first, and not cause a lot of conflicts.

PR #276, opened 2017-03-24
Cleanup (X-ray) event TOAs

matteobachetti:
RXTE, XMM, NICER  etc. data in a single load_event_TOA function.


matteobachetti:
@paulray I haven't deleted the current files, in order to check properly that all functionality is still there.  `test_photonphase` passes, which makes me think that we're not far. I added the possibility of a weight, to make it compatible with Gamma-ray data (it will be trivial to adapt the function to Fermi once you are done with your tests).
Addresses Issue #274

paulray:
That's quite nice @matteobachetti! Thanks!

One thing that occurred to me is that the config dictionary might also specify, on a per-mission basis, which FITS columns should be added to the TOAs as keyword arguments (e.g. weights, PHA, PI, ENERGY, or whatever).  Maybe it could have a list of (keyword, fits_column_name) pairs that would be used to read a fits column and associate it with each TOA as keyword=value.  That would clean up the current code a bit.

matteobachetti:
@paulray In principle, I agree. However, I think we have to use caution when relying too much on a fixed format for each mission. RXTE data formats have changed a lot, for example. I'm not sure that PI and PHA are used consistently. And different FTOOLS might create different columns. 
How do we deal with that?

paulray:
True, that is one of the reasons to have different code for each mission. We could add more mission-specific stuff if we needed (e.g. for RXTE, it could whether the data were and E_ mode or a GoodXenon more and do slightly different things.

However, I don't think we are trying to turn PINT into a general photon analysis tool. So, generally the idea is to limit the amount of meta data we add to each TOA to what is required for timing analysis.  This is mainly things like photon weights and energies. For NICER, there will also be a pulse height ratio that needs to be computed.

pint/event_toas.py
```diff
@@ -0,0 +1,136 @@
...
+        new_kwargs[col] = val
+
+    if weights is not None:
+        new_kwargs["weights"] = weights
+
+    hdulist.close()
+
+    obs, scale = _default_obs_and_scale(mission, timesys, timeref)
+
+    toalist = [None] * len(mjds)
```
paulray: Out of curiousity, why pre-allocate the `toalist` then fill it in by index, rather than creating an empty list and just using `.append`?

matteobachetti: I'm not sure it applies to this particular case, where the TOA objects are not standard Python variables, but since we are making this array grow it might be more efficient to specify how many entries it will have. I was planning to benchmark it before merging, anyway.

paulray: Cool. I thought it might be an efficiency argument like that, but which way is better wasn't really intuitive to me.

matteobachetti: I benchmarked it (with a few loops using `%timeit` in ipython). The solution I used here is consistently a little faster (~5%).

paulray:
This looks great to me now! Shall we merge?

matteobachetti:
@paulray please go ahead!

PR #277, opened 2017-03-25
Fix travis

paulray:


paulray:
Makes travis install and test from the installed version

PR #279, opened 2017-03-25
Fix prefix F2 unrecognised problem 

luojing1211:
This fix the F2 unrecognised problem. 

The issue is in the model builder. When it process the prefix parameter name, it actually get confused between F2 and FD2, since I used the .startswith() method.


paulray:
I tested it on the same 11yr data set and it works fine for me!

PR #280, opened 2017-03-26
Tox for tests

jellis18:
This is a better and more standard way of doing tests on the installed version of the package. Let me know what you think of this.


paulray:
Tox seems like a good thing to use. My question is that it still looks like it will run the tests with `python setup.py nosetests`. I am not sure that will use the installed version. I think that code runs the local version, but I don't really know what happens under the hood with setup.py nosetests.

jellis18:
Yea you are right. I was sure that it used the tox environment. What is the point of these things if they don't test on the installed versions! There has to be some way better than manually cd'ing into the tests directory but maybe not.

paulray:
I assume this is obsolete given the recent work by @aarchiba 
Should we close this?

PR #288, opened 2017-03-28
Added TOA method to select and unselect TOAs using boolean arrays

scottransom:
OK.  Check this out.  This makes a stack-based way of selecting groups of TOAs using boolean arrays.  It changes the .ntoas and .observatories attributes into properties.  And makes a simple method for allowing the selection.  There are examples of usage in fit_NGC6440E.py

### DIFF
--- a/examples/fit_NGC6440E.py
+++ b/examples/fit_NGC6440E.py
@@ -20,20 +20,20 @@ m = mb.get_model(parfile)
 # Read in the TOAs
 t = pint.toa.get_TOAs(timfile)
 
-# Examples of how to select some subset of the TOAs
-# The resulting new table must be grouped by 'obs', though!
+# Examples of how to select some subsets of TOAs
+# These can be un-done using t.unselect()
 #
 # Use every other TOA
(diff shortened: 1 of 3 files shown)


paulray:
I like!

Maybe just add a unit test that pushes a couple and pops a couple of selections.

scottransom:
Take a peek and if you like it, feel free to merge.

paulray:
properties are cool!

luojing1211:
It is super cool that we can apply multiple selection on TOAs. My question
is that it will undo select by the backwards order. Do we need to provide
the functionality for unselecting by telling the program which selection to
be unselected?

- Jing Luo

scottransom:
I don't think that would be super useful.  It would be kind of hard to figure out what has been selected in each version if you made a bunch of different selects.  probably better to .pop() all the way back to the beginning and select again.  But maybe I'm wrong.

paulray:
I'd wait on adding additional complexity until there is a particular use case driving the need to undo in some particular order different than what you get with this stack-based approach.

PR #289, opened 2017-03-28
Added Fermi event selection by MJD

paulray:
This is the Fermi TOA MJD selection part of PR #153


paulray:
Do not merge. This is not working yet.

PR #293, opened 2017-04-04
Fix glitch

luojing1211:
This is a fix for the glitch model.
* Add GLF2
* Add derivatives. 
* Add test for glitch.

### DIFF
--- a/pint/models/glitch.py
+++ b/pint/models/glitch.py
@@ -12,7 +12,7 @@ from .parameter import Parameter, MJDParameter, prefixParameter
 from .timing_model import TimingModel, MissingParameter
 from ..phase import *
 from ..utils import time_from_mjd_string, time_to_longdouble, str2longdouble, \
-    taylor_horner
+    taylor_horner, split_prefixed_name
 
 # The maximum number of glitches we allow
 maxglitches = 10  # Have not use this one in the new version.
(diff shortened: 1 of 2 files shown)


luojing1211:
Is "GLEP_" parameter fittable? It is kind of a toa selection parameter.

kerrm:
GLEP should absolutely be fittable, but it's true that it won't have a continuous derivative.

paulray:
I don't think GLEP is fittable in TEMPO/Tempo2. I think TEMPO prints out what value of GLEP would give GLPH = 0.0 after it does a fit that includes GLPH.

kerrm:
Yes, it's definitely not fittable in tempo2.  Nor is the glitch transient fitting particularly stable (fails when GLTD goes negative.)  But it's a well defined question in both the photon and TOA case: did the glitch happen between these two observations, or between these two other observations, or ...  For a glitch with just an F0 jump, this can be made generated with GLPH, but for anything else, there's a distinction, especially with short transients and densely sampled data.  MCMC will handle it more gracefully than an optimizer.

paulray:
Note that Appendix C of [Clark et al.](http://adsabs.harvard.edu/abs/2017ApJ...834..106C)  has a useful discussion of fitting glitch epochs.

paulray:
For the purposes of this PR, I think we should leave GLEP not fittable, at least not by fitters that require derivatives.  An MCMC can sample values of GLEP, I would think.  Adding fitting of GLEP can be deferred to the future, and maybe tracked with an issue.

For this PR, we just want the basic glitch machinery working and the par files output to be complete, including GLEP

luojing1211:
@paulray  Is it fixed?

paulray:
Yes, I wrote a par file and it came out with GLEP. I used it with Tempo2 and it worked great!

luojing1211:
Yeahhhhh! Let us merge it.

PR #294, opened 2017-04-05
update parameter print out control

luojing1211:
* Add a control flag for if the parameter will be printed out in the parfile.

### DIFF
--- a/pint/models/parameter.py
+++ b/pint/models/parameter.py
@@ -67,7 +67,8 @@ class Parameter(object):
         A function that sets the quantity property
     get_value:
         A function that get purely value from quantity attribute
-
+    print_as_parfile: bool
+        A flag tells if the parameter should be printed out as parfile.
     Attributes
     ----------
     quantity: Type depends on the parameter subclass, it can be anything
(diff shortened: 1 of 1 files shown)


paulray:
Just to document my comment from the telecon, I worry that this functionality could cause bugs, if a parameter that has a non-default value gets set to not printing and then someone writes out a par file. That par file would then not be able to reconstruct the model correctly

paulray:
Should we close this, or do we want to discuss it more?

scottransom:
@paulray So you think that if you print a model, you should just get everything, right?  If so, I agree with that.  You can always edit it later....

paulray:
Yes, and I don't see what problem this is solving. Is there an example of a parameter we don't want printed when we print as parfile?

scottransom:
Fair enough.  I'm totally fine with closing it.  In fact, I think we should  probably have tests (and a corresponding PR) to make sure that we have all of the relevant things in a model always printed!

paulray:
Good point @scottransom . A test that models can round trip to a par file would be good. Do we not have anything like this yet?  Great student project!

Issue #297, opened 2017-04-08
Error reading error on DECJ from par file


paulray:
As reported by @kerrm

The following ephemeris line:

```
 DECJ           -04:29:06.9716059            476.94611148516092061223
```

produces this error:

`IllegalSecondError: An invalid value for 'second' was found ('476.946111485'); should be in the range [0,60).`

I don't think this is correct, as the error is not strictly less than 60 arcseconds!

PR #298, opened 2017-04-09
Fix DECJ uncertainty more then 60 arcsecond problem

luojing1211:
This should fix the issue #297. But I am not sure if there is any parfile with DEC uncertainty in the format of '0:0:xx.xxxx'? 

@paulray Have you seen any parfile like this?

-Jing Luo


paulray:
@luojing1211 I searched all 5713 .par files on my computer (from MANY different sources) and non of them use a DECJ uncertainty with ":" in it, so I think it should always be interpreted as a simple floating point number of arcseconds.

As a side note, Tempo2 often produces par files with DECJ errors that look like this:
```
DECJ           -45:10:34.81100  19099631298631109784835146042022771335484276116063150918714795143707197181702597070494097091739463025361450901761261876054083020794361316684857344.00000000000000000000
```
So, this kind of absurd number should at least not cause PINT to crash. I'm not sure if this bug in Tempo2 was ever fixed, but numbers like that are in many of my par files.

luojing1211:
So right now. PINT takes the absurd number, but it would not keep the precision. If you think this change is ok. We can merge it.

paulray:
Why is only one line changed?  The same issue should apply to RAJ as well, where uncertainty is in seconds of hourangle and numbers >60 should also be allowed.

luojing1211:
The RAJ uncertainty is in the unit of hourangle seconds. I did not find an astropy unit match this propose. One I can think of is arcsecond/15.0 as unit. But it will not add to the quantity with a unit of hourangle. What do you think? Should we just define the unit of hourangle second?

paulray:
I'm not sure. It sounds reasonable to define the unit hourangle second.

luojing1211:
@paulray I changed the RAJ uncertainty unit to newly defined `hourangle_second`. It should be able to fix the problem. As long as the uncertainty is constantly in the unit of hourangle seconds with the format of a numerical value, it should work.

luojing1211:
Hi @paulray, can we merge this. I would not like a PR getting too big.

paulray:
Can you stick in a unit test to show that both cases work with the uncertainty greater than 60?

PR #300, opened 2017-04-10
Timing model composition

luojing1211:
@ALL, 

This is PR changes the current timingmodel style to composition style with the ability to handle the component order from the model builder. It is fully tested with python 2.7 and python 3. 

Adding removing components are simplified a lot. But not in the way of making a new TimingModel class.


luojing1211:
@jellis18 could you check what is wrong with Python 3? Why it is failing

jellis18:
@luojing1211, I'm not 100% sure why it is failing but I think that you may need to change the `getattr` method in `TimingModel` to do a  check like that used in the `__new__`  method of `observatory.py`.

pint/models/model_builder.py
```diff
@@ -129,54 +90,63 @@ def preprocess_parfile(self,parfile):
         pfile.close()
         return self.param_inparF
 
+    def get_all_categroies(self,):
```
paulray: categories is misspelled here

luojing1211:
I have updated the adding and removing a component part.  It is not in the way of making new TimingModel class. But replace the orderedDict with a new ordered list of components. This way we don't have to worry about if the information has been copied to the new TimingModel class.

Please let me know is this a better way to do it. 
If we agree this is the way. We can merge it.

paulray:
What is preventing this from being merged now? I'd like it merged so I could add some changes on top of it without making conflicts. @luojing1211 has requested reviews from @scottransom @demorest and @jellis18. Are you OK with merging?

luojing1211:
The parts need to be reviewed is add_components, change_component_order section. If it is ok for you, the reviewers, we can start merge it. Future improvement can be done by smaller PR.

scottransom:
Maybe you can walk us through the add_components and change_component_order code tomorrow?
One thing that kind of worries me is that I think we thought that this would simplify the code.  But it appears as if there are a lot more new lines rather than lines removed -- and that is even given that you are deleting a bunch of cache lines!  Any idea why that is?

luojing1211:
It looks ok now.

Issue #304, opened 2017-04-12
Error computing ICRS from Ecliptic coords


dlakaplan:
In the latest PINT (0.5.2+119.g8d31238.dirty), I was trying to compute the ICRS position from a model which is defined in terms of ecliptic coords:

```
In [7]: import pint.models as models
In [8]: PSR=mb.get_model('1641+8049.par')
In [9]: PSR.coords_as_ICRS()
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-45-e8057ed5a463> in <module>()
----> 1 model.coords_as_ICRS()

/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pint-0.5.2+119.g8d31238-py2.7-macosx-10.10-x86_64.egg/pint/models/timing_model.pyc in get_cached_result(*args, **kwargs)
     63             # without caching it.
     64             log.debug(" ... no cache found")
---> 65             return function(*args, **kwargs)
     66         return get_cached_result
     67 

/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pint-0.5.2+119.g8d31238-py2.7-macosx-10.10-x86_64.egg/pint/models/astrometry.pyc in coords_as_ICRS(self, epoch)
    370                              "Check your pint/datafile/ecliptic.dat file.")
    371 
--> 372         pos_ecl = PulsarEcliptic(lon=self.ELONG.quantity+dELONG, lat=self.ELAT.quantity+dELAT)
    373 
    374         return pos_ecl.transform_to(coords.ICRS)

UnboundLocalError: local variable 'dELONG' referenced before assignment
```

I think the fix is quite simple:
in https://github.com/nanograv/PINT/blob/master/pint/models/astrometry.py line 372 should be moved up into the `else` block above.  Or at least that works on my local copy (sorry I'm not set up for pull requests).

This was with a pretty generic par file, which I can supply if needed.

paulray:
I have other questions about that function.
* Does it work if the model has PMRA and PMDEC?
* Why does it need to check self.ECL.value? I presume that is needed for the transform_to statement later.  Is that not guaranteed to be set somewhere else? It seems strange for this utility method to have to check it.
* The fix seems to be to put the PulsarEcliptic line in both the if and the else, one with the dELAT/dELONG values and one without.

paulray:
[...]

PR #305, opened 2017-04-13
Fix ecliptical coordinate bug

luojing1211:
This is a fix for issue #304. 

* Unittest have not been added. 
@dlakaplan. Can you try it out?

### DIFF
--- a/pint/models/astrometry.py
+++ b/pint/models/astrometry.py
@@ -49,7 +49,7 @@ class Astrometry(TimingModel):
 
     def barycentric_radio_freq(self, toas):
         """Return radio frequencies (MHz) of the toas corrected for Earth motion"""
-        L_hat = self.ssb_to_psb_xyz(epoch=toas['tdbld'].astype(numpy.float64))
+        L_hat = self.ssb_to_psb_xyz(epoch=toas['tdbld'])
         v_dot_L_array = numpy.sum(toas['ssb_obs_vel']*L_hat, axis=1)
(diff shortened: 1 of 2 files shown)


dlakaplan:
Works for my test.  Thanks.

luojing1211:
Great. I am waitting for @paulray's further comment on the unittest.

luojing1211:
I agree double precision is enough for position and barycentric radio frequency. But I remember the reason why we truncate the tdbld here is numpy linear algebra did not accept the longdouble precision. I was thinking why do we have to do the steps to truncate the data and lose precision.
Do we get some computational speed benefit from using double?

scottransom:
In general, yes.  There is no hardware support for long double, and so those calculations take a lot more time.  That is especially true when you are using linear algebra codes which call libraries (like numpy does) -- they have accelerated versions of libraries for double precision, but not for long double.

And for almost all pulsar timing calculations (excepting mostly those to do with spin frequency and absolute times) long double is not needed.

So in general, I'd say that we want to be using double precision everywhere, unless we know for sure that we need long double.  And not the other way around.

luojing1211:
Ok, we can change it back?

scottransom:
Did you do some test that suggested that that change needed to be made?  If not, then yes, I suggest that you change that back.  (I'm pretty sure that I changed back another change like this that you made in toa.py a month or two ago).

luojing1211:
This is working. Merging.

PR #306, opened 2017-04-26
Change set_quantity_angle and set_quantity_mjd to accept long double input quantities.

kstovall:
Currently, if a long double quantity is passed to set_quantity_angle or set_quantiy_mjd (in models/parameter.py), it will result in a value error. This pull request fixes that.

### DIFF
--- a/pint/models/parameter.py
+++ b/pint/models/parameter.py
@@ -412,12 +412,16 @@ class floatParameter(Parameter):
     """
     def __init__(self, name=None, value=None, units=None, description=None,
                  uncertainty=None, frozen=True, aliases=None, continuous=True,
-                 long_double=False, **kwargs):
+                 long_double=False, unit_scale=False, scale_factor=None,
+                 scale_threshold=None, **kwargs):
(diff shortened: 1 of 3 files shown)


luojing1211:
Hi @kstovall, is there any case failing to accept longdouble for those two functions? Or do those changes improve the performance?

kstovall:
This was causing problems for me in getting QTIP working with PINT (through ltinterface.py). There were some values being passed in as long doubles and I would receive a ValueError. An alternative would be to look into whether everything actually need to be long doubles, but I looked at these functions and they convert their input to long double anyway, so why not just accept the case of a long double input as well?

luojing1211:
the check for numbers 
```
isinstance(val, numbers.Number):
```
should accept long double. That is strange.

kstovall:
Yes, it is indeed strange. I tried isinstance with a long double on two machines and got different answers. Thinking that numpy may be out-of-date, I updated and it and now numbers.Number seems to work for long double now. I guess this can be closed.

PR #312, opened 2017-06-13
Add pulse profile modeling classes.

kerrm:
These classes support photon-based light curve modelling, inherited from Fermi light curve modelling efforts.


paulray:
Can you add a blank __init__.py to the templates directory? I think that is needed for them to be imported as `pint.templates.lcfitters`, for example.

PR #321, opened 2017-07-13
Fix tempo clock correction file alinement. 

luojing1211:
Hi all, 
This is a quick fix for parks tempo style clock correction file alinement. TEMPO has very restrict rule for column alinement. However parks clock correction file has some lines did not follow the convention, which introduce a bug for reading the clock correction data. This is a fix in the file.


scottransom:
Is there a reason why this wasn't just merged?  I seriously don't see any issues with fixing blatant bugs via a commit if all of the tests continue to pass after the bugfix is made.....

luojing1211:
No there is no reason. Merging.

Issue #320, opened 2017-07-13
models.StandardTimingModel problem?


xuhengdada:
**When I was doing:**
m = models.StandardTimingModel
m.read_parfile("NGC6440E.par")
m.phase(tt)

**and I got a problem:**

---------------------------------------------------------------------------
UnitsError                                Traceback (most recent call last)
<ipython-input-63-9bb64dd30ad9> in <module>()
----> 1 m.phase(tt)

/usr/local/lib/python2.7/site-packages/pint-0.5.2+181.g6b8fbcd-py2.7-macosx-10.12-x86_64.egg/pint/models/timing_model.pyc in phase(self, toas)
    456         """Return the model-predicted pulse phase for the given TOAs."""
    457         # First compute the delays to "pulsar time"
--> 458         delay = self.delay(toas)
    459         phase = Phase(np.zeros(len(toas)), np.zeros(len(toas)))
    460         # Then compute the relevant pulse phases

/usr/local/lib/python2.7/site-packages/pint-0.5.2+181.g6b8fbcd-py2.7-macosx-10.12-x86_64.egg/pint/models/timing_model.pyc in delay(self, toas, cutoff_component, include_last)
    450 
    451         for df in self.delay_funcs[0:idx]:
--> 452             delay += df(toas, delay)
    453         return delay
    454 

/usr/local/lib/python2.7/site-packages/astropy/units/quantity.pyc in __array_ufunc__(self, function, method, *inputs, **kwargs)
    619         # consistent units between two inputs (e.g., in np.add) --
    620         # and the unit of the result (or tuple of units for nout > 1).
--> 621         converters, unit = converters_and_unit(function, method, *inputs)
    622 
    623         out = kwargs.get('out', None)

/usr/local/lib/python2.7/site-packages/astropy/units/quantity_helper.pyc in converters_and_unit(function, method, *args)
    481                                      "argument is not a quantity (unless the "
    482                                      "latter is all zero/infinity/nan)"
--> 483                                      .format(function.__name__))
    484             except TypeError:
    485                 # _can_have_arbitrary_unit failed: arg could not be compared

UnitsError: Can only apply 'add' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan)

**same problem in bash enviroment:**
pintempo --plot psr1.par.txt psr1.tim.txt
wrong:
[...]

PR #322, opened 2017-07-14
This is fix to get all the delays and phases with units

luojing1211:
As mentioned in Issue #320, Astropy 2.0 start to be very restricted to the units operation. This pull request is to fix it.

### DIFF
--- a/pint/__init__.py
+++ b/pint/__init__.py
@@ -23,7 +23,7 @@ dmu = u.def_unit('dmu', u.pc*u.cm**-3)
 
 # define equivalency for astropy units
 light_second_equivalency = [(ls, si.second, lambda x: x, lambda x: x)]
-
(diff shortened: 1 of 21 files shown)


paulray:
I've only had time to skim this, but in some of the cases the units look very "tacked on", being adding in the `return` statement for example, or doing a calculation with `x.value` and then sticking a unit on.  I think we should try to make sure that the computations are being done with units at a deeper level, wherever possible (e.g. why does `taylor_horner` need a .value when called?).

luojing1211:
Taylor_horner was designed to be a very general function deal with numerical value only. We can make it accept units, but the price is we have to put some unit detection functionality into it, or make a wrapper.

paulray:
I am merging this because people are not able to use PINT with astropy 2.0

PR #323, opened 2017-07-25
Noise model

luojing1211:
Hi all, 
This is a toy noise model for PINT.  I would like to get comment on it. Please check it out.


jellis18:
I don't mean to be a pain but I think this is moving in the wrong direction. In my opinion you never want to make the full N x N covariance matrix, especially for a diagonal matrix. It would probably crash your computer with the newest 1713 data set. For the red noise and ECORR it is best to model them just like the linear timing mode parameters with a design matrix. The only difference is that they have a gaussian prior piece that is set by the red noise amplitude/spectral index and the ecorr parameters. 

I would keep the EFAC/EQUAD scaling of the toa errors and treat the red noise ECORR differently. I don't know how to fit this into the whole PINT structure though...

jellis18:
We actually had this design question in `enterprise`. What you could do is instead of having `NoiseComponent` having `covariance_matrix_funcs`, it could instead have three lists `sigma_diag_func`, `basis_func`, `prior_func` where the first returns a diagonal matrix with the scaled errors, the second returns the "design matrices" for red noise and ECORR and the last returns the diagonal of the covariance of the gaussian prior on the amplitudes (corresponding to the aforementioned) basis functions. That way the fit could read in the scaled error bars and read in the needed components for red noise and ECORR. I could give some code examples tomorrow. As I say, this is very similar to how we handle this in `enterprise`.

luojing1211:
Hi Justin,

That are good suggestions. I had a short discussion with Paul D. We think a
robust covariance matrix class is a good idea. The class should smartly
store the covariance matrix elements and should be able to out put a
covariance matrix if needed. Of course the fitter can be adjusted
accordingly. What do you think?

-Jing

jellis18:
@luojing1211, I still don't know. Why would we even want to make or store a full covariance matrix if everything can be done so so much more efficiently? For example, a full GLS fit of the 11-year 1713 set takes nearly an hour in tempo using full covariance matrices and about 100 ms in tempo2 or libstempo using other methods. 

I see no problem having the capability of using a full covariance matrix but we definitely need the optimized fast version for general use. I'm working on this now in a [branch](https://github.com/jellis18/PINT/tree/noise_model) on my fork.

luojing1211:
I does look strange. T2EQUAD should not be there. I need to do some
investigation.

jellis18:
@luojing1211, you can look at the changes I've made [here](https://github.com/luojing1211/PINT/compare/noise_model...jellis18:noise_model?expand=1).

luojing1211:
Hi Justin,

In my branch, I have more stuff fixed after you branch out. You can do a
rebase on my branch. It should fix the issue.

-Jing Luo

jellis18:
OK @luojing1211. I have added your fork as a remote and I do `git pull luojing1211 noise_model` but is says it is up to date. Should I do something else?

luojing1211:
What is the last commit in my branch?
​

jellis18:
[This](https://github.com/luojing1211/PINT/commit/94be4bab6f4f612e3fd9fb81cd244983759cc82f) is the last commit.

luojing1211:
This commit should be the one fix the T2EQUAD unrecognized problem. Let me
do some more tests.

jellis18:
OK, the EFAC/EQUAD stuff works if I comment out the ECORR lines in the par file. Its just a problem when it has both EFAC/EQUAD and ECORR.

luojing1211:
Hi Justin,

I see where the bug is. In the model builder, it would add only one
component into the TimingModel class. So it will only take EcorrNoise. I
will fix this .

-Jing Luo

luojing1211:
I mean only one component from one category. For this case is the
timing_noise category.

-Jing Luo

luojing1211:
This pull request has been implemented by PR #334

PR #325, opened 2017-08-01
Bug fixes and tweaks.

kerrm:
Add some new capabilities for NICER, and fix some small bugs I found along the way.

The units in d_phase_d_toa could be done more gracefully.  I tried putting the correct units in the sample step, but apparently astropy isn't smart enough to figure out cycle/s == Hz, so I just kluged in Hz.

### DIFF
--- a/pint/models/timing_model.py
+++ b/pint/models/timing_model.py
@@ -505,7 +505,8 @@ class TimingModel(object):
             deltaT = time.TimeDelta(dt_array)
             copy_toas.adjust_TOAs(deltaT)
             phase = self.phase(copy_toas.table)
-            sample_phase.append(phase Use finite difference method.
+            sample_phase.append(phase)
+        #Use finite difference method.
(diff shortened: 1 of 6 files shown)


kerrm:
Looks like it will require some updates to a few other scripts (zima) to reflect the changed units.

paulray:
Do we need to have a discussion of whether phases should be unitless or have units of "cycles"? @luojing1211 what do you think?

kerrm:
It's somewhat ancillary to this question -- regardless of whether the
numerator is unitless or in cycles, the denominator is missing the time
unit.

kerrm:
OK, after some discussion with Paul, I think there's a better solution for the units in d_phase_d_toa, which I've implemented.  Using the conversion should work for the case where a user passes a step with non-second dimension.

kerrm:
Thank god it passes the tests.

PR #327, opened 2017-08-05
some new features for profile fitting

kerrm:
I don't understand why this branch shows it's 17 commits ahead of nanograv/master.  I've merged master and it says everything is up to date.  At any rate, only the last two commits are the relevant ones, if you want to cherry pick them.


paulray:
Yeah, I get the same thing will pull requests. Something about our github workflow could be improved. I think it is because I did a "squash and merge" for your last PR. Thus all those changes are already in PINT, but the individual commit hashes are not. So your new PR shows a ton of commits but the end result is only the few changes you put in recently.  Maybe a git expert like @scottransom can tell us how to tune the git workflow to make this cleaner?

paulray:
I'm going to try merging this without squashing and see if that is better.

PR #329, opened 2017-08-08
Fix infinite frequency problem

luojing1211:
This is a fix for the infinite frequency problem for FD model.

### DIFF
--- a/tests/test_fd.py
+++ b/tests/test_fd.py
@@ -5,6 +5,7 @@ import astropy.units as u
 import pint.residuals
 import numpy as np
 import os, unittest
+import copy
 from pinttestdata import testdir, datadir
 os.chdir(datadir)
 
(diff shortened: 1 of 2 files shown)


luojing1211:
@paulray Can you try it out now?

paulray:
I think the same fix needs to be applied in `d_delay_FD_d_FDX()`

paulray:
But, for my simple use case, this does seem to fix it. Thanks!

paulray:
Ideally, it would be good to have a unit test with an infinite freq TOA

luojing1211:
Hi Paul, 
I just pushed the fix for derivatives and added some infinite frequency test.

tests/test_fd.py
```diff
@@ -28,6 +29,19 @@ def test_FD(self):
         # Those two clock correction difference are causing the trouble.
         assert np.all(resDiff< 5e-6) , \
             "PINT and tempo Residual difference is too big. "
+    def test_inf_freq(self):
+        test_toas = copy.deepcopy(self.toas)
+        test_toas.table['freq'][0:5] = np.nan
```
paulray: I think this should be np.inf * u.MHz (i.e. it should not be np.nan and it should have units)

luojing1211: It is fixed now.

PR #332, opened 2017-08-08
Enable BIPM file selection

luojing1211:
This is a pull request that fixes the bipm file selection. It was only using bipm2015, but the pull request allows users to select other bipm files.

### DIFF
--- a/pint/observatory/observatory.py
+++ b/pint/observatory/observatory.py
@@ -164,7 +164,7 @@ def get_observatory(name, include_gps=True, include_bipm=True,
                       correction.
         include_bipm = Set False to disable TAI TT(BIPM) clock
                       correction.
-        bimp_version = Set the version of tt bimp clock correction files.
+        bipm_version = Set the version of TT BIPM clock correction files.
 
     .. note:: This function can and should be expanded if more clock
         file switches/options are added at a public API level.
--- a/pint/observatory/observatory.py
+++ b/pint/observatory/observatory.py
@@ -147,7 +147,8 @@ class Observatory(object):
         raise NotImplementedError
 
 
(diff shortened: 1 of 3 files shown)


pint/observatory/observatory.py
```diff
@@ -163,6 +164,7 @@ def get_observatory(name, include_gps=True, include_bipm=True):
                       correction.
         include_bipm = Set False to disable TAI TT(BIPM) clock
                       correction.
+        bimp_version = Set the version of tt bimp clock correction files.
```
paulray: Spelling error. This should read bipm_version

pint/observatory/topo_obs.py
```diff
@@ -50,8 +51,12 @@ def __init__(self, name, tempo_code=None, itoa_code=None, aliases=None,
                           values).  Default='tempo'
             include_gps = Set False to disable UTC(GPS)->UTC clock
                           correction.
-            include_bipm = Set False to disable TAI TT(BIPM) clock
+            include_bipm= Set False to disable UTC-> tt BIPM clock
```
paulray: * TT should always be capitalized
* This comment isn't very clear to me. The include_bipm flag sets what is done to convert TAI to TT, right? It has nothing to do with the UTC to TAI conversion. So, this docstring should say that if include_bipm is false, then then it will use TT = TAI+32.184s (Is that correct?  I think that must be what astropy.Time does for the TAI to TT conversion, but please check).  If it is true it will use TT=TT(BIPMYYYY). And, you might also include this hyperlink: http://www.bipm.org/en/bipm-services/timescales/time-ftp/ttbipm.html

luojing1211: How about the comments now?

Issue #273, opened 2017-03-24
Add GLS fitter to PINT


paulray:
Add GLS fitter than does the same thing as `tempo -G`, using pre-determined noise model parameters in a generalized least squares fitter.

jellis18:
I think we get a lot of ideas and copy some code from [libstempo](https://github.com/vallis/libstempo/blob/master/libstempo/libstempo.pyx). Take a look at the `_fit` function. All you really need is a way of reading in the noise values from the par file and the rest of the `_fit` function is the same.

luojing1211:
What is the difference between the EQUAD and T2EQUAD. Do we need to distinguish those two parameters in PINT?

jellis18:
@luojing1211, I'm not sure about EQUAD but we should be aware of all of the different variants. All versions of EFAC, (i.e. T2EFAC, TNEFAC) are the same. But T2EQUAD and TNEQ are both EQUADs but defined differently;

TNEQ is measured in log10 seconds and T2EQUAD is in microseconds, also they are defined as follows

TNEQ:
sigma_scaled = (EFAC^2\*sigma_old^2 + 10^(2\*TNEQ))^(1/2)

T2EQUAD:
sigma_scaled = EFAC \* (sigma_old^2 + (T2EQUAD\*1e-6)^2)^(1/2),

where sigma_old and sigma_scaled are the new and originally measured uncertainties.

In terms of ECORR, I believe the ECORR and TNECORR are the same and both measured in microseconds.

There are also lots of other components of IPTA/EPTA/PPTA data that PINT doesn't handle like the way that they do DM and the different noise parameters. We will have to talk with Lindley and others to figure it all out.

luojing1211:
PR #334 solved this issue. Closing...

PR #334, opened 2017-08-09
Noise model and GLS fitter

jellis18:
I have added functionality for the standard noise model (EFAC, EQUAD, ECORR, and red noise) and a corresponding GLS fitter. I have tested that this fitter recovers the WlsFitter functionality when no noise parameters are in the par file. I have also tested it against tempo2. The chi-squared values are 3813.79 (PINT GLS) and 3813.71 (libstempo GLS). I've also compared the postfit values and uncertainties against libstempo. In the plots below blue are the parameter differences  and uncertainty differences. By difference I mean abs(x_PINT-x_T2) / x_T2 where x is the parameter value or uncertainty. For all parameters this is less than 1%.

par_diff

dmx_diff

There are still a few things left to do with this. Some things could be cleaned up and some tests added. Also, at the moment we are not storing the red noise ML waveform, which we could do. Also I don't know how to get the "temponest" style parameters to be recognized in the par file. These are parameter like `TNRedAmp`, `TNRedGam`, etc. It looks like maybe PINT expects all par file parameters to be capitalized. I have the functionality in the red noise model to use these but the par file reader doesn't recognize them at the moment.

I can also add the full covariance matrix option as well...


paulray:
This addresses issue #273 

This is a great step for PINT!  I think that when this is merged, we should tag version 0.6.0, so that we can easily say that noise models were added as of version 0.6

jellis18:
OK, `GLSFitter.fit_toas` now has the `full_cov` option that will let one optionally use the full noise covariance matrix for the GLS fitting. As discussed elsewhere, I would recommend against using this in most cases because it could take a **very** long time for large data sets.

I've tested that it gives the same parameter estimates and uncertainties as the basis-prior method that I showed above. The plots below are the same as those above but now comparing the full covariance matrix method (blue) and the basis-prior method (orange).

par_diff_cov

dmx_diff_cov
 
The differences in parameter and uncertainty estimates between these two methods is negligible.

Now we need to figure out how to deal with `TNRedAmp`-like parameters. Anyone have any suggestions for this?

luojing1211:
I will look into the case sensitive problem, hopefully it will get solved
this afternoon.

-Jing

luojing1211:
@jellis18, when PINT reads the parfile it will alway convert the read in parameter name to upper case. So I would recommend when you define your parameters in PINT, use the upper case. But if you want case sensitized output parfile, we will have to modify the code.

luojing1211:
@Justin this is a quick fix for the case sensitivity problem. 
https://github.com/luojing1211/PINT/blob/a90379775c44b656ea687896d643ca389cf17da0/pint/models/parameter.py#L371
Since this is not a big commitment, you can add it to your PR directly.

jellis18:
OK, if any of the main developers want to take a look at this I think it is basically ready to go.

jellis18:
Merging!

jellis18:
Actually before I merge I will change some of the naming conventions to avoid confusion with the prior class.

jellis18:
Ok, I have changed the name from "prior" to "weight" when referencing the noise model. Merging now.

Issue #330, opened 2017-08-08
TDB times computed from spacecraft-based TOAs are wrong by a couple microseconds


paulray:
This is just a reminder that the TT to TDB conversion for spacecraft-based TOAs is still not quite correct. The results are off by a couple of microseconds because it isn't computing the topocentric term correctly.  This is supposed to be a dot product of the spacecraft barycentric velocity and the geocenter->observatory vector.  This isn't quite working as PINT currently lets astropy do the TT to TDB conversion with location=None. This needs to be fixed.

luojing1211:
PR #335 is merged. 
@paulray  can you test it out and see if PR #335 fixes this issue?

paulray:
I found this bug. It is in compute_TDBs() in toa.py
When it groups the TOA table by observatory, it assumes that all TOAs from the same observatory share a same location. This is NOT true for spacecraft TOAs!
Here is the offending snippet of code:
```        # Compute in observatory groups
        tdbs = numpy.zeros_like(self.table['mjd'])
        for ii, key in enumerate(self.table.groups.keys):
            grp = self.table.groups[ii]
            obs = self.table.groups.keys[ii]['obs']
            loind, hiind = self.table.groups.indices[ii:ii+2]
            site = get_observatory(obs)
            ### BUG BUG! This is wrong if the observatory is not at a fixed ITRF location!
            grpmjds = time.Time(grp['mjd'], location=grp['mjd'][0].location)
            grptdbs = site.get_TDBs(grpmjds, method=method, ephem=ephem)
            tdbs[loind:hiind] = numpy.asarray([t for t in grptdbs])
```

Now, how to fix it?

We don't want to revert to looping through every single TOA, calling t.tdb for each one. That would be horribly slow.

paulray:
Ah, the location attribute can be an array, one for each input time. That should solve it!

paulray:
Nope.  The docstring for `astropy.Time` says this:
```    
location : `~astropy.coordinates.EarthLocation` or tuple, optional
        If given as an tuple, it should be able to initialize an
        an EarthLocation instance, i.e., either contain 3 items with units of
        length for geocentric coordinates, or contain a longitude, latitude,
        and an optional height for geodetic coordinates.
        Can be a single location, or one for each input time.
```
but the code is this, which does not seem to support a list or array of locations:

```    
def __init__(self, val, val2=None, format=None, scale=None,
                 precision=None, in_subfmt=None, out_subfmt=None,
[...]

Issue #112, opened 2016-02-22
Add capability for global configuration information


paulray:
We need a capability to set global configuration variables.  Defaults could be read from some config or .pintrc file, for example, then any code should be able to set them as needed.  

These would be things that control the overall operation of PINT. Examples might include which clock correction files to use (although that might be better to be in the TOAs class), setting log levels, and other things I can't think of at the moment.  These should be parameters that aren't specific to the TOAs, TimingModel or Fitter classes, which of course can have their own configurations.  Partially, these could be a way to deal with par file parameters that need to get communicated to the other places (see [this wiki page](https://github.com/nanograv/PINT/wiki/Discussion-of-Control-Parameters))

demorest:
This is not directly addressed yet, but some of the framework started in #115 (pint.config module, inclusion of appdirs package) is meant to be extended to handle this type of thing.

paulray:
What if we do the following:
* Have a file in the pint distro that contains defaults for all global configuration settings
* Have pint.config read it in in some way so any PINT function can access it easily 
* Have a way to override any subset of the defaults by specifying a user configuration file, e.g. `pint.config.load('myconfig.rc')`
* And users should be able to override config files programmatically without needing to write a config file (e.g. `pint.config.settings['UNITS'] = 'TCB'`)

If that sounds good, we just have to decide exactly how to implement it.

Questions:
* What file format should the config file be in?  YAML or INI or Python?
* Where should the default config file be stored?  In the top level `pint` directory, in a subdir like `pint/etc` or `pint/rc`, or in `pint/datafiles` (which already exists)?
* Should users be able to have a file like `~/.pintrc` that holds their own defaults that are *automatically* read in?  Or, should users be able to edit the default configuration in the PINT distro (I vote no on this)?  I kind of think always having to load your own changes explicitly is good.  You may well have a pulsar.tim, pulsar.par, and pulsar.conf set of files for any particular pulsar you are working on, but that is fine.  If it reads a user-based .pintrc file it may be very hard to debug because you could forget what changes you'd made in there.

Notes:
* A very simple way to do this with YAML is shows here: http://btmiller.com/2015/04/11/simple-config-setup-python-yaml.html
* A discussion of various configuration file formats for Python codes is here: https://martin-thoma.com/configuration-files-in-python/

Discuss... @scottransom @cdeil @luojing1211

demorest:
Paul, a few thoughts on this:

[...]

PR #335, opened 2017-08-10
enable PINT to get tdb_tt correction from ephemeris

luojing1211:
This is a pull request to enable PINT read tdt-tt from ephemeris. It is not fully tested yet. If any developer wants to help test this PR, it would be great.

### DIFF
--- a/pint/solar_system_ephemerides.py
+++ b/pint/solar_system_ephemerides.py
@@ -163,3 +163,46 @@ def objPosVel(obj1, obj2, t, ephem):
         return obj2pv - obj1pv
     else:
         return PosVel(np.zeros((3,len(t)))*u.km, np.zeros((3,len(t)))*u.km/u.second)
+
+def get_tdb_tt_ephem(t, ephem, path=None, link=None):
+    """The is a function to read the TDB_TT correction from the JPL DExxxt.bsp
+       ephemeris file.
+       Parameter
+       ---------
+       t: Astropy.time.Time object
+           Observation time in Astropy.time.Time object format.
+       ephem: str
+           Ephemeris name.
+       Note
+       ----
+       Only the DEXXXt.bsp type ephemeris has the TDB-TT information, others do
+       not provide it.
+    """
+    ephem = ephem.lower()
+
+    if path is None:
+        if link is None:
+            link_str = ''
+        else:
+            link_str = link
+        is_load = _load_kernel_link(ephem, link=link_str)
+        if not is_load: # Link does not Try to load from data path.
+            path_str = ''
+            is_load = _load_kernel_local(ephem, path=path_str)
+        if not is_load:
+            raise ValueError("Can not load the ephemeris file '%s.bsp'. " % ephem)
+    else:
+        path_str = path
+        is_load = _load_kernel_local(ephem, path=path_str)
+        if not is_load:
+            raise ValueError("Can not load the ephemeris file '%s.bsp' from the"
+                             " local directory %s." % (ephem, path_str))
+    kernel = coor.solar_system_ephemeris._kernel
+    try:
+        seg = kernel[1000000000, 1000000001]
+    except KeyError:
+        raise ValueError("Ephemeris '%s.bsp' do not provide the TDB-TT correction.")
+
+    tt = np.longdouble(t.jd1) + np.longdouble(t.jd2)
+    tdb_tt = seg.compute(tt)[0]
+    return tdb_tt
--- a/pint/solar_system_ephemerides.py
+++ b/pint/solar_system_ephemerides.py
@@ -164,7 +164,7 @@ def objPosVel(obj1, obj2, t, ephem):
     else:
         return PosVel(np.zeros((3,len(t)))*u.km, np.zeros((3,len(t)))*u.km/u.second)
 
-def get_tdb_tt_ephem(t, ephem, path=None, link=None):
+def get_tdb_tt_ephem_geocenter(tt, ephem, path=None, link=None):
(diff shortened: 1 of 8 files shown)


pint/solar_system_ephemerides.py
```diff
@@ -163,3 +163,46 @@ def objPosVel(obj1, obj2, t, ephem):
         return obj2pv - obj1pv
     else:
         return PosVel(np.zeros((3,len(t)))*u.km, np.zeros((3,len(t)))*u.km/u.second)
+
+def get_tdb_tt_ephem(t, ephem, path=None, link=None):
+    """The is a function to read the TDB_TT correction from the JPL DExxxt.bsp
+       ephemeris file.
```
paulray: Could you add a hyperlink here in the docstring to where the TDB_TT part of the DExxxt ephemeris is defined?

paulray:
This is an excellent addition!

pint/solar_system_ephemerides.py
luojing1211: There seems no link for the tt-tdb ephemeris definition. There may be some pdf or txt documents. But the comments in the ephemeris have the definition of segments, which has part of TDB-TT.
If you do:
```
>>> from jplephem.spk import SPK
>>> kernel = SPK.open('de436t.bsp')
>>> print(kernel)
```
It should have the information you need.

paulray:
Does one of the tests compare the values computed from the ephemeris with what astropy gets when using the same ephemeris?  How different are those?  This may not need to be a unit test, but should be done.

paulray:
I'm concerned that this new method may be dropping the topocentric part of the TT to TDB correction.
When we do it with astropy:
```
               grpmjds = time.Time(grp['mjd'], location=grp['mjd'][0].location)
               grptdbs = grpmjds.tdb
```
It knows the location of the observatory, and the topocentric part of the correction is applied (see description [here](https://github.com/nanograv/PINT/wiki/Clock-Corrections-and-Timescales-in-PINT)).
But, the TT_TDB column in the ephemeris is evaluated for the Geocenter.  So, I think that this code will not get the topocentric part correct. Thus, I predict that if you compare these results for DE430 using astropy and the TT_TDB column, you will see a couple of microsecond difference.

luojing1211:
I am doing the test right now. The difference seem in the microsecond level. But I have to double check if I am doing the right thing.

luojing1211:
I have the difference between those two method. I can see the microsecond difference. 
screen shot 2017-08-10 at 11 10 04 pm
That means if we want to get this right, we need to have the part of topocentric correction. 

-Jing Luo

paulray:
In the docstring, you want to link to this document, page 6, which describes how this column is computed: https://ipnpr.jpl.nasa.gov/progress_report/42-196/196C.pdf

paulray:
Your test results are exactly what I expected. So, you just have to add that topocentric term as defined on the wiki page I linked to.  That is actually the part of the transformation that I'm having trouble with for spacecraft observatories, as I note in issue #330

luojing1211:
That means, we are lacking of the last term of Eq5 line2 from the paper. This should not be hard to do.

luojing1211:
I added the topocentric part. The plot below is the comparison between astropy computed tdb and ephemeris computed tdb. It looks like matching our expectation. 
screen shot 2017-08-11 at 2 53 00 pm

paulray:
Looks good! Can you say a little bit about how you did the comparison?  Were they using exactly the same ephemeris files? If the code is short maybe just post it in a comment?

luojing1211:
I used the same ephemeris. Here is the code for doing this.
```
par = 'datafile/B1855+09_NANOGrav_9yv1.gls.par'
tim = 'datafile/B1855+09_NANOGrav_9yv1.tim'
t = toa.get_TOAs(tim, ephem='DE436t')
t2 = toa.get_TOAs(tim, ephem='DE436t', tdb_from_ephem=True)
plt.plot(t2.get_mjds(), (t.table['tdbld']-t2.table['tdbld'])*86400.0, '+')
```
This PR modified the toa.py and topo_observatory.py. Paul D. and I think this function can be more generalized and the code can be cleaner. Maybe put all of them into observatory module.

paulray:
@demorest are you OK with merging this now, or do you think some code cleanup should be done first?

demorest:
hi Paul, there are a couple structural things I'm not totally happy with:

1. The addition of this option as yet another "top-level" argument to get_TOAs().  The more of these we add, the kludgier it gets (in my opinion anyways).  We really should have a better system for dealing with these sorts of options.

2. The addition of an "if-then" block inside TOAs.compute_TDBs() to deal with the option.  I think things like this should probably be handled inside Observatory.

Jing and I already talked about these some when he was here last week.  I think #1 might be a bigger issue, so maybe we should just ignore that for the time being and open a separate discussion about it.  #2 should probably be addressed here but I wanted to talk with you first about how to make this handle the spacecraft case cleanly; was hoping maybe we could do this on the next telecon.

Also I have a vague suspicion we could be making better use of the astropy time framework (rather than adding cases within PINT) for handling this TT-TDB conversion.  But I haven't really had a chance to look into it.

paulray:
@demorest yes I agree with your concerns.

Number 1 is definitely important to deal with, but is a bigger issue (as we've discussed in issue #112).

I'd be happy to discuss number 2 at a telecon. I should be able to make it on Wednesday this week. @luojing1211 can you put it on the agenda?

luojing1211:
​Yes, I already marked it.

-Jing ​

luojing1211:
Hi all, 
I have moved the method selection part to the observatory module. Please check it out.

PR #345, opened 2017-09-27
Bring toas to original order and add toas of observatory index

luojing1211:


scottransom:
Why is it necessary to reorder the TOAs?  Ordering by observatory helps with array calculations.  The index number of the TOA can be stored on the table easily to allow reordering back to original if needed.

Scott

PS:. I'm in NYC and can't be on the call today.  Sorry!

On September 26, 2017 10:52:14 PM EDT, Jing Luo <notifications@github.com> wrote:
>
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/nanograv/PINT/pull/345
>
>-- Commit Summary --
>
>  * add bring toas to original order and add toas of observatory index
>
>-- File Changes --
>
>    M pint/toa.py (4)
>
>-- Patch Links --
>
>https://github.com/nanograv/PINT/pull/345.patch
>https://github.com/nanograv/PINT/pull/345.diff
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/nanograv/PINT/pull/345

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

luojing1211:
​I asked myself the same question.  The issue I can image is that perhaps
our user have the assumption of TOAs are sorted in the order of .tim file,
since Tempo and Tempo2 are doing the same. It is simple to fix, but it
could potentially be a confusion for our users.

scottransom:
As long as we give them a way to re-order to the order in the .tim files I think that should be fine.  It makes sense to work with the TOAs in the way that is best for the software, not for the user.  And it is *easy* to re-order them in the original order if they want them that way.

luojing1211:
In that way we should have a very clear statement in our documentation about it. Another potential issue is after the sorting by the TOA index the observatory group will get disordered.

luojing1211:
I did the test. Sorting the table do breaks the groups. I think the best way to solve this problem is return a new copy of table. I will make the change on this PR.

scottransom:
Just a note:  the grouping by observatory is required for Solar System Shapiro delay calcs.

aarchiba:
I think if users are going to want a table they can manage t.sort('index').

PR #350, opened 2017-10-25
Change the delay jump to phase jump

luojing1211:
This is a pull request for changing the delay jump to phase jump.


luojing1211:
As we discussed in the PINT telecon, this branch is good to merge.

PR #356, opened 2017-11-20
BTX

luojing1211:
This is a pull request to add the BTX model. One big change is to Add FBX parameter for orbits. 
Unittest is not ready yet. But it is open to review now.


luojing1211:
Hi @jellis18, 
This branch fails at the gls fitter tests. The CHI2 becomes very small and post fit uncertainty get very big. I double checked the design matrix, it seems working fine. Can you help me to check this issue out? Thank you.

jellis18:
@luojing1211, I don't know why this is failing since it is using the DD model. I haven't followed these changes but the derivative for A1 must have changed somehow.

luojing1211:
@jellis18 , Actually, in this branch. I move the binary orbits to a new class for the future different parameterization implementations. In the code structure, orbits are now calculated by the orbits_cls attribute. That is the only change. I compare the derivative for A1 from the master branch and this branch. They look the same.

luojing1211:
@jellis18 This branch is fixed. Somehow I switched off the noise parameters in the par file.

PR #357, opened 2017-12-05
Improved event_optimize to allow reading FT2 file for raw Fermi FT1 files

paulray:
Added capability to event_optimize to specify an FT2 file, which will instantiate a Fermi observatory. Then, event_optimize should run fine on Fermi FT1 files that have not been barycentered or geocentered.


scottransom:
That sounds sweet!  How does it know the location of the observatory, though?  Do you have to explicitly supply an orbit file?

Also, if PINT is doing the barycentering/geocentering, how does that compare speed-wise to doing the geocentering first using the Fermi tools and then telling PINT to work with the geocentered events?

paulray:
Yes, the functionality was already there to read an orbit file. This just added an option to event_optimize to read that file, exactly as the fermiphase program already can.
I will try a speed comparison.

paulray:
I just (accidentally) added some commits to this PR that apply clock corrections in the pintbary script. I'm going to merge it since we need this functionality to work.

PR #358, opened 2017-12-11
Python 2/3

cdeil:
Some simple edits to make code work on Python 3 (no changes for Python 2).

### DIFF
--- a/examples/example_B1855.py
+++ b/examples/example_B1855.py
@@ -2,11 +2,9 @@
 import pint.models.model_builder as mb
 import pint.toa as toa
 import matplotlib.pyplot as plt
-import tempo2_utils
 import astropy.units as u
 from pint.residuals import resids
-import numpy as np
-import os, unittest
+import os
 import tempo2_utils
 
 # Using Nanograv data B1855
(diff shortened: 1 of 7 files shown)


paulray:
Looks good, but I don't understand why Python 3 took away the `file()` initializer for a file object. Seems like a strange thing to get rid of.

examples/example_B1855.py
```diff
@@ -15,10 +13,10 @@
 timfile = os.path.join(datadir, 'B1855+09_NANOGrav_dfg+12.tim')
 
 # libstempo calculation
-print "tempo2 calculation"
+print("tempo2 calculation")
```
paulray: How does this work without an import of `print_function` in Python 2?

cdeil:
> I don't understand why Python 3 took away the file() initializer for a file object. 

I think it was just part of cleanup to get closer to the goal in the [zen of Python](https://www.python.org/dev/peps/pep-0020/#id3)

> There should be one-- and preferably only one --obvious way to do it.

They took away a few things that are duplicated (in this case `file` and `open`) and unneeded
(see https://docs.python.org/3.0/whatsnew/3.0.html#builtins), causing some pain now, but making Python a better language for the long-term future.

examples/example_B1855.py
cdeil: In Python 2, what happens is that the `()` are ignored as a redundant set of parentheses:
```
$ python2
>>> print('spam')
spam
>>> print('spam', 'ham')
('spam', 'ham')
>>> print('spam', )
('spam',)
>>> ('spam')
'spam'
```
I can add the `from __future__ import print_function` line at the top if you want.

PR #359, opened 2017-12-18
Add from __future__ import absolute_import

cdeil:
This PR adds `from __future__ import absolute_import` at the top of each file. This makes Python 2 behave the same as Python 3, which can prevent some problems.

Concretely, I think the only problem with imports in the current pint codebase I found when adding this was in `pint/templates/lctemplate.py`, which had this "implicit relative import" before
```
from lcnorm import NormAngles
from lcprimitives import *
```
which doesn't work on Python 3, and which I rewrote as an "explicit relative import"
```
from .lcnorm import NormAngles
from .lcprimitives import *
```
which works the same on Python 2 and 3, guaranteed in the way the import was intended (i.e. a package `lcnorm` or `lcprimitives` outside `pint` in `site-packages` won't break `pint`.

### DIFF
--- a/pint/config.py
+++ b/pint/config.py
@@ -1,7 +1,7 @@
 # config.py
 
 # Functions related to PINT configuration
-from __future__ import absolute_import, print_function
+from __future__ import absolute_import, print_function, division
 import os
 from .extern import appdirs
 
--- a/pint/config.py
+++ b/pint/config.py
@@ -1,7 +1,7 @@
 # config.py
 
 # Functions related to PINT configuration
-
+from __future__ import absolute_import, print_function
 import os
 from .extern import appdirs
 
--- a/pint/erfautils.py
+++ b/pint/erfautils.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import, print_function
+from __future__ import absolute_import, print_function, division
 from . import utils
 import numpy as np
 import astropy.units as u
--- a/pint/erfautils.py
+++ b/pint/erfautils.py
@@ -1,3 +1,4 @@
+from __future__ import absolute_import, print_function
 from . import utils
 import numpy as np
 import astropy.units as u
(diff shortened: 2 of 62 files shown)


paulray:
Why not include division in all the future imports?

paulray:
@kerrm wrote the lcnorm and lcprimitives, so this comment is just to notify him

cdeil:
> Why not include division in all the future imports?

Because if you rely on int division somewhere, that might introduce issues for your analyses running on Python 2, and I wanted this PR to be a simple / safe change.

Long-term (or now if you're up to it), I think adding the division future import to have consistent division behaviour is the good solution to have a Python 2 / 3 compatible codebase that does the same in both versions. Let me know if you want it, and I could include it in this or a follow-up PR.

paulray:
@cdeil thanks for all your work on this!  The PINT stated policy is to write compatible code, by importing future division: <https://github.com/nanograv/PINT/wiki/Python-Compatibility>
So, we should never be relying on integer division in the code and if we are it is a problem. I'd be in favor of importing division in all PINT codes.

kerrm:
I've always followed the C convention here, and nevermind my thoughts on
changing the default behavior of an elementary operator.  So I have used
integer division and  the code would need to be carefully checked if adding
the future division import.  That having been said, I doubt it has come up
often if at all in these modules...

Cheers,
Matthew

cdeil:
@paulray - so what to do? Should I add `division` and if tests pass it can go in? or keep that for a future PR, possibly after adding a few more tests to increase test coverage in pint?

Are you or any PINT users producing and publishing results with Python 3 already?
If yes, then adding in `from __future__ import division` seems more urgent, since if there are any issues with int vs float division, some users would be already affected now anyways.

paulray:
II think since we have a design requirement to support Python 3 that we should add it now, since currently we will have different behavior in Python 2 and 3 and that should not be allowed to persist. If changing division breaks the Python 2 code, we should find and fix that ASAP.

cdeil:
Done in 4cc47b0

paulray:
I think this can be merged.  @kerrm  and @luojing1211 do you agree?

kerrm:
I just did a quick check through the lc* files, eventstats, and
event_fitter, and I don't see any obvious problems where integer division
is needed.

PR #365, opened 2018-02-05
Fix for photonphase to avoid ephem=None bug

paulray:
@luojing1211 changed the code in `adjust_TOAs` to reference `self.ephem`. This fails if you create a `TOAs` object and have not yet run `compute_posvels`, since that is where `self.ephem` is set. This PR makes sure that photonphase calls `compute_posvels` before `adjust_TOAs`.

Probably there should be a fix to `adjust_TOAs()` to not try to "recompute" the derived columns if they were not there in the first place.

### DIFF
--- a/pint/scripts/photonphase.py
+++ b/pint/scripts/photonphase.py
@@ -95,14 +95,12 @@ def main(argv=None):
     if len(tl) == 0:
         log.error("No TOAs, exiting!")
         sys.exit(0)
-    ts = toa.TOAs(toalist=tl)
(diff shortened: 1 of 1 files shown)


paulray:
Thanks to @sguillot for pointing out this issue!

luojing1211:
I agree with the "no recompute" idea. Also, should we consider to make self.ephem a property? If the user changes the ephemeris, the posvel column will be recomputed.

paulray:
@luojing1211 that sounds like a good improvement. For now I'm just going to merge this PR since we need a quick fix to allow our NICER codes to run.

PR #367, opened 2018-02-05
utils.time_to_mjd_string: Add special case for pulsar_mjd format.

demorest:
This function converts an astropy `Time` to a high-precision MJD string.  It will now check the `Time` object format and use the "pulsar_mjd" convention for days with leap seconds when appropriate.


luojing1211:
Just tested it. It works.

PR #371, opened 2018-02-14
Require astropy<3.0

demorest:
The automated tests with python3 seem to be picking up astropy 3.0 now.  It seems PINT is not working with astropy 3, so I've put a requirement for astropy<3.0 into requirements.txt.  Putting in this PR to check if the tests pass after this change.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 numpy>=1.11.0
 Cython>=0.25.2
-astropy>=2.0 
+astropy>=2.0,<3.0
 scipy>=0.18.1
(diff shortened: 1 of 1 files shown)


demorest:
This seems to be passing.  Any objections to merging it in?  I've noted the larger issue in #372 .

paulray:
I'm surprised that astropy bumped their major version so soon after going to 2.0. Was there some major API change they needed?  We do need to make sure that PINT is compatible with 3.0 sometime soon, but we can't break backwards compatibility with 2.0.

But I'm fine with merging.

PR #377, opened 2018-03-07
A few updates to photonphase

paulray:
This is a few small changes to photonphase, including removing the --fix option that was a hack for NICER data that is no longer needed.

### DIFF
--- a/pint/models/spindown.py
+++ b/pint/models/spindown.py
@@ -29,9 +29,6 @@ class Spindown(PhaseComponent):
                        unit_template=self.F_unit,
                        description_template=self.F_description,
                        type_match='float', long_double=True))
-        self.add_param(p.MJDParameter(name="TZRMJD",
-                       description="Reference epoch for phase = 0.0",
-                       time_scale='tdb'))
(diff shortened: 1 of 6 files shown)


paulray:
The PR is failing the travis-ci tests with the following error

```
======================================================================
ERROR: test_earth (test_solar_system_body.TestSolarSystemDynamic)
----------------------------------------------------------------------
  File "/home/travis/build/nanograv/PINT/tests/test_solar_system_body.py", line 31, in test_earth
    a = objPosVel_wrt_SSB('earth', self.tdb_time ,ephem = ep)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pint/solar_system_ephemerides.py", line 126, in objPosVel_wrt_SSB
    raise ValueError("Can not load the ephemeris file '%s.bsp'. " % ephem)
ValueError: Can not load the ephemeris file 'de434.bsp'. 
```

I tried these tests on my machine and it works fine. Also, I checked and de434.bsp is still available on the JPL FTP site. My guess is that something changed in astropy maybe? Any ideas?

PR #389, opened 2018-05-05
Added a function to utils.py 

hessej:
Added a function to utils.py that I made of the Pulsar signal simulator. It returns the parameters of a par file given a pulsar name and a file path to where the file is located.


luojing1211:
Hi, @hessej if this specific functionality is meant for Pulsar signal simulator. I would recommend putting it in the Pulsar Signal Simulator package instead of PINT. Another issue is that this change only processes the NanoGrave data release files, however, a general program like PINT and PSS should accept all types of ".par" file.  In this sense, PINT get_model() is sufficient.

hessej:
hi @luojing1211, It looks like I mixed up where this function was suppose to go. I thought it was suppose to go here but I talked to Jeff and it is actually suppose to go is PSS_utils in the simulator. Sorry for the mix up!

Issue #381, opened 2018-03-08
Support for WAVE parameters


paulray:
PINT does not support WAVE parameters:

```
WARNING: Unrecognized parfile line 'WAVEEPOCH 58074' [pint.models.timing_model]
WARNING: Unrecognized parfile line 'WAVE_OM 0.18150803212366 0' [pint.models.timing_model]
WARNING: Unrecognized parfile line 'WAVE1 -2.3580237569833 0.61257000451779' [pint.models.timing_model]
WARNING: Unrecognized parfile line 'WAVE2 0.17435474955813 -0.096065138705814' [pint.models.timing_model]
WARNING: Unrecognized parfile line 'WAVE3 -0.020362229321081 0.019154948126028' [pint.models.timing_model]
WARNING: Unrecognized parfile line 'WAVE4 0.0019083164479827 -0.0031306122851833' [pint.models.timing_model]
WARNING: Unrecognized parfile line 'WAVE5 -6.5280719905701e-05 0.00028346991238931' [pint.models.timing_model]
```

This should be added...

paulray:
@luojing1211 I'm thinking of adding this model. Is there a HOWTO somewhere that give a step-by-step guide to adding a model to PINT? I know the structure is described in the paper, but a simple guide on the wiki would be great for people who want to contribute new models.

There is a tiny bit here: https://github.com/nanograv/PINT/wiki/Code-Architecture but it really needs more detail on what a programmer needs to define for a new model (e.g. derivatives and such).

luojing1211:
Hi Paul,

I was thinking to add this part. So how about give an example code, and we
provide the guide from the example code.

-Jing Luo

paulray:
Is this working now? Can we close?

aarchiba:
There is `test_wave.py`; is WAVE now sufficiently tested to close this bug?

matteobachetti:
@aarchiba I tried to use extensively the WAVE parameters. If I'm not mistaken, they work well if you just take them at face value, but _fitting_ them leads to inconsistent results

abhisrkckl:
Can this be closed?

kerrm:
I think the status is still the same -- WAVE parameters work fine for an
input model, but you can't fit them.

paulray:
Is the problem that there are 2 parameters for each WAVE?

kerrm:
I think that's the crux.  It's also a "prefix parameter", which makes
things a little more complicated, but the pair is the big problem.

[...]

PR #392, opened 2018-05-21
Add WAVE parameters

matteobachetti:
This PR adds WAVE parameters (Resolves Issue #381 ).
I had to invent a new kind of parameter, that takes a pair of floats.
I don't know if the logical steps I did are all correct, so I would like some feedback to get this right. I tried the WAVE function on a local notebook and it seems to be doing the right thing, but I'm not sure this is still the case after I plug it into PINT.


paulray:
@luojing1211 having WAVE parameters in PINT would be great. Can you take a look at this and see if it looks good, or if you have suggested changes before merging?

luojing1211:
@matteobachetti, Can you rebase your PINT branch against the PINT master branch?

matteobachetti:
@luojing1211 Sorry for the late reply, I'm on vacation. I will try asap

matteobachetti:
@luojing1211 rebased.

matteobachetti:
@paulray there are tests failing. one seems to imply that TOAs class doesn't have the `ntoas` attribute anymore? Others seem unrelated instead. Do you have pointers?

luojing1211:
TOAs class should have it. But in the old API, we are passing TOAs.table
into the delay or phase function. Perhaps we need to check if there is any
place using the old API.

-Jing

paulray:
Yes, this is all due to the changed API.
You should now call `delay()` with a TOAs object, rather than TOAs.table
In your wave_delay you'll need to use `toas.table['tdbld']` instead of just `toas['tdbld']`

Issue #391, opened 2018-05-10
Add Parkes format


dlakaplan:
I believe the code below successfully processes Parkes format TOAs:

Note that I think the line:
```
    elif re.match(r"  ", line) and len(line) > 41 and line[41] == '.':
```

Should be:
```
    elif re.match(r" ", line) and len(line) > 41 and line[41] == '.':
```
(One space in the `re.match()`).

And why does the toa_format method have an argument fmt?  It doesn’t seem like it’s used.

I think a couple of other changes are needed to allow Parkes format.  It should read:
```
    elif re.match(r" ", line) and len(line) > 41 and line[41] == '.':
        return "Parkes"
    elif len(line) > 80 or fmt == "Tempo2":
        return "Tempo2"
```
(Having tempo2 before Parkes didn’t work since with a `\n` the Parkes lines had length 81, so they were caught by the tempo2 format.  Maybe do `line.strip()` or similar instead?)

And then this seems to work:
```
    elif fmt == "Parkes":
        """
        columns         item
        1-1             Must be blank
        26-34           Observing frequency (MHz)
        35-55           TOA (decimal point must be in column 42)
        56-63           Phase offset (fraction of P0, added to TOA)
        64-71           TOA uncertainty
        80-80           Observatory (1-character)
        """
        
        d["name"] = line[1:25]
        d["freq"] = float(line[25:34])
        ii = line[34:41]
        ff = line[42:55]
        MJD = (int(ii), float("0."+ff))
        phaseoffset=float(line[55:62])
        if phaseoffset != 0:
            raise ValueError(
                "Cannot interpret Parkes format with phaseoffset=%f yet" % phaseoffset)
        d["error"] = float(line[63:71])
        d["obs"] = get_obs(line[79].upper())
```

Sorry that I don't know how to do this as a pull request.

Issue #384, opened 2018-03-13
Add features to zima


paulray:
Should be able to give a .tim file as input to zima and have it create fake TOAs with that time sampling.

Also should be able to 'fuzz' out the times even in the regular spacing case to be more like a real observing cadence.

paulray:
This is done, if more features for zima are needed, another issue should be generated.

Issue #361, opened 2018-01-12
Possible enhancement: adding the cosine-test for testing periodicity


msjacob:
Hello,

I am working for CTA (https://www.cta-observatory.org/), an observatory of very high energy gamma-rays. We are working on a candidate for the CTA Science Tools which is gammapy (http://gammapy.org/), a python package associated to astropy and numpy.

In PINT, there are already statistical tests on periodicity like the H-test (https://github.com/nanograv/PINT/blob/master/pint/eventstats.py). There could be another interesting statistical tool for testing periodicity.

It is the cosine-test, a powerful test developed in 1994 by O. C. De Jager in his paper "On periodicity tests and flux limit calculations for gamma-ray pulsars". It was developed for gamma pulsars and is especially useful in the context of sparse data at very high energy.
It is particularly powerful for single peaked phasograms, for which we have in advance an idea of the position and width of the peak. It provides a way to test the presence of a signal located at a given phase phi_0 and for a given harmonic m.

Do you think it would be useful to have this test included in PINT ? Within the gammapy project, we tend to think that it would be more relevant to add this test to PINT than to gammapy because the cosine-test is in principle suitable for any wavelength.
If you think that such a test is interesting for PINT, I could provide a python implementation for it.

Best regards,
Marion Spir-Jacob (APC laboratory, CNRS, France)

paulray:
I'd be interested in @kerrm's take on this

kerrm:
It sounds like a Z^2 test with only odd phases, which would be appropriate if you knew the peak would occur at phase 0 (or shifted your data so that would be the case) and was approximately symmetric.  You'd win in this case since your chi^2 distribution would have fewer degrees of freedom.

I'm not sure how practical it would be.  If ones has a good idea of what the peak should look like, it might be better to fit a model light curve and then do a null test.  But it would be a one-line modification to the existing Z^2_m routine, so pretty easy to implement.

paulray:
This is now merged as the `cosm()` function in event_stats.py, so I'm closing this.

Issue #290, opened 2017-03-29
Improvements to residuals.py


paulray:
Two places where the `resids()` class could be improved:
* `get_PSR_freq()` should not compute the frequency from pulling out F0 from the model, but should use `m.d_phase_d_toa()` like zima.py does.
* If you are working in ipython or a notebook, it would be nice to fiddle with the TOAs (e.g. clipping some out using the new selection API) or the model (e.g. manually changing a parameter value) and have the residuals update. Currently you have to build a new residuals class.  Should we add a `resids.update()` method that recomputes everything? Or do people prefer the current way, where resids are basically immutable?

paulray:
This has some overlap with issue #100

jcolen:
Was this issue solved with my July 2 commit?

paulray:
I think it is, except maybe we want to expose the modelF0 option. Currently there is no way to use it. Maybe it could be a keyword argument when initializing a residuals object, which would default to False.

aarchiba:
Maybe a separate bug for usability in the notebook? Immutable objects are not particularly a problem as long as creating them is convenient and not too expensive.

PR #397, opened 2018-06-25
Assorted Issues and Bug Fixes

### LINKED ISSUE #330: TDB times computed from spacecraft-based TOAs are wrong by a couple microseconds

jcolen:
The new correction factor in computeTDBs can be used with the method='astropy_corrected' argument. Should this be relabeled? 
earth_location_itrf now returns None for spacecraft
Correction factor was compared with the old results from test_photonphase and the pulse phases agree to within 10e-6, pulse times agree to within 10e-7s
These changes together address Issue #330 

Added cosine test function as requested by Issue #361 

Added zima features - input tim file and data fuzzing as requested by Issue #384 

Added Parkes format loading as requested by Issue #391 

Enhanced residuals.py as requested by Issue #290 
Note: the change to get_PSR_freq() very slightly increases the differences from Tempo2 and caused some tests to fail. The thresholds have been updated after discussion with @paulray . In addition, the change significantly slows down calls to resids.calc_time_resids(), so it will be useful in the future to find ways to speed up model.d_phase_d_toas()

### DIFF
--- a/pint/eventstats.py
+++ b/pint/eventstats.py
@@ -135,6 +135,21 @@ def z2mw(phases,weights,m=2):
 
     return np.cumsum(s) * (2./(weights**2).sum())
 
+def cosm(phases,m=2):
+    """ Return the cosine test for each harmonic up to the specified m.
+        See de Jager et al. 1994 for definition
+    """
+
+    phases = np.asarray(phases)*TWOPI   #phase in radians
+    n = len(phases)
+
+    if n < 5e3:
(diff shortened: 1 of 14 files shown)


paulray:
Should we go ahead and merge?

Issue #271, opened 2017-03-24
MCMC Fitter


paulray:
It would be nice for PINT to have a Fitter that does an MCMC optimization, like event_optimize does, but for normal TOA data.

jellis18:
@paulray, I can work on this over the next few weeks. I think we should have a basic MCMC fitter class that is then composed of the MCMC algorithm. So the MCMC algorithm would be a separate class. That way we could easily switch out different algorithms instead of coding the whole thing to use just one.

luojing1211:
@jellis18 The fitter class is setting up the same way. But it is a inheritance not a composition. The Basic MCMC fitter can be inherited from Fitter class.

jellis18:
@luojing1211, yes I know that but the MCMC algorithm shouldn't be built in. In other words we shouldn't have an `EmceeFitter` and `MultiNestFitter`, etc. We should have one `MCMCFitter` (or maybe something else if we want to include nested sampling) that has a set method for different MCMC samplers. To do this we would need a `Sampler` class that is subclassed to different methods. We don't have to go this way but from what I've found, this kind of setup is more flexible.

paulthebaker:
I'm going to bump this issue, rather than start a new one.  Another way would be to have PINT include `get_likelihood()` and `get_prior()` methods.  These could be fed to the sampler of the user's choice.  

This is the way `enterprise` handles things.

paulray:
Good timing! @jcolen is thinking about this now.

PR #399, opened 2018-07-05
MCMC fitter

jcolen:
Added modular MCMC Optimization Fitter as discussed in Issue #271 . Fitter subclass is in pint.mcmc_fitter and a Sampler class that wraps around emcee is in pint.sampler. Example usage is in pint.scripts.event_optimize_MCMCFitter as well as in examples/MCMC_walkthrough.ipynb


jcolen:
Just added/improved examples for MCMC as requested in the telecon. MCMC_walkthrough.ipynb is more clear about why the displayed results aren't great, and a sample python script is in examples/fit_NGC440E_MCMC.py. I added some uncertainties to the parfile so that priors could be set - this has no effect on the other example fit program.

PR #401, opened 2018-07-11
Tzr

luojing1211:
This pull request implements the TZR absolute phase. The unit test is not provided yet.

### DIFF
--- a/pint/models/timing_model.py
+++ b/pint/models/timing_model.py
@@ -489,7 +489,7 @@ class TimingModel(object):
         # If the absolute phase flag is on, use the TZR parameters to compute
         # the absolute phase.
         if abs_phase:
-            if 'AbsPhase' not in list(self.component.keys()):
+            if 'AbsPhase' not in list(self.components.keys()):
                 log.warning("No absolute phase (TZRMJD, ...) model provided." +
                             " Returning the conventional phase.")
                 return phase
--- a/pint/models/timing_model.py
+++ b/pint/models/timing_model.py
@@ -477,7 +477,7 @@ class TimingModel(object):
             delay += df(toas, delay)
         return delay
 
-    def phase(self, toas):
+    def phase(self, toas, abs_phase=False):
         """Return the model-predicted pulse phase for the given TOAs."""
         # First compute the delays to "pulsar time"
         delay = self.delay(toas)
@@ -485,7 +485,23 @@ class TimingModel(object):
         # Then compute the relevant pulse phases
         for pf in self.phase_funcs:
             phase += Phase(pf(toas, delay))
-        return phase
+
+        # If the absolute phase flag is on, use the TZR parameters to compute
+        # the absolute phase.
+        if abs_phase:
+            if 'AbsPhase' not in list(self.component.keys()):
+                log.warning("No absolute phase (TZRMJD, ...) model provided." +
+                            " Returning the conventional phase.")
+                return phase
+            else:
+                tz_toa = self.get_TZR_toa(toas)
+                tz_delay = self.delay(tz_toa)
+                tz_phase = Phase(np.zeros(len(toas)) , np.zeros(len(toas)))
+                for pf in self.phase_funcs:
+                    tz_phase += Phase(pf(tz_toa, tz_delay))
+                return phase - tz_phase
+        else:
+            return phase
 
     def covariance_matrix(self, toas):
         """This a function to get the TOA covariance matrix for noise models.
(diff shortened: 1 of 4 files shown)


luojing1211:
@scottransom @paulray Could you review this PR. @paulray do you have a simple test case for this model?

paulray:
I'll compare it to my little hack in photonphase and let you know.

pint/models/timing_model.py
```diff
@@ -477,15 +477,31 @@ def delay(self, toas, cutoff_component='', include_last=True):
...
         phase = Phase(np.zeros(len(toas)) , np.zeros(len(toas)))
         # Then compute the relevant pulse phases
         for pf in self.phase_funcs:
             phase += Phase(pf(toas, delay))
-        return phase
+
+        # If the absolute phase flag is on, use the TZR parameters to compute
+        # the absolute phase.
+        if abs_phase:
+            if 'AbsPhase' not in list(self.component.keys()):
```
paulray: This  should be self.components.keys()
(It is missing the s in components)

paulray: Also, I loaded a par file using pint.models.get_model().  Even though it has TZR parameters, it is not adding an AbsPhase component.  Here is what is has:
```
In [3]: m.components
Out[3]: 
{'AstrometryEquatorial': <pint.models.astrometry.AstrometryEquatorial at 0x10ac87c50>,
 'Dispersion': <pint.models.dispersion_model.Dispersion at 0x10ac7bf10>,
 'SolarSystemShapiro': <pint.models.solar_system_shapiro.SolarSystemShapiro at 0x10ac96650>,
 'Spindown': <pint.models.spindown.Spindown at 0x10ac74450>}
```

paulray:
Still some bugs.
```
  File "/Users/paulr/src/PINT/pint/models/__init__.py", line 16, in <module>
    from .absolute_phase import AbsPhase
  File "/Users/paulr/src/PINT/pint/models/absolute_phase.py", line 57
    tz = toa.get_TOAs_list([TZR_toa,], include_bipm=clkc_info{'include_bipm'},
```
I think the curly braces should be brackets.

paulray:
Next, I think that `abs_phase=True` should be the default if there are TZR parameters in the model (i.e. if the AbsPhase component exists).

luojing1211:
Thank you, I am fixing it. Is the overall structure ok?

luojing1211:
I just realized passing the TOA information (ephemeris, clock correction flag... ) into the AbsPhase class requires changing the API of phase(). Currently, phase() accepts the TOAs.table. However, the setup of this PR requires the TOAs class. Is there any other way of avoiding it?

paulray:
Yeah, I've never liked having to figure out which functions want TOAs and which want TOAs.table. Having to pass TOAs.table really breaks the encapsulation of the TOAs class.  But, short of reversing that everywhere, I'm not sure the best solution in this case.

luojing1211:
The alternative is to append the TZR toa to the TOAs table and recompute the TDB and positions for all the TOAs again.

paulray:
Computing TDBs and positions is expensive, so I wouldn't want to do that again for all the TOAs.
I think @demorest was an advocate for passing TOAs.table (and maybe @scottransom too, I'm not sure), but it does lose metadata, as you point out. Perhaps we could revisit that discussion?

scottransom:
What would be the problem with changing everything to accept the TOAs class rather than just passing the table?  I don't think it would slow things down, given that we could do something like tbl = toa.table on the first line of each function that uses the table currently.

demorest:
I don't remember what (if any) rationale there was for passing the table rather than the TOAs object.  It may just be due to the history of the code.. I think we were originally simply using array-valued TDB Times for these arguments.  That has a nice simplicity but at some point I guess we needed more info from the table.  I agree that using the TOAs class makes more sense and could allow us to work towards something where the table is "hidden" inside TOAs and never directly accessed by outside code.  This makes it a bit more modular in case we want to change how the info is stored inside TOAs.

paulray:
Other than an API change, are there any negatives to changing all routines to accept `TOAs` instead of `TOAs.table`?  Since it is an externally visible API change, it is definitely good to do before version 1.0 (and before the paper is finalized), if we are going to do it, so we don't have that major inconsistency floating around.

scottransom:
I can't think of any issues that would be caused by passing a TOAs class instance rather than the toa table.  And I do think it is cleaner to do that.  So I think we should probably just make the changes (and put that in the paper if necessary).

luojing1211:
Ok, if there is no conflict we can think of, I will just start to change it.

PR #402, opened 2018-07-12
Change delay/phase API from accepting TOAs.table to TOAs class

luojing1211:
This PR changes the delay and phase function API from accepting TOAs.table to a more generalized container class TOAs class.


paulray:
This has to be considered carefully since it will break any code that uses PINT models. They will need to change `model.phase(toas.table)` to `model.phase(toas)` in their code!
I can fix this in NICERsoft, but what about Enterprise, or any other codes that call PINT?  How about the example notebooks and such?

luojing1211:
I can fix the notebook examples. However, I do not know what other software
package is using PINT.  We should update the tag and announce it to the
community. Anyway, I think this should be done when there are not a lot of
packages depending on PINT.

scottransom:
Yeah, I don't think this is probably a very big deal since I don't think many codes are using PINT (yet!).

paulray:
Yeah, the sooner we make this API change the better.  I assume this PR is required to get the TZRMJD solution to work?

luojing1211:
Yes.

luojing1211:
Hi all, since all the checks are passed. I am merging this PR.

bsipocz:
This indeed broke some downstream usage, e.g. I run into it in a totally unrelated PR for stingray: https://travis-ci.org/StingraySoftware/stingray/jobs/414440535#L1337

luojing1211:
Hi @bsipocz 
A simple fix should do it:
Change line
stingray/pulse/pulsar.py:720: in get_orbital_correction_from_ephemeris_file
```    
delays = m.delay(toalist.table)
```
to 
```    
delays = m.delay(toalist)
```

Issue #343, opened 2017-09-19
Fail when loading a UNITS TCB ephemeris


kerrm:
PINT currently doesn't complain when loading an ephemeris with "UNITS TCB".  The timing model parameters will be then be interpreted incorrectly by PINT.  Until TCB is supported internally, it should check for UNITS TCB when loading a model and issue a fatal error if encountered.

paulray:
I agree that this is important.  PINT should definitely fail if it finds UNITS TCB in a par file!

paulray:
@jcolen can you add this check in get_model()? It should probably call log.error() saying that TCB units are not yet supported by PINT, and raise an exception.

Issue #296, opened 2017-04-06
Add support for pulse numbers


paulray:
Tempo2 uses  `TRACK -2` to indicate that the absolute pulse numbers for each TOA are specified in the tim file using the `-pn` flag.  PINT should support this kind of usage as well.

PR #403, opened 2018-07-12
UNITS TCB and Pulse numbering

jcolen:
Produces error message for UNITS TCB (Issue #343)

Supports pulse numbering with TRACK -2 in parfile and -pn in timfile (Issue #296)
Pulse number is added as a column in the TOAs table if -pn is a flag in the timfile. Pulse numbers can be computed from a model using TOAs.compute_pulse_numbers(model) and are now written out as flags in tim files.


paulray:
My initial tests of this look good!  All it needs is a unit test.

paulray:
Just to document what we discussed, the next steps on this are:

- Make pulse number be a column in the TOAs table, instead of just a flag. But, it should enforce that either the flag is there or the column, but not both (to prevent inconsistent states).

- Add a method to TOAs() to compute the pulse number column, given a timing model

- Update the TOA writing function to write out the pulse numbers as flags on the TOA lines

jcolen:
@paulray Those steps were addressed in the most recent commit

PR #404, opened 2018-07-13
Changes to allow multiple clock correction files.

kerrm:
This is a simple hack that allows a list of clock corrections to be specified for a single observatory, and does that for ncyobs.


paulray:
@kerrm the tests are failing on this. Can you check out why?

kerrm:
OK, I think it's fixed now!

Issue #411, opened 2018-08-13
Tests failing when trying to download ephemerides


paulray:
On Travis, and on my local machine, the tests are now failing when they try to download de434 and de436.  This was fixed previously by hosting them on the nanograv server, but this seems not to be working anymore.  For example, this fails from the command line:
```curl -O https://data.nanograv.org/static/data/ephem/de436.bsp```

@nategarver-daniels did something happen to that server that is keeping us from accessing it?

nategarver-daniels:
I need to update the security certificate. I'll do that today.

nategarver-daniels:
The security certificate is updated, I'm going to close this issue.

PR #412, opened 2018-08-13
Fix a bug in computing TDB times for barycentric TOAs 

paulray:
A bug was introduced in the speedup of TDB calculation.  For TOAs that were already at the SSB, it still tried to add the topocentric part of the TT to TDB conversion. This fix makes compute_TDBs not do anything for TOAs that are already on the TDB timescale.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ language: python
 matrix:
     include:
         - python: "2.7"
-        - python: "3.5"
         - python: "3.6"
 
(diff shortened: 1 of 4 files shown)


paulray:
As noted in Issue #411 there seems to be a problem downloading de434.bsp and de436.bsp. I have removed those ephemerides from `test_solar_system_body.py`. When that issue is fixed, we should restore them.

PR #414, opened 2018-08-21
Add Molonglo Observatory Synthesis Radio Telescope (MOST). The positi…

fjankowsk:
…on is the one of the central pillar determined in the 2012 geodetic survey, with the elevation measured by us.


paulray:
Looks good.  I think it would be great to add the details of where the position came from as a comment in the code.  Would be nice if all the observatories had that too...

PR #417, opened 2018-08-24
Tkinter based PINT GUI

jcolen:
A PINT GUI that imitates and extends qtip without needing additional dependencies.


paulray:
I encourage people to try this and post here any suggestions for improvement.

scottransom:
That's really awesome already.  Nice job!

An absolute must-have feature is the ability to zoom in and out of the residual plots.  And possibly to be able to select groups of points using the mouse as well.

But this is great!  If the only real changes that you've made to PINT are the additions of a few files that help with the GUI, I think this should be merged basically as-is.  It doesn't hurt anything, and it is already usable!

Very cool.

jcolen:
Added zooming and point selection, as well as a summary of keyboard commands in the help menu

scottransom:
Just as an FYI, the zooming still doesn't seem to work for me.

jcolen:
Sorry about that, I had zooming done in a normal way and then changed it after a discussion with @paulray and didn't document those changes here. I've made some more changes that should hopefully make things more clear.

Now, the navigation toolbar just changes the plot view. It does not crop or delete points from the TOAs object. If you want to crop points from the TOAs object to perform a fit on a selection, just click and drag on the plot (without clicking anything on the navigation toolbar). Instead of zooming or panning, it will highlight the selected points in orange. You can then interact with them using the keyboard commands listed when you press "h".

To see whether you're using the navigation toolbar, look at the status message in the upper right. If it says "zoom rect" or "pan/zoom", you are just changing the plot view. If it says nothing, you can crop, delete TOAs, etc.

PR #422, opened 2018-09-05
Ephemeris option in PINTk

jcolen:


paulray:
This gives the user the ability to specify an ephemeris when calling pintk. If none is specified on the command line, it will use the EPHEM value from the par file. This should work fine, but also serves as a reminder that we've never really sorted out the 'configuration' information issue with PINT. But, that is something to work on later...

PR #423, opened 2018-09-05
A few ephem fixes

paulray:
This PR cleans up a bit the use of the `ephem` argument in the `TOAs` class.  But, even with this, there are too many places where it is specified and it may be possible to use it inconsistently. 
I think it might be better to have `compute_TDBs` and `compute_posvels` NOT take ephem as an argument, but always use `self.ephem`.  Then, `self.ephem` could be set either at initialization by `get_TOAs()` or by a separate method like `TOAs.set_ephem()`, which would invalidate any previously-computed TDBs and posvels.  Thoughts? @luojing1211

### DIFF
--- a/pint/models/absolute_phase.py
+++ b/pint/models/absolute_phase.py
@@ -42,16 +42,18 @@ class AbsPhase(PhaseComponent):
             self.TZRMJD.time_scale = 'tdb'
             log.info("The TZRSITE is set at the solar system barycenter.")
 
-        if self.TZRFRQ.value is None:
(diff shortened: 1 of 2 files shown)


luojing1211:
That is definitely a lot cleaner. Should we do this to the clock correction flags as well?

paulray:
I think the clock corrections look OK for now. I don't see how to simplify them in the current framework.  What are you suggesting there?

paulray:
This also corrects #424

PR #428, opened 2018-09-12
Fixes for photonphase

### LINKED ISSUE #427: Bug in photonphase
I just noticed that the --barytime option in photonphase is wrong.  The code looks like this:
```
            tdbs = np.asarray([t.mjd for t in ts.table['tdb']])
            data_to_add['BARY_TIME'] = [tdbs,'D']
```
This is almost certainly not what was meant to happen. This gives the photon times converted to the TDB timescale, but *without* the solar system delays applied.
I think the correct answer would be to use the `get_barycentric_toas()` method from the timing model.
@kerrm do you concur?

paulray:
This fixes #427 so that the BARY_TIME column is now computed with the solar system delays included. It also fixes a bug in the code that added columns to the output FITS file. This worked for adding one column but failed when you tried to add multiple columns (e.g. PULSE_PHASE and BARY_TIME). This PR also reduces the verbosity in places by changing some `log.info()` to `log.debug()`

### DIFF
--- a/pint/event_toas.py
+++ b/pint/event_toas.py
@@ -65,12 +65,12 @@ def _get_columns_from_fits(hdu, cols):
 def _get_timesys_and_timeref(hdu):
     event_hdr = hdu.header
     timesys = event_hdr['TIMESYS']
-    log.info("TIMESYS {0}".format(timesys))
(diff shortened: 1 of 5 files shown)


paulray:
@tenoto This fixes the bug you found...

PR #436, opened 2018-10-10
spacecraft TOAs

paulray:
Added a spacecraft observatory class that is associated with spacecraft-topocentric TOAs. Such TOAs are identified by an observatory code STL_GEO. Each TOA line must have flags telx,tely,telz for the spacecraft GCRS position in km, and flags vx,vy,vz for the spacecraft GCRS velocity in km/s.


juliadeneva:
I fixed a couple of small things that were failing the checks.

PR #438, opened 2018-10-24
Fixed pintk bugs

juliadeneva:
Fixed 2 pintk bugs: empty tim file being written by clicking the 'Write tim' button under the residuals plot; and pintk reverting to the default DE421 ephemeris when 'Reset' was clicked instead of continuing to use the ephemeris specified in the par file.

### DIFF
--- a/pint/models/timing_model.py
+++ b/pint/models/timing_model.py
@@ -95,6 +95,8 @@ class TimingModel(object):
             description="Tracking Information"), '')
         self.add_param_from_top(strParameter(name="EPHEM",
             description="Ephemeris to use"), '')
+        self.add_param_from_top(strParameter(name="UNITS",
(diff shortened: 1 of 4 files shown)


paulray:
Can you add UNITS TDB to all par files written by TimingModel.as_parfile() as well?

juliadeneva:
Done, plus fixed a zima bug.

paulray:
Thanks for all these fixes!  For --fuzzdays, can you change the default back to 0.0 and just put the part where it applies the fuzz in an if statement `if args.fuzzdays > 0.0:`.  For some applications, adding a 0.01 day fuzz to the TOAs by default would not be good.

juliadeneva:
Ok, done

PR #442, opened 2018-11-08
pintk: Fix compatibility with latest matplotlib.

bastianbeischer:
NavigationToolbar2TkAgg was renamed to NavigationToolbar2Tk in matplotlib 3.0

### DIFF
--- a/pint/pintk/plk.py
+++ b/pint/pintk/plk.py
@@ -16,7 +16,11 @@ except ImportError:
     import tkinter.filedialog as tkFileDialog
     import tkinter.messagebox as tkMessageBox
 import matplotlib as mpl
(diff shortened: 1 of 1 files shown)


paulray:
Thanks for the fix! We really need to add pintk to the test suite...

bastianbeischer:
@paulray - Do you want me to add pintk to the test suite? I think a separate PR would be better. Also I have no experience with the PINT test suite...

paulray:
Hi, @bastianbeischer it would be great if you put in a PR with a test, but I really wrote that comment more as a reminder to me. I just haven't found time to do it.  The test could be very simple, along the lines of test_zima.py, just to show that it runs and imports all that it needs without crashing. That kind of thing would have caught this.

PR #445, opened 2018-11-21
Add nustar

matteobachetti:
I made a simple verification of the barycentering, by comparing the results with the official results by`barycorr`. There is a difference of about ~1 us, which is tiny compared to other problems with NuSTAR's clock.


paulray:
Looks great! Thanks for the checks!

PR #451, opened 2018-12-10
Fix polyco due to an API update

luojing1211:
This is a fix for the polyco generate code due to the API update.


paulray:
Looks good to me!

PR #459, opened 2018-12-22
add **kwarg to get_toas()

luojing1211:
This PR enables the **kwarg functionality of get_toas() function.

### DIFF
--- a/pint/toa.py
+++ b/pint/toa.py
@@ -32,7 +32,7 @@ JD_MJD = 2400000.5
 
 def get_TOAs(timfile, ephem=None, include_bipm=True, bipm_version='BIPM2015',
              include_gps=True, planets=False, usepickle=False,
-             tdb_method="default"):
+             tdb_method="default", **kwargs):
     """Convenience function to load and prepare TOAs for PINT use.
 
     Loads TOAs from a '.tim' file, applies clock corrections, computes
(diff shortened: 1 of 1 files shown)


luojing1211:
@scottransom @demorest @paulray, Do you think this PR will create any problems?

paulray:
What does it do?  Are the kwargs used somehow?

luojing1211:
It allows the users to pass the optional parameters as a dictionary.

demorest:
You don't need `kwargs` to pass optional parameters via a dictionary.  You can use the `**` operator, for example this works fine in current PINT:
```
params = {'ephem':'DE435', 'include_gps':False}
toas = pint.toa.get_TOAs('J0613-sim.tim',**params)
```
What `kwargs` does is let you pass keyword params with names that are not in the list of standard parameters.  Without `kwargs` this raises a "got an unexpected keyword argument" error.

I don't really understand what this PR is intended to accomplish.  Jing, can you describe a bit more?  Unless we have a specific need for it, I guess it seems better to me to not use `kwargs` here because it will suppress errors in cases where there is a typo in a paramer name, etc.

luojing1211:
I did not know the **operator can be used like that way. I guess my problem is solved and close the PR.

PR #460, opened 2018-12-22
Polyco fix

luojing1211:
This PR fixes the polyco phase unit issues


luojing1211:
I think this is ready to merge. Adding Polyco test will be in another PR.

PR #462, opened 2019-01-10
Astrop3.1 compatibility

luojing1211:
Astropy 3.1's time format has API changes. This pull request allows the pulsar_mjd work with astropy3.1


luojing1211:
@demorest, Could you double check if this is ok to merge?

PR #464, opened 2019-01-27
Wideband TOAs

luojing1211:
This is the initial pull request for Wideband TOAs.

### DIFF
--- a/pint/models/dispersion_model.py
+++ b/pint/models/dispersion_model.py
@@ -25,7 +25,11 @@ class Dispersion(DelayComponent):
     """
     def __init__(self):
         super(Dispersion, self).__init__()
+        self.add_param(p.boolParameter(name='DMDATA', value=0,
+                                       description="Flag for using the DM data"
+                                       " from the Wideband TOAs."))
         self.dm_value_funcs = []
+        self.dm_value_derivs = {}
 
     def dispersion_time_delay(self, DM, freq):
         """Return the dispersion time delay for a set of frequency.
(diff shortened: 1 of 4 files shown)


paulray:
What's the status on this PR? Can PINT process wideband TOAs?

pennucci:
I'm not sure, @luojing1211 should comment, as it's quite old.  The good news is that I'm pretty sure the implementations we have in enterprise and tempo will offer a reliable way to cross check the results from PINT.

paulray:
We have not discussed this PR in a long time and now it has accumulated conflicts. @luojing1211 do you plan to get it working so we can merge?

pennucci:
I have no idea about the status, but at some point soon this will get 
picked up again, since it will be critical for the 15-year data set.

luojing1211:
I will clear the conflict. However, we would have to make wideband fitting work. This will be my next PINT goal.

paulray:
I'm kind of a fan of test-driven development. Tim, can you put in a simple set of .par/.tim files in the examples directory that work with TEMPO/Tempo2, then we can write a PINT script that should replicate that and we'll know when it works...

pennucci:
Just to be clear; you want me to open a PR in which I added a set of WB 
.par/.tim files that work with tempo?  Tempo2 has no knowledge of the WB 
DM measurements and, in fact, my tempo changes have not yet been merged, 
so the tests would have to use a clone of my tempo repo.

paulray:
Ah, so no standard code yet processes wideband TOAs?  OK.
Maybe just send them to Jing and he can add them to his PR. Or maybe he already has an example set of files that he is working with?

pennucci:
Indeed, the merge is still waiting to happen.  I believe Jing has some 
sample files.  But there are many more.

luojing1211:
@pennucci I think a set of small test .tim and .par files will not hurt, and it will be great if you can also send me the result from your code as some text files. So that we can validate the PINT results without installing tempo.

scottransom:
@pennucci Are you saying that TEMPO can't properly process WB TOAs now?  I thought it could.  If not, that needs merged ASAP (in TEMPO).  Should I ping David Nice or @demorest about this?  Or are there still issues that need handled?

demorest:
I think I said I'd merge Tim's tempo updates but haven't done it yet..

luojing1211:
This is a question about the Wideband API. In the TEMPO scheme, we have a flag in the par file, DMDATA. For PINT, we can read the DM values for .tim file without any problem. But do we want to have a separate fitter, like Wideband TOA fitter for it?

pennucci:
Technically tempo reads the DM values from the .tim values without a 
problem, too.  DMDATA is a switch/toggle for /using/ those DM 
measurements in the likelihood, so I think you'll need an analogous 
toggle in PINT, e.g., use_wideband = True.  Strictly speaking, the 
wideband likelihood (at least the one we developed) is a generalized 
form of the timing likelihood.  So a wholly separate code structure may 
not be needed, but that's a choice for the coders.

luojing1211:
This is my understanding of wideband toa fitting. 

DM values from .tim file are like TOAs. Param_nonDM are the parameters that are not related to the DM model (e.g., F0, F1,..) and Param_DM parameters are the parameters related to the DM model (e.g., DMX_0001), so I have this plot

wideband_fitting

Is this interpretation correct?

luojing1211:
@pennucci @scottransom @demorest  Is the interpretation above correct?

pennucci:
Sorry for the delay.  I was hoping others might chime in.  This looks 
reasonable to me.  I'd have to look more closely at the tempo code to 
see if this is how it's implemented there.  In enterprise, the design 
matrix is not modified, but rather the DM data points serve to contrain 
the priors on DMX params.  Here, I would think all 
d_dm_value/d_param_non_DM(TOA) terms would be zero (or inf?), and 
d_dm_value/d_param_DM(TOA) are finite only if the DM data correspond to 
that DMX bin/value.  The left side of the design matrix, I think stays 
the same as usual.

luojing1211:
@pennucci How do you compute the degree of freedom? Do you count the DM values as degree of freedom?

pennucci:
@luojing1211 Yes, they should be counted!

luojing1211:
This PR is continued in the PR #699

PR #466, opened 2019-02-06
Add ground-based gravitational-wave observatory locations

mattpitkin:
I'm thinking about using the PINT barycentring routines in a Python-based GW search code, so it would be useful for PINT to contain the locations of the current ground-based gravitational-wave observatories.


paulray:
Looks good to me! I'm not sure why some of the travis builds are timing out.

luojing1211:
I restart the building and it passed tests now.
-Jing

paulray:
I'm a little dubious about using 'geo' as an alias. That could easily get confused with the geocenter. Could we remove that one alias?

PR #485, opened 2019-07-30
Update Fermi FT2 handling

paulray:
This PR does a couple small things:
* Uses the new SC_VELOCITY column in Fermi FT2 files (if present), instead of trying to compute the velocities as the gradient of the position.
* Adds a track mode in LCFitters.fit_position() to handle cases where you have two peaks at 0.5 in phase apart.
* A couple of tiny fixes like deprecated "normed" keyword and stuff like that.


paulray:
This fixes a travis issue as well, that was causing builds to fail.

PR #488, opened 2019-08-07
Timing utils

luojing1211:
This PR adds the functionality of adding/deleting parameters via building new timing model from the parfile lines (a list of par file lines)


aarchiba:
I think it's worth mentioning that this PR also introduces the ability to easily do F tests!

aarchiba:
Is rebuilding the par file from lines really the best we can do for modifying model objects? Wouldn't it be nice to be able to add and remove parameters from a model using methods of the model object?

Alternatively, for the F test, might it be more appropriate to freeze and unfreeze model parameters? It doesn't permit comparison of models with non-nested parameter sets - you can ask whether F2 or proper motion provide statistically significant improvements over the base model, but not whether one is better than the other. At least not with the F test. Bayesian and Bayesian-adjacent model comparison techniques do exist, so alternatives other than freezing and unfreezing parameters are valuable.

pint/models/model_builder.py
```diff
@@ -70,8 +70,15 @@ def preprocess_parfile(self, parfile):
         """
         param = {}
         repeat_par = {}
-        pfile = open(parfile, 'r')
-        for l in [pl.strip() for pl in pfile.readlines()]:
+        if isinstance(parfile, str):
+            pfile = open(parfile, 'r')
+            plines = pfile.readlines()
+        elif isinstance(parfile, (list, tuple)):
```
aarchiba: Lists and tuples are not the only iterable types. Perhaps astropy's isiterable, or just try doing a for loop over it and catch the exception?

pint/models/model_builder.py
```diff
@@ -70,8 +70,15 @@ def preprocess_parfile(self, parfile):
         """
         param = {}
         repeat_par = {}
-        pfile = open(parfile, 'r')
-        for l in [pl.strip() for pl in pfile.readlines()]:
+        if isinstance(parfile, str):
+            pfile = open(parfile, 'r')
+            plines = pfile.readlines()
```
aarchiba: Do you still need the file object after this? Why not just open it in a with statement and pull out the lines?

pint/models/model_builder.py
```diff
@@ -90,7 +97,8 @@ def preprocess_parfile(self, parfile):
         self.param_inparF = param
         for key in repeat_par.keys():
             self.param_inparF[key + str(1)] = self.param_inparF.pop(key)
-        pfile.close()
+        if isinstance(parfile, str):
+            pfile.close()
```
aarchiba: Closing files that were opened for reading is not essential, which is good because if an exception is raised this never gets called.

pint/models/model_builder.py
```diff
@@ -70,8 +70,15 @@ def preprocess_parfile(self, parfile):
         """
         param = {}
         repeat_par = {}
-        pfile = open(parfile, 'r')
-        for l in [pl.strip() for pl in pfile.readlines()]:
+        if isinstance(parfile, str):
```
aarchiba: The usual python custom is to accept a filename or a file-like object; this is easily added here. In fact, iterating over a file-like object will yield its lines, so this falls directly into the case below.

pint/models/model_builder.py
```diff
@@ -241,16 +249,12 @@ def get_model(parfile):
     """A one step function to build model from a parfile
         Parameters
         ---------
-        name : str
-            Name for the model.
-        parfile : str
-            The parfile name
+        parfile : str or str list
```
aarchiba: "list of str" is the standard way of specifying this.

pint/models/timing_model.py
```diff
@@ -162,6 +163,22 @@ def components(self,):
                 comps[cp.__class__.__name__] = cp
         return comps
 
+    @property
+    def categories(self):
+        categories = []
+        if six.PY2:
+            type_list = super(TimingModel, self).__getattribute__('component_types')
```
aarchiba: is __getattribute__ really needed here? It's *extremely* arcane. Shouldn't there be a better interface to getting the list of component types?

pint/models/timing_model.py
```diff
@@ -774,14 +792,42 @@ def designmatrix(self, toas, acc_delay=None, scale_by_F0=True, \
...
+        elif isinstance(parfile, (list, tuple)):
+            plines = parfile
+        else:
+            raise ValueError("Acceptable .par file format: `filename` and "
+                             "`.par file line list.`")
+        self.read_parfile_string(plines)
+        if isinstance(parfile, str):
+            pfile.close()
+
+    def read_parfile_string(self, par_strings):
```
aarchiba: Is there any need for a separate function here rather than just making read_parfile accept both types?

pint/models/timing_model.py
```diff
@@ -774,14 +792,42 @@ def designmatrix(self, toas, acc_delay=None, scale_by_F0=True, \
...
+           par_strings: list
+               List of parfile line strings.
+        """
         checked_param = []
         repeat_param = {}
         param_map = self.get_params_mapping()
         comps = self.components
-        pfile = open(filename, 'r')
-        for l in [pl.strip() for pl in pfile.readlines()]:
+        for l in [pl.strip() for pl in par_strings]:
```
aarchiba: I recommend (writing and) using a generator that yields only non-comment non-blank stripped lines. I bet this isn't the only place we'd use such a thing. Particularly if the comment marker was a parameter.

pint/timing_utils/core.py
```diff
@@ -0,0 +1,134 @@
...
+        self.model = deepcopy(self.input_model)
+
+    def save_changes(self):
+        # TODO add check if the model is saved
+        self.model_stack.append(deepcopy(self.model))
+
+    def undo_changes(self):
+        self.model = self.model_stack.pop(-1)
+
+    def add_params(self, params):
```
aarchiba: Why not operate by freezing parameters in the largest model in use?

pint/timing_utils/finalize_timing
```diff
@@ -0,0 +1,1212 @@
+# #! /usr/bin/env python
+#
+# # Script to call when finalizing timing for nanograv pulsars
```
aarchiba: Why is this file being checked in all commented out?

pint/timing_utils/statistics.py
```diff
@@ -0,0 +1,50 @@
...
+
+       Return
+       ------
+       The probability that the improvement in chi2 is due to random chance
+       (i.e. a low probability means that the new fit is quantitatively better,
+       while a value near 1 means that the new model should likely be rejected).
+       If the new model has a higher chi^2 than the original model, returns
+       value is -1.
+
+       Note
```
aarchiba: Should probably also mention that the models must be *nested*, that is, that model 1 must be obtainable from model 2 by freezing a certain number of the parameters of model 2. In fact it might make sense to replace one of the parameters dof_1, dof_2 by frozen_parameters to encourage users to be aware of this restriction.

pint/timing_utils/timing_setup.py
```diff
@@ -0,0 +1,32 @@
...
+# """
+#
+# from .toa import get_TOAs
+# from .models import get_model
+#
+#
+#
+# def check_components(model, required_components=['astrometry', 'spindown']):
+#     """Check if a timing model has sufficent components for high precision
+#     pulsar timing.
```
aarchiba: Better to explain what this does and doesn't do. For example, is a binary model required? What about a DM model? Anything to do with the Solar System? I can do high-precision timing with nothing but a spin-down model if I'm using already-barycentered X-ray data on an isolated pulsar. What about noise components?

pint/models/model_builder.py
luojing1211: > I think the ability to easily run F tests is very useful, as is the ability to take par files apart and put them together differently, but wouldn't it be simpler to just operate by freezing one or more parameters at their best values?

That would be easier. However, since the model (e.g., the binary model) is selected dependent on the parameters that are given, rebuilding the model ensures that PINT uses the proper model.

aarchiba: Well, that's sort of the problem: Because model selection is so opaque, how can you be sure the F test is valid if you rebuilt your TimingModel from scratch? If the model with the new parameters doesn't specialize to the one you started with when you fix certain parameters, the F test will give nonsense results. If the larger model does specialize correctly, then freezing the parameters will give you exactly what you expect.

If you're unsure whether an astrometry model that permits proper motion but with the proper motion fixed to zero is the same as a model with no proper motion, then you had better not apply the F test!

Issue #491, opened 2019-08-08
Ability to change PEPOCH, etc.


mtlam:
Given a new epoch, the code should be able to change the spin parameters accordingly. Similar functionality should be done for POSEPOCH and probably also DMEPOCH.

luojing1211:
Just a short double-check:
Do you mean a function to change the related parameters when changing the epoch, based on the parameter value new?

mtlam:
Yes. So for PEPOCH, if you change a new one, it should set a new F0, F1, etc. (including higher-order terms if available). DMEPOCH I guess does the same but with DM derivatives (DM001). And then POSEPOCH needs to rotate things according to the binary model.

So my thought is that there should be a function that would work like `model.change_epoch(57000)` and then the output would be the new parameters and PEPOCH, etc., set to 57000.

luojing1211:
I have a question about the epoch time scale. For instance, PEOCH we have:
```
phase(t) = P0 + F0 * (t - PEPOCH - delay(t)) + 1/2 * F1 * (t - PEPOCH - delay(t)) ^2 
nu0(t) = d phase(t) /d (t - PEPOCH - delay(t)) = F0 +  F1 * (t - PEPOCH - delay(t)) 
``` 
If we change PEPOCH to PEPOCH', the new F0, let us call it F0', would be: 
```
F0' = nu0(PEPOCH') =  F0 + F1 * (PEPOCH' - PEPOCH - delay(PEPOCH')) + ...
```
My question is what the the time scale of PEPOCH' ? Is it a time at one observatory. If yes, which observatory should we choose? This is important for delay(PEPOCH') term.

paulray:
Since it is evaluating the phase model, it must be MJD(TDB) transferred to the pulsar via all the delay terms, right?  If we ever support TCB, then it would be that instead.

scottransom:
The "etc" also should include binary orbital phase (T0 or TASC updating)

rossjjennings:
@luojing1211's current implementation of this for for PEPOCH doesn't work on a model object without TOAs. For example, if I run

```
import pint
from pint import toa, models, residuals, fitter
from astropy.time import Time

model = models.get_model('NANOGrav_12y/par/J1713+0747_NANOGrav_12yv2.gls.par')
t0 = Time(56000, scale='tdb', format='mjd')
model.change_pepoch(t0)
```
I get the following error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
[...]

PR #496, opened 2019-08-14
Move pepoch

luojing1211:
This PR adds the functionality of change pepoch (issue #491 ). It will change the related parameters: F0, F1, F2... as well. 
Here is a preliminary test of function.  A unittest will be added in the new commitment. 
```
m = get_model("J0023+0923_NANOGrav_11yv0.gls.par")
t = get_TOAs("J0023+0923_NANOGrav_11yv0.tim")
max_diff = []
epochs = np.linspace(40000, 70000, 1000)

for epoch in epochs:
    m2 = deepcopy(m)
    m2.change_pepoch(epoch, t, m2.delay(t))
    f2 = WlsFitter(t, m2)
    diff = f2.resids.time_resids - f.resids.time_resids
    max_diff.append(np.abs(diff).max())
``` 
The original PEPOCH is 56567.0
The new pepochs vs max residual difference are plot below:
epoch_diff

### DIFF
--- a/pint/models/spindown.py
+++ b/pint/models/spindown.py
@@ -92,7 +92,7 @@ class Spindown(PhaseComponent):
         else:
             phsepoch_ld = time_to_longdouble(self.PEPOCH.quantity)
 
-        dt = (tbl['tdbld'] - phsepoch_ld)*u.day - delay
+        dt = (tbl['tdbld'] - phsepoch_ld) * u.day - delay
 
         return dt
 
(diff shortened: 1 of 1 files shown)


mtlam:
Seems like a reasonable tolerance to me.

scottransom:
Just to make sure that I understand...  the plot is showing that the new resids, after PEPOCH have changed, are consistent with the original ones to a couple ns, right?  If so, that seems fine to me.

luojing1211:
That is the case. Those are the maximum absolute difference between the residuals before and after changing pepoch.

mtlam:
Even less, 0.2 ns, which I thought was totally reasonable.

PR #498, opened 2019-08-27
WIP: Packaging

aarchiba:
This branch is meant to make PINT behave better with respect to python's (messy) packaging environment, so that it can be more reliably built, distributed, and installed. This includes making it work with "tox", a tool to systematically run the entire test suite in fresh virtualenvs for all supported versions of python.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 include versioneer.py
 include pint/extern/_version.py
+include pyproject.toml
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,8 @@
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = [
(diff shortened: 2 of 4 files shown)


aarchiba:
I have to admit I don't understand what's failing here.

On 2.7 and 3.6 it seems like maybe it's having trouble getting the support files? I don't think anything I did should have affected that.
On 3.5 it seems like there's a bogus numpy version problem - maybe a version conflict between a pre-installed numpy (no longer needed but presumably in the travis setup) and the one pip installs?

aarchiba:
Incidtentally with the tox-travis plugin tox can handle producing clean installs that have exactly the packages specified in the requirements files at each stage (e.g. it knows you need numpy to even run setup.py). This simulates installation on people's strange setups.

PR #500, opened 2019-08-27
Warnings cleanup

aarchiba:
Reduce the chatter during testing and actual use by ignoring known-unknown par file entries and cleaning up FutureWarnings. 

Ideally one would capture warnings as part of the testing process and show them only on failed tests; one could also write tests to ensure warnings are correctly emitted when, say, typos appear in par files. Unfortunately astropy's log mechanism defeats nosetests' attempts to manage log output and I haven't yet figured out how to make it behave.


aarchiba:
Okay, these Travis failures are definitely not my fault. It can't find ephemeris files?

Issue #191, opened 2016-11-26
Don't download datafiles in setup.py


cdeil:
You're currently doing downloading data files in `setup.py`:
https://github.com/nanograv/PINT/blob/master/setup.py#L43
```python
# TODO: need to check if datafile have all the datafile inplace. We do not want
# to download those files everytime we reinstall PINT
# Download data files
data_urls = [
        "ftp://ssd.jpl.nasa.gov/pub/eph/planets/bsp/de405.bsp",
        "ftp://ssd.jpl.nasa.gov/pub/eph/planets/bsp/de421.bsp",
        "ftp://ssd.jpl.nasa.gov/pub/eph/planets/bsp/de430t.bsp",
        "http://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/earth_latest_high_prec.bpc",
        "http://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/pck00010.tpc",
        "http://naif.jpl.nasa.gov/pub/naif/generic_kernels/lsk/naif0012.tls",
        "http://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/de-403-masses.tpc"
        ]
data_files = []
data_dir = 'datafiles'
for u in data_urls:
    os.system("wget -N -c -P pint/%s %s" % (data_dir, u))
    data_files.append(os.path.join(data_dir,u.split('/')[-1]))
```

Like the `TODO` comment says, that's no good, and I'd like to help implement something better.

The question is how exactly to do it.

How about using a shell environment variable `PINT_DATA` instead that the user has to set, that points to the directory where the files shall be located.
And then to provide a `download_data` function in `pint`, and load data via `PINT_DATA` from examples and the tests?

That's what we do in Gammapy for now. It has the advantage that it decouples the data from your software (`setup.py`), so it can also be done e.g. if you publish pint on PyPI some day and someone wants to use the 200 MB of data files with an installation coming from there.

Thoughts?

demorest:
Hi @cdeil , we purposefully removed a PINT environment variable a while back, I don't think we want to put it back in (at least, I don't).  For background see discussion in #115 .  These are a few different types of data PINT needs.  Some (like IERS data) are always being updated and we use the astropy download system for these.  The ones currently in setup.py should only need to be downloaded once.  Changing the download scheme for these would be fine, but I think it would be good to have a system-wide install (rather than per-user) by default.

paulray:
[...]

Issue #383, opened 2018-03-13
A way to pre-download data files for offline work


paulray:
We should provide a script or wiki page with instructions for how to download data files that are needed, so that astropy won't need to download them. This will allow people to work offline or from machines that can't download from JPL for whatever reason.

NihanPol:
Hi,
I am trying to run Enterprise using PINT on a remote cluster which does not have an internet connection. As a result, PINT cannot install the IERS files using astropy. I would really appreciate a way to be able to download these ancillary files with the installation of PINT with pip, if that's possible.

aarchiba:
Some files are cached by astropy, others (I think) by PINT directly, and those will be in different places. Unfortunately the set of all ephemeris files is a bit big for direct inclusion in github, although Git Large File Service might be workable. But the easiest would be to write a script that downloads all(?) up-to-date ephemerides to cache, another that does this and creates a package, a third that installs such a package, and finally a way to tell astropy and PINT not to worry if they can't see the internet to check for updates. This is not as complicated as it sounds, and astropy should have a similar thing for the same reason.

I think, if we were ever to support conda builds, I think they are more able to accommodate large binary blobs like this as part of a package. This would spare users having to download these large files on first use.

aarchiba:
Incidentally solving this problem might also resolve some erratic Travis failures like those that afflict PR #500 .

aarchiba:
Let me draw your attention to the astropy pull request:

https://github.com/astropy/astropy/pull/9182

It should be easy to write up a script that clears the astropy cache (or not), runs our test suite, and dumps out both the URLs needed and a ZIP file that can be loaded into astropy on another machine. This should cover most of the file requests from astropy; you can of course trigger any additional requests you like before dumping out the ZIP file.

Now, PINT's need for external data occurs only the first time you run setup.py on a fresh git clone (as far as I can tell). So that shouldn't be what's giving you headaches on airplanes.

paulray:
I think it checks and downloads IERS files if they have been updated, whenever you do a coordinate system conversion that needs the Earth rotation.

aarchiba:
Does it always check them? I thought that the data for times (far enough?) in the past never changed, so there would be no reason to update the IERS data unless you were working with a time that was more recent than the last update. I know there are temporary values for the time since the last release and the near future, but those could be kept separately.

aarchiba:
[...]

Issue #376, opened 2018-03-07
Could get_TOAs() be faster?


xuhengdada:
This may costs several minutes.

Forming the TOAs from a .tim file requires lots of time. The reason for this may be using the functions in astropy.coordinates to produce the barycentric positions and velocities.

So is there any methods to make this process faster? Writing C code to achieve this? 

It takes little time in tempo2 to form BATs.

paulray:
Can you provide an example .tim file and how long it is taking to load?

xuhenghb:
oh yeah, loading the NGC6440E.tim file in the examples files of PINT costs about 241 seconds. 

I also test other .tim file, the B1855+09 NANOGrav 11yr timing data, which has 5618 toas, and this takes about 249s. 

So the number of ToAs shouldn’t be the reason.

luojing1211:
That is wried, I have never experienced this before. Can you profile the get_TOAs() method?

scottransom:
yeah...  that is some kind of strange issue with your system I would think.  Loading the NGC6440E.tim file on my machine is nearly instantaneous!  I just ran "python examples/fit_NGC6440E.py" and the fill script, including starting python, reading the TOAs, doing the fit, showing the results, and making the plots, only takes about 2 sec total.

paulray:
Yes, for me files of that size are plenty fast.
I am running into issues loading millions of TOAs when processing photon data on bright sources. This does finish as long as I run on a machine with enough memory, but it could certainly use some speedup.  One issue that could be fixed is that a TOA() object is instantiated for every single photon, then destroyed once the TOAs.table is created.  It would be better to have a shortcut to make the table directly in that case.

xuhenghb:
my machine is mac. Just now I run get_TOA() on a linux machine, same problem……sigh

Here is profile result. the result is amazing, the real time looks small, but the actual time is so long, I'm totally confused!

## Profile Result:
INFO: Applying clock corrections (include_GPS = True, include_BIPM = True. [pint.toa]
INFO: Evaluating observatory clock corrections. [pint.observatory.topo_obs]
INFO: Applying GPS to UTC clock correction (~few nanoseconds) [pint.observatory.topo_obs]
INFO: Applying TT(TAI) to TT(BIPM) clock correction (~27 us) [pint.observatory.topo_obs]
INFO: Computing TDB columns. [pint.toa]
INFO: Computing positions and velocities of observatories and Earth (planets = False), using DE421 ephemeris [pint.toa]
[...]

PR #502, opened 2019-08-29
Clean up handling of downloaded files

aarchiba:
It turns out none of the data files we were downloading were used for anything. I have removed them and the downloading machinery. In future any files that do not change can be downloaded via `astropy.utils.download_file(url, cache=True)`, which caches them persistently and system-wide. Files that are updated remain an unsolved problem (both with and without this PR). The best current option is probably `astropy.utils.download_file(url, cache=False)`, which downloads them once per session.

Old message was:

The most important improvement here is that setup.py will actually stop, signal an error, if the data files are missing and it can't install them. This should produce a fast clean failure when trying to install  the package or produce a release rather than mysterious errors somewhere down the line.

The list of files and their MD5 hashes is also now a text file rather than being buried in the middle of setup.py, so other tools can get at them, in case someone wanted to write a script to download them.

If the Travis failures are related to a specific server, or if we have users who can't access US military servers, this could be made to allow a list of mirrors for each one; doesn't matter where they get it as long as the MD5 sum matches. (Well, okay, someone could create a fake file with the same MD5 sum if they wanted to.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,14 +4,6 @@ language: python
 
 matrix:
     include:
-        - python: "2.7"
-        - python: "3.5"
-        - python: "3.6"
-        - python: "3.7"
-        - python: "3.6"
-          name: "Tests that don't need the ephemeris server"
-          script:
-            - (cd tests ; nosetests -v -a '!ephem_server' --with-coverage --cover-package=pint --ignore-files=.\*DISABLED.py)
         - script: wget https://data.nanograv.org/static/data/ephem/de405.bsp
           name: "NANOGrav ephemeris server accessible"
           install: []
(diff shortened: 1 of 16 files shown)


aarchiba:
Relevant to #191 #383 #376

paulray:
The `datafiles.txt` does not include any `deXXX.bsp` files. I guess those are cached by astropy in `~/.astropy/cache/download/py3`. So, if you want to work offline you need to trigger astropy to download those files first. Is that right?  Should `setup.py` trigger astropy to download at least the ephemeris files that we use in tests?

paulray:
Ah, I see now that is what your astropy PR is meant to deal with.

aarchiba:
This work is basically not the right solution because of #504 - this is designed for files like de421 that never change but are too big or too awkward to include at installation. This is not the problem here - we have (at least some) files that are continuously updated, and therefore need to be re-downloaded whenever the user's data demands it.

aarchiba:
Closes #191

aarchiba:
Okay, it may have gotten lost in the commits but there are now two new Travis jobs: one that does nothing but check that the NANOGrav ephemerides are available, so you can know that's why your other testing runs are going to fail, and another Travis job that runs all the tests that *don't* require the ephemeris server. Admittedly this isn't all that many. But the combination should help reassure contributors that all those failures aren't all their fault.

aarchiba:
Closes #504

aarchiba:
Closes #505

PR #516, opened 2019-09-10
Fix location of pint import for readthedocs

aarchiba:


aarchiba:
Sorry for the chaos, I'm not used to readthedocs and it doesn't build docs quite the same way I do on my local machine. Maybe tox could help?

PR #524, opened 2019-09-15
Convert all code to "black" formatting

aarchiba:
PINT source code is written by expert astronomers who may not be expert software developers. This has resulted in some unfortunate code formatting. Future commits will presumably also be quite variable in their formatting. The "black" code formatter (any color you like, as long as it is black) systematically reformats the code into its own format, which is simple, uniform, and well thought-out. This PR both runs all the current code through black, reformatting everything, and also installs pre-commit hooks so that "black" is run whenever one tries to commit new code.

This giant patch reformatting everything does complicate "git blame", but the tool "git-hyper-blame" exists for working around exactly this.


aarchiba:
Because this touches almost every file, it will probably flunk flake8-diff badly. This is because we have a lot of non-formatting flake8 failures, and the diff includes nearly everything. Nevertheless using "black" simply eradicates many classes of flake8 failure, and having it in pre-commit ensures that those mistakes cannot recur.

paulray:
How do you deal with all the conflicts that this big change will incur?  Any PR against the previous version will be a big mess.  Maybe if they run black on their code it will take care of that?

aarchiba:
Our backlog of PRs is mostly very old and busily bitrotting. (I've been going through and trying to merge or close them.) But if someone wants to revive one, there is a description here:

https://engineering.depop.com/implementing-python-black-on-a-legacy-codebase-35b37f10ce18

Specifically
```
git merge develop -s recursive -X ours
```
is supposed to choose the developer's version when there's a conflict with the blackened code, resolving that merge conflict. Or if you're just merging in that one giant commit, you can just manually choose your own versions for all the conflicts. Blackening your own code before committing should then bring the whole code base into standard. 

The author of black also claims that when you go to merge a PR that has conflicts there are useful instructions on de-conflicting.

aarchiba:
My experience incorporating new changes from master into this PR is that because the changes here can be easily redone by ``black``, I could just without much thought choose the version of each hunk from master, then re-blacken the final result. Manually merging the other way - moving a PR past this - shouldn't be any harder. I'm pretty sure the above command automates the "just choose my real change over the reformatting" process.

And goodness is it nice not to have to ever think about where to break a line.

paulray:
Yes, we should discuss this. If other devs agree that black is a good idea (I just read about it and I'm in favor of it), then we should get this merged soon, so all new development can be based on it.

One thing I noticed in the test failures is the python style check. I think we should turn off all the complaints for variables that are not in lowercase. The names it was complaining about all make sense to me (e.g. TIMEZERO, d_delay_astrometry_d_ELONG).

Issue #525, opened 2019-09-15
np.float128 misleading and inefficient


aarchiba:
A number of places in the code we use the dtype np.float128. This is misleading, in that while float128 takes up 128 bits of space, it offers only 80 bits of precision. It is also inefficient: on 64-bit platforms, it is most efficient for computation to pack 80-bit floats into 128 bits so they align better with loads. But on 32-bit platforms, numpy's default type for 80-bit floats is np.float96, because there is no advantage to alignment and a considerable disadvantage to cache use to pad with four more bytes of zeros. We should not be using np.float96 or np.float128 but np.longdouble. 

The sole advantage to these more-specific types is that on machines with no 80-bit support in hardware, np.longdouble==np.float but np.float96 and np.float128 simply don't exist and thus any attempt to use them raises an error. To solve this we could create a type pint.extended_float that raises an error on import if there is no extended float type and otherwise provides whatever there is with at least eighteen decimal places.

luojing1211:
Hi @aarchiba, I am revising the PITN paper now. There are several places I used np.float128. Should I change this to 80-bit floats as well? Or should I keep np.float128 for this version?

aarchiba:
In the paper I think we should be very clear we are using hardware 80-bit floats. Unless you are mentioning API details that need np.float128, I think I wouldn't mention them at all. I should probably make a quick pull request that just introduces a high-precision float type, checks it supports enough precision for us, and then uses that throughout the code everywhere we now say np.float128 or np.longdouble. 

Of course, we actually use a mixture of 80-bit floats and the astropy Time object's "double-double" numbers (an implicitly-added pair of double-precision numbers). The Time objects support more precision but are slower and don't support all sorts of useful operations.

PR #527, opened 2019-09-15
Make and use type extended_precision

aarchiba:
Replace float128 and longdouble with a new type `extended_precision`,
which is checked before construction.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ matrix:
           name: "Documentation build"
           python: "3.7"
           after_success: []
-    allowed_failures:
+    allow_failures:
         - script: flake8-diff
 sudo: required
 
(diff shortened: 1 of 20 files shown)


aarchiba:
Other reasonable names:

* extended_precision - a little unclear that we mean a data type but human-readable (this is what my fingers came up with when I was going for extended_float)
* extended_float - very clear, not confusable with the C default type
* longdouble - clear enough, we're probably never going to use anything different from c long doubles (unless there's a way to get 80-bit floats on MSVC?), the phrase "longdouble" appears in a number of function names
* np.longdouble - that is, don't define a name at all just use numpy longdoubles throughout. This has the disadvantage of not necessarily raising an exception immediately if on a platform that doesn't have extra precision, but "import util" will fail and is probably enough.

aarchiba:
Closes #525

PR #528, opened 2019-09-16
Incorporate example notebooks into documentation

aarchiba:
This gives us usage examples. They're not great, they need updating, but they're there. Also include running them as part of Travis, so that we notice if we break them (they were all broken).


aarchiba:
Notebooks in git tend to have big and awkward diffs. I haven't done anything about this. One option would be to have ``examples`` contain notebooks with empty output (done with ``jupyter nbconvert --clear-output``, possibly in a pre-commit hook). In theory ``nbsphinx`` is able to run notebooks to fill in the output while building the docs, though this is awkward as the doc builder will then maybe crash if we do something wrong in the code.

aarchiba:
Closes #529 turns out that helps quite a lot; once it makes it into the mainline branches will pull in the cache from the mainline install and so won't need to hit the servers unless you need new files downloaded.

pennucci:
@aarchiba, not sure this belongs here, but for the IPTA student week this year I put together a reasonably involved intro-timing activity that uses PINT and serves also as a PINT introduction, which occupied the students for quite some time.  A while ago I sent it to Jing, but I'm not sure he looked at it.  I'd be very happy to contribute it to the official repo if the PINT people like it; it's a notebook with a parfile and two TOA files: https://github.com/pennucci/PINT_timing_activity.  I just updated it to reflect some of the dozens of changes you've recently made!

pennucci:
@aarchiba, not sure this belongs here, but for the IPTA student week this year I put together a reasonably involved intro-timing activity that uses PINT and serves also as a PINT introduction, which occupied the students for quite some time.  A while ago I sent it to Jing, but I'm not sure he looked at it.  I'd be very happy to contribute it to the official repo if the PINT people like it; it's a notebook with a parfile and two TOA files: https://github.com/pennucci/PINT_timing_activity.  I just updated it to reflect some of the dozens of changes you've recently made!

PR #532, opened 2019-09-18
Setup.py and packaging modernization

aarchiba:
This does basically what #535 does but also moves the PINT code to a src/ directory. This layout has several advantages, including that any data files we want included can simply be dropped in alongside the code. More importantly it helps separate the installed version of PINT from the un-installed code that is sitting there in the directory people are probably running their development from. But I recognize that moving all the source code to a new location is disruptive to all developers!


aarchiba:
A few references:
https://blog.ionelmc.ro/2014/05/25/python-packaging/
https://blog.ionelmc.ro/2014/06/25/python-packaging-pitfalls/

Issue #533, opened 2019-09-19
pintk's graphical interface 


hahareally:
Hi,
I tried the example files of PINT like below, but pintk's graphical interface can not appear(attachment). virtualenv is not used. PINT version is pint-0.5.4+246.gd22c479
$pintk J0613-sim.par J0613-sim.tim
$uname -a 
Linux M 3.19.8-100.fc20.x86_64 #1 SMP Tue May 12 17:08:50 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
gh

hahareally:
STARTING LOADING OF PULSAR J0613-sim.par
prefit_model.as_parfile():
PSR                            J0613-0200
EPHEM                               DE421
UNITS                                 TDB
RAJ                      6:13:43.97767520 1 0.00000353933036693595
DECJ                    -2:00:47.22538000 1 0.00008785648095574063
PMRA                   1.8277959527833962 1 0.01764671502858701
PMDEC                 -10.338557735985926 1 0.03426481576151733
PX                     1.1163892630643213 1 0.1113633013982899
POSEPOCH            54890.000000000000000
F0                  326.60056202349007573 1 4.67708045e-12
F1              -1.0229704624634172473e-15 1 3.5895689268283532237e-20
PEPOCH              54890.000000000000000
PLANET_SHAPIRO                          N
DM                  38.777534445346590455 1 0.0006317482436224768
DM1                                   0.0
DMEPOCH             54890.000000000000000
BINARY BT
PB                  1.1985125752225728535 1 1.274978299e-11
PBDOT                                 0.0
A1                     1.0914342120578364 1 7.292248660874e-08
A1DOT                                 0.0
ECC                 5.286463595509009e-06 1 3.8222268559199e-07
EDOT                                  0.0
T0                  53113.955095598856088 1 0.00266858155449336328
OM                  47.681519597443860595 1 0.8015680450545875635
OMDOT                                 0.0
GAMMA                                 0.0
NE_SW                                 0.0
SWM                                   0.0
TZRMJD              54401.720778357950278
TZRSITE                                 1
TZRFRQ                             1432.0

Number of TOAs:  1462
Number of commands:  2
Number of observatories:  1 ['gbt']
MJD span:  53216.127 to 55639.377
gbt TOAs (1462):
  Min error:     0.146 us
  Max error:     38.8 us
  Mean error:    3.35 us
  Median error:  1.54 us
  Error stddev:  4.53 us

[...]

PR #534, opened 2019-09-19
Fix API change in pintk (.resids is now .Residuals)

paulray:
Addresses #533


paulray:
This passes all checks. The checks are much improved now, thanks to @aarchiba !
Strangely, it failed on Python 3.6 the first time when it tried to download IERS A, but when I re-ran just that test it worked.  
I think this is ready to be merged.

aarchiba:
Closes #533 
(This message - specifically with "Closes" - should make github close the referenced bug.)

Issue #497, opened 2019-08-22
 test_wrong_path fail


ktzhao:
ERROR: test_wrong_path (test_observatory.TestObservatory)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/k/psrsoft/PINT/tests/test_observatory.py", line 82, in test_wrong_path
    _ = site.clock_corrections(self.test_time)
  File "/home/k/psrsoft/PINT/pint/observatory/topo_obs.py", line 192, in clock_corrections
    format=self.clock_fmt, obscode=self.tempo_code))
  File "/home/k/psrsoft/PINT/pint/observatory/clock_file.py", line 59, in read
    return cls._formats[format](filename, **kwargs)
  File "/home/k/psrsoft/PINT/pint/observatory/clock_file.py", line 95, in __init__
    mjd, clk, self.header = self.load_tempo2_clock_file(filename)
  File "/home/k/psrsoft/PINT/pint/observatory/clock_file.py", line 108, in load_tempo2_clock_file
    f = open(filename,'r')
IOError: [Errno 2]  No such file or directory : '/home/k/psrsoft/tempo2/T2runtime/clock/fake2gps.clk'

luojing1211:
How did you run the tests?

-Jing

scottransom:
@luojing1211 Note that @clp3ef and I have also recently seen this fake2gps.clk issue when running tests.  It is happening when running all the nosetests as per the instructions.

scottransom:
And note that fake2gps.clk does not seem to be provided in TEMPO2 anymore.  That might be the issue.

luojing1211:
Since this is the test for the wrong path, it should raise an error. It will catch the RuntimeError, but the error message gives an IOError. I think that is why.

paulray:
Ah, I found this issue, which I'm having too.  Can we fix just by catching the right Exception? I'm not sure fake2gps.clk ever existed.

aarchiba:
Since you can reproduce it, can you tell who is asking for it? pytest with the --pdb will drop you into the debugger wherever the IOError was raised and you can walk up and down the stack trying to figure out who wants it and why. Whatever wants it needs to be fixed.

ktzhao:
I don't know if it is necessary for test the wrong path.
But it(1 test faill) will confuse beginners.

paulray:
[...]

PR #544, opened 2019-10-02
Fix missing files test to handle unset TEMPO2 variable

aarchiba:
Straighten out annoying file not found error that makes tests fail.


paulray:
Maybe add the astropy>=3.2 to requirements.txt to this PR as well?

aarchiba:
> Maybe add the astropy>=3.2 to requirements.txt to this PR as well?

That's a tiny PR so I'd rather keep it separate. But yes it should absolutely go in. (Maybe before I manage to fix all the bugs I put in this one...)

aarchiba:
I sort of think we should have whole categories of tests that are run whenever tempo or tempo2 are found to be installed. I don't know that we can put them in Travis - automatedly installing tempo and tempo2 may not be feasible - but at least some of our users will have them, and systematically running these tests would help catch some of our mistakes. 

In fact I think we have some of these but they're under "validation" and not run?

aarchiba:
Fixes #497

luojing1211:
We have some Tempo2 generated results saved in the test_data directory. The
validate is not a part of our Travis test.
-Jing

aarchiba:
> We have some Tempo2 generated results saved in the test_data directory. The validate is not a part of our Travis test. -Jing

Yes, I saw that. I bet many of those tests now fail, since we're not running them.

I would like to see an adapter through which we could run tempo/tempo2 that cached the results. Then we could check in the results to git, and regenerate them on a machine that does have tempo/tempo2 whenever we expect those results to change. On a machine without tempo/tempo2, the tests would just pull in the cached results and do the comparison. Using the wrapper would make it easy to write the tests, running tempo/tempo2 as needed.

PR #545, opened 2019-10-02
Astropy 3.1 doesn't work. Require 3.2.

### LINKED ISSUE #543: Test failures
I just updated my python install/packages and now I'm getting test failures for every test.  Here is the first one (~every test fails this way):
```
platform darwin -- Python 3.6.9, pytest-5.1.2, py-1.8.0, pluggy-0.13.0
rootdir: /Users/paulr/src/PINT-nanograv
collected 28 items / 47 errors

==================================== ERRORS ====================================
_____________________ ERROR collecting tests/test_B1855.py _____________________
tests/test_B1855.py:2: in <module>
    import pint.models.model_builder as mb
../PINT/src/pint/models/__init__.py:14: in <module>
    from .spindown import Spindown
../PINT/src/pint/models/spindown.py:19: in <module>
    import pint.toa as toa
../PINT/src/pint/toa.py:23: in <module>
    from .observatory import Observatory, get_observatory
../PINT/src/pint/observatory/__init__.py:10: in <module>
    from . import observatories
../PINT/src/pint/observatory/observatories.py:9: in <module>
    itrf_xyz=[882589.65, -4924872.32, 3943729.348])
../PINT/src/pint/observatory/topo_obs.py:95: in __init__
    self._loc_itrf = EarthLocation.from_geocentric(*xyz)
/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/astropy/coordinates/earth.py:253: in from_geocentric
    return super().__new__(cls, struc, unit, copy=False)
/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/astropy/units/quantity.py:379: in __new__
    value = value.astype(float)
E   TypeError: Cannot cast array from dtype([('x', '<f8'), ('y', '<f8'), ('z', '<f8')]) to dtype('float64') according to the rule 'unsafe'
```

Here is some version information (on macOS):
* platform darwin -- Python 3.6.9, pytest-5.1.2, py-1.8.0, pluggy-0.13.0
* py36-astropy @3.1_0 (active)
* py36-numpy @1.17.2_0+clang37+gfortran+openblas (active)
* py36-scipy @1.3.1_1+gfortran+openblas (active)

My guess is that astropy.quantity just got more picky about casts.

aarchiba:
On python2 astropy 2.0 is still fine (which is good because it's the
last version that will exist).

Fixes #543


paulray:
Shouldn't requirements.txt be modified as well?

aarchiba:
> Shouldn't requirements.txt be modified as well?

Ah. Um. Well.

In fact, as I understand it, requirements.txt is meant for setting up development environments; to set up a normal install the requirements should be in setup.cfg. So I moved all the normal-install requirements there and moved requirements_dev.txt to requirements.txt. And I left a symlink behind so people running pip install -r requirements_dev.txt would get what they expected.

I'm still not totally sure I did this right, but:

- https://packaging.python.org/discussions/install-requires-vs-requirements/
- https://caremad.io/posts/2013/07/setup-vs-requirement/

aarchiba:
Fixes #543

PR #548, opened 2019-10-05
Determinism

aarchiba:
Try to nail down the problems with determinism so our MCMC notebook stops changing.


aarchiba:
This is waiting on https://github.com/dfm/emcee/issues/313

aarchiba:
Now the notebook comes out identical every time I run it through tox.

aarchiba:
We might be better off using jupytext or other magic to avoid checking the notebooks in to version control.

aarchiba:
I started working on using jupytext but ran into https://github.com/mwouts/jupytext/issues/339

PR #549, opened 2019-10-06
Fix buggy time conversion code

aarchiba:
Update: essentially all precision problems intrinsic to PINT are under control at the nanosecond level or better, as far as I know and hypothesis can find. The remaining xfail tests are almost all pointing out upstream problems, whether due to astropy bugs or ERFA limitations. The astropy bugs are fixed in the current 3.2.2 release and should be fixed in the final 2.0.16 release which will appear in a few months. (2.0.15 was meant to be the final if-theres-a-bug-tough-luck python2 release but in light of the bug revealed here and other worries the mainainers decided to delay stopping bug fixes.)

The patch also dramatically simplifies the precision-conversion functions in pint.utils by providing new astropy.time.Time formats mjd_string and pulsar_mjd_string that read and write full-precision strings, as well as mjd_long and pulsar_mjd_long that write but do not read long doubles (reading long doubles is prevented by a frustrating limitation of astropy, under discussion for removal in astropy 4). Conversion of existing code to replace time_to_* and time_from_* with the new formats has not been carried out yet.

Was:

We have a number of severe time conversion problems, producing errors of, in some cases, 60s or a day. Others may just be off by a microsecond. I am fixing them where I can but there are a number I have not tracked down yet. This PR tracks those with xfail statements; use
```
pytest --runxfail -v tests/test_precision.py
```
to see the damage.

Most but not all of the problems center around days with leap seconds, and some of the problems appear to originate from within our pulsar_mjd implementation (possibly but I hope not from within ERFA).

### DIFF
--- a/src/pint/pulsar_mjd.py
+++ b/src/pint/pulsar_mjd.py
@@ -26,19 +26,52 @@ was observing the sky at the moment a leap second was introduced.
 
 .. _leap_smear: https://developers.google.com/time/smear
 """
-from __future__ import absolute_import, division, print_function
 
+import astropy._erfa as erfa
+import astropy.time
+import astropy.units as u
 import numpy as np
+import six
+from astropy._erfa import DJM0
 from astropy.time.formats import TimeFormat
+from astropy.time import Time
 
-from pint.utils import (
-    jds_to_mjds,
-    jds_to_mjds_pulsar,
-    mjds_to_jds,
-    mjds_to_jds_pulsar,
-    mjds_to_str,
-    str_to_mjds,
-)
+try:
+    maketrans = str.maketrans
+except AttributeError:
+    # fallback for Python 2
+    from string import maketrans
+
+# FIXME: can we make this exception raise on install as well as on use?
+if np.finfo(np.longdouble).eps > 2e-19:
+    raise ValueError(
+        "This platform does not support extended precision "
+        "floating-point, and PINT cannot run without this."
+    )
+
+
+__all__ = [
+    "Time",
+    "PulsarMJD",
+    "MJDLong",
+    "PulsarMJDLong",
+    "MJDString",
+    "PulsarMJDString",
+    "fortran_float",
+    "longdouble2str",
+    "str2longdouble",
+    "data2longdouble",
+    "time_from_longdouble",
+    "time_to_longdouble",
+    "time_from_mjd_string",
+    "time_to_mjd_string",
+    "mjds_to_str",
+    "str_to_mjds",
+    "mjds_to_jds",
+    "jds_to_mjds",
+    "mjds_to_jds_pulsar",
+    "jds_to_mjds_pulsar",
+]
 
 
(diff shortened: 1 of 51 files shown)


src/pint/pulsar_mjd.py
```diff
@@ -56,25 +59,38 @@ class TimePulsarMJD(TimeFormat):
...
             # Seems like there should be a ERFA routine for this..
-            h = safe_kind_conversion(numpy.floor(f*24.0), dtype=int)
-            m = safe_kind_conversion(numpy.floor(numpy.remainder(f*1440.0,60)),
-                              dtype=int)
-            s = numpy.remainder(f*86400.0,60)
-            self.jd1, self.jd2 = erfa.dtf2d('UTC',y,mo,d,h,m,s)
+            # There is: erfa.d2tf. Unfortunately it takes a "number of
+            # digits" argument and returns some kind of bogus
+            # fractional-part-as-an-integer thing.
+            # Worse, it fails to provide nanosecond accuracy.
```
scottransom: @aarchiba So do you think this is a SOFA/ERFA bug?  Since I'm on the SOFA board, if you can give me a simply example of the numerical issues, it *will* get fixed.  We just did that over the last year for another small issue or two.

aarchiba: I'm not totally sure whether I'm interpreting the ERFA output correctly, but if you want I can put together some tests that give astropy._erfa some significant exercise and report any bugs I come up with. Translating them back into C to be sure they're really SOFA bugs and not somewhere in the wrappers would be... more work. 

Is there documentation of the level of accuracy I should expect from various ERFA/SOFA routines, over what range of inputs?

scottransom: There is a lot of SOFA documentation.  And given that the times are stored internally with two doubles, the *intent* is to have very high precision where possible -- certainly much better than a microsecond.  I'd have to dig around to see what is expected overall.  But if there are definite instabilities at the ns level -- especially if they can be fixed easily, then those should definitely be fixed.

aarchiba: Okay, well, a quick check shows that a roundtrip through d2tf and tf2d gives nanosecond accuracy for ndp=9 but does *not* give nanosecond, or even ten-nanosecond, accuracy for higher ndp. No errors or warnings are emitted. See 3cc6a8a.

aarchiba:
<sigh /> Travis has gotten grumpy again. Even *apt-get* is failing to pull stuff across the network. At least it's usually fairly obvious (! instead of x next to runs).

aarchiba:
TODOs:

- mjd_decimal and pulsar_mjd_decimal
- day_frac_longdouble? Does day_frac work for longdoubles?
- Convert code base to eliminate time_to functions
- Get rid of time_from_mjd_string (can't get rid of time_from_longdouble, but maybe use day_frac_longdouble?)

There are other places precision could be mishandled - Parameters that are supposed to handle long doubles spring to mind - but I think that this is enough for one patch.

PR #560, opened 2019-10-16
Fix AbsPhase bug

### LINKED ISSUE #559: Bug with photonphase ('AbsPhase' object)
Using photonphase on NICER data and with a simple par file (F0, F1, PEPOCH, POSEPOCH, ...), and I get the following error:

```
  File "/Users/sebastien/Softwares/anaconda3/lib/python3.6/site-packages/pint-0.5.4+360.ga4b33ea-py3.6.egg/pint/models/timing_model.py", line 1103, in __getattr__
    return super(Component, self).__getattribute__(name)
AttributeError: 'AbsPhase' object has no attribute 'PEPOCH'

During handling of the above exception, another exception occurred:

  File "/Users/sebastien/Softwares/anaconda3/lib/python3.6/site-packages/pint-0.5.4+360.ga4b33ea-py3.6.egg/pint/models/timing_model.py", line 1110, in __getattr__
    % (self.__class__.__name__, name)
AttributeError: 'AbsPhase' object has no attribute 'PEPOCH'.

During handling of the above exception, another exception occurred:

  File "/Users/sebastien/Softwares/anaconda3/bin/photonphase", line 11, in <module>
    load_entry_point('pint==0.5.4+360.ga4b33ea', 'console_scripts', 'photonphase')()
  File "/Users/sebastien/Softwares/anaconda3/lib/python3.6/site-packages/pint-0.5.4+360.ga4b33ea-py3.6.egg/pint/scripts/photonphase.py", line 192, in main
    iphss, phss = modelin.phase(ts, abs_phase=True)
  File "/Users/sebastien/Softwares/anaconda3/lib/python3.6/site-packages/pint-0.5.4+360.ga4b33ea-py3.6.egg/pint/models/timing_model.py", line 580, in phase
    toas
  File "/Users/sebastien/Softwares/anaconda3/lib/python3.6/site-packages/pint-0.5.4+360.ga4b33ea-py3.6.egg/pint/models/absolute_phase.py", line 89, in make_TZR_toa
    PEPOCH = self.PEPOCH.quantity.value
  File "/Users/sebastien/Softwares/anaconda3/lib/python3.6/site-packages/pint-0.5.4+360.ga4b33ea-py3.6.egg/pint/models/timing_model.py", line 1117, in __getattr__
    % (self.__class__.__name__, name)
AttributeError: 'AbsPhase' object has no attribute 'PEPOCH'.
```

I'm not sure what can cause this, but this problem does not occur for other different pulsars (with more complex par files).

paulray:
This fixes #559 

It also makes the pintempo script a bit smarter


aarchiba:
Fixes #559

paulray:
OK, added some tests.  If travis finishes OK then I think this can be merged

tests/test_photonphase.py
```diff
@@ -36,5 +37,40 @@ def test_result():
...
+    photonphase.sys.stdout = saved_stdout
+
+
+@pytest.mark.skipif(
+    "DISPLAY" not in os.environ, reason="Needs an X server, xvfb counts"
+)
+def test_AbsPhase_exception():
+    "Verify that passing par file with no TZR* parameters raises exception"
+    with pytest.raises(ValueError):
+        saved_stdout, photonphase.sys.stdout = photonphase.sys.stdout, StringIO("_")
```
aarchiba: please don't ever import sys through another module. It just gives you the same sys as import sys, and it means this fails mysteriously if the other module doesn't do "import sys". 

More generally this mess can be avoided by using pytest's "capsys" fixture - it captures stdout and you can read it out, and it restores stdout even if an exception occurs in your test.

tests/test_photonphase.py
```diff
@@ -36,5 +37,40 @@ def test_result():
...
+    saved_stdout, photonphase.sys.stdout = photonphase.sys.stdout, StringIO("_")
+    cmd = "{0} {1}".format(eventfile_nicer, parfile_nicer)
+    photonphase.main(cmd.split())
+    lines = photonphase.sys.stdout.getvalue()
+    v = 0.0
+    for l in lines.split("\n"):
+        if l.startswith("Htest"):
+            v = float(l.split()[2])
+    # Check that H-test is greater than 725
+    log.warning("V:\t%f" % v)
```
aarchiba: No need for this, pytest will tell us the value of v if it is smaller.

tests/test_photonphase.py
paulray: Ah, I don't know where this idiom came from. I just copied it from the other command line tool tests.

src/pint/scripts/pintempo.py
```diff
@@ -43,7 +43,13 @@ def main(argv=None):
     log.warning(m.params)
 
     log.info("Reading TOAs")
-    t = pint.toa.get_TOAs(args.timfile)
+    use_planets = False
+    if m.PLANET_SHAPIRO.value:
+        use_planets = True
+    model_ephem = "DE421"
+    if m.EPHEM is not None:
+        model_ephem = m.EPHEM.value
```
aarchiba: Why are these defaults set up in pintempo? Are they different from the defaults get_TOAs would assume? Wait, why does loading TOAs depend on the par file in use?

tests/test_photonphase.py
paulray: Yeah, I don't know why that is there. Must have come from initial debugging of the tests

src/pint/scripts/pintempo.py
paulray: Because TOA objects compute the posvels at each TOA time. This requires knowing what the ephemeris is and whether the planet positions are needed. These are specified in the par file, as a consequence of the unfortunate mixing of model parameters and 'control' parameters in TEMPO par files.

aarchiba: We should probably discuss what can be done to clean up this mess. For example, we might have the par file reader return two objects, one for the "control" parameters and one for the actual model parameters. But obviously it's not something you can fix in this PR.

aarchiba:
There are a lot of bad idioms baked into the existing tests, and a huge amount of code duplication. But this isn't the place to fix them, beyond not adding any more problems.

The PR looks good to me.

src/pint/scripts/pintempo.py
paulray: Yes, this has been the subject of lots of previous discussion, e.g. https://github.com/nanograv/PINT/wiki/Discussion-of-Control-Parameters

But we haven't come up with a good solution.

PR #562, opened 2019-10-16
Add git blame config to ignore the blackening

aarchiba:


paulray:
Maybe add the necessary git config command to the comment?

aarchiba:
Good point. I'll squash and do that before merging. I was also going to add it to the docs as part of this PR.

PR #563, opened 2019-10-17
Change WAVE delay sign to match tempo2; add real WAVE test.

kerrm:
Commit line says it all.  Somehow, the WAVE delay convention was opposite to that used in tempo2.  I made that trivial change and tested it with radio and gamma-ray data.  I also updated the test to make use of unittest and to actually test the residuals in a meaningful way.


paulray:
This is failing `test_compare_get_model_new_and_old_all_parfiles` because one has the SolarWindDispersion component and the other does not. @aarchiba do you know why this is?
It may be related to the fact that there are two NE_SW parameters in the par file.

paulray:
As it turns out, `get_model()` did not make a `SolarWindDispersion` component when presented with two NE_SW lines, but `get_model_new()` did. I guess we should discuss how duplicated parameters should be handled. Should it just overwrite the value, or throw an exception?  For sure it should not just ignore them and not create the model component, as `get_model()` currently does.

paulray:
I believe @matteobachetti added the WAVE model initially. Matteo, did you check the sign convention?

aarchiba:
> As it turns out, `get_model()` did not make a `SolarWindDispersion` component when presented with two NE_SW lines, but `get_model_new()` did. I guess we should discuss how duplicated parameters should be handled. Should it just overwrite the value, or throw an exception? For sure it should not just ignore them and not create the model component, as `get_model()` currently does.

The intention is that get_model_new should replace get_model; there are a few more changes I need to make to ensure that it does the right thing in all cases but it already has much more robust handling of, ah, anomalous .par files.

matteobachetti:
Hi @paulray & @aarchiba, I'm not sure. I remember testing it on a dataset, making sure that the residuals really improved, but I might have caught a lucky case.

tests/test_model_wave.py
```diff
@@ -1,25 +1,36 @@
...
+    def setUpClass(cls):
+        cls.m = pint.models.get_model(parfile)
+        cls.t = pint.toa.get_TOAs(timfile, ephem="DE405", include_bipm=False)
+
+    def test_vela(self):
+        print("Test RMS of a VELA ephemeris with WAVE parameters.")
+        rs = pint.residuals.Residuals(self.t, self.m).time_resids
+        rms = rs.to(u.us).std()
+        emsg = "RMS of " + str(rms.value) + " is too big."
+        assert rms < 350.0 * u.us, emsg
```
aarchiba: Is it worth including a version without WAVE and confirming that the RMS is worse?

matteobachetti: It would be great!

kerrm: With this example, without the waves there are many, many wraps, and the rms will be catastrophic.  So in that sense it's a fairly stringent test since the waves need to be working and working well to pass the test.  Of course, there is always room for more tests.

Based on the email exchange I've implemented Waves as a PhaseComponent and slightly tweaked the test.  I plan to submit this as a separate PR, so this one can be rejected/deleted, imo.

aarchiba:
Catastrophic is good, perversely! We want the test to fail for sure if the waves aren't working right.

tests/test_model_wave.py
aarchiba: When the new one is ready to upload, mention that it supersedes this one (by number so GitHub will make the link) and we can close this one.

aarchiba:
Closing as superseded.

PR #567, opened 2019-10-24
Update to enable plotting astropy Quantities

paulray:
I updated pintempo and the PINT walkthrough notebook to use the astropy feature for enabling Quantity support in matplotlib. This makes plotting residuals with error bars much nicer, without needing a bunch of `.value` stuff.


aarchiba:
Looks good to me!

PR #569, opened 2019-10-25
Add IFUNCs; change WAVES implementation to PhaseComponent

kerrm:
This PR supersedes PR (https://github.com/nanograv/PINT/pull/563).

This PR adds IFUNC functionality (basically a time series of phase offsets interpolated onto TOAs).  Some items remain TODO, but I wanted to go ahead and submit the bones before I got too out of sync with master.

IFUNC and an updated WAVES model are implemented as PhaseComponents, which allows conversion between barycentric time and phase trivially (multiplication by F0).  This is necessary for compatibility with tempo2 in the case of pulsars with substantial spin evolution.

The testing suites are minimal: I suggest that if/when adding derivatives to allow the fitting of WAVES, a nice and easy test would be to manually zero out a few of the WAVE terms once added in, then see if the fitter can get back to the original RMS.


aarchiba:
The Wls-related error is because I renamed them in a recent PR. The other I'd better investigate.

kerrm:
It's a Fermi pulsar/ephemeris, so there is no DM.  (Although curiously, a DMEPOCH...)  Could that be related to the model builder error?

kerrm:
I was testing with an ephemeris with a ridiculously large WAVE signal (which can happen due to covariance with F0/F1 etc.) and I found that I needed to include the Roemer delay to reproduce the tempo2 behavior.

paulray:
The tests are failing to download IERS_A_URL.  Those should be rewritten to try IERS_A_URL first, then IERS_A_URL_MIRROR if that fails.

paulray:
@aarchiba I cannot replicate the test_precision failures on my machine.  Maybe it is specific to something in the travis configuration.

I suggest we merge this and sort out that test failure independently since it has nothing to do with this PR, I believe.

Issue #558, opened 2019-10-15
In TOA line, a flag can overwrite a positional value


paulray:
I just got a NRT TOA file whose lines look like this:
```
FORMAT 1
 nuppi_55804_1821-24_155239.F8Tp 1263.05500000 55804.81748030375163339 0.48700 ncyobs -i NUPPI -r ROACH -gof 1.32588 -tpl NUPPI_1484_smoothed.prof -freq 1484.0 -bw 512.0 -tobs 3718.6 -orig_nbin 2048 -orig_nsub 61 -orig_nchan 128 -obsnchan 144 -nsub 1 -nchan 8 -sublen 3718.6 -chanwidth 64.0 -snr 115.9 
 nuppi_55804_1821-24_155239.F8Tp 1324.48300000 55804.81748029492493046 0.40100 ncyobs -i NUPPI -r ROACH -gof 1.31246 -tpl NUPPI_1484_smoothed.prof -freq 1484.0 -bw 512.0 -tobs 3718.6 -orig_nbin 2048 -orig_nsub 61 -orig_nchan 128 -obsnchan 144 -nsub 1 -nchan 8 -sublen 3718.6 -chanwidth 64.0 -snr 115.9 
 nuppi_55804_1821-24_155239.F8Tp 1386.78300000 55804.81748028949086660 0.44300 ncyobs -i NUPPI -r ROACH -gof 1.26227 -tpl NUPPI_1484_smoothed.prof -freq 1484.0 -bw 512.0 -tobs 3718.6 -orig_nbin 2048 -orig_nsub 61 -orig_nchan 128 -obsnchan 144 -nsub 1 -nchan 8 -sublen 3718.6 -chanwidth 64.0 -snr 115.9 
```

Notice that one of the flags is `-freq`.  In `parse_TOA_line()`, it builds a dictionary of the positional values (`obs`,`freq`,`errror`) as well as all the flags. In this case the `-freq` flag silently overwrites the `freq` so the TOA gets 1484.0 MHz as the frequency instead of the correct 1263.055 MHz.  

I'm not sure if this counts as a PINT bug or a user error, but it should not happen.  Perhaps we should make "freq", "obs", and "error" reserved words that can't be used as flags?  Or we should change the names of those dictionary items to reduce the chance of a conflict (e.g. "_freq").

Thoughts?

aarchiba:
[...]

PR #570, opened 2019-10-26
Doc/Notebook improvements

### LINKED ISSUE #555: tim file commands must be UPPERCASE
I tried using "include" in a .tim file to read in an included file. This works fine in Tempo2, but in PINT the parser does the comparison as cmd == "INCLUDE" so it requires the command to be all uppercase. Is this the desired behavior?  Should we support lower case commands, like Tempo2 does? Note that TEMPO seems to require all uppercase.

paulray:
This is a few doctoring and notebook improvements.
Also makes TOA line command parsing be case insensitive: Fixes #555


paulray:
Also includes a partial fix for #558. At least it will raise an exception instead of quietly doing the wrong thing.  Full fix would still be useful.

paulray:
OK, this passes and is ready to merge

paulray:
OK, the black failure is wrong. I tried black on the files it specifies and no changes are needed.  Maybe there is some version difference?

paulray:
I think this can be merged now. Then @kerrm can merge with his PR and it should test without errors (except for the bizarre black thing)

paulray:
@aarchiba is it OK to ignore the black failures in travis and merge this now?  I can't replicate them on my machine:
```
mcxr2 : 35>black --check src tests examples
All done! ✨ 🍰 ✨
153 files would be left unchanged.
```

kerrm:
19.10b?  how old and busted!  The glorious future over here has 19.3b0!

PR #575, opened 2019-10-31
WIP: Fix buggy TOA writing

paulray:
This fixes a bug in `format_toa_line()` where TOAs that were barycentric got the UTC correction applied before being written. This is incorrect. The TOA line should have UTC times for all observatories, EXCEPT for Barycenter, when they should have TDB times.

This current fix works, but we may want to discuss a better solution. Should Observatories provide their own method for making the time in a TOA line?  What about the TT times in "spacecraft" observatories? 

Also, I will add a unit test to round trip a BAT TOA to a file and make sure it works.


paulray:
This bug seems to have come in when some combination of @aarchiba and @clp3ef  changed `format_toa_line()` from using `utils.time_to_mjd_string()` to using `toa_str = Time(toatime, format="pulsar_mjd_string", scale="utc")`.  The new code assumes that all MJDs in .tim files are UTC, which is not true.  Obviously, this exposes our lack of test coverage for the other cases.  I will try to remedy that.

paulray:
We also need to make sure that trying to write event TOAs from Fermi, NICER, RXTE, etc. fails in some sensible way. 
Currently the only spacecraft TOAs that can be written to a .tim file are SpacecraftObs.
The others (Fermi, NICER, RXTE, ...) can't unless they are barycentered, in which case the observatory is Barycenter and not NICERObs, FermiObs or whatever.  We could make it fail, or we could have them translated into SpacecraftObs at writing, thus embedding the observatory position on the TOA lines.

paulray:
I found that the new string formatting also broke writing TEMPO format .tim files because the string could be longer than 20 characters, which is all that is allowed in the Princeton format. I included a bug fix for that too.

PR #577, opened 2019-11-03
fix for random_models to work in Python3

clp3ef:
small fix that lets random_models (and therefore pintk) work in Python3. The issue was dictionary .values() is no longer accepted as an array that a scalar can be multiplied by. Replaced with conversion to list and then numpy array.

### DIFF
--- a/src/pint/random_models.py
+++ b/src/pint/random_models.py
@@ -53,7 +53,7 @@ def random_models(
     # scale by fac
     log.debug("errors", np.sqrt(np.diag(cor_matrix)))
     log.debug("mean vector", mean_vector)
-    mean_vector *= fac
+    mean_vector = np.array(list(mean_vector)) * fac
     cov_matrix = ((cor_matrix * fac).T * fac).T
 
     toa_mjds = fitter.toas.get_mjds()


clp3ef:
.tox is failing on test_glitch.py with the message "ERROR: InvocationError for command /home/travis/build/nanograv/PINT/.tox/py37/bin/coverage run -m pytest (exited with code -6)"
Does anyone know why? this single-line change doesn't touch any glitch code, I don't know what could be breaking.

paulray:
I have no idea why this is failing, but since all the other tests pass, I'd say we can merge this now and sort out that unrelated error in a separate PR

paulray:
In addition, it seems that random_models always runs.  For some situations it might not be desirable. What about adding a radio button to enable/disable plotting the random models?

clp3ef:
I actually tried to get that to work during the summer and had an unreasonably difficult time with it. I gave up after a while and it’s still definitely something I’d like to implement, but it isn’t a quick/easy add on like I thought it would be.

paulray:
tox is failing on my local machine as well. Some of our tests, and the tox environment seems to be incredibly flaky and unpredictable. I'm going to merge this anyway

aarchiba:
If the travis job gets stuck like this, someone who has write access to the repository can restart the job. If it's a real problem it'll come up again but this error seems highly suspicious - Travis can be flaky in various ways.

aarchiba:
If you don't have write access, you can make some minor doc fix and add it to your PR to trigger another build.

PR #586, opened 2019-12-11
Update gps2utc.clk

paulray:
Updates clock file.  Also re-enables event_optimize tests.


paulray:
Note that I got a couple of errors in travis on this PR. They were resolved just by restarting the build, so our travis tests are really fragile.  Some of the errors were from the chime data tests, but I think the root cause was the download of IERS data from cache_IERS.

PR #587, opened 2019-12-12
Fix bug in eventstats and add test

paulray:
This fixes a bug in sig2sigma().

I added a test, but it is not great, and really we should add tests for many of the other routines in this file.


paulray:
@aarchiba @luojing1211 These test failures are in either test_ell1h.py or test_fbx.py and seem to occur randomly. They have nothing to do with my changes. Somehow those tests are extremely fragile. Any idea what is going on?

luojing1211:
This is because the relative differences between the analytical derivative and numerical derivative are a little bit bigger than the old tolerance (0.00104 vs 0.001). I found this is interesting, py2.7 does not report the bug.

paulray:
So, do we increase the tolerance or fix the algorithm so they match more closely?

luojing1211:
I think we can do it now before we have a better way to test the derivative precision.

paulray:
@luojing1211 when you say "I think we can do it now", what is "it"?  I'd like to not have those failing tests.

paulray:
This PR now includes fixes for using pulse numbers either from a par file or computed internally by PINT. I know I should have kept them separate.. sorry.

paulray:
As soon as issue #593 is fixed, I'll add a good unit test for pulse numbers, which I think was not there before.

paulray:
OK, I'm going to merge this so that master will be passing all the tests.
The ex_pn.py in examples shows that pulse numbering is working but currently crashes due to issue #593 
I'll try to be better about making smaller PRs going forward.

Issue #550, opened 2019-10-08
Breaks with astropy 4.0dev


aarchiba:
I think the problem has to do with different handling of units? Specifically I think units pass through more functions than they used to - I saw np.interp(number, number, number*u.us)*u.us have a unit of u.us2. It seems like *everything* breaks.

aarchiba:
The following CHANGES.rst entry explains:
```
- For numpy 1.17 and later, the new ``__array_function__`` protocol is used to
  ensure that all top-level numpy functions interact properly with
  ``Quantity``, preserving units also in operations like ``np.concatenate``.
  [#8808]
```
This means we have to go through and do a careful units audit. Which will be good for us.

paulray:
This sounds difficult to do while maintaining backwards compatibility.

aarchiba:
Actually the Right Way is what we should have been doing all along: never pass Quantities to functions that are not explicitly Quantity-aware. The code that breaks assumes that (for example) np.interp *strips* units from its inputs, which it does in pre-4, but in 4 it just passes them through. If we just convert to the unit we want, take value, call the function, and multiply by the resulting unit, we're sorted. If the function was right before, it's just a matter of sticking .value on the end of its arguments; if the code didn't explicitly request particular units before calling such a function, it probably gave the wrong answer when presented with milliseconds instead of seconds (say).

When I talk about a careful units audit, I mean that anywhere there is a .value not immediately preceded by a .to(), there's a large chance there is a lurking bug just waiting to bite users. Someone (me I guess) should go through systematically and fix all of those.

Such an audit might not be necessary for the 4.0 update because almost any code that relies on pure numpy stripping units will immediately cause a unit conversion problem. This means about 2/3 of our tests fail, but possibly with the addition of some actual unit (rather than integration) tests it should be pretty easy to track down the problems, and very easy to tell when they're all gone.

paulray:
I don't quite get why that is the "Right Way".  What makes a function explicitly Quantity-aware?
If functions are migrating towards all supporting quantities, shouldn't we go that direction too, and not strip and re-add units with every function call?

I guess it would be nicer if there were a way to ask a numpy/scipy/astropy function if it is Quantity-aware and only do the .value hack if it is not.

scottransom:
I agree with @paulray   That is not the right way at all.  It is *certainly* not how the Astropy units maintainers want you to write code.  Please do *not* change all the code to do this.  We should talk to Martin van Kerkwijk about this.

[...]

PR #599, opened 2020-02-12
Fix der test

luojing1211:
This is a simple fix for the current derivative failure.

### DIFF
--- a/tests/test_ell1h.py
+++ b/tests/test_ell1h.py
@@ -61,7 +61,7 @@ class TestELL1H(unittest.TestCase):
         testp = tdu.get_derivative_params(self.modelJ1853)
         delay = self.modelJ1853.delay(self.toasJ1853)
         # Change parameter test step
-        testp["H3"] = 5e-1
+        testp["H3"] = 6e-1
         testp["H4"] = 1e-2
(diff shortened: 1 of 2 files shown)


luojing1211:
The tox failing is due to astropy4.0 changes. The only thing I changed it the numerical derivative step size. I think this should get merged and then fix astropy 4.0 issue. What do you think @scottransom @paulray @demorest

paulray:
We don't yet support astropy 4.0. See #550 
So, can you change requirements.txt to say astropy<4.0 in this PR?

luojing1211:
I don't know why tox somehow uses astropy 4.0.  PR #594 is aiming to fix the astropy 4.0 issue.

paulray:
Yeah, I can't figure out why tox is installing astropy 4.0.  The requirements_dev.txt clearly specifies astropy<4.0.  I think @aarchiba made the tox test in travis. Anne, any ideas?

luojing1211:
@paulray if you think this PR is ok, I will just merge it. TOX is really another issue.

paulray:
Yeah, that is fine, but it won't fix the main branch test failures, which was the point.

luojing1211:
OK, I will fix that separately.

PR #605, opened 2020-02-13
Allow change_pepoch() with no TOA argument

rossjjennings:
Removes an unnecessary use of the `toa` argument to `models.Spindown.change_pepoch()` that prevented it from being used on a model object without associated TOAs (#491). Also adds a unit test for the `change_pepoch()` function.

The new unit test checks for equality between two `Quantity` objects, which is maybe too strict (it may fail if the order of floating-point operations changes), but it passes as is.

### DIFF
--- a/src/pint/models/spindown.py
+++ b/src/pint/models/spindown.py
@@ -146,8 +146,6 @@ class Spindown(PhaseComponent):
             new_epoch = Time(new_epoch, scale="tdb", precision=9)
         else:
             new_epoch = Time(new_epoch, scale="tdb", format="mjd", precision=9)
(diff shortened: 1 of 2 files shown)


rossjjennings:
I'm not sure why the `tox` build is failing. How is it different from the other Travis builds?

paulray:
For some reason that I can't figure out, the tox build pulls in astropy 4.0, even though all of our specs call for astropy<4.0. That breaks the build.  We need @aarchiba to help with this.  
In the meantime, we could just disable that by commenting it out.

luojing1211:
On the other hand, I am trying to get astropy 4.0 work. But I agree, we should disable tox for the moment.

PR #607, opened 2020-02-19
Work around np.longdouble bug.

mhvk:
Unfortunately, it seems np.longdouble(string) looses precision if string is a numpy string scalar or array rather than a python string.  This adjusts the test to not avoid that bug, but does not actually check whether the bug introduces problems elsewhere in the code.

NOTE: NEEDS FURTHER CODE CHECKS BEFORE MERGING!

See https://github.com/numpy/numpy/issues/15608

cc @aarchiba, since this is partially a change in `Time` string formats now always returning a numpy scalar. Definitely need to check whether anywhere numpy strings are fed directly into long doubles.

### DIFF
--- a/tests/test_precision.py
+++ b/tests/test_precision.py
@@ -363,9 +363,13 @@ def test_time_from_longdouble_utc(format, i_f):
 def test_time_to_longdouble_close_to_time_to_mjd_string(format, i_f):
     i, f = i_f
     t = Time(val=i, val2=f, format=format, scale="utc")
-    assert (
-        abs(np.longdouble(time_to_mjd_string(t)) - time_to_longdouble(t)) * u.day
-    ).to(u.ns) < 1 * u.ns
+    tld = time_to_longdouble(t)
+    tstr = time_to_mjd_string(t)
+    # NOTE: have to add str() here, because of a numpy bug which treats
+    # numpy string type differently from python str.
+    # See https://github.com/numpy/numpy/issues/15608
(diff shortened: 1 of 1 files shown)


mhvk:
I tried looking through the code to see if this bug could have bitten elsewhere, but don't think so.

Aside: not clear why the tests are not running...

paulray:
Yeah, why are the travis tests not running?

paulray:
Ah, now the tests are showing up.  They did run. 
The code needs to be blackened.
https://travis-ci.org/nanograv/PINT/builds/652613794?utm_source=github_status&utm_medium=notification

mhvk:
@paulray - oops, that shouldn't have happened in #594 - see #608 for a fix.

paulray:
Let me know when you think this should be merged

mhvk:
@paulray - not sure. While I think this is an isolated bug that is fixed here, I'd prefer @aarchiba to have a look first, as she may have a better idea whether the numpy bug exposed here could also affect other parts of PINT.

In any case, I rebased so that now again only the `tox/astropy-4.0` run should be failing.

mhvk:
I've rebased on master so that #612 is included. I think there should only be one failure left (unrelated). I'm also fairly convinved it is just safe to merge this - I was worried the longdouble/string interaction somehow affected the other bits, but @luojing1211's discovery of the missing `location` showed that was wrong.

paulray:
The only failure is the photonphase one, so I'm merging this and I'll fix that issue in my own PR

PR #608, opened 2020-02-19
Blackening of previous PR

mhvk:
I should not have let that test failure in #594 go unaddressed. 

p.s.  Must admit I think these changes go rather against the spirit of PEP8, but good to stick to a standard regardless. EDIT: having looked at black, I now definitely understand why one would go with it, even if I don't particularly like the outcome here!


paulray:
Yeah we discussed it for a while with @aarchiba and decided that for multiple developers it was nice to have an automatically-applied standard style.

mhvk:
OK, this now passes all but the astropy 4.0 tests, so does indeed revert the non-black-ness introduced in #594.

Issue #610, opened 2020-02-19
Inconsistent defaults for ECL


rossjjennings:
Currently, the value used for the obliquity of the ecliptic is determined by the `obliquity` attribute of the `PulsarEcliptic` class. The default value specified in `ecliptic.dat` is equivalent to IERS2010, but when building a model with an `AstrometryEcliptic` component, the IERS2003 value (compatible with Tempo2) is used unless `ECL` is explicitly specified. We should have a consistent default. Probably this means changing the default for `AstrometryEcliptic` to IERS2010, but we might also go with IERS2003 to preserve Tempo2 compatibility.

rossjjennings:
Also, the setting as currently implemented is global across all `PulsarEcliptic` instances. This can lead to some very surprising behavior. For instance, it means that once you call `get_psr_coords()` on *any* model object where `ECL` has not been explicitly specified, `PulsarEcliptic.obliquity` changes to the IERS2003 value for the rest of the session -- the global default is overridden.

rossjjennings:
I think the best way of dealing with this would be to allow different `PulsarEcliptic` objects to have their own `obliquity` values. This matches how astropy treats parameters of other coordinate frames, like the equinox of `BarycentricTrueEcliptic`, and would allow things like reading in multiple par files with different `ECL` values in the same session.

paulray:
Yes, it seems wrong to have the obliquity as a class attribute instead of a property of an instance.

luojing1211:
The reason to make Obliquity a class method is the transformation function accepts a class instead of an instance.  I am not sure if this has been changed in the newer astropy version.

rossjjennings:
Closing now that #613 is merged.

PR #611, opened 2020-02-20
Add ability to change POSEPOCH

rossjjennings:
This is a major step toward solving #491. I've added a `change_posepoch()` function to the `AstrometryEquatorial` and `AstrometryEcliptic` classes, along with a basic test for it that covers the ecliptic case (in tests/test_change_epoch.py). The new function reuses epoch-updating functionality from `get_psr_coords()`. For that reason, I made another change here that is, in retrospect, logically independent: the astropy coordinate object returned by `get_psr_coords()` now includes the proper motion, and `get_params_as_ICRS()` now relies on it, so everything is in the same place. I added a test in tests/test_astrometry.py to make sure this didn't break coordinate conversions.


scottransom:
This is great, @rossjjennings !  Thanks so much!  I've been wondering about the PM aspects of astropy coordinates.  While you were looking at the astrometry code, did you get a sense of how hard it would be to completely switch over to using astropy PM stuff rather than our explicit PM functionality?

rossjjennings:
@scottransom The `params_as_ICRS()` method of `AstrometryEcliptic` was already using astropy proper motion internally. The main explicit PM functionality left is the epoch-updating functionality. Astropy 3.0 has its own `apply_space_motion()` for this, but we can't use it as long as we continue to support python 2.7.

rossjjennings:
The test for the `change_posepoch()` function was failing on Python 2.7 because the attribute I used to extract the proper motion in the test wasn't available in Astropy 2. I just added a commit that should fix that, so all the tests should pass except the ones in the tox build that are still failing due to Astropy 4.

rossjjennings:
@scottransom @luojing1211 @paulray What are your thoughts on merging this? I have some thoughts for building on it (related to #610 for instance), so it would be nice to get it merged soon. It's also substantial enough that we might want to discuss it more first, though.

scottransom:
I've done a quick look and I think it looks pretty good.  I'd be happy for it to be merged if @luojing1211  and @paulray are good.  Might be good for @demorest  to take a peek as well.

paulray:
I agree that it looks good.  Nice to see a test included!

luojing1211:
This looks good, I will merge it.

PR #612, opened 2020-02-20
Fix astropy4.0 timecolumn

luojing1211:
This PR works around the astropy time column lost location information problem. In astropy 4.0, astropy Time classes in the astropy table are stored as Time column, as default. This PR forces it make an object column instead.

### DIFF
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -743,7 +743,7 @@ class TOAs(object):
             self.table = table.Table(
                 [
                     np.arange(len(mjds)),
-                    mjds,
+                    table.Column(mjds),
                     self.get_mjds(),
                     self.get_errors(),
                     self.get_freqs(),
--- a/tests/test_toa.py
+++ b/tests/test_toa.py
@@ -1,13 +1,14 @@
 import unittest
-
 import astropy.units as u
 from astropy.time import Time
-import astropy.units as u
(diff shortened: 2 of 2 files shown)


scottransom:
Does this allow multiple different locations to be in the table?  We need to make sure that is still the case....

luojing1211:
This PR actually forces astropy4.0 to do what we have before. So it should. But I will add more tests on it.

paulray:
Yeah, the issue is probably that a Time() object can hold an array of times, but in that case it can't hold a different location for each time. We need to make sure that we keep each Time() object independent.  I haven't looked at your code, but maybe it does the right thing?

luojing1211:
Hi all, 
I think this PR is ready to get merged. @paulray @scottransom @demorest, could you check it out?

paulray:
Why are the tox tests still failing?

luojing1211:
@paulray There is more than one reason that caused tox failure. This is the biggest one. The other fix is done from PR #607 and I believe you have a photonphse fix as well. If we merge those three, then tox test should be fine.

paulray:
OK, maybe merge this and #607 then I'll rebase my PR and see if we can get it all passing!

luojing1211:
@scottransom What do you think?

scottransom:
@luojing1211   That's what I think!  :-)

PR #613, opened 2020-02-21
Make obliquity a QuantityAttribute

rossjjennings:
This addresses #610. Rather than make the `PulsarEcliptic.obliquity` a normal instance attribute, it makes it a `QuantityAttribute`, thereby using the same mechanism as Astropy's built-in coordinate frames like `BarycentricTrueEcliptic` (https://docs.astropy.org/en/stable/coordinates/frames.html#defining-frame-attributes). This means that printing a `PulsarEcliptic` object now shows the corresponding obliquity:
```
<PulsarEcliptic Coordinate (obliquity=84381.4059 arcsec): (lon, lat) in deg
    (244.34767784, -10.07183903)
 (pm_lon_coslat, pm_lat) in mas / yr
    (0.46, -7.16)>
```
It also means that it's easy to transform to and between `PulsarEcliptic` frames with different obliquities. `coords.transform_to(PulsarEcliptic)` uses the default obliquity value, and you can also transform to a `PulsarEcliptic` frame with a different obliquity value if you want by doing things like `coords.transform_to(PulsarEcliptic(obliquity=84381.406*u.arcsec))`. To make this more convenient, I added an optional parameter `ecl` to the `PulsarEcliptic` constructor, so you can also do things like `coords.transform_to(PulsarEcliptic(ecl="DE403"))`.

I've also changed the default `ECL` in `astrometry.py` to `IERS2010` to match the global default set in `ecliptic.dat`. This means that models based on par files with out `ECL` set (which is most of them that I've seen) will now use the IERS2010 value rather than IERS2003. This will affect comparisons with Tempo2.

### DIFF
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -459,7 +459,7 @@ class AstrometryEcliptic(Astrometry):
         the position at the given epoch.
         """
         try:
-            PulsarEcliptic.obliquity = OBL[self.ECL.value]
+            obliquity = OBL[self.ECL.value]
         except KeyError:
             raise ValueError(
                 "No obliquity " + str(self.ECL.value) + " provided. "
@@ -476,6 +476,7 @@ class AstrometryEcliptic(Astrometry):
             broadcast = numpy.ones_like(epoch)
 
         pos_ecl = PulsarEcliptic(
+            obliquity=obliquity,
             lon=self.ELONG.quantity + dELONG,
             lat=self.ELAT.quantity + dELAT,
             pm_lon_coslat=self.PMELONG.quantity * broadcast,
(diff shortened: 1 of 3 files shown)


rossjjennings:
It looks like the Travis CI build hasn't even started for this one? I have no idea why.

rossjjennings:
This was failing on Python 2.7 because the way I implemented the `ecl` parameter accidentally relied on [PEP 3102](https://www.python.org/dev/peps/pep-3102/)! The latest commit should fix that.

rossjjennings:
All tests are now passing except for the remaining Astropy 4.0 bugs in the tox build.

paulray:
That tox error in photonphase is a mystery!  I can't replicate it by running pytest or by running those command manually. It only happens when I run "tox -e py37".  Something is really odd about tox.

paulray:
Ah! It isn't tox.  When you run photonphase under astropy 4.0 it fails. I'll try to figure out why.

paulray:
This seems all good to me.  I'm fine with merging if others are. The build failures should all be resolved in master now.

rossjjennings:
Rebased on master just so I could see all the tests pass for once :smiley:

Issue #615, opened 2020-02-22
u.cycle is insane and causes astropy 4 to fail


paulray:
See my email to the list, but I think we should make Phases unitless instead of using u.cycle which has many unexpected behaviors. Currently it causes one of the tests to fail when using astropy 4.0 since the unit propagates into photonphase in a way it didn't use to

rossjjennings:
@paulray -- I've tried to reproduce the weird behavior you wrote about in your email, and found that it isn't the default behavior in astropy 4.0. In an astropy 4.0 environment, if I just open up `ipython` and run
```
import astropy.units as u
int(20*u.cycle)
```
I get a `UnitConversionError`. But if I `import pint.models` before the second line, I get `125` as output, with no error.

rossjjennings:
Aha! I think the culprit is `u.set_enabled_equivalencies(u.dimensionless_angles())` [here](https://github.com/nanograv/PINT/blob/a4b33eae5fce21a6010e8665bfeecf9caec8cf96/src/pint/models/stand_alone_psr_binaries/DDK_model.py#L13). This is the only place where it appears outside a `with` block.

paulray:
Glad this is fixed!  But, I think we should have a discussion at some upcoming PINT telecon about whether it might be better to remove the u.cycle units on Phases.

rossjjennings:
Should we close this now that #616 is merged?

paulray:
I still think we should hold it open pending a discussion on whether to stop using u.cycle at all for pulsar phases.

paulray:
Fix is #685

Zac-HD:
FYI you can [put a comment in your PR description which auto-closes linked issues](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) :slightly_smiling_face:

PR #616, opened 2020-02-23
Fix u.cycle bugs on astropy 4.0

rossjjennings:
This is the result of following up on the weird behavior @paulray noticed in #615. Certain operations on quantities with the `u.cycle` unit were causing them to be implicitly made dimensionless, i.e. converted to radians by multiplying by 2π. This turns out not to be an astropy 4.0-specific behavior, but it was causing bugs on astropy 4.0 because the `u.cycle` unit was being propagated further than it had been. I found that the weird behavior could be traced to [this line](https://github.com/nanograv/PINT/blob/a4b33eae5fce21a6010e8665bfeecf9caec8cf96/src/pint/models/stand_alone_psr_binaries/DDK_model.py#L13), which sets the `u.dimensionless_angles()` equivalency globally.

This PR gets rid of this behavior by deleting the offending line. However, doing this caused quite a few test failures due to the resulting stricter unit checks. The follow-up commits fix these failures, which turned out to have four different causes:

1. Some of the scripts, including `photonphase.py`, had relied on `np.where()` silently stripping the `u.cycle` unit from its arguments. I fixed these instances by explicitly stripping the units with a line like `phases=phases.to(u.cycle).value` before the call to `np.where()`.
2. Some derivative-calculating functions in the binary model classes relied on the `u.rad` unit being treated as equivalent to dimensionless units. I fixed these instances by wrapping the offending code in a `with u.set_enabled_equivalencies(u.dimensionless_angles())` block.
3. The phase-calculating functions in the `IFunc` and `Wave` model components explicitly converted dimensionless quantities (in radians) into the `u.cycle` unit. I fixed these by going directly to cycles, rather converting from radians.
4. The delay-calculating function in the `SolarWindDispersion` model component took the inverse cosine of a dimensionless `Quantity`, which in astropy 4.0 results in a `Quantity` with the `u.rad` unit. Without `u.dimensionless_angles()`, this lead to unit incompatibilities later. I fixed this by converting the dimensionless `Quantity` into a regular `np.ndarray` before taking the inverse cosine.

With these changes, all of the tests now pass on astropy 4.0! So I added a final commit removing astropy <4.0 requirement from `requirements.txt`, `requirements_dev.txt`, and `setup.cfg`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy>=1.11.0
-astropy>=2.0, <4.0
+astropy>=2.0
 scipy>=0.18.1
 jplephem>=2.6
 matplotlib>=1.5.3
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -3,8 +3,8 @@ pip>=9.0.1
 coverage>=4.3.4
 Sphinx>=1.8.5; python_version < "3.0"
 Sphinx>=2.2; python_version >= "3.0"
-astropy>=2.0,<4.0; python_version < "3.0"
-astropy>=3.2,<4.0; python_version >="3.0"
+astropy>=2.0; python_version < "3.0"
+astropy>=3.2; python_version >="3.0"
 #astropy-helpers>=1.3
 sphinx-rtd-theme>=0.1.9
 coveralls>=1.1
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ package_dir =
 include_package_data = True
 install_requires =
     astropy>=2.0; python_version < "3.0"
-    astropy>=3.2, <4.0; python_version >= "3.0"
+    astropy>=3.2; python_version >= "3.0"
     numpy>=1.11.0
     scipy>=0.18.1
     jplephem>=2.6
--- a/src/pint/mcmc_fitter.py
+++ b/src/pint/mcmc_fitter.py
@@ -291,7 +291,8 @@ class MCMCFitter(Fitter):
         """
         phases = self.model.phase(self.toas)[1]
         # ensure all positive
-        return np.where(phases < 0.0 * u.cycle, phases + 1.0 * u.cycle, phases)
+        phases = phases.to(u.cycle).value
+        return np.where(phases < 0.0, phases + 1.0, phases)
 
     def lnposterior(self, theta):
         """
@@ -641,7 +642,8 @@ class CompositeMCMCFitter(MCMCFitter):
             print("Showing all %d phases" % len(phases))
         else:
             phases = self.model.phase(self.toas_list[index])[1]
-        return np.where(phases < 0.0 * u.cycle, phases + 1.0 * u.cycle, phases)
+        phases = phases.to(u.cycle).value
+        return np.where(phases < 0.0, phases + 1.0, phases)
 
     def get_template_vals(self, phases, index):
         if self.templates[index] is None:
--- a/src/pint/models/ifunc.py
+++ b/src/pint/models/ifunc.py
@@ -11,7 +11,7 @@ class IFunc(PhaseComponent):
     """This class implements tabulated delays.
 
     These mimic a tempo2 feature, which supports piecewise, linear, and sinc
-    interpolation.  The implementation here currently only supports the 
+    interpolation.  The implementation here currently only supports the
     first two formulae.
 
     For consistency with tempo2, although the IFuncs represent time series,
@@ -29,7 +29,7 @@ class IFunc(PhaseComponent):
     0 == piecewise (no interpolation)
     1 == sinc (not supported)
     2 == linear
-    
+
     NB that the trailing 0.0s are necessary for accurate tempo2 parsing.
     NB also that tempo2 has a static setting MAX_IFUNC whose default value
     is 1000.
@@ -130,5 +130,5 @@ class IFunc(PhaseComponent):
         else:
             raise ValueError("Interpolation type %d not supported.".format(itype))
 
-        phase = ((times * u.s) * self.F0.quantity * 2 * np.pi).to(u.cycle)
+        phase = (times * u.s) * self.F0.quantity * u.cycle
         return phase
--- a/src/pint/models/solar_wind_dispersion.py
+++ b/src/pint/models/solar_wind_dispersion.py
@@ -62,7 +62,7 @@ class SolarWindDispersion(Dispersion):
             rsa = tbl["obs_sun_pos"].quantity
             pos = self.ssb_to_psb_xyz_ICRS(epoch=tbl["tdbld"].astype(np.float64))
             r = np.sqrt(np.sum(rsa * rsa, axis=1))
-            cos_theta = np.sum(rsa * pos, axis=1) / r
+            cos_theta = (np.sum(rsa * pos, axis=1) / r).to(u.Unit("")).value
             ret = (
                 const.au ** 2.0
                 * np.arccos(cos_theta)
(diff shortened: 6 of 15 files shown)


paulray:
Nice work!  

BTW, I think we should delete htest_optimize.py from PINT. It is a quick hack I made and isn't worth maintaining.

rossjjennings:
OK. I've been bold and added a commit removing it.

luojing1211:
Do we need htest functionality in PINT?

rossjjennings:
What do we mean by htest functionality?

luojing1211:
@paulray mentioned the htest_optimize.py is a quick hack. But should we have a more serious htest code in PINT? or should it be independent?

rossjjennings:
@luojing1211 I understood that much. I just don't know what "htest code" is supposed to do.

paulray:
Given that PINT is really useful for processing event mode data from many spacecraft, it is approppriate that we have H-test calculation routines (among others) available in pint.eventstats. However, the htest_optimize.py script was a quick hack that I wrote that wasn't being maintained and so I suggested deleting it.  So, with that deletion, I think we are in fine shape.  If someone else has feature requests it is fine to put them in as issues.

paulray:
Also, note that other higher-level manipulations of event data that make use of PINT are in packages such as NICERSoft and HENDRICS. This include TOA generation and pulsation searches.

luojing1211:
Let us remove it in a separate PR, so we can better track on it.

rossjjennings:
OK, I took the removal out of this PR.

paulray:
I guess we need to fix this conflict now that #618 has been merged.

rossjjennings:
I just rebased on master, skipping the commit that modified `htest_optimize.py`. That should take care of it.

rossjjennings:
The Python 2.7 build is now failing for a reason I don't understand. One of the precision tests failed -- it looks like it might just have randomly chosen a rare example that it happens to break on?

paulray:
Yeah, I don't understand it either. @aarchiba wrote those tests, so maybe she can help?
It says it is Flaky, but maybe one of the runs just timed out (as it tried to download an external file or something)? It suggests setting deadline=none but I don't know where that would be done.

paulray:
Or sometimes if you just re-run the test it passes!

rossjjennings:
Just made a do-nothing change to re-trigger the Travis build.

rossjjennings:
OK, still failing. So it's not *that* Flaky...

rossjjennings:
@paulray Did you manage to reproduce it locally? I don't have a working Python 2.7 environment with PINT installed.

paulray:
So, I just installed a py27 virtualenv and tried it two ways, with "make test" and with "tox -e py27".  I noticed a couple of errors/warnings in the install:

```
/Users/psray/src/PINT/.tox/py27/lib/python2.7/site-packages/hypothesis/extra/pytestplugin.py:61: HypothesisWarning: 
        You are using Pytest version 3.6.4.  Hypothesis tests work with any test
        runner, but our Pytest plugin requires Pytest  4.3 or newer.
        Note that the Pytest developers no longer support this version either!
        Disabling the Hypothesis pytest plugin...
    
  warnings.warn(PYTEST_TOO_OLD_MESSAGE % (pytest.__version__,), HypothesisWarning)
```

and 

```
ERROR: astropy 2.0.16 has requirement pytest<3.7,>=2.8, but you'll have pytest 4.6.9 which is incompatible.
```

Then, the "make test" passed the precision tests but had one fail in test_observatory:
```
../test_observatory.py:90: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../env/pint27/lib/python2.7/site-packages/pint/observatory/topo_obs.py:227: in clock_corrections
    clock_file, format=self.clock_fmt, obscode=self.tempo_code
../../../../env/pint27/lib/python2.7/site-packages/pint/observatory/clock_file.py:61: in read
    return cls._formats[format](filename, **kwargs)
../../../../env/pint27/lib/python2.7/site-packages/pint/observatory/clock_file.py:103: in __init__
    mjd, clk, self.header = self.load_tempo2_clock_file(filename)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

filename = '/Users/psray/src/tempo2-paulray/T2runtime/clock/fake2gps.clk'

    @staticmethod
    def load_tempo2_clock_file(filename):
        """Reads a tempo2-format clock file.  Returns three values:
        (mjd, clk, hdrline).  The first two are float arrays of MJD and
        clock corrections (seconds).  hdrline is the first line of the file
        that specifies the two clock scales connected by the file."""
>       f = open(filename, "r")
E       IOError: [Errno 2] No such file or directory: '/Users/psray/src/tempo2-paulray/T2runtime/clock/fake2gps.clk'
```
This is due to me having TEMPO2 installed, but I'm not sure why we have a test that fails looking for fake2gps.clk.

The tox run passed all tests.

So, I can't replicate the error either way.  It may be due to the fact that hypothesis is being disabled.  We may want to tweak the versions of packages being installed under py27.

paulray:
Ah, looking at the test_observatory fail, it looks like it is supposed to fail, but the test is looking for OSError and it is triggering IOError instead. Perhaps something having to do with me having Tempo2 installed is causing the difference.

paulray:
OK, I fixed test_observatory in my branch so it will test for both OSError and IOError.
That will remove the one issue.
But, I couldn't quite figure out a set of versions of hypothesis, pytest, astropy that all play happily together. astropy 2 wants an early version of pytest, while hypothesis wants a very recent version.  This makes it difficult to have consistent tests on py27.

paulray:
Should we merge this?  I wonder if for the time being we should restore the astropy<4.0 specifications?  Or are we really ready to allow astropy 4?

rossjjennings:
I finally got a chance to try to reproduce the test failure here in a working Python 2.7 environment today, and I can't. All the tests pass on this branch on my machine. Examining the Travis output more closely leads me to believe `test_time_from_mjd_string_accuracy_vs_longdouble()` was failing to meet its `hypothesis`-imposed deadline because astropy was fetching something over the network when it was first called, and doing that took longer than expected.

paulray:
I concur. That must be the issue.  Any idea how to extend the deadline?

rossjjennings:
We could turn off the deadline on this particular function by applying the `@settings(deadline=None)` decorator from `hypothesis`. It might not just be this function, though.

paulray:
Seems like it is worth a shot

rossjjennings:
The relevant part of the `hypothesis` documentation is [here](https://hypothesis.readthedocs.io/en/latest/settings.html).

Also, in response to your earlier comment, removing the astropy < 4 requirement is part of this PR, so it hasn't made it to the main branch yet.

rossjjennings:
I've added a commit extending the deadline to 1 second (from 200 milliseconds). Hopefully that fixes it. I'm still not sure why the issue only affects this branch, and only after I rebased. It can't be an astropy 4.0 thing, since it only happens on Python 2.7.

rossjjennings:
On whether we're really ready to allow astropy 4.0, I'd like to point out that all the Python 3 Travis builds of this branch are already using it, and have been passing for quite a while (even as Python 2.7 failed).

rossjjennings:
The extended deadline worked! All tests are now passing!

paulray:
Sweet!  I guess we can just take the plunge and enable astropy 4.0.  We can always back off if we find an issue.

scottransom:
Yeah.  Let's take the plunge and allow Astropy4 and merge this.

PR #617, opened 2020-02-24
Add test for change_posepoch() in equatorial coordinates

rossjjennings:
This fixes an oversight in #611. The test I added there only covered the case of ecliptic coordinates, and in fact the test failed when run on a par file with astrometry specified in equatorial coordinates. I've added the equatorial version of the test and made some small changes so that it passes.


rossjjennings:
This one is a fairly simple change and should be easy to merge. The changes I made to `AstrometryEquatorial.get_psr_coords()` are in order to
1. Make it always include proper motion in the returned coordinate object (like `AstrometryEcliptic` does now), and
2. Fix a bug where it overwrites the whole `Parameter` objects `model.RAJ` and `model.DECJ` with `Quantity` objects.

paulray:
Sounds good to me. Merging.

PR #621, opened 2020-02-28
Fixed wsrt observatory to use Tempo2 clock files

paulray:
Also fixed Tempo2 clock file reader to be more robust


paulray:
Should add a test, but I don't want to make travis depend on having Tempo2 installed.

rossjjennings:
For a test, maybe you could just add a clock file that might cause problems to `tests/datafile` and test reading it in?

paulray:
Good idea @rossjjennings. Done.

paulray:
Note that to successfully use WSRT with PINT, you will need to pull the latest Tempo2, since I had to patch the $TEMPO2/clock/wsrt2gps.clk file as it had a couple of lines with just "0" on them, which broke PINT's parser.

paulray:
I'm done with this and is passes all tests.  Is it OK to merge?

PR #624, opened 2020-02-28
Test against Python 3.8

rossjjennings:
Python 3.8 is now the default for new conda environments and we should probably be testing against it. I've run the tests in a local Python 3.8 environment, and they pass, but with more warnings than under 3.7.


paulray:
Sheesh, new versions keep coming fast and furious! Doesn't anyone value stability?

PR #630, opened 2020-03-07
Move parameter validations from setup() to validate()

luojing1211:
This PR moves the parameter validation from setup() to validate(). This will allow us to set up the timing model first (e.g., register derivatives), without error raise and then validate the parameters. 
One can do
``` 
tm = TimingModel('name',  [Astrometry(), Spindown()])
tm.setup() # setup the derivatives. In old method, this would fail since 'RA' and 'Dec' are not set.
tm.RA.value = '19:59:48'
tm.DEC.value = '20:48:36'
tm.F0.value = 1.6 * u.ms  
tm.validate() # Check parameters.
```
Or
```
tm = get_model('pulsar.par', validate=False)
tm.add_component(Jump())
tm.setup() # This would fail in the old setup()
tm.JUMP1.value = 0.3
tm.validate()
```

### DIFF
--- a/src/pint/models/absolute_phase.py
+++ b/src/pint/models/absolute_phase.py
@@ -42,7 +42,7 @@ class AbsPhase(PhaseComponent):
         )
 
     def setup(self):
-        pass
+        super(AbsPhase, self).setup()
 
     def validate(self):
         super(AbsPhase, self).validate()
--- a/src/pint/models/absolute_phase.py
+++ b/src/pint/models/absolute_phase.py
@@ -43,7 +43,7 @@ class AbsPhase(PhaseComponent):
 
     def setup(self):
         pass
-    
+
     def validate(self):
         super(AbsPhase, self).validate()
         # Check input Parameters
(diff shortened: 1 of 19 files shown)


paulray:
Should this or #585 be reviewed first? Is there a way to merge them into a single PR? It seems like they are tightly related.

paulray:
Why do a couple of the classes call super.validate() but the rest don't?  Is there something different about the ones that call validate on their superclass?

luojing1211:
PR #585 should be merged after this one. I thought about putting them into on PR, however, they are not totally related, and it is easier for me to run the tests. 

About the super().validate comments, some component is inherited from a base component which has validation function. That is why it is necessary to run super().validate. But I can add them to all of the components.

paulray:
Add .is_valid attribute to a timing model so you can check if it has been validated.
Also, have model.phase() verify is_valid before running

luojing1211:
As discussed in the telecon, I added validate() to all components. this can be merged.

PR #631, opened 2020-03-10
Add logo files

paulray:


paulray:
These are the final logo files in a variety of sizes and formats. Once this is merged, we can reference them from the various .rst files.

rossjjennings:
Once we have a follow-up PR to actually add the logos to the documentation, maybe we can close #444.

Issue #584, opened 2019-11-26
What is the minimum numpy required?


mhvk:
In testing in our `scintillometry` package, the runs with `numpy` at 1.14 stopped working with #549 by @aarchiba - I think this is not a problem and we needed to update our minimum version of numpy anyway, but I could't find what PINT's minimum numpy is (or, rather, IIRC, `requirements.txt` gives 1.11 or so and `requirements_dev.txt` doesn't list it).

mhvk:
Just to be a bit clearer: the use of `np.nditer` as a context manager is only possible since numpy >=1.15. That does seem a reasonable minimum version, but it would be good to be explicit about this (and have a travis run that *tests* with the minimum supported version!).

paulray:
Anyone know how to make a travis run that sets all the packages at the *minimum* supported version?

mhvk:
@paulray - over at astropy, we've just switched over to setting up environments with `tox` (which I see PINT has started to use as well), and in `tox.ini`, there's:

https://github.com/astropy/astropy/blob/424b3fb21f8cdf61db7fccb16ceee4c8bc99a80e/tox.ini#L49-L53

which then gets used in `.travis.yml` as:

https://github.com/astropy/astropy/blob/424b3fb21f8cdf61db7fccb16ceee4c8bc99a80e/.travis.yml#L79-L88

PR #638, opened 2020-03-12
Updates travis/tox to test at minimum supported versions

### LINKED ISSUE #632: Test against minimum allowed versions of packages
We should have at least one Travis build set up that runs PINT with the minimum allowed version of each of our dependencies. This will make us notice when a change breaks on an older version of (say) numpy or astropy, and prevent issues like #584 from cropping up. If we do find something that breaks on an older version, we can always bump up the minimum supported version of that package, which will reduce the chances of getting a broken PINT install.

As per the discussion on #584, we should bump the minimum numpy version to at least 1.15. In trying to get a working python 2.7 environment, I ran into problems with astropy 2.0.9 -- it looks like there were some bugs fixed in 2.0.15 that PINT now relies on. So on python 2.7 we should require astropy>=2.0.15. Running a travis build with minimum versions of everything will catch more issues like this that would otherwise hide from the tests.

paulray:
This updates our travis/tox testing to run tests on the minimum supported versions of numpy and astropy to ensure that new features don't break support for the older versions.
In order to make this work, I had to bump up the minimum version for numpy to 1.15.*

Fixes #632 and #584

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,6 +84,15 @@ tests/datafile/fermitest.png
 tests/datafile/photontest.fits
 tests/datafile/photontest.png
 tests/datafile/test_parfile_write.par
+tests/datafile/J0030+0451_chains.png
+tests/datafile/J0030+0451_post.par
+tests/datafile/J0030+0451_post.png
+tests/datafile/J0030+0451_pre.png
+tests/datafile/J0030+0451_prof_post.txt
+tests/datafile/J0030+0451_prof_pre.txt
+tests/datafile/J0030+0451_results.txt
+tests/datafile/J0030+0451_samples.pickle
+tests/datafile/J0030+0451_triangle.png
 fermitest.fits
 fermitest.png
 photontest.fits
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,7 @@ de-403-masses.tpc
 
 .idea
 .cache
+.vscode
 
 # Test output/intermediate files
 tests/datafile/fake_testzima.tim
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@
 
 language: python
 
+compiler: gcc
+
+os: linux
+
 matrix:
     fast_finish: true
     include:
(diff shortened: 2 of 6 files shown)


paulray:
Note that our use of travis and tox is a bit of an unholy mess. It would be nice to use tox for all our test runnings, following the nicely organized example of astropy.  This should be a separate issue/feature request.

rossjjennings:
It would be nice to have a Python 2 version of this test as well. I suspect it will show that we need to bump the minimum version of astropy to 2.0.15.

paulray:
True I'll add a Python2 version

paulray:
OK, I have tried but I can't make the python2 test work in travis-ci.  So, it is commented out of .travis.yml. If you want to run it on your local machine you can do `tox-2.7 -e oldestdeps2`.  That runs fine on my machine. But, it is not that interesting since the oldest deps under Python 2.7 are very close to the current versions, since astropy got stuck at 2.x under Python 2.7.

Therefore I propose that this be merged now. This bumps up the minimum versions for astropy and  numpy and puts in place a test that will ensure that (under Python 3.x at least) we don't introduce changes that break under the oldest supported versions.

paulray:
@scottransom or @luojing1211 if you are happy with this, please merge.

luojing1211:
So which version is the last py2.7 supported version? The original plan was PINT 0.7 the next one, should be the last one.

paulray:
That still sounds like a good plan to me.  When we release version 0.7, the one that matches the paper, we can then move forward and say that versions 0.8 and forward to NOT try to support Python 2.

luojing1211:
Should we merge this PR later, since it does not test PY2.7

paulray:
No, it DOES test Python 2.7, as it always did. What is missing is @rossjjennings request to ALSO test Python 2.7 with the **minimum** supported versions of numpy and astropy.  But, we've never had that anyway. This PR at leasts gives us that capability for Python 3.

luojing1211:
Ok, I saw the tests. I think we can merge it.

Issue #637, opened 2020-03-11
Add literature reference to this docstring


paulray:
https://github.com/nanograv/PINT/blob/dd333f67bce123d9b8863860ca0d850912ea9129/src/pint/models/solar_wind_dispersion.py#L19

PR #641, opened 2020-03-13
added solar wind reference

mtlam:
Solar wind reference added to the docstring to resolve issue #637. There are older ones but this is a rather thorough treatment.

### DIFF
--- a/src/pint/models/solar_wind_dispersion.py
+++ b/src/pint/models/solar_wind_dispersion.py
@@ -19,6 +19,10 @@ class SolarWindDispersion(Dispersion):
     The model is a simple spherically-symmetric model that varies
     only in its amplitude.
 
+    References
+    ----------
+    Madison et al. 2019, ApJ, 872, 150; Section 3.1
+
     """
 
     register = True
--- a/src/pint/models/solar_wind_dispersion.py
+++ b/src/pint/models/solar_wind_dispersion.py
@@ -22,6 +22,7 @@ class SolarWindDispersion(Dispersion):
     References
     ----------
     Madison et al. 2019, ApJ, 872, 150; Section 3.1
+    Edwards et al. 2006, MNRAS, 372, 1549; Setion 2.5.4
 
     """
 
@@ -54,6 +55,12 @@ class SolarWindDispersion(Dispersion):
     def solar_wind_delay(self, toas, acc_delay=None):
         """Return the solar wind dispersion delay for a set of frequencies
         Eventually different solar wind models will be supported
+
+        Implements equations 29, 30 of Edwards et al. 2006,
+        where their rho is given as theta here
+
+        rvec: radial vector from observatory to the center of the Sun
+        pos: pulsar position
         """
         if self.SWM.value == 0:
             tbl = toas.table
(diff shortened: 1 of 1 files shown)


paulray:
Thanks! That is helpful, but maybe I was too quick in adding my little issue.
My overall goal is to have equations in the mapped to specific literature references. So, while that one is good at the top level, could you also add some comments down around line 68 explaining that the equations being implemented here are Eqns (30) and (31) of Tempo2 paper II?

mtlam:
It's not immediately clear to me how the equations in `solar_wind_delay()` relate to those in the TEMPO2 II paper. The DM is proportional to ρ/(r sin ρ). Is θ defined differently? It sort of looks like it's the other side of the triangle but again it's not immediately clear. If this is okay, I can add some more comments in.

luojing1211:
Feel free to make PR for this.
-Jing

paulray:
This is a PR, @luojing1211 . What are you saying, exactly?  Are you saying the equations might be wrong?  How do they compare to the Tempo2 code, and how does that match the Tempo2 paper?

luojing1211:
Oops, I thought this is an issue. I can not really tell from my email.

mtlam:
I've not looked at the TEMPO2 code, just the definition they give, which I know has been verified by others. I'm not sure what `rsa` is supposed to stand for but it looks like the observatory-Sun vector, which is used to get the radial distance `r`. The variable `pos` presumably means pulsar-Observatory-Sun, which indeed means that θ is not the same as ρ, which is the pulsar-Sun-observatory angle. So I'd have to work through it to see if it matches. If anything, I can add in what those quantities mean in comments.

paulray:
@mtlam were you going to put in those comments on the definitions of the quantities before we merge this?

mtlam:
Yes, at some point... hopefully shortly. I'm still not sure what `rsa` means and was going to ask @luojing1211.

mtlam:
After digging in tempo2 upon suggestion, I found:
`/* Note rsa = vector from observatory to centre of Sun */`
It's used in a number of places. However, I would like to rename this (maybe some others but I'd have to look back), especially since it only appears in this block of PINT code. Can someone else confirm that is okay? I'll add in the appropriate comments, etc.

mtlam:
I have added some more references, renamed one variable, and added a little more documentation to hopefully make it clear that it does in fact relate to the equations given in the TEMPO2 paper.

paulray:
Hmm, lots of test failures even though this is just a few variables renamed. I think this is all due to astropy 4.0.1.

mtlam:
I think that's probably likely. The only variable I ended up renaming was `rsa` in the end; I just added comments to what the new one, `pos`, `theta` represented.

paulray:
OK, my suggestion is to split up these two issues.  Can  you add <4.0.1 to the astropy lines in requirements*.txt and setup.cfg and see if that makes all your tests pass. Then we should create a separate issue on astropy 4.0.1 breaking PINT.

paulray:
Actually, I forgot that the master branch already has the astropy<=4.0 so you can just rebase on master.

paulray:
Sorry to post again, but I should not have said rebase. It is bad to rebase after already pushing a branch. What you want to do is merge master into your PR, or we could just merge this PR and that will do it.  In fact, that is the easiest thing to do so I'll just merge this now.

demorest:
@paulray I already started #651 about astropy 4.0.1 issues.

Issue #642, opened 2020-03-14
Parameters that aren't there: default value vs. UNSET vs. AttributeError


rossjjennings:
Currently, writing PINT code for a situation where a model parameter may or may not exist is unnecessarily difficult, because the absence of the parameter might be handled in one of several different ways: it might be absent from the model object, so that asking for it raises an `AttributeError`, present, but with the value `UNSET`, so that asking for it as a `value` or `quantity` gives `None`, or present with a default value (usually 0). We should try to standardize this as much as possible. 

I would propose getting rid of `UNSET` entirely, so that looking for a parameter that hasn't been set always raises `AttributeError` unless there is a sensible default value. This might break enough existing code that it would be a pain to implement, though.

luojing1211:
Raise AttributeError will make the current code broken. The reason for a default value parameter is that some model that has some parameters we general does not set, but it is used in the calculation. For instance, in DD model, we do not use DR or DTH very often, but they are used in the calculation with zero value.

rossjjennings:
A particularly annoying instance of this that I ran into is `A1DOT`: If a `DD` binary model doesn't have an `A1DOT`, it is treated as 0, but if an `ELL1` model doesn't have an `A1DOT`, it is `UNSET`.

luojing1211:
It happens here. DD model sets the A1DOT in the model. But ELL1 may not.  But I don't think it is the problem of default value or unset. It is the convention about how we use them is not clear. https://github.com/nanograv/PINT/blob/0f6a1c933de85c235a6d6bd273b0b81492739db3/src/pint/models/binary_dd.py#L80

rossjjennings:
@luojing1211 I agree that it's nice to have some parameters be 0 by default. It's less nice when it only works for some parameters, and you have to sift through the code to figure out which. And having to remember which are `UNSET` and which will raise `AttributeError` is just annoying.

luojing1211:
By the way, where does this AttributeError coming from? Parameters are not added to the timing model?

rossjjennings:
It happens a lot for prefix parameters. For example, if `PBDOT` is not present, it will be unset (so that `model.PBDOT.value is None`), but if `FB1` is not present, it will raise an `AttributeError`. I think there are other cases, too.

luojing1211:
PBDOT should get default value 0. So the set up is PBX and FBX are from two classes for orbit, thus they can not show up together. If one uses PB, FBx would not be there.

rossjjennings:
[...]

Issue #639, opened 2020-03-13
Stand-alone binary model retains 10 day default FB0


rossjjennings:
I ran into the following alarming state of affairs while trying to add a function to change the binary epoch. 

Try loading `J0023+0923_NANOGrav_11yv0.gls.par` from the `tests/datafile` directory:
```
In [1]: from pint import models
In [2]: model = models.get_model("tests/datafile/J0023+0923_NANOGrav_11yv0.gls.par")

In [3]: model.FB0.quantity
Out[3]: <Quantity 8.33872155e-05 1 / s>

In [4]: model.binary_instance.FB0
Out[4]: <Quantity 1.1574e-06 1 / s>
```
That last value (approximately 1/(10 days)), is a default set by the base class for the stand-alone binary models (`pint.models.stand_alone_psr_binaries.PSR_BINARY`). It is being retained by the `binary_instance` object even though `FB0` is specified in the par file! Right now it's going unnoticed because whenever PINT looks for `FB0` it looks in the first place (`model.FB0`), and finds the correct value, as specified in the par file. But if someone were to pull the "stand-alone" binary model out and ask it for the value of `FB0` (i.e., `binary_instance.FB0`), they would mysteriously get the wrong answer.

We really shouldn't have this default at all -- it's a reasonable enough value that it could get past a cursory inspection and get used in later computations, even though it has nothing to do with user input or the contents of any par file. At the very least, it needs to be overwritten when the par file actually specifies `FB0`.

rossjjennings:
I plan on putting together a solution for this soon that should also address #633.

paulray:
Holy cow! That is scary!

demorest:
FWIW, I still think we should get rid of the "stand-alone" binary models entirely.

rossjjennings:
Why do we have the "stand-alone" binary models to begin with?

luojing1211:
Even we fixed the default problem, the cache is still there. The reason for it is that someone may just use the binary model for something else (like orbit searching).

rossjjennings:
For a minute last night I thought it would be easy to solve this and #633 together (hence my earlier comment), but I was wrong, so you can disregard my first comment on this. I do still need (or at least want) this to be solved before I add the code to change the binary epoch, and I'm willing to work on it myself, but the solution is not obvious.

luojing1211:
The solution is obvious, but it is very long and it will be a big PR. After PINT version 0.7. I am happy to work with you to solve it since I coded most of them.

[...]

PR #643, opened 2020-03-14
Add ability to change binary epoch

rossjjennings:
This is the last major piece of #491. For now I've implemented it at the "PINT wrapper class" level; i.e., you can't do it with a stand-alone binary model, for reasons related to #639. I ran into some annoyances around missing parameters that I detailed in #642. I could deal with them, but not without producing some fairly messy code -- take a look at the bit that gets `PB` and `PBDOT`. 

I'm also ignoring orbital frequency/period derivatives higher than the first in computing the new epoch. While this omission can lead to small inaccuracies, it doesn't affect models that don't have these higher derivatives (which is most of them), and even for the ones that do, the results should be fine as long as the model is refit afterwards.

I should also point out that I had to override the `change_binary_epoch()` function for ELL1 models, and a lot of code is duplicated between the two implementations. I don't know a good way to reduce this duplication, so I've left it as is.


rossjjennings:
I considered using the `model.binary_instance.orbits()` function, which takes into account all orbital frequency derivatives, to compute the orbital phase as part of finding the new epoch, but I also needed the inverse of this function, which is currently not implemented and would be quite a bit trickier to implement.

rossjjennings:
The initial test failures were due to PINT not recognizing the H3 line in one of my test par files. Maybe that needs to be its own issue. For now I've just commented out the line.

paulray:
There is an important question here.  Should these change epoch functions be such that the residuals before and after the change are identical (to machine precision)?  Many people may use them as such.  But, often it is useful to do the epoch change approximately then re-fit to get the precise values.  So, where possible, the unit tests should verify that the residuals don't change to some strict tolerance. Where that is NOT possible, the code should certainly emit Precision Loss warning of some kind, and maybe that should be an error unless the user specifies some approximate_ok=True flag or something like that.  

Note that form some of the rapidly evolving orbits, using the first order approximation may not come out very close at all.

rossjjennings:
@paulray I discussed this a little with @scottransom at the meeting, and came away with the impression that having some version of this functionality soon is more important than having it be perfect. But I agree that the more guarantees we can make around accuracy and precision, the better.

Do you have examples of rapidly evolving orbits that I could write tests against?

paulray:
Yes, I agree that something is better than nothing, but **definitely** have it emit a warning.

Here is a par file with a lot of orbital evolution:
```
PSR                            J2339-0533
EPHEM                               DE421
RAJ                     23:39:38.74105423
DECJ                    -5:33:05.10815967
PMRA                                4.147
PMDEC                             -10.311
PX                                    0.0 0 0.0
POSEPOCH           55792.0000000000000000
F0                   346.7133792204550673 1 2e-11
F1              -1.6942169221776437339e-15 1 3e-19
PEPOCH             55792.0000000000000000
START 54682
FINISH 58826
BINARY ELL1
A1                     0.6116660540086037 1 3.97439427207e-06
FB0             5.9938735816315316055e-05 1 1.1635526271371187127e-13
TASC               55791.9182103258863208 1 5.4132986898782e-07
EPS1                                  0.0 0 0.0
EPS2                                  0.0 0 0.0
FB3             -1.7795323933159306739e-33 1 1.8815636923779602566e-35
FB2             1.4553150100638518123e-26 1 3.3303916978609588776e-28
FB1               7.03725434744961224e-19 1 5.4332669868139948757e-21
FB7             -1.1468823953863174394e-62 1 1.5582489154671737184e-64
FB6             -4.2593887032115926134e-56 1 1.4278485036990097178e-57
FB5             5.3220305161746596376e-48 1 6.9531355833051827763e-50
FB4             -2.912644103931864879e-41 1 8.3840111451374732775e-43
FB9             -8.8475561365212462225e-78 1 1.0662311609509164363e-79
FB8              5.250316158222745923e-70 1 7.0874959224021488374e-72
FB10            2.2044791630302574554e-85 1 5e-86
FB11            -1.335419320262546109e-92 1 1e-93
FB12            2.807404079260208171e-100 1 1e-101
TZRMJD             56552.1042084326638773
TZRSITE                               bat
TZRFRQ                                0
UNITS TDB
```

scottransom:
@paulray That parfile is one which I do not think we should handle internally at all.  That's because polynomials are so specific to their ranges of validity.  And changing TASC for that model dramatically changes all of that stuff.  (Note:  what I would do for that is to record all my absolute rotation numbers for that model, change TASC and refit demanding that the rotation numbers between TOAs are the same.  And even that, given what order the polynomial is, it could fail).

I suggest that we be able to handle PBDOT and OMDOT (and maybe XDOT).  And anything beyond that we punt on, do the "DOT"s and say that to get a good fit the user needs to use absolute phases and a re-fit.  And we should *always* recommend that the user refits after the EPOCH changes.

paulray:
Yep, I know that parfile is a nightmare!  So, I think it is fine to handle a subset of cases, as long as make sure that users won't think PINT is doing the right thing on par files that it can't do the right thing on.

Issue #556, opened 2019-10-15
How do we handle IERS_B for current (or future) TOAs


paulray:
As of now (2019 Oct 15), the file `http://hpiers.obspm.fr/iers/eop/eopc04/eopc04_IAU2000.62-now` ends on 2019 Sep 15.  If I run PINT on my data, I get this exception:
```
  File "/Users/paulr/env/pint/lib/python3.6/site-packages/pint/erfautils.py", line 28, in get_iers_b_up_to_date
    raise ValueError("IERS B data not yet available for MJD {}".format(mjd))
ValueError: IERS B data not yet available for MJD 58769.70234516895
```
So, how should we handle current or future (e.g. fake) TOAs?  The quantities in the IERS_B table are need for any `TopoObs` to compute the position and velocity at a given time. Is there an extrapolated product we could use?

luojing1211:
This could be an astropy issue? Maybe we should make an issue on astropy.

-Jing

paulray:
No, this doesn't seem to have anything to do with astropy.  This is all in erfautils.py.

luojing1211:
It looks like the IERS A table is a solution. It updates daily and provides the prediction for the next year. 
see : https://astroplan.readthedocs.io/en/latest/faq/iers.html and https://www.iers.org/IERS/EN/Publications/Bulletins/bulletins.html

aarchiba:
Currently we use only IERS B values, because they are at least in principle
final. The code should download the most recent IERS B table whenever you
supply data more recent than is available, but B does not contain
predictions and is only issued about once a month. So if you have really
recent data, we don't yet have a very good answer for you.

Also, not all of PINT uses this recent table; certain conversions are not
readily under our control and may use a different table.

Anne

aarchiba:
The IERS A table includes values predicted by a month or two; it could be used. There will also be circumstances where people want to do calculations going forward for which there might not even be predicted values, or observatory clock corrections. Obviously in such a situation one should not expect microsecond accuracy but it is useful to be able to use it at all. We can and do emit warnings but we already emit so many warnings in ordinary circumstances that users just ignore them; they are ineffective.

Let me suggest that it be possible to set flags on TOAs, Model, and Residual objects to indicate that the results contain extrapolated or otherwise approximate answers. These would then be included in outputs of things like par files or nicely-formatted timing tables, so that no one could accidentally use extrapolated values in precision work.

luojing1211:
[...]

PR #644, opened 2020-03-18
Switch to using astropy for posvels

paulray:
This is Work In Progress.

I'm making this PR so people can easily check it out and test it.

See the email to the pint list.

The only change here is that it uses astropy to compute the PosVels instead of the manual code in erfautils.py


paulray:
The idea here is to fix #556

paulray:
**Question**: What does Tempo2 do?
**Answer**: 
In eop.C it loads the file $TEMPO2/earth/eopc04_IAU2000.62-now
That file comes from here: <http://hpiers.obspm.fr/eop-pc/index.php?index=FTP&lang=en>
These data are "C04 combined EOP series consistent with ITRF 2014"
This dataset is described in this paper: <http://hpiers.obspm.fr/iers/eop/eopc04/C04.guide.pdf>
It is valid from 1962 to "now" where now is one month before the current date.
This looks like it is a specially reprocessed dataset and not precisely the same as IERS B, but I think it should be quite close (TODO: check that).
If you ask Tempo2 for a value beyond the table, it looks like get_EOP() just gives you the last value in the EOP table and emits a warning.

demorest:
I was going to reply to the email about what tempo/tempo2 do.  As @paulray says tempo2 uses that IERS C04 file.  At least for modern times I think this is effectively identical to Bulletin B, I spot-checked a few recent dates and they were the same.

Tempo itself does not explicitly use any specific data source, it is up to the people who maintain `ut1.dat` to decide what goes in it.  In recent years, this is me, and I've been using the IERS C04 data as well, with no rapid or extrapolated values.  So this should be consistent with tempo2.

paulray:
Next question: What does astropy use?
For IERS_A and IERS_B, you can just look at the URLs:
```
In [13]: IERS_A_URL                                                                                                                                                                                                  
Out[13]: 'ftp://cddis.gsfc.nasa.gov/pub/products/iers/finals2000A.all'

In [14]: IERS_B_URL                                                                                                                                                                                                  
Out[14]: 'http://hpiers.obspm.fr/iers/eop/eopc04/eopc04_IAU2000.62-now'
```
Notice that the IERS_B is exactly what Tempo2 uses!

And for the "IERS_A", you can find a description of finals2000A.all here:
https://datacenter.iers.org/versionMetadata.php?filename=latestVersionMeta/9_FINALS.ALL_IAU2000_V2013_019.txt
In part, it says "finals2000A.all contains the values from IERS Bulletin A for x/y pole, UT1-UTC, LOD, dX, dY, their errors and predictions for next 365 days for x/y pole, UT1-UTC, dX and dY as well as the values from IERS Bulletin B for x/y pole, UT1-UTC, dX, dY at daily intervals since 02. January 1973.
Celestial pole offsets (dX, dY) are related to the IAU2000A precession/nutation theory."
So  it is not really IERS_A, but rather a mix depending on the date.  We should check to see how close old dates are.

But what does `IERS_Auto` do? That code is kind of complicated, but I think in general it should use IERS_B (that is the eopc04 data) where available and then tack on the extrapolation from IERS_A when needed and it makes sure to download the latest files when needed.  It can optionally return status codes that are defines like this:
```
# Status/source values returned by IERS.ut1_utc
FROM_IERS_B = 0
FROM_IERS_A = 1
FROM_IERS_A_PREDICTION = 2
TIME_BEFORE_IERS_RANGE = -1
TIME_BEYOND_IERS_RANGE = -2
```

So, we just need to make sure we understand which IERS table astropy is using, and then confirm that for older times we get the same answers as Tempo2, which we should since it seems astropy is interpolating into the same table!

paulray:
Just adding a bit more on Tempo2. This is from the docs, but I think the C05 should be C04. That must have changed since this was written:

> 4.2 Earth Orientation Parameters
To compute the Roemer delay, the position of the observatory must be known. This depends not only on the Earth’s orbit, but on the Earth’s orientation and rotation. The necessary parameters are obtained by interpolation of the “C05” series of Earth Orientation Parameters (EOPs) from the IERS. The file $TEMPO2/earth/README specifies the web address for downloading the latest EOPs. The user may optionally select to emulate the algorithm of tempo (which neglects polar motion and uses an out of date precession/nutation model) for transforming the observatory coordinates to the celestial frame, using the T2CMETHOD parameter; in this case $TEMPO2/clock/ut1.dat (in the same format as the corresponding file for tempo) is used.

So, for a proper comparison make sure T2CMETHOD is IAU200B in all your Tempo2 par files!

I have written a script to compare various astropy methods against Tempo2. Plots coming shortly.

paulray:
OK, I wrote a couple of codes. One is a C code that calls tempo2's `get_EOP()` to get their dut1 and xp, yp (polar motion coordinates). The other is a python script that uses astropy IERS tables to get the same quantities. These have 3 sources (IERS_A, IERS_B, IERS_Auto).  The attached 2 figures show the comparison.  It all looks pretty good. As expected (since they are from the same source), Tempo2 and astropy's IERS_B match precisely (except for slight differences in the extrapolation). But, we want to be able to get reasonable values going forward using the IERS A data and predictions. This is supposed to be exactly what IERS_Auto does. It is supposed to fold in the IERS_B values where they exist into the IERS_A table then use that. However, as you can see, the are some small discrepancies during the time when IERS B is available, which I don't understand.  If anyone understands this better, let me know (@demorest @mhvk).  What I hope is that we can just use IERS_Auto and be done with it. Thoughts?

[compare_dut1.pdf](https://github.com/nanograv/PINT/files/4361274/compare_dut1.pdf)
[compare_xp.pdf](https://github.com/nanograv/PINT/files/4361275/compare_xp.pdf)

mhvk:
@paulray - this is encouraging. One thing that may not be quite obvious about IERS_Auto is that, where possible, it replaces the IERS B values as listed in the IERS A tables with IERS B values taken from the actual IERS B files. Unfortunately, these are not always the same (we tried to find out but never got a clear answer; indeed, worse, I would have thought IERS B values would stay consistent from issue to issue, but that is also not true: apparently, some amount of reanalysis is done, and this affects even older values). Might this explain the differences? It may be good to just pick a specific time point and simply look up the actual values.

p.s.  There is also a long-standing issue to improve interpolation. Right now, like tempo, astropy interpolates linearly between entries, but one can do better; see https://github.com/astropy/astropy/issues/1803. Anyway, since right now it is the same, that should not affect your comparisons.

mhvk:
Another postscriptum: we have a specific issue about possible combinations of IERS A and B a user might want (https://github.com/astropy/astropy/issues/9227; raised following comments by @aarchiba). PINT is definitely one of the drivers for our quest for precision, so requests for improvements are definitely welcome (and PRs even more, though as @aarchiba found, it is not always so easy...)

paulray:
> where possible, it replaces the IERS B values as listed in the IERS A tables with IERS B values taken from the actual IERS B files

Yeah, that is what I thought was happening, and why I'm puzzled that there are differences between IERS_Auto and IERS_B during the times when IERS_B is valid.

paulray:
Looking at the plots, the differences are of order 100 us in DUT1.  Since this is the rotation of the Earth, this should make a difference in the observatory position of 100 us * (rotational velocity of Earth), which I think is about 4 cm, so that is quite small.

mhvk:
Another hitch might be that astropy replaces the IERS B values with those from the IERS B file stored under astropy/utils/iers/data; is that the same?

Indeed, IIRC, this was one of the complaints of @aarchiba! clearly we should allow the user to download their own more up to date IERS B file and use that. 

If these files are not the same, I think right now it is possible if you open your own favourite IERS B file first thing (before anything else in astropy), and then open IERS_Auto (can check if need be).

mhvk:
The differences suggest it may not matter in practice, but we might as well understand it! And, if my hunch above is right, I think it should be possible to have a more reliable work-around than the order-of-loading above (and of course fix it in astropy eventually).

paulray:
Ding! Ding! Ding! @mhvk wins a prize!  The difference between IERS_Auto and IERS_B is just the reprocessing that happens between releases of the C04 product. IERS_B was getting a new version of that file, while IERS_Auto was using the one embedded in astropy.  I could not figure out in the IERS_Auto API how to tell it to use a fresh IERS_B, so I just replaced the version inside astropy in my virtualenv. Now running my script shows everything that IERS_Auto matches both IERS_B and Tempo2 for the period when IERS_B is available then nicely transitions to IERS_A and then extrapolates. In other words IERS_Auto does the *right thing*.  And those small differences are just what you get from the reprocessing they do between releases.

So, PINT can happily use IERS_Auto and all we need to do is figure out how to prod it to update its IERS_B table.

scottransom:
Nice @paulray and @mhvk !  That's great!

paulray:
The key astropy code to fix is this in iers.py:
```
    def _substitute_iers_b(cls, table):
        """Substitute IERS B values with those from a real IERS B table.
        IERS-A has IERS-B values included, but for reasons unknown these
        do not match the latest IERS-B values (see comments in #4436).
        Here, we use the bundled astropy IERS-B table to overwrite the values
        in the downloaded IERS-A table.
        """
        iers_b = IERS_B.open()
```
Here is the link in context:
https://github.com/astropy/astropy/blob/eb9445eba6fa399ccfde2314a5e7006becb7d1b3/astropy/utils/iers/iers.py#L795

This needs an additional option to be able to specify that it should read the IERS_B_URL instead of just using the local version.  If we could just stick in there  `file=download_file(IERS_B_URL,cache=True)` that should do it, I think, but that needs some more logic so that is an option and it would be better to not use cache=True but rather check it the cached file is older than 30 days and only do the download if so.  Perhaps the IERS_B class generally should have an initialization option specifying whether to try to download the URL rather than just going to the internal version?  Then `IERS_Auto` could give that option to `IERS_B.open`, maybe something like `IERS_B.open(update=True)`

mhvk:
So, as a work-around `IERS_B.open()` will not do anything if a file was already loaded, so if one preloads it with the right file, it will work. In a new session, this would be:
```
from astropy.utils import iers
iers_b = iers.IERS_B.open(iers.IERS_B_URL)
iers_auto = iers.IERS_Auto.open()
```
From then on, I think it should be fine.

Of course, within PINT, you cannot know whether the user has already loaded `IERS_Auto()` before importing pint, but one can do something along the lines of
```
IERS_Auto.iers_table = None
IERS_B.open(iers.IERS_B_URL)
IERS_Auto.open()
```
to force a recalculation. (Though this particular stanza forces the user to download the IERS B time everytime PINT starts, which would seem overkill...)

mhvk:
p.s. Essentially, in astropy, we need an `IERS_B_Auto`, or force `IERS_Auto` to also update `IERS_B` as needed.

paulray:
OK, so maybe in `pint.__init__` we should add a call to 
```
from astropy.utils import iers
iers.earth_orientation_table.set(None)
```
to ensure we are using IERS_Auto for our computations. This same place could do a check of the age of the IERS_B cached file and download a fresh copy if it is old.  I'll try to implement that.

mhvk:
@paulray - hey, I had forgotten about introducing the flexibility of `earth_orientation_table`! The plan sounds good. It might be good to emit a once-only warning if the table is being updated.

src/pint/erfautils.py
```diff
@@ -6,18 +6,43 @@
...
 
 from pint.pulsar_mjd import Time
 from pint.utils import PosVel
 
-__all__ = [
-    "get_iers_b_up_to_date",
-    "gcrs_posvel_from_itrf",
-    "astropy_gcrs_posvel_from_itrf",
-]
+__all__ = ["get_iers_b_up_to_date", "gcrs_posvel_from_itrf"]
```
mhvk: Include `get_iers_up_to_date` (or replace `get_iers_b_up_to_date` with it)

paulray:
OK, the only test that is still failing is the oldest dependencies test:
https://travis-ci.com/github/paulray/PINT/jobs/318587227

It seems to be something with the derivatives. Anyone have ideas? @luojing1211 ?

luojing1211:
This is super interesting, I will take a look.

mhvk:
@paulray - this looks really nice. Obviously, arbitrarily increasing tolerances is worrying! I think part of the problem is that the test involved seems to be at a level so high that one has little idea where the actual problem originates. It may be good to open an issue to greatly expand unit tests... Having seen @aarchiba's work on [testing astropy's `Time` class](https://github.com/astropy/astropy/pull/9532), I've also become more convinced that it should be a mixture of well thought-through tests and using `hypothesis` to check that assumptions about what the code does actually hold.

paulray:
Yes, I completely agree. Setting well justified testing tolerances is a separate issue and an important one. I think we should merge this PR to get IERS_Auto into PINT, and open up that issue on testing. I haven't looked into `hypothesis` yet, so I don't know what it does, just that in PINT it is responsible for a lot of test failures that are hard to comprehend, especially all the flaky errors that were caused by timeouts.

luojing1211:
I  was trying to track down why this change causes an increase in derivative comparisons. I have not found anything yet. It is so hard to track down the sources. On the other hand, I asked myself how much tolerance should we assume for the derivative functions? TEMPO/TEMPO2 do not give super precise derivative functions, are we overkill? I don't have the answer yet.

paulray:
Can we merge this now? It is useful for all people using recent TOAs and we've spawned issues for the testing problems that were discovered.

luojing1211:
Is there a test with a TOA which happened "right now"? Like TOA(astropy.time.now()) just make sure it will not fail when we have very recent TOAs.

paulray:
Ah, good idea!  I'll add that.  In fact, IERS_Auto has predictions up to a year in the future as I recall.

mhvk:
¡Hurray!

PR #646, opened 2020-03-24
Fix jump toa select

luojing1211:
This PR fixes the bug mentioned in #596 . Before, PINT can not work with the JUMPs with their key value in the upper case.  This PR fixes the issue. I saw PINTK modified that part of code. @clp3ef @scottransom can you make sure this PR does not break PINTK?


clp3ef:
Hi Jing, 

I just checked, and these changes don't cause any issues in pintk as far as I can tell. Thanks for checking!

luojing1211:
Glad to know! I think we can merge this. Please also check out the add
component PR, see if that improves you JUMP adding.
-Jing

Issue #600, opened 2020-02-12
Make DMX_parse.py equivalent in PINT


scottransom:
Make equivalent functionality from tempo utils in PINT.

scottransom:
This is the same as #490

Issue #490, opened 2019-08-08
Equivalent dmxparse.py script


mtlam:
As in TEMPO, which has a dmxparse.py script which will report reduced DMX-parameter uncertainties with respect to the systematic offset from the mean DM, it would be useful for a utility which would take the parameters and the covariance matrix and produce similar output. This should be done within the function calls rather than requiring a script to redirect to a text file. Personal preference is to do output the same format as TEMPO with a *.dmx filename extension as default.

bshapiroalbert:
I've done a little work on this and have some functions written that use the PINT `toa` and `fitter` objects, and the covariance matrices to get the DMX values and uncertainties in the same way that `dmxparse` does, at least as far as I can tell. I've compared results on a few test cases and it seems to be consistent with results from `dmxparse`. I've attached a python script (as a text file cause it doesn't like .py files) that has the two functions in it here, if it's helpful for adding this feature in. [brent_utils.txt](https://github.com/nanograv/PINT/files/4199750/brent_utils.txt)

PR #647, opened 2020-03-25
Added dmxparse function to utils

bshapiroalbert:
Added a dmxparse function to PINT utils that does the same things as the current dmxparse function with TEMPO outputs, but using the PINT fitter class. Output is a list of the mean subtracted DMX values and a list of the DMX variance errors. Optional output writes a text file of the same style as dmxparse traditionally output. This addresses issues #600 and #490 as part of the updates to get PINT ready for the NANOGrav 15 year dataset.

This also includes a test for the dmxparse function in the test_utils.py test that I was having issues getting to pass. Any comments on formatting, the test, etc. are appreciated.


src/pint/utils.py
```diff
@@ -434,3 +435,113 @@ def show_param_cov_matrix(matrix, params, name="Covaraince Matrix", switchRD=Fal
...
+            dmx_epochs += 1
+    # Get DMX values (will be in units of 10^-3 pc cm^-3)
+    DMX_keys = []
+    DMXs = []
+    DMX_Errs = []
+    DMX_R1 = []
+    DMX_R2 = []
+    DMX_center_MJD = []
+    for ii in range(dmx_epochs):
+        ii += 1
```
scottransom, line 463: This is kind of a dangerous line here since ii is already being defined in the range() loop above.  If you want ii to start with 1, then do for ii in range(1,dmx_epochs+1)

bshapiroalbert: Thanks Scott, I'll fix that up too and generally try to fix it up if there are other issues.

bshapiroalbert:
All check passed, and the `exec` commands have been removed, as well as the poor indexing noted by Scott above. I appreciate any other comments on the code or fixes that should be done, just let me know, but otherwise I think it's working and should be ready to go.

paulray:
Ack, somehow this PR now is reverting all the recent documentation changes I made.
Can you remove all the .md and .rst files from this PR? I'm not sure the best way to do this, however.

bshapiroalbert:
Ah, oops that's no good, sorry, I must have forgotten to update my branch after your last PR Paul, I will fix that now, I think I probably just need to update my branch and then push those changes...

paulray:
Much cleaner, thanks!

bshapiroalbert:
Okay, I think those issues are fixed now, the only files that are changed are the correct ones, utils.py and test_utils.py, and all tests are passing. Let me know if there are any other issues though!

src/pint/utils.py
```diff
@@ -434,3 +435,115 @@ def show_param_cov_matrix(matrix, params, name="Covaraince Matrix", switchRD=Fal
...
+    DMX_R2 = []
+    DMX_center_MJD = []
+    for ii in range(1, dmx_epochs + 1):
+        if ii < 10:
+            ext = "000%s" % (ii)
+        elif ii >= 10 and ii < 100:
+            ext = "00%s" % (ii)
+        elif ii >= 100 and ii < 1000:
+            ext = "0%s" % (ii)
+        DMX_keys.append("DMX_%s" % (ext))
```
paulray, line 471: This can be written more concisely like this:
```
    DMX_keys.append("DMX_{:04d}".format(ii))
```
and similarly for the next few lines that use `ext`

src/pint/utils.py
```diff
@@ -434,3 +435,115 @@ def show_param_cov_matrix(matrix, params, name="Covaraince Matrix", switchRD=Fal
...
+    DMX_center_MJD = np.array(DMX_center_MJD)
+
+    # now get the full parameter covariance matrix from pint
+    # NOTE: we will need to increase all indices by 1 to account for the 'Offset' parameter
+    # that is the first index of the designmatrix
+    fparams = fitter.get_fitparams().keys()
+    params = []
+    for p in fparams:
+        params.append(p)
+    params = np.array(params)
```
paulray, line 492: I don't think you need to loop here. How about just this:
```
params = np.array(list(fparams))
```

src/pint/utils.py
```diff
@@ -434,3 +435,115 @@ def show_param_cov_matrix(matrix, params, name="Covaraince Matrix", switchRD=Fal
...
+    m = np.identity(n) - np.ones((n, n)) / float(n)
+    cc = np.dot(np.dot(m, cc), m)
+    DMX_vErrs = np.zeros(n)
+    # We also need to correct for the units here
+    for i in range(n):
+        DMX_vErrs[i] = np.sqrt(cc[i, i])
+    # Check we have the right number of params
+    if len(DMXs) != len(DMX_Errs) or len(DMXs) != len(DMX_vErrs):
+        print("ERROR! Number of DMX entries do not match!")
+        exit()
```
paulray, line 520: I think we should avoid print statements and exits in PINT code.  The following might be better:
```
from astropy import log #  Put at beginning of file
log.error("ERROR! Number of DMX entries do not match!")
raise RuntimeError('Number of DMX entries do not match!')
```

src/pint/utils.py
scottransom: Nice.  I was going to make the same comment, but you beat me to it.

bshapiroalbert:
Thanks for the comments, I really appreciate them. They have been added and the changes pushed.

paulray:
Looks ready to merge to me!

bshapiroalbert:
I've changed the output format to a dictionary similar to #656 as suggested earlier today. This returns all the same information that is output in the text file, so you get the same output just two different ways now.

As an aside, there is one minor difference here from the tempo dmxparse function, which is that values of DMXF1/2 are not returned, because they are not parameters in the timing model. I think this is because they are not defined as parameters in the dispersion_dmx class, but I was wondering if there was any particular reasoning for this. As the return is a dictionary and those values are not needed for the calculation, it would be easy to add this in later, I was just wondering.

Issue #602, opened 2020-02-12
PINT should be able to output whitenened residuals


scottransom:
Given a red noise model, PINT should be able to output the whitened resids the same way that TEMPO does in the resid2.tmp  (print_resid does the subtraction of reddened and whitened residuals)

demorest:
I will work on this, probably starting next week.  I'd originally thought it made sense for `GLSFitter` to output the noise realizations back to the relevant `NoiseComponent` classes in the timing/noise model.  However, since the noise realization depends on both a specific model and specific TOAs, I now think it makes more sense to add this to `Residuals`.  So `Residuals` will get some extra attributes that hold the noise realization for each noise component type after running a GLS fit.  Let me know if anyone sees any issues with this approach.

demorest:
Addressed in #648

PR #648, opened 2020-03-26
Residual whitening

demorest:
This addresses issue #602, allowing for output of whitened residuals.  I still need to add some tests, so don't merge yet.  But let me know if you see any issues with the approach:   After running a GLS fit, the resulting `Residuals` will have a new attribute `noise_resids` that is a dictionary containing the residuals corresponding to each noise component.  Computing "whitened" residuals is then done by subtracting the red noise component from the full residuals.


demorest:
Well, I added a test but the results are weird:
- in python 2.7 everything passes.
- in python 3.5 everything except the new GLS test passes.
- in python 3.6, 3.7, 3.8, the GLS test passes but a number of unrelated things fail.

The test compares PINT-whitened residuals with a pre-computed set of tempo-whitened residuals.  These agree well (to within 10 ns RMS and 50 ns max difference) except in the python 3.5 test.  In that case, I was able to reproduce the failure locally, the difference between the two is a large 1 us peak-to-peak parabola.  Must be somehow related to different versions of dependencies but it's such a simple calculation that it's hard to see what could be going wrong (that would not also cause many other tests to fail).  Any ideas?

The python >=3.6 failures look like some new astropy incompatibility.  The Travis test was using astropy 4.0.1 which just came out a few days ago.  Everything passed on my system using python 3.6 and astropy 4.0.

demorest:
Discussion from this morning: Change PINT python requirement to 2.7, >=3.6.  Temporarily change astropy requirement to <=4.0 until the 4.0.1 issues are fixed.  I'll put these in later today.

demorest:
For the record, the versions of numpy (1.18.2) and scipy (1.4.1) were identical in the failing py3.5 environment and the working py3.6 one.  Aside from astropy (3.2.3 vs 4.0) I don't see any other package differences that sound like they should be meaningful.

demorest:
OK, seems passing now.  I'll merge this later today unless there are any final comments.

PR #655, opened 2020-04-06
New option and bug fix to photophase.py

juliadeneva:
photonphase.py: Added a --addorbitphase option that will add a column ORBIT_PHASE to the photon event file. Fixed a bug that made the options --barytime and --absphase have no effect unless --addphase was also used.


src/pint/scripts/photonphase.py
```diff
@@ -168,6 +174,14 @@ def main(argv=None):
         )
         raise ValueError("Model missing AbsPhase component.")
 
+    if (args.addorbphase is not None) and (not hasattr(modelin,'binary_model_name')):
```
paulray, line 177: The default for args.addorbphase is False not None

Issue #652, opened 2020-04-06
config.datapath() did not return None when file is missing


luojing1211:
config.datapath() should return None when the file is missing, according to the docstring, however it raises an error. This makes the searching for bipm clock correction file fail. What is the right way to do it? None vs Error? 

https://github.com/nanograv/PINT/blob/78596924eb36dbb31e74b8fafb53b4c9cd7c08bb/src/pint/config.py#L36

demorest:
As I just said on the call, I'd vote for leaving the behavior of `config.datapath()` as it is, ie raise an error rather than return None.  We should change the docstring to correctly describe this behavior though.

paulray:
I just ran into this problem when loading TOAs with `bipm_version="BIPM2017"`. That is supposed to try PINT first then fall back to $TEMPO2/clock if that is available. Instead the exception causes it to die. So `bipm_fullpath()` in topo_obs.py needs to be fixed to use try/except if we keep this, as @demorest suggested.

paulray:
I made this fix in my copy and it works fine. I guess I'll make a PR for it at some point

scottransom:
If it fixes the issue, feel free to to just push it up as a bugfix

paulray:
Oops, I changed my mind. Looking at where `datapath()` is used in PINT, I think it should return None instead of raising an exception.

paulray:
This has been dealt with.

PR #663, opened 2020-04-16
Make datapath return None if file not found

paulray:
This is to address issue #652

### DIFF
--- a/src/pint/solar_system_ephemerides.py
+++ b/src/pint/solar_system_ephemerides.py
@@ -52,7 +52,12 @@ def _load_kernel_link(ephem, link=None):
     if ephem in _ephemeris_hits:
         # If we found it earlier just pull it from cache
         coor.solar_system_ephemeris.set(_ephemeris_hits[ephem])
-        log.info("Set solar system ephemeris to link {}".format(_ephemeris_hits[ephem]))
+        # Don't use log.info since this is using a cached version. No need to say it again
+        log.debug(
+            "Set solar system ephemeris to (cached) link {}".format(
+                _ephemeris_hits[ephem]
+            )
+        )
         return
 
     # FIXME: is link supposed to be a URL for the file or a directory?
@@ -67,7 +72,9 @@ def _load_kernel_link(ephem, link=None):
             coor.solar_system_ephemeris.set(ephem_link)
             _ephemeris_hits[ephem] = ephem_link
             log.info(
-                "Set solar system ephemeris to link {}".format(_ephemeris_hits[ephem])
+                "Set solar system ephemeris to link:\n\t{}".format(
+                    _ephemeris_hits[ephem]
+                )
             )
             return
         except (ValueError, IOError) as e:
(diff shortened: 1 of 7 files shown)


paulray:
Should really add a unit test to verify that datapath does what we expect when the file is found and when it is not.

demorest:
So the reason I didn't want to do this is that all the places in the code where `datapath()` is used now need to be updated to check for `None` and do something sensible in that case.  Previously, when `datapath` raised an exception we would at least get a comprehensible error message.  Now we'll get some cryptic message about not being able to do some operation on `NoneType`.

demorest:
So the reason I didn't want to do this is that all the places in the code where `datapath()` is used now need to be updated to check for `None` and do something sensible in that case.  Previously, when `datapath` raised an exception we would at least get a comprehensible error message.  Now we'll get some cryptic message about not being able to do some operation on `NoneType`.

paulray:
Yeah, I thought that too, but several of them are like this:
```
    search_list = [custom_path, datapath(ephem_bsp)]
```
I think this kind of construct will work fine if it returns `None` but would need to be rewritten if it  raises an exception.

The other places in special_locations.py and topo_obs.py already had logic that was expecting `None` as a return value, e.g.:
```
        fullpath = datapath(fname)
        if fullpath is not None:
            return fullpath
```

On the other hand in pulsar_obliquity.py, the situation you describe is correct:
```
OBL = load_obliquity_file(datapath("ecliptic.dat"))
```
So, I think either way there needs to be some rewriting.  I guess I could go either way.

demorest:
Looks like four files use `datapath`:  special_locations.py, topo_obs.py, pulsar_ecliptic.py, solar_system_ephemerides.py.  The first two check for `None` but the latter two don't.  Maybe it's same amount of work either way, but if we are going with the `None` approach those two should be fixed up in this PR.

paulray:
Concur. We should choose the more "pythonic" way, which generally is to not force people to put error checking in their code and just try to do things. But when that does cause an exception, we really don't want a super cryptic exception about trying to operate on None.  I'll think about it a bit and make the PR consistent one way or the other.

src/pint/solar_system_ephemerides.py
```diff
@@ -109,12 +109,18 @@ def _load_kernel_local(ephem, path):
...
+        search_list.append(datapath(ephem_bsp))
+    except FileNotFoundError:
+        # If not found in datapath, just continue. Error will be raised later if also not in "path"
+        pass
     for p in search_list:
         if os.path.exists(p):
             # .set() can accept a path to an ephemeris
             coor.solar_system_ephemeris.set(ephem)
-
+    # Question(paulr): How does this work? It looks like this function should always get to this exception!
```
paulray, line 122: @demorest Can you look at this? I don't understand how it ever worked. It looks like it goes through the search_list and calls `.set(ephem)` for any path that exists, but then it should fall to the raise. How did it not always raise an exception?

paulray:
OK, I made this all consistent, with `datapath()` raising an exception if the file is not found.
There is one place in solar_system_ephemerides.py that I really don't understand. See my question to Paul above.

src/pint/solar_system_ephemerides.py
demorest: yeah that looks pretty busted to me.  I think you're right that it will always raise an exception.  looks like that function is only called from within a try/except that ignores the exception.. probably should understand this and clean it up.

paulray: OK, thanks! I'll give a shot at cleaning it up and add that to this PR

paulray: OK, I pushed some fixes. I don't think load_kernel_local ever worked, it was just hidden by try/except blocks.

paulray:
I think this is OK to merge now

PR #666, opened 2020-04-21
Add print_summary to fitter

paulray:
After loading TOAs you can do `ts.print_summary()` which is quite handy in a notebook.  After I did a fit in PINT in a notebook I wanted a nice way to see a summary of the fit. This PR adds `ftr.print_summary()` which does just that. It also adds a bunch of utilities that are translated from presto.psr_utils to support Quantities.

This also adds a rms_weighted() method to compute weighted RMS residuals, which is usually what you want instead of std().

Plus a few other test and documentation fixes.

Introduces a dependency on the uncertainties package, which makes very nice pretty printed parameter values with uncertainties. Too bad it doesn't support Quantity yet.

### DIFF
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -506,8 +506,8 @@ def dmxparse(fitter, save=False):
         DMX_Errs.append(
             getattr(fitter.model, "DMX_{:04d}".format(ii)).uncertainty_value
         )
-        dmxr1 = getattr(fitter.model, "DMX_{:04d}".format(ii)).value
-        dmxr2 = getattr(fitter.model, "DMX_{:04d}".format(ii)).value
+        dmxr1 = getattr(fitter.model, "DMXR1_{:04d}".format(ii)).value
+        dmxr2 = getattr(fitter.model, "DMXR2_{:04d}".format(ii)).value
         DMX_R1.append(dmxr1)
         DMX_R2.append(dmxr2)
         DMX_center_MJD.append((dmxr1 + dmxr2) / 2)
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -4,12 +4,14 @@ from __future__ import absolute_import, division, print_function
 import re
 from contextlib import contextmanager
 from copy import deepcopy
-
 import astropy.units as u
 from astropy import log
+import astropy.constants as const
 import numpy as np
 import six
-from six import StringIO
+import scipy.optimize.zeros as zeros
+
+from io import StringIO
 
 __all__ = [
     "PosVel",
@@ -26,6 +28,18 @@ __all__ = [
     "interesting_lines",
     "show_param_cov_matrix",
     "dmxparse",
+    "p_to_f",
+    "pferrs",
+    "weighted_mean",
+    "ELL1_check",
+    "mass_funct",
+    "mass_funct2",
+    "pulsar_mass",
+    "companion_mass",
+    "pulsar_age",
+    "pulsar_edot",
+    "pulsar_B",
+    "pulsar_B_lightcyl",
 ]
 
 
@@ -378,7 +392,7 @@ def interesting_lines(lines, comments=None):
         yield ln
 
 
-def show_param_cov_matrix(matrix, params, name="Covaraince Matrix", switchRD=False):
+def show_param_cov_matrix(matrix, params, name="Covariance Matrix", switchRD=False):
     """function to print covariance matrices in a clean and easily readable way
 
     :param matrix: matrix to be printed, should be square, list of lists
@@ -387,7 +401,7 @@ def show_param_cov_matrix(matrix, params, name="Covaraince Matrix", switchRD=Fal
     :param switchRD: if True, switch the positions of RA and DEC to match setup of TEMPO cov. matrices
 
     :return string to be printed"""
-    output = StringIO.StringIO()
+    output = StringIO()
     matrix = deepcopy(matrix)
     try:
         RAi = params.index("RAJ")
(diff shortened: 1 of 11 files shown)


paulray:
Here is a sample of what the output looks like:
```
In[12] : ftr.print_summary()
Fitted model using weighted_least_square method with 18 free parameters to 932 TOAs
Prefit residuals 1.0039414007367866 us, Postfit residuals 1.0021121050070239 us
Chisq = 1070.943 for 914 d.o.f. for reduced Chisq of 1.172

PAR                   Prefit                  Postfit            Units
============== ==================== ============================ =====
PSR                      J0740+6620                              None
EPHEM                         DE421                              None
POSEPOCH                      57785                              d 
PX                          1.61727                     1.62(19) mas 
RAJ                   7h40m45.7913s     7h40m45.7913s +/- 1.8e-08 arcsec 
DECJ                 66d20m33.4943s    66d20m33.4943s +/- 9.3e-09 arcsec 
PMRA                       -10.3043                  -10.305(13) mas / yr 
PMDEC                      -30.8872                  -30.887(18) mas / yr 
PLANET_SHAPIRO                    1                              None 
NE_SW                             0                              1 / cm3 
SWM                               0                               
DM                          14.9647                 14.96466(16) pc / cm3 
DM1                     0.000221932                   0.00022(5) pc / (cm3 yr) 
DMEPOCH                       57785                              d 
DM2                     1.81045e-05                    4(7)×10⁻⁵ pc / (cm3 yr2) 
PB                          4.76694              4.7669446170(5) d 
PBDOT                    1.2666e-12               1.27(29)×10⁻¹²  
A1                          3.97756               3.97755650(17) ls 
M2                         0.221961                    0.222(10) solMass 
SINI                       0.999548                  0.99955(14)  
TASC                        56155.4            56155.36847105(8) d 
EPS1                   -5.63281e-06                -5.63(4)×10⁻⁶  
EPS2                   -1.84618e-06              -1.846(26)×10⁻⁶  
TZRMJD                      57792.9                              d 
TZRSITE                      ncyobs                              None
TZRFRQ                      1552.33                              MHz 
F0                          346.532        346.5319964636176(12) Hz 
F1                     -1.46396e-15            -1.46396(5)×10⁻¹⁵ Hz / s 
PEPOCH                        57785                              d 

Derived Parameters:
Period = 0.0028857364116591463 s +/- 1.0161743356857107e-17 s
Pdot = 1.2191088448465029e-20 +/- 4.256345862178073e-25
Characteristic age = 3.75e+09 yr (braking index = 3)
Surface magnetic field = 1.9e+08 G
Magnetic field at light cylinder = 7.158e+04 G
Spindown Edot = 2.003e+34 erg / s (I=1e+45 cm2 g)

Parallax distance = 618.9±71.6 pc

Binary model BinaryELL1
Conversion from ELL1 parameters:
ECC = (7.97±0.04)×10⁻⁶
OM  = 225.00±0.31
T0  = 56155.36852990(11)
Checking applicability of ELL1 model -- 
    Condition is asini/c * ecc**2 << timing precision / sqrt(# TOAs) to use ELL1
    asini/c * ecc**2    = 0.000252 us 
    TRES / sqrt(# TOAs) = 0.0328 us 
    Should be fine.

Mass function = 0.0029733880(4) Msun
Companion mass min, median (assuming Mpsr = 1.4 Msun) = 0.1964, 0.2300 Msun
From SINI in model:
    cos(i) = 0.030(5)
    i = 88.28(27) deg
Pulsar mass (Shapiro Delay) = 1.6944112951922108
```

luojing1211:
That looks very nice. Also, the utility functions are very useful.

-Jing

paulray:
In this code, I'm looping through parameters like this:
`for pn in self.get_allparams().keys():`
but that gives a semi-random ordering of parameters. Is there a way that gives a sensible order for the parameters (e.g. grouped by Component)?

paulray:
Also note that `np.isclose()` does not support units until numpy 1.17, so one of my tests fails. I fixed that with an xfail, but we really should quit supporting python 2 soon.  As soon as we tag 0.7, we can clean all the python2 crud out of the master branch, then keep a different branch for python2 support (releases 0.7.x can be bug fixes on that branch).

luojing1211:
Yea, the dictionary does not give you the right order, since they are harshed. One thing I can think of is by the component category. Given a list of print order of the components, one can print out by a particular order. The components actually have a function 'print_par()' that defines the print order for the parameters in one component.

src/pint/utils.py
```diff
@@ -486,8 +506,8 @@ def dmxparse(fitter, save=False):
         DMX_Errs.append(
             getattr(fitter.model, "DMX_{:04d}".format(ii)).uncertainty_value
         )
-        dmxr1 = getattr(fitter.model, "DMX_{:04d}".format(ii)).value
-        dmxr2 = getattr(fitter.model, "DMX_{:04d}".format(ii)).value
+        dmxr1 = getattr(fitter.model, "DMXR1_{:04d}".format(ii)).value
```
paulray, line 509: @bshapiroalbert In this PR, I included this bug fix in dmxparse(). It was reading DMX_nnnn instead of DMXR1_nnnn and DMXR2_nnnn. I'm not sure how this ever worked? Did you verify that the results from this matched Paul's dmxparse?

bshapiroalbert: @paulray Thanks for fixing that, that's embarrassing that I didn't catch that. I did check this version against Paul's dmxparse and got consistent results. The R1 and R2 aren't used unless you want to plot something or print an output file, they're not used when actually doing the mean subtraction on the DMX values (DMX_nnnn) or getting the errors, since there's no error on R1/2. But that will give you incorrect MJDs for the DMX bins....

paulray: Yeah, I noticed it when I tried to plot my DMX time series :-)

paulray:
> Yea, the dictionary does not give you the right order, since they are harshed. One thing I can think of is by the component category. Given a list of print order of the components, one can print out by a particular order. The components actually have a function 'print_par()' that defines the print order for the parameters in one component.

@luojing1211 yes this is true. I see that `as_parfile()` uses this and ends up with a sensible order. It would be nice to refactor this code some so that one could get a list of parameter names in a  sensible order from one method. Then `as_parfile()` could use that method for its work.

Issue #665, opened 2020-04-19
PowellFitter is broken


paulray:
I discovered this while working on issue #664

If you make a one line change to examples/fit_NGC6440E.py from WLSFitter to PowellFitter, it crashes.  This should be fixed.

Also, if scipy can do it, PowellFitter should set the covariance matrix.

paulray:
As an addendum to this, there should be an integration test in the test suite that does the NGC6440E fit with all available fitters and ensures that get compatible fits and uncertainties.

Issue #597, opened 2020-01-30
Warning: Scaled PBDOT


paulray:
Just wanted to point out a warning.  See this snippet of a par file:
```
BINARY        ELL1
PB            0.193098396006713           1    1.95113233487243e-11
PBDOT         1.23
```
This is interpreted as PBDOT = 1.23E-12, since the E-12 scaling is implied (I hate this, but it is for TEMPO backwards compatibility). 
Anyway, when you have this model, it does some very non-intuitive things:
```
In [20]: m.PBDOT                                                                                                      
Out[20]: floatParameter( PBDOT           1.23              (1e-12) frozen=True)

In [21]: m.PBDOT.quantity                                                                                             
Out[21]: <Quantity 1.23 1e-12>

In [22]: m.PBDOT.value                                                                                                
Out[22]: 1.23

In [23]: m.PBDOT.quantity.to(u.dimensionless_unscaled)                                                                
Out[23]: <Quantity 1.23e-12>

In [24]: m.PBDOT.quantity.to(u.dimensionless_unscaled).value                                                          
Out[24]: 1.23e-12
```

The fact that m.PBDOT.value returns 1.23 is scary!  If someone tries to use PBDOT in a calculation that way, they will get an error.  It is not intuitively obvious to convert to u.dimensionless_unscaled, to say the least!

So, should we make PBDOT not use the Quantity scaling and just get converted into a proper number so the value would be 1.23E-12 without scaling?  Thoughts?

Issue #668, opened 2020-04-22
PINT reports wrong number of degrees of freedom


paulray:
`fitter.resids.dof` returns an incorrect value and thus `fitter.resids.chi2_reduced` is also wrong. It seems that dof does not account for the hidden phase parameter.

For the trivial fit_NGC6440E.py example, here is what PINT gives:
```
Fitted model using weighted_least_square method with 5 free parameters to 62 TOAs
Prefit residuals 1090.580262221985 us, Postfit residuals 21.182038051610704 us
Chisq = 59.574 for 57 d.o.f. for reduced Chisq of 1.045
```
and here is Tempo2
```
RMS pre-fit residual = 1090.569 (us), RMS post-fit residual = 21.182 (us)
Fit Chisq = 59.57	Chisqr/nfree = 59.57/56 = 1.06368	pre/post = 51.4865
Number of fit parameters: 6
Number of points in fit = 62
Offset: -0.000227858 3.95351e-05 offset_e*sqrt(n) = 0.0003113 n = 62
```

scottransom:
yeah, this has been known and discussed for a while.  I thought there was already an issue for it, but apparently not!  Thanks for bringing it up.  I think it is an easy fix, but we have to make sure that it is changed in all the relevant places.

PR #670, opened 2020-04-23
Fix RAJ, DECJ uncertainties and a couple other bugs

### LINKED ISSUE #664: Incorrect uncertainties on RAJ, DECJ?
I've been working with PINT and was making a pretty-printer for the results of a fit. In doing so I found that the uncertainties reported on RAJ and DECJ are VERY different than Tempo2, by many orders of magnitude. Something is wrong.  Attached here is a little notebook that fits the data in pint/examples and looks at the uncertainties, compared to a Tempo2 fit of the same.  
(You will have to change the paths and delete the .print_summary() cell if you want to re-run it since that is only in my branch).

To summarize here, tempo2 gives:
```
PARAMETER       Pre-fit                   Post-fit                  Uncertainty   Difference   Fit
---------------------------------------------------------------------------------------------------
RAJ (rad)       1.63071752883066          1.63071752883548          8.8177e-11    4.8146e-12    Y
RAJ (hms)       06:13:43.9756752           06:13:43.9756753         1.2125e-06    6.6205e-08   
DECJ (rad)      -0.0351355401430826       -0.0351355400811907       1.7101e-10    6.1892e-11    Y
DECJ (dms)      -02:00:47.22538           -02:00:47.22537           3.5273e-05    1.2766e-05   
```

PINT gives compatible fitted values, but the RAJ uncertainty is 2.4e-14 rad and the DECJ uncertainty is 4.6e-14 rad.

Any ideas?

[PositionTest.ipynb.zip](https://github.com/nanograv/PINT/files/4496886/PositionTest.ipynb.zip)

paulray:
This is a small patch to fix #664 

This is the way to do it if you want to leave the parameters with different units for `par.quantity` and `par.uncertainty`

As discussed in the #664 thread, the alternative is to change the units of par.uncertainty to be the same as the parameter.


paulray:
Since this PR is still open, I added a fix for #668

paulray:
This is just a few lines of changes and fixes 2 bugs. @luojing1211 @scottransom @demorest are you OK merging this?

scottransom:
Fine with me.  I'd really like to see the uncertainty unit issue cleaned up as well, though (including fixing the 1e-12 thing for PBDOT).  But until then, I agree these should go in.

paulray:
OK, I think I fixed PBDOT scaling with the last commit.  It should scale PBDOT properly upon reading but the PBDOT Quantity will always have units that are not scaled.

paulray:
This should also fix #597

paulray:
To show an example like that pointed out in #597, this now works correctly.
I added this line to J0613-sim.par:
`PBDOT 7`
```
In [2]: m = pint.models.get_model("J0613-sim.par")                                                                                            
INFO: Parameter PBDOT's value will be scaled to 1e-12  [pint.models.parameter]

In [3]: m.PBDOT                                                                                                                               
Out[3]: floatParameter( PBDOT           7e-12             () frozen=True)

In [4]: m.PBDOT.quantity                                                                                                                      
Out[4]: <Quantity 7.e-12>

In [5]: m.PBDOT.value                                                                                                                         
Out[5]: 7e-12

In [7]: import astropy.units as u                                                                                                             

In [8]: m.PBDOT.quantity.to(u.dimensionless_unscaled)                                                                                         
Out[8]: <Quantity 7.e-12>

In [9]: m.PBDOT.quantity.to(u.dimensionless_unscaled).value                                                                                   
Out[9]: 7e-12

```

scottransom:
sweet.....  :-)

paulray:
Oops, last commit is not just a typo fix. It cleans up a bit of cruft that is not needed now that a parameter's units aren't actually modified when scaling is done. Just the value is scaled.

luojing1211:
@paulray, why the commit number is so big. Is this rebased to the master?

paulray:
Yes, I merged in the master branch

luojing1211:
I see. That is why my commits on the add_components show here.

paulray:
I also just added another commit that fixes the Powell fitter.  
@scottransom this is a strange one.  It was caused by `fitter.set_params()`.
It is called with a dictionary of keys and values. Normally the values are all floating point numbers. For some reason on like the 10th call to set_params() in the Powell fitter, the values change to Quantity (but dimensionless).  This causes par.value = v to barf.  So I added code to `fitter.set_params()` to check if the values were Quantities and use v.value in that case.

Strange but it works.  My only concern is that there could be a performance penalty if set_params is called many times by some fitters.

paulray:
This now also fixes #665

scottransom:
@paulray I remember seeing that before with early tests of the Powell fitter.  It must be some strange conditional part of a loop in the Scipy routine.  Very bizarre....

paulray:
Bizarre indeed. Since this PR fixes a bunch of bugs now, I'm merging it so future work can be based of this. I think we are very close to ready to tag 0.7!

Issue #672, opened 2020-04-24
dmxparse requests


paulray:
Hi @bshapiroalbert 
I have a couple requests for `dmxparse()`.

1. Could you make it work even if there is no covariance_matrix? In that case it could just use the single parameter uncertainties, and maybe emit a `log.warning()` that it is doing that?
2. I haven't tried it but the code looks like it requires that the DMX bins be continuous from an index of 1 to the maximum. But, PINT supposed to support DMX ranges that aren't continuous (e.g. if you delete a DMX bin). So, this code should work even if you have just `DMX_001`, `DMX_002`, `DMX_004`, for example)

scottransom:
@paulray How would we get the values without a covariance matrix?  I think that is basically *required* by the DMX methodology as there is a huge covariance with the average DM (and maybe a few other things).  Would like to hear from @demorest about this.  Note:  I totally agree with your #2, though.

paulray:
Maybe it is a bad idea, but I figured that DMX_nnnn should have a single parameter .uncertainty that could be used if the covariance matrix is not available. I think it is (or should be) illegal to had DM not frozen when you fit for DMX.  I assume that would be a problem either way.

bshapiroalbert:
I can definitely do number 2 right now. For number 1, I can have it output a `log.warning` and then have it try to just return any initial values of DMX instead that are already listed in the input par file, like the `DMX_nnnn.uncertainty` and `DMX_nnnn.value` and if it hasn't been fit for it should just return zero, right? But it wouldn't actually be the numbers that you want as Scott mentioned.

paulray:
Cool, sounds good! I think there is value in being able to plot the DMX time series even if you have just read it in and there are either no errors or the errors are just the uncertainties read from the par file. Doing that, with a warning, seems preferable to just having it fail.

PR #674, opened 2020-04-24
Dmxparse update

bshapiroalbert:
Requests for updates to dmxparse function from #672 have been included. The changes are as follows:
1. If there is no covariance_matrix attribute, the function will return the mean subtracted dmx values from the initial `fitter.model` as well as the initial errors from `fitter.model`. It will also print a warning that it is doing so (let me know if you'd like the warning to say something else). If there are no errors, then the DMX_vErrs should be a list of `None`'s. 
2. The DMX bins do not need to be in any particular order, and the number of values in the index of the DMX bin should not matter, e.g. DMX_0001, DMX_0002, DMX_0004 will be parsed appropriately, and DMX_01, DMX_02, DMX_04 should also work.


paulray:
I hate to ask one more thing, but since you are working on the code...

I think it would be more "PINT-like" to have the arrays returned include units on them. I think this is a tiny change. Since you grab the values as par.value, when you are building the dictionary of arrays to return, you can just multiply each by the appropriate par.unit.

bshapiroalbert:
Sound good @paulray, I'm fine with that. I'll add that in soon.

bshapiroalbert:
I've updated the `dmxparse` function so that if you put in a model that does not have any DMX bins in it, it will throw a Runtime error saying that's why it's failing. I have also (I think) figured out how to get the function to run if a few DMX bins are not fit for as discussed by @paulray. The returned values on the errors in this case in the array are nan, and if you print the results to the output file it also prints nan. It will also tell you if there are any bins that are not fit for, and that it's doing the appropriate masking. I also added an additional test for some of this coverage. It passed all the tests I could think of that seemed plausible. Let me know if anybody has any comments on this or finds a way to break it that I haven't thought of.

paulray:
I tried a couple of my problem cases and it worked great. Thanks for the fixes! I think I'll merge this.

bshapiroalbert:
Sounds great! Let me know if you find other ways it needs to be improved.

PR #675, opened 2020-04-24
Return random models in random_models

clp3ef:
Random_models used to only return residuals and evenly spaced TOAs for plotting the random models. This is a small edit to return a list of the actual models themselves in addition to the residuals and TOAs, and an edit to pintk to allow it to work with the additional return parameter.


paulray:
This looks handy. Ready for it to be merged?

clp3ef:
Yes, so long as no one has any issues with it, it’s ready to merge.

PR #676, opened 2020-04-25
GLS fixes

demorest:
Fixes bugs in GLS fitter when noise model not present, discussed in #664


paulray:
Yay! This works great. I did my position test and now GLS fitter works with or without full_cov=True.

Issue #659, opened 2020-04-13
@paulray - this looks really nice. Obviously, arbitrarily increasing tolerances is worrying! I think part of the problem is that the test involved seems to be at a level so high that one has little idea where the actual problem originates. It may be good to open an issue to greatly expand unit tests... Having seen @aarchiba's work on [testing astropy's `Time` class](https://github.com/astropy/astropy/pull/9532), I've also become more convinced that it should be a mixture of well thought-through tests and using `hypothesis` to check that assumptions about what the code does actually hold.


paulray:
As part of working on PR #644 I found that I had to increase the testing tolerances in a few places, including the PosVel comparison between PINT and Tempo2, and in `test_fbx.py` check of FB1 derivative, which only failed for certain versions of numpy/astropy.

Here is what @mhvk had to say:

@paulray - this looks really nice. Obviously, arbitrarily increasing tolerances is worrying! I think part of the problem is that the test involved seems to be at a level so high that one has little idea where the actual problem originates. It may be good to open an issue to greatly expand unit tests... Having seen @aarchiba's work on [testing astropy's `Time` class](https://github.com/astropy/astropy/pull/9532), I've also become more convinced that it should be a mixture of well thought-through tests and using `hypothesis` to check that assumptions about what the code does actually hold.

_Originally posted by @mhvk in https://github.com/nanograv/PINT/pull/644#issuecomment-612902869_

Zac-HD:
:wave: hi all, I'm not an astrophysicist but I *am* particularly interested in open source and in testing scientific code.  So if getting a Hypothesis core dev to review or discuss some tests would ever be useful, ping me and let me know how I can help!

Along with precision tests, I highly recommend getting Hypothesis to generate *any* valid data (i.e. including unusual formats) and just calling your code.  No assertions needed, in my experience you'll find a bunch of crashes, quickly improve your input validation, and probably fix some bugs too.

Finally, are there any any applications or calculations where you can generate inptu data, feed it to both PINT and TEMPO, and assert that the results should be identical?  This is a scarily powerful testing trick which most of us only dream of :smile:

paulray:
Thanks @Zac-HD ! We welcome suggestions. I personally don't understand Hypothesis at all. It got added to our tests by @aarchiba and I haven't yet figured out what it is doing or how to use it. Also, it seems to be evolving so our tests now generate a ton of deprecation errors. So, a hypothesis tutorial would be great.

Zac-HD:
[...]

PR #698, opened 2020-05-13
Translate test fixture to decorator

Zac-HD:
This pattern is safer for Hypothesis, as the context is refreshed each time the inner test function is invoked instead of just once for all inputs.  (if it *is* safe to share the fixture, it can be declared as `@pytest.fixture(scope="module")` to say so and Hypothesis won't complain)

Ref #659, [hypothesis release notes](https://hypothesis.readthedocs.io/en/latest/changes.html#v5-6-0).  CC @paulray; this fixes the seven Hypothesis-related warnings you had.


paulray:
Thanks for this fix!

PR #699, opened 2020-05-13
Wideband toa

luojing1211:
This is a work based on an old wideband toa PR #464, which is deviated from the master branch too much. The idea behind is treating the independent DM value as part of the data for fitting DM parameters. I am trying to put several things together. 

* A residual class for DM values
* A fitter that can take TOA residual class and DM residual class can fit them together. 
* Do a limited generalization of this method for the future's development.


bshapiroalbert:
Hey @luojing1211, I was playing with this branch a bit and found that If I try to use the par and tim files for a pulsar with red noise (e.g. J1747-4036, J1744-1134, etc.) from the 12.5 year wideband release, I get an error about input array dimensions when trying to initialize the wideband fitter. I've attached a screenshot of the cell that shows what I'm running and the resulting error. I'm fairly sure that this was the most up-to-date version of this branch, and it works fine for every pulsar without red noise in the par file that I've tested. Let me know if you need any other info or anything to help fix this. Thanks!
PINT_WB_RN_Fitter_Error

luojing1211:
This is good feedback. That is what I need!

luojing1211:
Ok, I did not add the red noise design matrix for dm data. It should just padding zeros.

luojing1211:
@paulray  Can you take a look at the clock correction tests. I did not touch anything about it.

paulray:
This is a line that I modified recently.  It works on my computer with my install of TEMPO2. I'd like to merge this and if it still fails I'll troubleshoot.  OK?

PR #702, opened 2020-05-15
Fix buggy TimeDelta comparisons

paulray:
I was playing with Time and TimeDelta when I found these buggy comparisons.  There are 2 problems:
1. The comparison did not take the absolute value of the TimeDelta
1. The comparison did not have units on the quantity being compared.

I think the second issue is really an astropy bug.  This should not work but does:
```
In [111]: td                                                                                                                                                              
Out[111]: <TimeDelta object: scale='tdb' format='jd' value=-4.01624200563333e-09>

In [112]: td<1                                                                                                                                                            
Out[112]: True
```
This allowed me to compare a TimeDelta to "1" without specifying the units. Is it 1 day, 1 second?  This should have failed with some kind of units exception. Do you agree @mhvk ?

### DIFF
--- a/docs/examples/example_dmx_ranges.py
+++ b/docs/examples/example_dmx_ranges.py
@@ -25,7 +25,7 @@ m1855.remove_component("DispersionDMX")
 
 # Build the new component with maximum DMX bin of 15 days ensuring that there are TOAs above and below 1000 MHz in each gin
 mask, dmx_comp = pint.utils.dmx_ranges(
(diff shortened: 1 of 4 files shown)


paulray:
Note that I have not checked the rest of the test suite. There may be other examples of this.

mhvk:
Definitely a bug! Raised at https://github.com/astropy/astropy/issues/10355

Issue #692, opened 2020-05-08
pintk - on/off button for random models


clp3ef:
Currently, pintk calculates and plots 12 random models every time a fit is performed. The random models take ~15 seconds to calculate, lengthening the time taken for each fit considerably. An on/off button in the pintk GUI which toggles this calculation and the resulting plotting would be very useful.

paulray:
This is one of the issues that @champagne-cmd is going to take a look it.

PR #704, opened 2020-05-16
Randomfit checkbox

champagne-cmd:
Added checkbox to pintk GUI to turn random model plotting on and off.

### DIFF
--- a/src/pint/phase.py
+++ b/src/pint/phase.py
@@ -83,3 +83,12 @@ class Phase(namedtuple("Phase", "int frac")):
 
     def __sub__(self, other):
         return self.__add__(other.__neg__())
+
+    def __mul__(self, num):
+        # for multiplying phase by a scalar
+        product = (self.int + self.frac) * num
+        ff, ii = numpy.modf(product)
+        return Phase(ii, ff)
+
+    def __rmul__(self, num):
+        return self.__mul__(num)
--- a/src/pint/phase.py
+++ b/src/pint/phase.py
@@ -86,8 +86,9 @@ class Phase(namedtuple("Phase", "int frac")):
 
     def __mul__(self, num):
         # for multiplying phase by a scalar
-        product = (self.int + self.frac) * num
-        ff, ii = numpy.modf(product)
+        ii = self.int * num
+        ff = self.frac * num
+        # if frac out of [-0.5, 0.5) range, Phase() takes care of it
         return Phase(ii, ff)
 
     def __rmul__(self, num):
--- a/src/pint/pintk/plk.py
+++ b/src/pint/pintk/plk.py
@@ -682,9 +682,14 @@ class PlkWidget(tk.Frame):
         self.plkCanvas.draw()
 
     def plotErrorbar(self, selected, color):
+        """ 
+        For some reason, xvals will not plot unless unitless. 
+        Tried using quantity_support and time_support, which plots x & yvals,
+        but then yerrs fails - cannot find work-around in this case.
+        """
         self.plkAxes.errorbar(
-            self.xvals[selected].reshape([-1, 1]),
-            self.yvals[selected].reshape([-1, 1]),
+            self.xvals[selected].value,
+            self.yvals[selected],
             yerr=self.yerrs[selected],
             fmt=".",
             color=color,
--- a/src/pint/pintk/plk.py
+++ b/src/pint/pintk/plk.py
@@ -209,6 +209,43 @@ class PlkFitBoxesWidget(tk.Frame):
         log.info("%s set to %d" % (par, self.parVars[par].get()))
 
 
+class PlkRandomModelSelect(tk.Frame):
+    """
+    Allows one to select whether to fit with random models or not
+    """
+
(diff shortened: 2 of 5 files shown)


paulray:
Fixes #692

scottransom:
Nice!  :-)

paulray:
Thanks Chloe! One little tip is that this PR includes the changes from your other PR. This means it can't be merged without merging in those changes as well. Not a big deal here since both are ready to be merged. But, in the future it is a bit better if you make a new feature by starting with a new branch off of master. Then each feature PR can be merged without affecting the other.
See here: https://github.com/nanograv/PINT/wiki/Git-Workflow and here: https://nanograv-pint.readthedocs.io/en/latest/contributing.html for some tips.

champagne-cmd:
Ah, my apologies @paulray, I thought I had branched off the master. I will be more mindful next time!

paulray:
No worries! I just tried it and the checkbox works great!
I notice that the random models only actually get plotted if the X-axis is MJD or Rounded MJD. Would it be hard to make them appear for other X-axes (at least "Year")?

paulray:
Also, since you are editing pintk code, could you look into this deprecation warning? Presumably it is just a tiny fix:
```
/Users/paulr/src/PINT/src/pint/pintk/plk.py:695: MatplotlibDeprecationWarning: Support for passing a (n, 1)-shaped error array to errorbar() is deprecated since Matplotlib 3.1 and will be removed in 3.3; pass a 1D array instead.
```

champagne-cmd:
I fixed the deprecation warning @paulray. I will need to study the code more to figure out the plotting with other variables, but I will see what I can do.

paulray:
Wow, I love the simplification!  That makes me wonder if the rest of the `reshape([-1,1])`'s need to be there? I had never seen that particular usage and had to look up what -1 meant to `reshape()`. If they need to be reshaped 1-d arrays, then why not use `reshape(-1)` instead? Just curious.

champagne-cmd:
I believe reshape(-1) creates a (1, N) array, whereas reshape([-1, 1]) creates an (N, 1) array. The x and y (and error arrays formerly) take in (N, ) arrays or (2, N) arrays. So reshape(-1) wouldn't work for the x and y values.

paulray:
Hmm. When I do reshape(-1), I get a 1-d array of length N, i.e. the shape is (N, ). I don't really know why that code ever was using 2-d arrays with one dimension being length 1 (either (N,1) or (1,N)), but I haven't looked at any of that code so maybe there was some reason. Anyway, the plotting now works great with no warnings.  Fantastic work!

scottransom:
Yeah, why are reshapes required at all?  I don't understand that...  I would have thought that everything was a 1-D array.

champagne-cmd:
For the x values, it seems like the units attached are causing problems. When I remove reshape, it throws an error saying 'only dimensionless scalar quantities can be converted to Python scalars'. So perhaps the reshape serves to remove units as well as ensure the array is the correct dimensions?

luojing1211:
The correct way to remove unit is Quantity.value(). reshape may remove the unit, however, when numpy fully supports the unit, it will fail.

paulray:
Ah! I didn't realize the goal was to remove the units!  Those reshapes have been there from the beginning with Jonathan wrote the code.  The "correct" way is to enable `quantity_support()` from astropy.visualization. Then you don't have to remove units at all. See: https://docs.astropy.org/en/stable/visualization/matplotlib_integration.html#plotting-quantities

luojing1211:
That is the correct way to do it, instead of removing the unit.

champagne-cmd:
Huh, I used quantity_support and the x and y values now don't need reshape. However, now the y error is throwing 'matplotlib.units.ConversionError: Failed to convert value(s) to axis units: []' 
I'm still not sure how to fix this...

paulray:
Hmm. Presumably `yvals` and `yerrs` need to have compatible units.  Could one of them have lost their units?

champagne-cmd:
They both seem to have units of us.

paulray:
Just a guess, but maybe you called `quantity_support()` in this routine, so it wasn't enabled when the axes were created?  I think you want to call it only once, so I'd put it at the top of plk.py just after matplotlib is imported. That should make it a 'global' setting.

scottransom:
Yeah, that might be it. However, what happens if we change the axes (i.e. MJD to year or orbital phase).  Does that need to completely reset the axes?

paulray:
Beats me! :-)

champagne-cmd:
Nope, I called it inside the routine and at the top of the file, it still gives me the same error. I will try to do some more digging.

paulray:
Worst case you can just resort to .value as Jing suggested

champagne-cmd:
Ok, I used .value in plotErrorBars since I couldn't figure out a work-around with quantity_support or time_support. Also, I modified the random model plotting so that it would plot with MJDs and years. While playing around with that, I noticed there was a small bug in the orbitalphase() function in pulsar.py -- the return was incorrect for the case where a non-binary pulsar was used, so I amended that.

Issue #640, opened 2020-03-13
Built-in F-test for parameter significance


mtlam:
Part of NANOGrav's workflow, and I believe useful to others more broadly and thus should probably go into PINT, is the ability to test the addition (or removal) of a timing model parameter or parameters. Therefore, I'd suggest it either takes a value or list of values, which I suppose could be `Parameter` objects but preferably for ease strings. I'm not sure where this function would live within PINT or even what the convention would be (it probably should return a boolean?)

This is performed via a standard F-test which will compare the number of degrees of freedom (`get_dof()`) and the chi-squared value (`calc_chi2()`) for each model. From these, it will calculate the F statistic. The function should take some value (e.g., often called `alpha`) that will act as the significance to compare against the p-value (in the NANOGrav datasets, we use 0.0027, i.e., "3σ"). The value could be set with a default and made optional.

Example usage:

    model.Ftest("XDOT") #check if XDOT is significant
    model.Ftest("XDOT", remove=True) #check if it is preferred to remove XDOT
    model.Ftest(["M2", "SINI"]) #check if adding M2 and SINI is significant
    model.Ftest("XDOT", alpha=0.05) #check if XDOT is significant, but at the "2σ" level

For the F-test, one can use `scipy.stats` in a fairly straightforward manner, e.g., https://stackoverflow.com/a/21503346; see the note on using `stats.f.sf` rather than `stats.f.cdf`. I am happy to discuss fuller code on this end, though this is probably the easiest part of the implementation of the test into PINT.

paulray:
This is clearly a good functionality to support! However, I don't think it can be implemented as a method of a TimingModel object.  It requires re-fitting after modifying the model, so it needs a Fitter and a TOAs object as well.  Given that, it probably makes sense to make a higher level convenience function that takes a model, fitter, and TOAs and a list of parameters to test.
So, maybe `pint.Ftest(model, toas, fitter, "XDOT")`?  As this is quite separable from the details/internals of any of those classes, do you want to take a crack at writing this?

scottransom:
Yeah, I think this is correct what @paulray wrote.  Although I think it could also be done as a part of the fitter class itself (for WLS and GLS instances).  That's because there is already a base model and TOAs assigned.

[...]

PR #705, opened 2020-05-18
Add F-test to fitter

bshapiroalbert:
I've written an F-test for the fitter class so that the user can add/remove parameters to the timing model and get back the significance of having the additional parameter in the timing model from an F-test. This addresses Issue #640 (at least in part). I have also added a test to the `test_fitter.py` that tests the adding/removing of an FD parameter to the timing model. This PR needs some discussion as to what the inputs/outputs and/or additional functionality of the F-test should be. Comments and suggestions are welcome.


scottransom:
@bshapiroalbert Thanks for this!  I have one suggestion:  maybe factor your ftest method so that it calls an ftest() function that could co in utils.py.  That function could be like the PRESTO one that just computes the Ftest and explains how it works (and has as input the chi^2s and DOFs).  That routine is potentially useful for other purposes, so it is nice to have around.  Then your fitter method could call that function after doing all the checks etc that you are doing.

bshapiroalbert:
Hey @scottransom, I can definitely do that easily. I'll make sure to put that in (and fix the failing tests) shortly.

paulray:
Looking good!  For at least the test_utils.py, could you add an `assert np.isclose(ft,xxx)` where xxx is the analytically computed correct value for the Ftest given those input values?  If you want to be a bit more extensive, you could add more tests that show that it handles various edge cases correctly.

bshapiroalbert:
Thanks @paulray , I'll add that assert in at least for those values and maybe add some edge cases if I have some time.

bshapiroalbert:
> For at least the test_utils.py, could you add an `assert np.isclose(ft,xxx)` where xxx is the analytically computed correct value for the Ftest given those input values?

Hey @paulray, maybe I need to work on looking into this a bit more, but I think the analytical integral of of the Ftest is possible but not really feasible to do by hand. I used those values for the test because they represent realistic values for what might get put in, but integrating the F-distribution CDF for more than a few degrees of freedom by hand is extremely intense (as far as I can tell). I can do the integral by hand if I use degrees of freedom that are more along the lines of 2,3,4 but not ~4000...Or maybe I am just missing some trick to doing this integral? I think that asserting something we know analytically is probably better, but let me know your thoughts.

scottransom:
If we wanted one or two "truth" values for the Ftest inegral, we could compute with Mathematica or Maple (or even one of the python extended precision packages) to 20 or 30 decimal places.  That would get us plenty of precision for a test like this.  And it could be hard-coded in the test.

bshapiroalbert:
@scottransom what are some of the python extended precision packages you had in mind? I don't have access to Mathematica or Maple right now (I don't have a license for it, and I can't use the computers at WVU since we can't go on campus). Would that computation with a python package have a user coded F-distribution and CDF? I assume we wouldn't want to use the `scipy` f distribution since that's what we're using to do the F-test, and that wouldn't be independent I don't think.

scottransom:
Maybe SymPy?  I have access to Mathematica and I suspect that others do as well, so we could go that route for a couple hardcoded test values if you want.

bshapiroalbert:
Okay, I can try to take a look at SymPy then and maybe we can discuss more on the next PINT call. Not sure I can get that done and hardcoded before then anyways right now.

paulray:
Initially, I think it would be fine to use the scipy result. At least that way we'll know if something ever changes to break it. It doesn't prove that it is right at the moment, but is better than nothing!

paulray:
Here is the Mathematica code for it: https://reference.wolfram.com/language/ref/FRatioDistribution.html

scottransom:
So I poked around with SymPy today.  It is really cool.  There is arbitrary precision floating point math in it (using .evalf()).  And the F-distribution is built in:
https://docs.sympy.org/latest/modules/stats.html
Not sure this is super necessary, but it is good to know that we can easily do some really precise stuff if we need to.

scottransom:
Or, using the mpmath module, mpmath.betainc()  (with regularized=True keyword set) might even be easier.

demorest:
It might be worth using `fdtrc()` rather than `1.0 - fdtr()` for higher precision.

scottransom:
Totally agree with @demorest.  We should be using `fdtrc()`.  Below is an example showing where it makes a difference (in the tail of the distribution).

```
In [68]: fdtrc(1.0, 50, 10000.0/1.00001)                                                  
Out[68]: 2.961635592779673e-59

In [69]: fdtr(1.0, 50, 10000.0/1.00001)                                                   
Out[69]: 0.9999999999999999

In [70]: 1.0-fdtr(1.0, 50, 10000.0/1.00001)                                               
Out[70]: 1.1102230246251565e-16
```

bshapiroalbert:
I've made a few updates to the F-test now. I'm now using `fdtrc()` in the F-test in `utils.py`. The test for the F-test in `test_utils.py` also has the return of `fdtrc()` to check hardcoded in. In the `ftest()` that's part of the `fitter` class, I have it set to `deepcopy` the fitter and just adjust the model thathas been copied instead of the initial model. Instead of removing parameters from the model, I now set the value and uncertainty to zero, and then freeze them so they are not fit for. When adding parameters for the `ftest()`, the function now also checks to see if the parameter is already part of the model, but with values of zero and set to be frozen (as discussed with some binary parameters on the last call, like `XDOT`). If they're already there, it will just unfreeze them so those parameters get fit for. If they're not there already, it will add the parameter to the appropriate input component of the timing model as done before. I think that address all of the comments from last time, but please do let me know if you have any more, or find cases that break the test where it shouldn't break.

Issue #707, opened 2020-05-19
Add version check to pickle file


paulray:
I suggest that the PINT version number be embedded in the .pickle.gz files of TOAs. Then when it is read, it can be checked against the current version and if it is old, it won't use the pickle but will regenerate from the .tim file.

demorest:
I've never really liked the TOA pickling, because of issues like this.  The main reason we do it is because the initial TOA reading/processing was really slow early in PINT.  Is this still the case now?  Are there any use cases that would be seriously affected if we did not pickle TOAs by default?

demorest:
PS, the version check is of course a good idea if we want to continue to use pickled TOAs!

paulray:
I'm fine with having use_pickle=False by default. But, I do use pickles a lot. For example, when I'm writing a notebook to do analysis with PINT, I tend to "Restart and Run All Cells" a lot as I'm hacking away at the code.  This process goes a lot faster if it just has to re-read the pickle, so I don't want to get rid of it completely.

scottransom:
Pickling can still be *extremely* useful.  Especially for single photon timing (for example).  Here is loading of the J1713 12.5yr TOAs with and without pickle.  The difference is over a minute without, and 2.5 sec with.  I think it will be very easy to fix this up.  Although I do think we need better tests that will compare TOAs objects loaded with and without pickling and make sure they are identical.

```
In [1]: import pint.toa as toa                                                            
INFO: Running get_iers_up_to_date() to update IERS B table [pint.erfautils]

In [2]: %time t = toa.get_TOAs("J1713+0747_NANOGrav_12yv3.tim", usepickle=False)          
INFO: Applying clock corrections (include_GPS = True, include_BIPM = True) [pint.toa]
INFO: Observatory arecibo, loading clock file 
	/home/sransom/python_venvs/py3/lib/python3.8/site-packages/pint/datafiles/time.dat [pint.observatory.topo_obs]
INFO: Applying observatory clock corrections. [pint.observatory.topo_obs]
INFO: Applying GPS to UTC clock correction (~few nanoseconds) [pint.observatory.topo_obs]
INFO: Observatory arecibo, loading GPS clock file 
	/home/sransom/python_venvs/py3/lib/python3.8/site-packages/pint/datafiles/gps2utc.clk [pint.observatory.topo_obs]
INFO: Applying TT(TAI) to TT(BIPM) clock correction (~27 us) [pint.observatory.topo_obs]
[...]

PR #712, opened 2020-05-20
Pickle fixes

### LINKED ISSUE #710: Problem with pickled TOAs
There is a bug in pickled TOAs. When you create a TOAs object like this:
```
In[18]: traints = pint.toa.get_TOAs(traintim,ephem="DE421",include_bipm=False,include_gps=False,planets=False,usepickle=False)
INFO: Applying clock corrections (include_GPS = False, include_BIPM = False) [pint.toa]
INFO: Applying observatory clock corrections. [pint.observatory.topo_obs]
INFO: Doing astropy mode TDB conversion [pint.observatory]
INFO: Computing PosVels of observatories and Earth, using DE421 [pint.toa]

In [19]:  traints.clock_corr_info
{'include_bipm': False, 'bipm_version': 'BIPM2015', 'include_gps': False}
```
it has a `traints.clock_corr_info` that is a dictionary with several entries.  But when you read it from a pickle file, that dictionary is empty.
```
In [27]: traints = pint.toa.get_TOAs(traintim,ephem="DE421",include_bipm=False,include_gps=False,planets=False,usepickle=True)
INFO: Reading TOAs from pickle file [pint.toa]
INFO: Reading pickled TOAs from '1937/radio/training.tim.pickle.gz'... [pint.toa]

In [28]: traints.clock_corr_info
{}
```

This causes a crash when you try to compute phases with `abs_phase=True`

paulray:
Fixes #710 and #707 

Currently emits an error message if the PINT version has changed since the pickle file was created, but does not stop or delete the pickle file. That is left to the user. I think this is sufficient for now.  Certainly better than it used to be.  People might be upset if we summarily delete their pickle files.

Also allows "CC " to start a comment line.  Found this in an EPTA .tim file.


paulray:
This was worse than I thought. The pickle files did not properly restore ephem or planets either.  Those are now fixed.

PR #714, opened 2020-05-24
Added error check

paulray:
Fix for Joe Swiggum. print_summary was failing when SINI was UNSET


paulray:
The test failure here is that `pip install -r requirements.txt` is failing under Python 2.7. We *really* need to drop Python 2.7 support!

paulray:
@swiggumj this is the fix you need.

PR #716, opened 2020-05-29
Succinct benchmark file. Updated run_profile to output pdf.

champagne-cmd:
This benchmarking code allows you to choose the functions you want to see and run multiple files at once, providing a more specialized version of the run_profile code. I also changed the run_profile code to output a pdf rather than a png file per @paulray's request.


champagne-cmd:
Added parser file to display clean summary of profiler output and hid all logging output to terminal. This benchmark suite requires the installation of 'py-cpuinfo' and 'pandas' packages (details in docstrings of high_level_benchmark.py and parser.py).

champagne-cmd:
As soon as it finishes on travis-ci this PR will be ready for review @paulray.

paulray:
Looks great!  Here is the output of the high level benchmark suite, if anyone wants to see it.
Now we can start digging in to the big time hogs in more detail and see if we can speed them up.

Also the PDF profiler output is much nicer and space efficient!
I'm not clear on exactly what all the numbers on the boxes and lines are. I need to figure that out.

How do I generate the HTML view of the profile?

```
Processor running this script: Intel(R) Core(TM) i5-4570S CPU @ 2.90GHz
Python version: 3.7.7
SciPy version: 1.4.1, AstroPy version: 4.0, NumPy version: 1.18.2
PINT version: 0.7+10.g2236882

*******************************************************************
OUTPUT FOR BENCH_LOAD_TOAS.PY:
Total Time: 18.991 s
                Function Time(s)
    Construct TOA Object   2.750
   Construct TOAs Object   5.949
 Apply Clock Corrections   5.189
            Compute TDBs   1.720
         Compute Posvels   3.612

*******************************************************************
OUTPUT FOR BENCH_CHISQ_GRID.PY:
Total Time: 128.454 s
         Function Time(s)
 Get Designmatrix  75.486
    Update Resids  14.797
       Cho Factor   0.014
        Cho Solve   0.016
  Select TOA Mask  18.129

*******************************************************************
OUTPUT FOR BENCH_CHISQ_GRID_WLSFITTER.PY:
Total Time: 124.845 s
         Function Time(s)
 Get Designmatrix  79.639
    Update Resids  15.154
              svd   0.257
  Select TOA Mask  10.414

*******************************************************************
OUTPUT FOR BENCH_MCMC.PY:
Total Time: 30.726 s
```

champagne-cmd:
Running the line 'cprofilev -f /path/to/outputfile' in the terminal will run it. I put it in comments at the bottom of run_profiler.py. The html viewer doesn't seem to have a good mechanism to stop itself once it provides you the link to view (I've had to manually kill it each time), so I didn't put it in the new benchmark suite as you would only be able to open one of the output files when running the benchmark suite. Right now I delete the output files once I've displayed the summary, I can change that if you'd like @paulray so one can analyze indiviual benchmarks in detail.

luojing1211:
@champagne-cmd and @paulray it would be great if it also displays how many TOAs we are using, and how big is the design matrix.

champagne-cmd:
@luojing1211 the bench codes now do that when run through run_profile. The high level bench suite suppresses all output from individual bench codes.

luojing1211:
@champagne-cmd Just a quick question, in the test of Geting Designmatrix, could you compute the overhead time? In the current setup, each parameter's derivative is stored in a list. If a .par file has a lot of parameters,  like DMXs, PINT will do a lot of for loops just compute the design matrix. I think that is why it is so slow.

paulray:
@luojing1211  How do you mean to define "overhead" time?  Do you mean the total time spent in `model.designmatrix()` minus the time spent in `d_phase_d_param()`?  I believe that overhead is small, particularly in the case where the number of TOAs is much larger than the number of free parameters. But this is something you can look at in the profile output, especially with the HTML viewer or in the .pdf output.

luojing1211:
The "overhead" time I mean is the time that python spend on the for loops, which is slower than other languages. If we have 100 free parameters, PINT will call at least 200 for loops since the derivative functions are stored at {'param': [deriv_func,]}. This design is for the case that one parameter is used by two model components.

paulray:
You mean the for loop at line 942 of timing_model.py?

luojing1211:
yes, and line 951. For phase derivatives, it will add one more loop, there is the d_phase_d_delay part.

champagne-cmd:
I can look into it @luojing1211. I'm not sure if you can specify loops with the profiler, you may just have to put timing statements on either side of the two you want, but I will investigate further.

luojing1211:
Thank you @champagne-cmd. It would be very useful.

champagne-cmd:
@luojing1211 for the loop in d_delay_d_param, it takes up almost all the time for that function, so looking at that function will basically give you the loop time. For the loops in d_phase_d_param (lines 928 and 942), they each only took about 2 seconds total when I ran the bench_chisq_grid.py. I don't see a great way to time them other than placing timing statements around them. I can then accumulate the time in another file and parse it (which will add close to 5 seconds to runtime). I guess to avoid the extra time during normal operations, you would keep them commented out until you want to use them. Would you like me to implement the loop times in a future PR, and would you like to see the times in the run_profile file?

luojing1211:
Thank you. If it is too much trouble to implement it, we can do it later. I
don't think it is critical. When I think about it, the most important
metrics are 1. runtime for each function, 2. how many times each function
suppose to be called (This can be estimated from the input data size, for
instance, how many parameters are fitted in the test suite.) 3. how many
times each function is actually being called. I think you already have the
most.

-Jing

paulray:
I'm not sure I understand this discussion.  @luojing1211 you are worried about loop overhead right?  That should just be the difference between the total time spent in `d_phase_d_param()` minus the time spent executing the actual `phase_deriv_funcs` and `delay_deriv_funcs` right? Maybe at the telecon you can explain in more detail?

luojing1211:
That was my initial worry. But, ultimately, we would like to avoid unnecessary for loops. That is why I change my mind saying the overhead time is not critical but the number of unnecessary for loops matters. For instance, the d_phase_d_delay can be cached when constructing the design matrix.

PR #719, opened 2020-06-01
Fix PHASE bug and add -padd support

### LINKED ISSUE #717: Support -padd flag
It seems that TEMPO and Tempo2 support the -padd flag on their TOA lines, which applies a phase offsets to TOAs.  PINT does not support this. It does support the -to flag, which is the same but for TIME. It also does support the PHASE command in TOA files so most of the functionality is there.  Probably this is an easy fix.

I found this when trying to process EPTA data that used the -padd flag in their TOA files.

paulray:
Fixes #717


paulray:
Oops, I accidentally included the commits from #715

paulray:
This PR now includes added clock correction info for the EFF observatory. It now uses TEMPO2 eff2gps.clk.  Note that this will not work properly until my PR to Tempo2 is merged: https://bitbucket.org/psrsoft/tempo2/pull-requests/57/fix-eff2gpsclk/diff

There are two bad lines in eff2gps.clk that confused our clock file parser (@demorest @scottransom ) so it was giving 0.0 instead of the correct clock correction.  This isn't really PINT's fault since the file is clearly malformed.  But maybe we should add a check to our clock file parser to verify that the times are in numerical order...

PR #724, opened 2020-06-08
Fixes for pulse numbers and TRACK -2

paulray:
Fix some bugs and inefficiencies in pulse number calculations and usage with TRACK -2


paulray:
@scottransom this PR now also updates the name `InclinationPrior` to `RandomInclinationPrior`

paulray:
I do not know why test_solar_system_body.py started failing. It does not fail with `pytest` or `tox -e py37` on my system so it is some issue with the travis-ci version of jplephem or something like that.

paulray:
I guess that one of these ephemeris files must be bad in the travis builds `["de405", "de421", "de434", "de430", "de436"]`.  That is what `test_solar_system_body` is trying to use with this expression: `a = objPosVel_wrt_SSB("earth", self.tdb_time, ephem=ep)`

PR #726, opened 2020-06-11
Refact component type

luojing1211:
This Pull Request re-facts the way TimingModel stores its components and the APIs to gather the information from the same type of components. The user interface has changed as little as possible. 
Things are changed:
* 'XXComponent_list' becomes a class called "ModelSector", where the components of a type are stored in ModeSector.component_list. 
* The current ModelSector has 3 subclasses 'DelaySector', 'PhaseSector', 'NoiseSector', The functions like 'delay()' and 'phase()' are moved to the corresponding ModelSector class.
* In the TimingModel, the mode sectors are stored in the .model_sectors as a dictionary.  The key is the component type name, 'XXComponent', and the value is the ModelSector class.
* One can access the other model sector functions from TimingModel, Component, and one of the Modelsector class. All other APIs have kept the same. 
* ModelSectors are defined in a separate file. 

Benefits:
* More modulized design and keeps TimingModel class clean. 
* Easy to expend new type of component. The overall function for a new type of component can be defined in the ModelSector, instead of adding methods to the TimingModel class. 
* Remove the name 'XXComponent_list'


luojing1211:
Hi all, could you please take a look at this PR. I would like to move it forward for the wideband TOA development.

src/pint/models/model_sector.py
```diff
@@ -0,0 +1,510 @@
...
+    ---------
+    components: list of `Component` sub-class object.
+        Components that are belong to the same type, for example, DelayComponent.
+
+    Note
+    ----
+    The order of the component in the list is the order a component get computed.
+    """
+
+    _methods = tuple()
```
aarchiba, line 28: What is `_methods` supposed to contain? Presumably a list of some subset of the methods provided by objects in this sector? Not all methods, one assumes? A name that described which methods should be listed would be useful.

It also looks like this has to be populated manually, by listing the methods - it would be preferable to create a decorator one could attach to a method that wanted to be listed in its class' _methods list.

If order doesn't matter, a set or a frozenset would be appropriate here, especially if you're using intersection operators on it.

src/pint/models/model_sector.py
```diff
@@ -0,0 +1,510 @@
...
+    def component_names(self):
+        return [cp.__class__.__name__ for cp in self.component_list]
+
+    @property
+    def component_classes(self):
+        return [cp.__class__ for cp in self.component_list]
+
+    def get_quantity_funcs(self, func_list_name):
+        """List of all model sector functions that contribute to the final
+        modeled quantity.
```
aarchiba, line 81: I don't understand what this is?

src/pint/models/model_sector.py
```diff
@@ -0,0 +1,510 @@
...
+            for k, v in getattr(cp, deriv_dict_name).items():
+                deriv_funcs[k] += v
+        return dict(deriv_funcs)
+
+
+class DelaySector(ModelSector):
+    """ Class for holding all delay components and their APIs
+    """
+
+    _methods = (
```
aarchiba, line 101: It looks like some of these are methods and others are properties?

src/pint/utils.py
```diff
@@ -1391,3 +1392,38 @@ def FTest(chi2_1, dof_1, chi2_2, dof_2):
             log.warning("Models have equal degrees of freedom, cannot preform F-test.")
         ft = False
     return ft
+
+
+def get_component_type(component):
```
aarchiba, line 1397: Does this return the actual type object? Is it "type" in the python sense you mean here?

src/pint/utils.py
```diff
@@ -1391,3 +1392,38 @@ def FTest(chi2_1, dof_1, chi2_2, dof_2):
...
+    Note
+    ----
+    Since a component can be an inheritance from other component We inspect
+    all the component object bases. "inspect getmro" method returns the
+    base classes (including 'object') in method resolution order. The
+    third level of inheritance class name is what we want.
+    Object --> component --> TypeComponent. (i.e. DelayComponent)
+    This class type is in the third to the last of the getmro returned
+    result.
+
```
aarchiba, line 1414: No need for blank lines at the end of a docstring.

tests/test_model_manual.py
```diff
@@ -50,7 +51,7 @@ def test_component_categories(model):
 
     """
     for k, v in model.components.items():
-        assert model.get_component_type(v) != v.category
+        assert get_component_type(v) != v.category
```
aarchiba, line 54: This makes it look like here "type" means the same thing as "category"?

tests/test_model_sector.py
```diff
@@ -0,0 +1,37 @@
...
+    TimingModel,
+    Component,
+)
+from pint.models.model_sector import (
+    ModelSector,
+    DelaySector,
+)
+
+
+class TestSector:
```
aarchiba, line 21: It seems worth checking that the code correctly rejects sectors that share "methods", that all listed methods in existing sectors actually exist, and that the "external_sector_map" actually does what it's supposed to.

Issue #635, opened 2020-03-10
Fitters should update START and FINISH


paulray:
When a fit is run in PINT the output model should have START and FINISH set properly, so when you write out a par file that information is saved.

paulray:
It should also write out parameters that allow Tempo2 to properly process a PINT par file. Most important is "UNITS TDB" but others could be useful as well, like "MODE 1"

paulray:
@champagne-cmd maybe this is an easy one to do?

paulray:
Fixed by #772

PR #727, opened 2020-06-11
Set UNITS parameter from par file

paulray:
This one line change makes PINT actually save the UNITS value from the par file.  Currently there is a UNITS parameter on the top level TimingModel, but it is always "UNSET". This is not good, and it means that `m.as_parfile()` does not write out UNITS TDB, so if you feed a PINT output par file to Tempo2, it sees no UNITS and assumes UNITS TCB, which is wrong.


paulray:
This fixes a bit of what is discussed in #635

PR #732, opened 2020-06-18
Fixed F-test bug where unset parameters would not get assigned values

bshapiroalbert:
When adding parameters to the timing model, if the `parameter` already exists in the `model` but is `UNSET`, just unfreezing it does not assign it a value and the fitter throws errors in this case. Since the input is a predefined `parameter` with a value already there, that value now gets assigned, in addition to unfreezing it. There is also now a short list of parameters to check where the default value should not be zero, so if the input value is zero, it will output a warning and then replace it with a non-zero, hardcoded value.

I believe this is ready to be merged unless anybody has any other comments or suggestions regarding this fix.

### DIFF
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -498,7 +498,7 @@ class Fitter(object):
             parameters from the timing model.
         full_output : Bool
             If False, just returns the result of the F-Test. If True, will also return the new
-            model's residual RMS (us), chi-squared, and number of degrees of freedom of 
+            model's residual RMS (us), chi-squared, and number of degrees of freedom of
             new model.
 
         Returns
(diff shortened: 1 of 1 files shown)


paulray:
How does this fix the UNSET issue?  `m.PAR.value == 0.0` is False when the par is UNSET, so I don't see how this fixes that situation?  I looks like it fixes freeing an existing parameter whose value is 0.0 but shouldn't be.

paulray:
Oh wait, I think I see from your comments above. The new instance of Parameter that gets pass *will* have a value that is not UNSET.  I get it now.

bshapiroalbert:
Hey @paulray, yeah that's pretty much it. What it does is that it checks to see if the parameter that's being added needs to have a non-zero initial value. If it does, it checks to see if the value that the user input for that parameter is zero. If it is, it logs the warning, and changes the user input value of the parameter to be added. Then, no matter what, it sets the value of the parameter to be added to the user input value.

PR #734, opened 2020-06-19
Update Residuals class

paulray:


paulray:
* Added `subtract_mean` kwarg to `Residuals()`. It defaults to True so it is backwards compatible. If you set it False, the mean is not subtracted from the residuals.

* Made `use_weighted_mean` and `subtract_mean` be attributes of the `Residuals` class, which are set at initialization. This way you don't have to keep specifying those option each time you call `resids.update()`, for example.  Of course, if you want to change this behavior, you'd need to instantiate a new `Residuals` object.

* Removed `set_pulse_nums` kwarg from `Residuals`. I couldn't understand what it was really supposed to do, and it was clearing out the delta_pulse_number column, which is bad since that column includes and -padd phase offsets or PHASE commands built into the TOAs. I think this was added by @clp3ef.  It was only used in the pintk GUI.  We need to revisit how to properly implement adding phase offsets in the GUI.

* Tiny fix for undefined variable in solar wind dispersion model

* Fixed a bunch of fragile usage in the tests where kwargs were assigned positionally.

PR #741, opened 2020-07-07
Added weighted rms to optional F-test output

bshapiroalbert:
I've added one more output, the weighted rms of the residuals, for the `fitter` F-tests when `full_output=True`. This is needed for NANOGrav timing, and is also an interesting value that others might like if they want all of the additional output from the F-test. Let me know if you have any comments, questions. This was a very small change and assuming all tests pass it should be ready to merge.


paulray:
As the list grows, I wonder if is would be better/safer to return a dictionary of keys/values instead of a long tuple where the user just has to "know" the order of the values and there is a chance for confusion?

bshapiroalbert:
Hey @paulray. I was wondering about that too. It would certainly make it easier to add parameters in the future as well huh. I will change the output to a dictionary later this afternoon. Thanks!

bshapiroalbert:
Have changed the output to be a dictionary. Any further comments/thoughts welcome.

PR #743, opened 2020-07-09
Fixed division error in python3 

ktzhao:
n/2 becomes a float in python3, may cause an error in next step:' ax[r][c].hist()'

### DIFF
--- a/pint/scripts/event_optimize.py
+++ b/pint/scripts/event_optimize.py
@@ -313,7 +313,7 @@ class emcee_fitter(Fitter):
                     hval = hm(phss[good])
             htests.append(hval)
             if ii > 0 and ii%2==0 and ii<20:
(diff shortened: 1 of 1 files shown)


paulray:
Rather than changing `a/b` to `int(a/b)`, the preferred way would be to use integer division: `a//b`.
Also, I'm not sure why there are conflicts. Make sure your base branch is updated with nanograv/master.

ktzhao:
In file “ pint/scripts/event_optimize.py” , line 383: 
“              r, c = ((ii - 2) / 2) / 3, ((ii - 2) / 2) % 3
                ax[r][c].hist(...) “
in python2.x, it is ok. But in python3.x, r and c will become float. They need to be integer.
 I am not sure which way is better.

PR #744, opened 2020-07-09
fixed division error in python3

ktzhao:


paulray:
Looks good, thanks for that fix!

ktzhao:
Thank you！

PR #747, opened 2020-07-14
Avoid unnecessary 'More than one component' warning

### LINKED ISSUE #627: "More than one component made use of par file line" warnings on NANOGrav par files
Whenever I load a NANOGrav par file using PINT, I get warnings like the following:

```
WARNING: More than one component made use of par file line 'T2EFAC -f L-wide_PUPPI   1.507': [('ScaleToaError', 'EFAC1'), ('ScaleToaError', 'EFAC2'), ('ScaleToaError', 'EFAC3'), ('ScaleToaError', 'EFAC4')] [pint.models.timing_model]
WARNING: More than one component made use of par file line 'T2EQUAD -f L-wide_PUPPI   0.25518': [('ScaleToaError', 'EQUAD1'), ('ScaleToaError', 'EQUAD2'), ('ScaleToaError', 'EQUAD3'), ('ScaleToaError', 'EQUAD4')] [pint.models.timing_model]
WARNING: More than one component made use of par file line 'ECORR -f 430_PUPPI   0.00601': [('EcorrNoise', 'ECORR1'), ('EcorrNoise', 'ECORR2'), ('EcorrNoise', 'ECORR3'), ('EcorrNoise', 'ECORR4')] [pint.models.timing_model]
```

It looks like the noise model components are stepping on each other's toes. In particular, these warnings aren't the result of bad/corrupted par files or unrecognized par file lines, but seem to be symptoms of a problem internal to PINT. Why is this happening and what would we need to do to get rid of it?

rossjjennings:
Fixes #627. The issue there was that, when reading in a par file containing mask parameters, PINT would see the first line containing a specific kind of parameter as pertinent to all the parameters of that kind, causing it to complain (for no good reason) that they all made use of that line. For example, in a par file with multiple JUMPs that weren't explicitly numbered, the parameters JUMP1, JUMP2, etc. would all try to use the first JUMP line (and would fill themselves in with the appropriate values). For subsequent JUMP lines, PINT would recognize them as duplicates, and so treat them as if they started "JUMP2", "JUMP3", etc., so there was no real issue -- the JUMP2, etc. parameters would overwrite the original, incorrect values with new, correct ones.

This PR is a one-line change that avoids the whole situation by making it so that only the first `maskParameter` of a given kind (JUMP1 in the example) is considered to match a line starting with the un-numbered name (JUMP). Since the duplicates all get numbers added to them (the second JUMP becomes JUMP2, etc.), there's no need to have them match un-numbered lines, and doing so just creates the confusing warning.

### DIFF
--- a/src/pint/models/parameter.py
+++ b/src/pint/models/parameter.py
@@ -1421,7 +1421,7 @@ class maskParameter(floatParameter):
     def name_matches(self, name):
         if super(maskParameter, self).name_matches(name):
             return True
-        else:
+        elif self.index == 1:
             name_idx = name + str(self.index)
             return super(maskParameter, self).name_matches(name_idx)
 


paulray:
Was this discussed at the telecon?  Is it fine to merge?

scottransom:
We did discuss it.  But I asked @luojing1211 to take a look at the implementation.  The "fix" causes a change in behavior to the name_matches() method.  But does that match with the intent of that method?  There is no docstring, so I'm not exactly sure what the method is really supposed to be checking for.

rossjjennings:
About the `name_matches()` method: it's not used anywhere outside of the various versions of `from_parfile_line()` in  `parameter.py`. And it's not really part of the public API either. There are a bunch of things like that in the par file parsing logic. In an ideal world they would all have unit tests as well as docstrings, but I don't know if anyone has the time to write such things.

paulray:
It if is not public, maybe rename it to start with "_" or something, and also add a docstring?

rossjjennings:
Well, it's mostly that I think it's not really intended to be public, since it's only used in par file parsing logic. It does have auto-generated documentation [here](https://nanograv-pint.readthedocs.io/en/latest/api/pint.models.parameter.html?highlight=name_matches#pint.models.parameter.maskParameter.name_matches).

PR #749, opened 2020-07-16
Added modelutils.py: convert_to_ecliptic(), convert_to_equatorial()

swiggumj:
Added two convenience utils (in modelutils.py to address a circular import issue in utils.py) to convert astrometry model components from equatorial to ecliptic and vice versa.

### DIFF
--- a/src/pint/modelutils.py
+++ b/src/pint/modelutils.py
@@ -1,13 +1,16 @@
+import logger
 from pint.models.astrometry import AstrometryEquatorial, AstrometryEcliptic
 
 
-def convert_to_equatorial(model):
+def convert_to_equatorial(model,force=False):
     """Converts Astrometry model component, Ecliptic to Equatorial
 
       Parameters
       ----------
-      model
+      model: `pint.models.TimingModel` object
           current model with AstrometryEcliptic component
+      force: boolean, optional
+          will force conversion even if an equatorial component is already present
 
       Returns
       -------
@@ -15,35 +18,51 @@ def convert_to_equatorial(model):
           new model with AstrometryEquatorial component
     """
 
-    if "AstrometryEcliptic" in model.components:
+    if not ("AstrometryEquatorial" in model.components) or force:
+        if "AstrometryEquatorial" in model.components:
+            logger.warning(
+                "Equatorial coordinates already present but re-calculating anyway"
+            )
 
-        c = model.coords_as_ICRS()
-        a = AstrometryEquatorial()
+        if "AstrometryEcliptic" in model.component:
 
-        a.POSEPOCH = model.POSEPOCH
-        a.PX = model.PX
+            c = model.coords_as_ICRS()
+            a = AstrometryEquatorial()
 
-        a.RAJ.quantity = c.ra
-        a.DECJ.quantity = c.dec
-        a.PMRA.quantity = c.pm_ra_cosdec
-        a.PMDEC.quantity = c.pm_dec
+            a.POSEPOCH = model.POSEPOCH
+            a.PX = model.PX
 
-        model.add_component(a)
-        model.remove_component("AstrometryEcliptic")
+            a.RAJ.quantity = c.ra
+            a.DECJ.quantity = c.dec
+            a.PMRA.quantity = c.pm_ra_cosdec
+            a.PMDEC.quantity = c.pm_dec
+
+            model.add_component(a)
+            model.remove_component("AstrometryEcliptic")
+
+            model.setup()
+            model.validate()
+
+        else:
+            raise AttributeError(
+                "Requested conversion to equatorial coordinates, but no alternate coordinates found"
+            )
 
     else:
-        pass
+        logger.debug("Equatorial coordinates already present; not re-calculating")
 
     return model
 
 
-def convert_to_ecliptic(model):
-    """Converts Astrometry model component, Equatorial to Ecliptic
+def convert_to_ecliptic(model,force=False):
+      """Converts Astrometry model component, Equatorial to Ecliptic
 
       Parameters
       ----------
-      model
+      model: `pint.models.TimingModel` object
           current model with AstrometryEquatorial component
+      force: boolean, optional
+          will force conversion even if an ecliptic component is already present
 
       Returns
       -------
@@ -51,25 +70,37 @@ def convert_to_ecliptic(model):
           new model with AstrometryEcliptic component
     """
 
-    # Initial validation to make sure model is not in a bad state?
 
-    if "AstrometryEquatorial" in model.components:
+    if not ("AstrometryEcliptic" in model.components) or force:
+        if "AstrometryEcliptic" in model.components:
+            logger.warning(
+                "Ecliptic coordinates already present but re-calculating anyway"
+            )
+        if "AstrometryEquatorial" in model.components:
 
-        c = model.coords_as_ECL()
-        a = AstrometryEcliptic()
+            c = model.coords_as_ECL()
+            a = AstrometryEcliptic()
 
-        a.POSEPOCH = model.POSEPOCH
-        a.PX = model.PX
+            a.POSEPOCH = model.POSEPOCH
+            a.PX = model.PX
 
-        a.ELONG.quantity = c.lon
-        a.ELAT.quantity = c.lat
-        a.PMELONG.quantity = c.pm_lon_coslat
-        a.PMELAT.quantity = c.pm_lat
+            a.ELONG.quantity = c.lon
+            a.ELAT.quantity = c.lat
+            a.PMELONG.quantity = c.pm_lon_coslat
+            a.PMELAT.quantity = c.pm_lat
 
-        model.add_component(a)
-        model.remove_component("AstrometryEquatorial")
+            model.add_component(a)
+            model.remove_component("AstrometryEquatorial")
+
+            model.setup()
+            model.validate()
+
+        else:
+            raise AttributeError(
+                "Requested conversion to ecliptic coordinates, but no alternate coordinates found"
+            )
 
     else:
-        pass
+        logger.debug("Ecliptic coordinates already present; not re-calculating")
 
     return model
--- a/src/pint/modelutils.py
+++ b/src/pint/modelutils.py
@@ -2,7 +2,7 @@ import logging
 from pint.models.astrometry import AstrometryEquatorial, AstrometryEcliptic
 
 
-def convert_to_equatorial(model, force=False):
+def model_ecliptic_to_equatorial(model, force=False):
     r"""Converts Astrometry model component, Ecliptic to Equatorial
 
     Parameters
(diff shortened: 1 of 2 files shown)


paulray:
@luojing1211 After he adds and removes the components, is there any need to call `.setup()` or `.validate()`, or is that all taken care of?

Joe, can you add a unit test?  It should test both versions. There are both types of par files already in the test data directory. At minimum it should check that a working model is returned. Ideally it should check that the residuals don't change (by much?).

Also the docstring could be improved.  Below is an example. The first line is a short 1-line description. Then there is a longer description, then the lists of Parameters and Returns.
```
"""Compute initial DMX ranges for a set of TOAs
    
    This is an alternative algorithm for computing DMX ranges

    Parameters
    ----------
    divide_freq : Quantity, MHz
        Requires TOAs above and below this freq for a good DMX range
    offset : Quantity, days
        The buffer to include around each DMX range. Warning, may cause bins to overlap?!?
    max_diff : Quantity, days
        Maximum duration of a DMX bin
    verbose : bool
        If True, print out verbose information about the DMX ranges including par file lines.

    Returns
    -------
    mask : bool array
        Array with True for all TOAs that got assigned to a DMX bin
    component : TimingModel.Component object
        A DMX Component class with the DMX ranges included
    """
```

Thanks for doing this!

luojing1211:
We should do validate and setup, so the derivatives will have added. Another issue that I heard from Joe is that PINT does not have a duplicate component category check if using add_component().  We only have it in the model builder.

swiggumj:
Working on the unit test now... Let me know if the updated docstrings look any better.

swiggumj:
@luojing1211 I can raise that issue too. I had hoped I could get away with completing this PR first, but it's probably good to get into the habit of raising issues immediately when I find them.

luojing1211:
@swiggumj For issues, we should do it immediately, someone may fix it in the next 10 minutes.

src/pint/modelutils.py
```diff
@@ -2,9 +2,17 @@
...
 
 def convert_to_equatorial(model):
-    """Converts pulsar's PulsarEcliptic coordinates to ICRS,
-    adds the AstrometryEquatorial model component with the
-    correct values, and removes AstrometryEcliptic component.
+    """Converts Astrometry model component, Ecliptic to Equatorial
+
+      Parameters
+      ----------
+      model
```
luojing1211, line 9: After the parameter name, giving a required type would be helpful, like:
```
model: `pint.models.TimingModel` object
```

swiggumj:
@paulray I noticed there are tests for 1855/1613 (par files with ecliptic/equatorial coords, respectively). Would it be ok to add unit tests there for these utilities, or should I create a separate test script?

src/pint/modelutils.py
```diff
@@ -2,9 +2,17 @@
 
 
 def convert_to_equatorial(model):
-    """Converts pulsar's PulsarEcliptic coordinates to ICRS,
-    adds the AstrometryEquatorial model component with the
-    correct values, and removes AstrometryEcliptic component.
+    """Converts Astrometry model component, Ecliptic to Equatorial
+
+      Parameters
```
luojing1211, line 7: I am not sure if this will cause any indentation error when doc building. I think the best would be like
```
"""
Convert Astro.....

Parameters
---------------
model: `pint.models.TimingModel` object
    Timing model with astrometry Ecl.....

Return
---------
    New timing model with eqtor
"""
```

luojing1211: The mark down did not line up. But if you remove the ``` magic will happen.

src/pint/modelutils.py
swiggumj: I can certainly do that, although I made this decision by going and finding/copying other examples like this in existing code. So it might be good to impose a standard if anyone else, like me, looks for examples this way.

src/pint/modelutils.py
swiggumj: I don't think I understand this comment. I ran black, but it sounds like you're talking about something else.

paulray:
> @paulray I noticed there are tests for 1855/1613 (par files with ecliptic/equatorial coords, respectively). Would it be ok to add unit tests there for these utilities, or should I create a separate test script?

I'm fine with adding your tests to those test scripts.

src/pint/modelutils.py
paulray: I've been trying to be consistent using numpydoc style:
https://numpydoc.readthedocs.io/en/latest/example.html#source

luojing1211: Sorry for lacking context, there are two parts. 

1. It is the indentation of the docstrings.  
```
"""
  Parameters
  ---------------
"""
```
 vs.

```
"""
Parameters
---------------
"""
```
2. It is about the github comment box. 
if I write the docstring using numpy styles here in this comment. It will look like: 

"""
Convert Astro.....

Parameters
---------------
model: `pint.models.TimingModel` object
    Timing model with astrometry Ecl.....

Return
---------
    New timing model with eqtor
"""
However, if we use "```" outside of """, which indicates a code block, it will look like 
```
"""
Convert Astro.....

Parameters
---------------
model: `pint.models.TimingModel` object
    Timing model with astrometry Ecl.....

Return
---------
    New timing model with eqtor
"""
```

src/pint/modelutils.py
```diff
@@ -0,0 +1,59 @@
...
+        a.RAJ.quantity = c.ra
+        a.DECJ.quantity = c.dec
+        a.PMRA.quantity = c.pm_ra_cosdec
+        a.PMDEC.quantity = c.pm_dec
+
+        model.add_component(a)
+        model.remove_component("AstrometryEcliptic")
+
+    else:
+        pass
```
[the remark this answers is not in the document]
swiggumj: I've addressed this now.

src/pint/modelutils.py
swiggumj: This has been addressed.

swiggumj:
I ended up just making a stand-alone unit test.

src/pint/modelutils.py
swiggumj: Hmmm I think I forgot about this. Let me double-check to see if it's fixed...

swiggumj: Ok, I think I've fixed this in the latest commit. Doesn't look like it's showing up quite yet...

src/pint/modelutils.py
```diff
@@ -0,0 +1,105 @@
+import logging
+from pint.models.astrometry import AstrometryEquatorial, AstrometryEcliptic
+
+
+def convert_to_equatorial(model, force=False):
```
luojing1211, line 5: I think a better name is "pulsar_ecl_to_equatorial" or "pulsar_ecliptic_to_equatorial" if this function only handles from pulsar ecl to equatorial. If David's galactic coordinate PR get merged, the user will get confused about this function.

src/pint/modelutils.py
```diff
@@ -0,0 +1,105 @@
...
+            model.setup()
+            model.validate()
+
+        else:
+            raise AttributeError(
+                "Requested conversion to equatorial coordinates, but no alternate coordinates found"
+            )
+
+    else:
+        logger.debug("Equatorial coordinates already present; not re-calculating")
```
luojing1211, line 52: Maybe a warning.

src/pint/modelutils.py
```diff
@@ -0,0 +1,105 @@
...
+
+    return model
+
+
+def convert_to_ecliptic(model, force=False):
```
luojing1211, line 57: Same but this is from equatorial to pulsar ecliptic.

src/pint/modelutils.py
```diff
@@ -0,0 +1,105 @@
...
+            model.setup()
+            model.validate()
+
+        else:
+            raise AttributeError(
+                "Requested conversion to ecliptic coordinates, but no alternate coordinates found"
+            )
+
+    else:
+        logger.debug("Ecliptic coordinates already present; not re-calculating")
```
luojing1211, line 103: Same as the last one.

tests/test_modelutils.py
```diff
@@ -0,0 +1,103 @@
+import logging
+import os
+import unittest
+
+import astropy.units as u
+import numpy as np
+
+import pint.models.model_builder as mb
```
luojing1211, line 8: I don't think this import is needed.

tests/test_modelutils.py
```diff
@@ -0,0 +1,103 @@
...
+import pint.toa as toa
+import test_derivative_utils as tdu
```
luojing1211, line 10: Same for this one.

src/pint/modelutils.py
swiggumj: To be clear, I'm going with `model_x_to_y` (where x and y stand for equatorial and ecliptic, or vice versa).

src/pint/modelutils.py
swiggumj: Fixed.

src/pint/modelutils.py
swiggumj: Fixed (see above).

src/pint/modelutils.py
swiggumj: Fixed.

tests/test_modelutils.py
swiggumj: It is used in the test, so I'll leave it for now.

tests/test_modelutils.py
swiggumj: Now I'm a bit confused about which imports you're referring to, but I removed `tdu` since it is definitely not used.

tests/test_modelutils.py
luojing1211: I see. This is the old method. we can do 
```
```
And use get_model

swiggumj: Please specify which one. All are currently being used in the test...

tests/test_modelutils.py
```diff
@@ -0,0 +1,103 @@
...
+    def setUpClass(cls):
+        # J0613 is in equatorial
+        cls.parfileJ0613 = os.path.join(
+            datadir, "J0613-0200_NANOGrav_dfg+12_TAI_FB90.par"
+        )
+        cls.timJ0613 = os.path.join(datadir, "J0613-0200_NANOGrav_dfg+12.tim")
+        cls.toasJ0613 = toa.get_TOAs(
+            cls.timJ0613, ephem="DE405", planets=False, include_bipm=False
+        )
+        cls.modelJ0613 = mb.get_model(cls.parfileJ0613)
```
luojing1211, line 32: mb.get_model() should be get_model(), if we do 
```
```
So does for the other ones.

swiggumj: Ok, I've fixed both of these. FYI, I copied/pasted this from another test function, so it might be good to check these and be consistent. I can do that soon.

PR #752, opened 2020-07-16
Addgalactic

dlakaplan:
An attempt to add utilities for returning `astropy` `SkyCoord` objects in Galactic coordinates including proper motions.

I have put in some basic unit tests.  These do
* a check of the basic proper motion corrections using the `astropy` and `PINT` routines in native coordinates
* a check of the conversions for t=now and some time in the future using `astropy` and `PINT` routines

the precisions for these comparisons are a bit arbitrary.  I found that most were good to ~1 uas, but some were worse (~100 uas).  Not sure of the origin of that discrepancy.

### DIFF
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -20,6 +20,7 @@ from pint.models.parameter import (
 )
 from pint.models.timing_model import DelayComponent, MissingParameter
 from pint.pulsar_ecliptic import OBL, PulsarEcliptic
+from pint.utils import add_dummy_distance, remove_dummy_distance
 
 astropy_version = sys.modules["astropy"].__version__
 mas_yr = u.mas / u.yr
@@ -263,30 +264,43 @@ class AstrometryEquatorial(Astrometry):
     def get_psr_coords(self, epoch=None):
         """Returns pulsar sky coordinates as an astropy ICRS object instance.
 
+        Parameters
+        ----------
+        epoch: `astropy.time.Time` or Float, optional
+            new epoch for position.  If Float, MJD is assumed
+
+        Returns
+        -------
+        position
+        ICRS SkyCoord object optionally with proper motion applied
+
         If epoch (MJD) is specified, proper motion is included to return
         the position at the given epoch.
 
-        If the ecliptic coordinates are provided,
         """
         if epoch is None or (self.PMRA.value == 0.0 and self.PMDEC.value == 0.0):
             dRA = 0.0 * u.hourangle
             dDEC = 0.0 * u.deg
             broadcast = 1
             newepoch = self.POSEPOCH.quantity
+            return coords.SkyCoord(
+                ra=self.RAJ.quantity + dRA,
+                dec=self.DECJ.quantity + dDEC,
+                pm_ra_cosdec=self.PMRA.quantity * broadcast,
+                pm_dec=self.PMDEC.quantity * broadcast,
+                obstime=newepoch,
+                frame=coords.ICRS,
+            )
         else:
-            dt = (epoch - self.POSEPOCH.quantity.mjd) * u.d
-            dRA = dt * self.PMRA.quantity / numpy.cos(self.DECJ.quantity.radian)
-            dDEC = dt * self.PMDEC.quantity
-            broadcast = numpy.ones_like(epoch)
-            newepoch = Time(epoch, format="mjd")
-        return coords.SkyCoord(
-            ra=self.RAJ.quantity + dRA,
-            dec=self.DECJ.quantity + dDEC,
-            pm_ra_cosdec=self.PMRA.quantity * broadcast,
-            pm_dec=self.PMDEC.quantity * broadcast,
-            obstime=newepoch,
-            frame=coords.ICRS,
-        )
+            if isinstance(epoch, Time):
+                newepoch = epoch
+            else:
+                newepoch = Time(epoch, format="mjd")
+            position_now = add_dummy_distance(self.get_psr_coords())
+            position_then = remove_dummy_distance(
+                position_now.apply_space_motion(new_obstime=newepoch)
+            )
+            return position_then
 
     def coords_as_ICRS(self, epoch=None):
         """Return the pulsar's ICRS coordinates as an astropy coordinate object."""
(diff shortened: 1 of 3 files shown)


dlakaplan:
On this topic, is there a reason the `SkyCoord` objects returned by PINT don't have the `obstime` set?  I have found that in order to use the native `astropy` proper motion routines they need both `obstime` (which we have) and distance (which we don't), but I don't see the disadvantage to including the time for use down-stream.  @paulray ?  @luojing1211 ?

luojing1211:
Astrometry module is one of the first models we write back to 2012. At that time, astropy was not very robust. I have no problem to integrate it as long as it passes our tests.

dlakaplan:
Thanks, @luojing1211 .  I agree that astropy has changed a lot recently.  
Are there any specific tests that I should look at? I did some quick searches for relevant code but didn't see anything.  I didn't see _any_ proper motion tests until I added one, but maybe I just missed them.

luojing1211:
I am mostly worried about precision. I think our end to end residual precision test would be the strongest. But adding the test of new vs old method if a good idea.

dlakaplan:
Which residual test do you mean?  The way I view using these (and maybe my view is limited):
* If you are going to barycenter residuals keep all of the systems (time & coordinates) internal to PINT
* But, it can be useful to output astropy native formats for other purposes. What I _wouldn't_ do is go PINT -> astropy -> PINT

In any case, the routines to output `SkyCoord` are there anyway.  All I'm suggesting is adding a new coordinate system (which as far as I can tell isn't a native system used by PINT) and adding `POSEPOCH` to the `obstime`.  

As far as precision goes, I am not convinced that everything is done identically, since there are changes at the ~microarcsec level between the different methods.  However, I don't have a *truth* for comparison.

dlakaplan:
In fact I can see areas where the PINT implementation is *not* correct:
```
            dt = (epoch - self.POSEPOCH.quantity.mjd) * u.d
            dRA = dt * self.PMRA.quantity / numpy.cos(self.DECJ.quantity.radian)
            dDEC = dt * self.PMDEC.quantity
```
doesn't take into account the fact that the DEC changes over the course of the time period.

luojing1211:
That is interesting. Could you be more specific about DEC changes? Is there any other DEC drifting other than PMDEC?

paulray:
Indeed that is incorrect. In most use cases the error will be very small, of course.  But, what is the right way to do it?  If you make it into a SkyCoord with PM and epoch, then have astropy update the coordinates to the requested epoch?

dlakaplan:
No, but imagine an infinitesimal change in the Dec from Dec0 -> Dec0+dDec.  Then the new infinitesimal coordinate update should be computed with this new position.  

I don't think the change is large.  And I have certainly made the same approximation myself.  I think the way astropy does it is to convert into a 3D cartesian frame where the velocities are inertial and applying them.  This is why a distance is needed.  It's also what I did a while ago (https://ui.adsabs.harvard.edu/abs/2002ApJ...571..447K/abstract).

Look at the attached code.  It starts a source at (RA,Dec)=(0,0) and applies a proper motion of (100 mas/yr,100 mas/yr).  It computes the results using `astropy` and `PINT`.  After 500d they differ by 0.0009689192879343939 mas.

```
from astropy.time import Time
from astropy.coordinates import SkyCoord
import numpy as np

t0=Time(55000,format='mjd')
t1=t0 + 1000*u.d

c0 = SkyCoord(ra=0*u.deg,
              dec=0*u.deg,
              pm_ra_cosdec = 100*u.mas/u.yr,
              pm_dec = 100*u.mas/u.yr,
              obstime = t0,
              distance=1*u.kpc,
              )

c1 = c0.apply_space_motion(new_obstime = t1)

# PINT method
dt = (t1 - t0)
dRA = dt * c0.pm_ra_cosdec / np.cos(c0.dec)
dDEC = dt * c0.pm_dec
c1_PINT = SkyCoord(
            ra=c0.ra + dRA,
            dec=c0.dec + dDEC,
            )

print('Astropy Separation: {}'.format(c1.separation(c0).to(u.mas)))
print('Naive separation: {}'.format((np.sqrt(c0.pm_dec**2 + c0.pm_ra_cosdec**2)*(t1-t0)).to(u.mas)))
print('PINT separation: {}'.format(c1_PINT.separation(c0).to(u.mas)))
print('PINT to Astropy separation: {}'.format(c1_PINT.separation(c1).to(u.mas)))
```
gives:
Astropy Separation: 387.1896025722623 mas
Naive separation: 387.19057149160716 mas
PINT separation: 387.1905714915502 mas
PINT to Astropy separation: 0.0009689192879343939 mas

dlakaplan:
@paulray : I think that is better, but as I mentioned you also need to supply a (dummy) distance to get it to work

dlakaplan:
I just added some routines in `utils` to add/remove dummy distances to make these calculations easier.  I don't know if that's the best place to add them.

I was also a bit limited in the coordinate frames that I looked at (ICRS, Galactic, PulsarEcliptic).  Probably there is a more elegant way of handling these but I got somewhat bogged down in the SkyCoord class hierarchy.

paulray:
So, can this be merged now?
If we merge, I suggest we also open an issue suggesting the use of astropy SkyCoords more fully in astrometry, so they handle the proper motions correctly.

dlakaplan:
> If we merge, I suggest we also open an issue suggesting the use of astropy SkyCoords more fully in astrometry, so they handle the proper motions correctly.

I think I looked at most of those places in `models/astrometry.py` and updated them, but this should be done more carefully/comprehensively

swiggumj:
Seems nothing has happened on this since last Wednesday and I haven't gotten a response to my post on gitter. Would this help kick off the Travis run again (https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit)? @luojing1211 @paulray

paulray:
I did trigger a new build, but it still fails.  It is running `jupyter nbconvert --execute --inplace examples/*.ipynb` and there seems to be a 30s timeout that causes the MCMC walkthrough to abort. I don't know if that notebook takes longer to run than it used to, or if the 30s timeout is new behavior. But, @scottransom is working on MCMC stuff now and also having trouble with the notebook tests, so I suggest we merge this and sort out the MCMC stuff separately.

Issue #733, opened 2020-06-19
Solar wind dispersion broken


paulray:
https://github.com/nanograv/PINT/blob/3486bb7ede296008d868e727d3859c33e335a1cd/src/pint/models/solar_wind_dispersion.py#L111

This line uses the variable `AUdist`, which is not defined anywhere, so it fails when you try to fit.

paulray:
I suppose it should be replaced with `const.au` and the unit tests should be updated to make sure that solar wind fitting works.

paulray:
In addition, when I fit for NE_SW, I get a very different answer than Tempo2 gets, e.g. Tempo2 gives 4.9, while PINT gives -1.8.  I suspect the equations are wrong.

luojing1211:
Is the derivative wrong or the delay function is wrong?

paulray:
Perhaps both?

paulray:
If I use the tempo2 "fake" plugin to make TOAs from a simple timing model with NE_SW 10 and then plot the residuals in PINT, I get this plot. Ugh. It looks like the solar wind in PINT is making both positive and negative contributions!

![fakeSW10](https://user-images.githubusercontent.com/1771428/85332209-87d97380-b4a5-11ea-97f7-51047873fc00.png)

paulray:
@scottransom maybe this would be good for your student to look into since he's already doing the tropospheric delay?

zimmyneutron:
@paulray To clarify, you are using tempo2 fake plugin for data, then using PINT to calculate the residuals between the model and the fake.tim data without fitting?

paulray:
Yes, I think that is what I did to make that plot.

PR #755, opened 2020-07-21
Solar Wind Model

zimmyneutron:
Fixing undefined `AUdist` in the derivative function, and the improper behavior of the fitter described in #733


paulray:
This is OBE. Closing...

Issue #757, opened 2020-07-30
Error when "copy"ing WidebandTOAFitter object


bshapiroalbert:
When trying to run F-tests on the WidebandTOAFitter, the `ftest()` function copies the `fitter` with `copy.deepcopy()` so that the tests are on a copy of the `fitter` and do not change the original. But I'm getting an error in the `copy.deepcopy()` line in the F-test when trying to run then on the WidebandTOAFitter:

`TypeError: __new__() missing 2 required positional arguments: 'derivative_quantity' and 'quantity_unit'`

The full error report is attached, but I'm not really sure how to fix this error.

<img width="862" alt="WidebandTOAFitter_copy_error" src="https://user-images.githubusercontent.com/33037158/88957466-e8aa5780-d26c-11ea-8fc4-26e0ada65ec0.png">

luojing1211:
I will fix it in a little bit.

PR #759, opened 2020-08-04
Fix wideband fitter copy

luojing1211:
This fixes the issue of #757


paulray:
Can you update CHANGELOG.md before merging this PR?  The CHANGELOG doesn't mention that wideband TOAs are now supported.

PR #761, opened 2020-08-07
Gui mod

champagne-cmd:
Miscellaneous bug fixes and functionality improvement to `pintk`, including:

- fixes to keypress options so that 'Shift' doesn't unintentionally activate an option

- fixes to click-and-drag so that not capturing any toas won't throw `pintk` into an endless loop of errors

- `pintk` now supports multiple coloring schemes for the TOAs

- DispersionDMX defaults to the off/unchecked state to avoid overcrowding the `pintk` window upon start-up

- keypress option 'p' to print TOA info now displays TOA flags

- `.tim` file `JUMPs` no longer display as `gui_jumps`


champagne-cmd:
@paulray there seems to be an issue with a notebook - MCMC_walkthrough, in the fit_toas function. I feel like I've gotten issues with this one before... but I'm not sure how to fix this:
`ValueError: Initial state has a large condition number. Make sure that your walkers are linearly independent for the best performance`

scottransom:
@champagne-cmd This is caused by one of the things I pushed up yesterday and I'm not understanding why.  I've been looking into it and will try to get it fixed soon.  Sorry about this!

PR #762, opened 2020-08-10

bshapiroalbert:
I have updated the F-tests in the `fitter` so that they are compatible with the `WidebandFitter` and subsequent `CombinedResiduals` classes. The test automatically figures out if you're using a `WidebandFitter` or `GLS/WLSFitter` and get the appropriate values of chi-squared for the actual F-test. If using the `full_output` option with a `WidebandFitter`, it will also return the time and DM chi-squared values separately (as well as the total value) and the weighted and unweighted rms of the DM residuals. If using a `GLS/WLSFitter`, output remains unchanged from previous version. Documentation and tests have been updated as well. (Also had an error when initially committing this, and thought I had two commits, "Updated F-test to work with WidebandFitter" and also "black code", but it only read that second one. Everything got pushed correctly though.) Comments, questions, concerns, etc. welcome.


bshapiroalbert:
Looks like the failed tests were on the example notebooks and the documentation build, but otherwise it seems to be ready to merge pending other comments.

Issue #736, opened 2020-06-25
New SciPy version breaking Travis-CI


champagne-cmd:
Some change in the latest version of SciPy (1.5.0) is causing Travis tests to fail, specifically when testing the notebook `understanding_fitters.md`. The error says `numpy.any()` is being called on an unsupported type when calling `powfit.fit_toas()`. Excluding SciPy version 1.5.0 in `requirements.txt` and `setup.cfg` bypasses the issue, but we should figure out what changed in SciPy to cause the error.

PR #764, opened 2020-08-10
SciPy Version Fix

champagne-cmd:
Fix for Travis error involving SciPy v. >1.5.0 - `np.any` no longer accepts `Quantity` arrays. Wrapper function `minimize_func` now returns the `.value` of the result. ( addresses #736 )


paulray:
The build failure is something about a docstring in polycos.py, which isn't modified in this PR. Not sure what is going on, but it shouldn't prevent merging this.

paulray:
Chloe, do you have any idea why this scipy issue happened?  It is strange since `np.any()` is part of numpy, not scipy and `Quantity` is defined in astropy.  Also I believe numpy has been evolving to have better support for `Quantity`.

champagne-cmd:
I'm afraid I don't know @paulray , I skimmed through the release notes for SciPy 1.5.0 but didn't see anything that immediately stuck out as being relevant. 
Also, this failure is occurring with #762 as well, so perhaps a pull request before yesterday caused the docstring error? I looked at the docstring though and it has closing quotes, so I'm not sure what the problem is there either.

paulray:
I think I fixed the docstring in `nanograv/master`, so can you merge that into your PR?  That should fix the build failure

paulray:
I thought about this a bit and I think a better fix may be to ensure that `resids.chi2` is not a quantity.  Right now it sometimes it, but sometimes isn't.  It is computed by `calc_chi2()` in either the `Residuals` or `WidebandDMResiduals(Residuals)` class.  Those can return `np.inf` which is not a Quantity, or a Quantity with dimensionless units, or a pure number (in the case when the model has a noise component and the chi2 is computed by the GLSFitter). So, I think that in the two `calc_chi2()` routines is where it should ensure that chi2 is a pure number, not a Quantity.

champagne-cmd:
@paulray when I implement that, the test_model_ifunc.py fails because it uses chi2.value. Should this test be modified? Every other test passes that is suppose to pass.

paulray:
Yes, I'd remove the .values from that test.

champagne-cmd:
Closes #736

PR #765, opened 2020-08-11
Fix wideband toa bugs

luojing1211:


luojing1211:
This PR should be able to go. I will add it to the changelog and merge it.

Issue #654, opened 2020-04-06
PINT incorrectly handles par files with mixed ecliptic and equatorial coords


paulray:
This is related to #304 

I made 4 par files using different combinations  of equatorial and ecliptic coordinates for the position (POS) and proper motion (PM). When both POS and PM are the same coordinate system all works fine. When POS is equatorial and PM is ecliptic, PINT runs but gives incorrect answers. When POS is ecliptic and PM is equatorial, PINT print some warnings and raises an exception:
```
WARNING: Unrecognized parfile line 'ELONG          192.7997211876664938878800371 1  0.00000102102277960435' [pint.models.timing_model]
WARNING: Unrecognized parfile line 'ELAT           -9.9435557450403477515801787 1  0.00000664803199093257' [pint.models.timing_model]
pint.models.timing_model.MissingParameter: Astrometry.RAJ
```
For what is it worth, Tempo2 does not work correctly for either of the mixed system par files, but does not crash. It just reads the values and interprets them incorrectly.

So, I propose that PINT should raise an exception when the coordinate system for the POS is different from the PM.  And, it should probably cue off the POS coordinate system. So, if RA is specifies thes PMELONG or PMELAT should raise an exception, and if ELONG is specified then PMRA or PMDEC should raise an exception.

paulray:
As a note, TEMPO does the **right thing**, as @demorest predicted:
```
% tempo -f TDB_ECLPOS_EQUPM.par TOAs_NUPPI_1484_F4Tp.tim
 TEMPO v 13.000 Princeton/ATNF Pulsar Collaboration
 Data from TOAs_NUPPI_1484_F4Tp.tim,   Input parameters from TDB_ECLPOS_EQUPM.par                                                                                                                                            
ERROR: cannot mix ecliptic and equatorial coordinates

% tempo -f TDB_EQUPOS_ECLPM.par TOAs_NUPPI_1484_F4Tp.tim
 TEMPO v 13.000 Princeton/ATNF Pulsar Collaboration
 Data from TOAs_NUPPI_1484_F4Tp.tim,   Input parameters from TDB_EQUPOS_ECLPM.par                                                                                                                                            
ERROR: cannot mix ecliptic and equatorial coordinates
```

paulray:
Note that Tempo2 now does the *right thing*, as of this commit:
https://bitbucket.org/psrsoft/tempo2/commits/adb29f71a9f945d3773373df78929a0f63f608ca

[...]

PR #767, opened 2020-08-12
Check for Mixed Coord. Systems in Par File

champagne-cmd:
Adds a check when reading in `.par` file to make sure POS and PM coordinate systems are the same.

### DIFF
--- a/src/pint/models/model_builder.py
+++ b/src/pint/models/model_builder.py
@@ -177,7 +177,8 @@ class ModelBuilder(object):
         return prefixs
 
     def build_model(self, parfile=None, name=""):
-        """Read parfile using the model_instance attribute.
+        """Read parfile using the model_instance attribute. Throws error if 
+           mismatched coordinate systems detected.
         Parameters
         ---------
         name: str, optional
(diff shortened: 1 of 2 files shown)


champagne-cmd:
@paulray two tests will fail in `test_model_manual` because they use a par file with mixed coordinate systems, `J1744-1134.basic.ecliptic.par`. It has `RAJ`, `DECJ`, and `ELAT` commented out and `ELONG` uncommented, and `PMRA` and `PMDEC`. I would comment out `ELONG` and uncomment `RAJ` and `DECJ`, but the file's name seems to suggest it is meant to be an ecliptic file. How should I modify this?

paulray:
I think you should uncomment ELAT, so it has ecliptic position and equatorial proper motion. This *should* throw an exception, and that seems to be what is expected, since the tests is marked "xfail".

luojing1211:
Hi @champagne-cmd, it looks good. I would give more explanation to the error message. 
For example, "Cannot have Ecliptic PM with Equatorial POS in par file", I would say, "Cannot have Ecliptic poper motion parameters (PMELONG/PMELAT) with Equatorial position parameters (RAJ/DECJ) in par file". You can also dynamically give the conflicted parameter names in the message, so the user would know which parameters are conflicting.

champagne-cmd:
@paulray the two tests I was talking about are `test_compare_get_model_new_and_old_all_parfiles` and `test_get_model_roundtrip`, which aren't marked "xfail". The former is testing the `get_model_new` function, and the latter is testing the `as_parfile` function. I think maybe they are using this `.par` file when they shouldn't be, because other tests that are marked "xfail" use the `.par` file. Perhaps putting this `.par` file in the `bad_trouble` array is a solution.

paulray:
I see, sorry! I agree that this bad par file (with mixed ECL and EQU coordinates) should be in the bad_trouble list.  That should solve it.

champagne-cmd:
Fixes #654

PR #771, opened 2020-08-13
Default set POSEPOCH w/ warning

swiggumj:
I removed the conditional `if self.PMRA.value != 0.0 or self.PMDEC.value != 0.0:` in astrometry self.validate() functions so that POSEPOCH is set to be PEPOCH (if not none) by default. I also added a warning that is triggered when this happens.

### DIFF
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -233,17 +233,17 @@ class AstrometryEquatorial(Astrometry):
         for p in ("RAJ", "DECJ"):
             if getattr(self, p).value is None:
                 raise MissingParameter("Astrometry", p)
-        # If PM is included, check for POSEPOCH
-        if self.PMRA.value != 0.0 or self.PMDEC.value != 0.0:
-            if self.POSEPOCH.quantity is None:
-                if self.PEPOCH.quantity is None:
-                    raise MissingParameter(
-                        "AstrometryEquatorial",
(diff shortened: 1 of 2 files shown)


paulray:
Your tests are failing, but I think the problem is a poorly designed test. I think in test_model_manual.py you it should set `tm.PEPOCH` at line 166. If it is not set, then `tm.validate()` *should* fail, in my opinion.

swiggumj:
Ok I'll have a look at that. Sounds like you're saying I don't need to implement entirely new unit tests though? :)

paulray:
Well, you don't have to, but it certainly wouldn't hurt to add a short tests that verifies the new behavior you are trying to add.

swiggumj:
Sorry! I'm still getting the hang of conventions like this.

paulray:
It isn't really a convention. Just something that helps me. I've done a lot of looking at PINT tests to see why they are failing and often I can't figure out what exactly they are trying to test, so I figured we should try to make sure there is some explanation in the code.

swiggumj:
I think this is ready to go, unless we need to discuss more next week on the PINT call. @paulray @luojing1211

paulray:
Cool. I'm having a Zoom with Chloe this afternoon on her last day. I'll talk over her two PRs and plan to merge them, and this one at that time.

PR #772, opened 2020-08-13
START/FINISH Updated and Displayed

champagne-cmd:
The `START` and `FINISH` parameters in the .par files are now incorporated as frozen `MJDParameters`, displayed in .par file output, and updated after fitting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - Added ability to disable subtracting mean from residuals
 - Added track_mode to Residuals to select pulse number tracking without needing the model to have TRACK -2
 - Added support for wideband-TOA fitting (Pennucchi 2019).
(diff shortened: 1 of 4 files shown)


paulray:
Do you also need to remove START and FINISH from ignore_params?
Also, if you have time to add a unit test that show that they get updated that would be great!

paulray:
Oh, and please add a line to CHANGELOG

champagne-cmd:
I think they still need to be in `ignore_params` so that they register as "interesting lines" like `UNITS` and `EPHVER` and are addressed in `read_parfile` in the same fashion. I will add the test and modify the CHANGELOG.

paulray:
Oops, I see a couple of tests failed.  Probably small issues.  You can run them on your local machine one at a time with `pytest -s -v tests/test_fitter.py`, for example.  The "-s" makes in print out stdout and "-v" makes it more verbose about what test it is running.

PR #776, opened 2020-08-24
Make solar wind use dispersion function

luojing1211:
This PR does restructure the solar wind model to use the dispersion base class's functions. It does not intend to change any calculations. It is still missing useful tests. 

@zimmyneutron I think it would be better to have your fix based on this version. But feel free to test this branch against your version.


luojing1211:
This implemented by the wideband TOA branch. I will close this.

PR #779, opened 2020-08-28
Reblacken in light of updates to black style

### LINKED ISSUE #778: Black 20.8 reformats docstrings
The new version of `black` pulled in by `pip install` reformats docstrings and a few other things (that were badly formatted to start off), so `master` no longer passes the blackening test.

aarchiba:
Closes #778

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -20,3 +20,6 @@
 
 # Blacken code base
 ef448d5271e7ce4a2de912dd2a76a629e7a569cb
+
+# Re-blacken source code with black 20.8b1
+268f6de6a6668981984729ecffb74461a0b37de0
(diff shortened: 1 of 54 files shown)


paulray:
Does this mean we need to change requirements to black>=20.8?  I don't think we should be using a beta version, however.

aarchiba:
Well, the changes are generally things like rearranging incorrectly formatted docstrings, and I think the change in format is a good thing. And `black` is not a dependency to use PINT, just to work on the code. But I think pinning a version, in `requirements_dev.txt` and in Travis, might simplify the process - we could then bump the version in the same PR that re-blackens.

matteobachetti:
There are many open PRs. Will this create a truckload of conflicts?

aarchiba:
> There are many open PRs. Will this create a truckload of conflicts?

We ran into this with the original blackening. If you touch part of the code that is affected by the re-blackening changes, yes, you'll have a conflict; it's generally easy to resolve by running black over the PR code but we did try to merge or close as many PRs as we could before the original blackening.

aarchiba:
#780 provides a probably better alternative to this.

paulray:
OK, for now, I merged #780. This means the master should pass tests and the current PRs are all good. When we are ready, we can do this re-blackening and then enforce black>20.8.
Question: Is code blackened with 20.8 backwards compatible with earlier versions (since I understand it is just docstrings)?

PR #783, opened 2020-09-24
Updates to model.compare function

almcewen0:
Aside from some small aesthetic changes and a docstring update, this PR updates model.compare to return information with varying levels of verbosity. Specifically, options for verbosity include "max", "med", "min", and "check". These return in the following ways:

                "max"      - print all lines from both models whether they are fit
                                   or not (note that nodmx will override this); DEFAULT
                "med"      - only print lines for parameters that are fit
                "min"       - only print lines for fit parameters for which
                                   diff_sigma > threshold_sigma
                "check"   - print significant changes with astropy.log.warning, not
                                   as string

The threshold_sigma is now included as an optional input (default = 3). Verbosities which return strings (i.e. not "check") will also include asterisks (*) for parameters that are more uncertain in the non-reference model, and exclamation points (!) when the uncertainties in either model are larger than threshold_sigma.


paulray:
Hmm. I worry about the epoch changing in this code.  Does that happen on a  copy of the model, or might this function modify a model it was called on?  That  would be a bad side effect.

Also, we should discuss the adding of START and FINISH as top level parameters. Were they not already included in PINT?  Is this being done the right way?

almcewen0:
> Hmm. I worry about the epoch changing in this code. Does that happen on a copy of the model, or might this function modify a model it was called on? That would be a bad side effect.

Good point. My thought was just to make sure that we align the two in time to suppress warnings for parameters that might change over long spans of time. So perhaps I should just save the various initial EPOCHs and then reset them at the end.  
 
> Also, we should discuss the adding of START and FINISH as top level parameters. Were they not already included in PINT? Is this being done the right way?

I'm not too sure what you mean by this... perhaps there is somebody else who would have a better answer.

paulray:
I think I'd feel safer if you made a copy of the model object and changed the epoch on that, rather than modifying and un-modifying the object that is passed in.

almcewen0:
Screen Shot 2020-10-07 at 5 31 15 PM
here is an example of the astropy.log warning issue: the ELONG value is changed in the first iteration of the loop, but the astropy warning for it shows up in the second iteration alongside of the ELAT warning. I also included the test I was running.

paulray:
I suspect the problem is a logic problem and not strange behavior of the logging.
Can you post the par file used in the above test?

Also, there is some dubious python in the code above.   I've never seen anyone use type(None) before and I think you should not compare to None with = or != but rather should use `is None` or `is not None`. But, I don't think either of those are causing the problem.

Also, it makes a lot of use of `param.value` that probably should be avoided -- `param.quantity` has units for a  reason.
Finally is your definition of "significant" 3 sigma?  If so, then your test should probably change the parameter by 3.1 sigma. If you use 3.0 sigma then a rounding error could cause it to not be "significant".

Hopefully one of those things fixes the issue!

almcewen0:
[J0613-0200_NANOGrav_11yv0.gls.par.txt](https://github.com/nanograv/PINT/files/5349889/J0613-0200_NANOGrav_11yv0.gls.par.txt)
okay, I think that I have fixed the changes you mentioned. I also added some extra debugging lines in the compare function to help diagnose, and it seems there is still some issue with the log output. will attach specific screenshots in a second...

almcewen0:
Screen Shot 2020-10-08 at 12 52 12 PM
here is the new loop I am running. I included a little more printing to help separate the output, and I made the changes you mentioned. also, within the actual compare function I am testing, I added some print statements to frame the log.warning call:
Screen Shot 2020-10-08 at 12 50 37 PM
Note that newstr is the string containing the comparison line for par. however, in the output, while the exclamation point shows up for the change in PX properly, the warning for it does not arise until after the loop completes. The framing print statements suggest that there is no error in the warning call. 
Screen Shot 2020-10-08 at 12 52 36 PM
It's cut off here, but the while the warnings show up closer to the place I would expect them to, the reference changes to both ELONG and ELAT (the next parameter in the loop).

paulray:
Curious. 

When you `print()`, the output goes to the stdout stream, while when you use `log.warning()` the output goes to the stderr stream. What we are seeing I think it that they have different buffering, so the output of each will always come out in order, but may be interleaved in funny ways.

Try adding `sys.stdout.flush()` before each of the `log.warning()` calls. Does that fix it?

If not, then add `log.handlers[0].flush()` after each `log.warning()` call.

I think that will fix it, but the basic message is that it is not safe to send output to two different streams and expect them to come out interleaved as you might expect.  You could change your `print()` statements to go to stderr and that should also fix it.

paulray:
Hi, so the master is passing CI tests just fine.
This PR now just fails the Python 3.6 tests.
I suggest you delete the "<4.1" from astropy requirements and push that.
We can merge and then see if Py3.6 still fails and figure out what to do about that then.

paulray:
The CI build is not quite done, but it looks like it is passing everything except Python 3.6. I really don't understand why. It doesn't seem like it could be due to any changes in this PR. So, I suggest we merge and if master starts failing travis, we troubleshoot that then.

Issue #773, opened 2020-08-21
Error in pintk while plotting orbital phase


abhisrkckl:
Hello,

When I use orbital phase as the X axis for the plot in pintk, I get the following exception:

Exception in Tkinter callback
Traceback (most recent call last):
  File "/usr/lib/python3.6/tkinter/__init__.py", line 1705, in __call__
    return self.func(*args)
  File "/home/abhimanyu/.local/lib/python3.6/site-packages/pint/pintk/plk.py", line 376, in updateChoice
    self.updatePlot()
  File "/home/abhimanyu/.local/lib/python3.6/site-packages/pint/pintk/plk.py", line 759, in updatePlot
    self.plotResiduals(keepAxes=keepAxes)
  File "/home/abhimanyu/.local/lib/python3.6/site-packages/pint/pintk/plk.py", line 801, in plotResiduals
    xmin, xmax = xmin.value, xmax.value
AttributeError: 'numpy.float64' object has no attribute 'value'

This problem does not occur for any other plot in pintk.

Issue #686, opened 2020-05-02
Should use real orbit phase calculation


paulray:
https://github.com/nanograv/PINT/blob/f9ee990aa09f94d566816929d6b071e5e7a5c5c2/src/pint/pintk/pulsar.py#L161

This code should use PINT's built-in ability to compute orbit phases, if possible.
Of course it really should wait for #598 to be fixed first

Issue #598, opened 2020-02-05
provide a nice interface for orbit phase


luojing1211:
Currently, to get the orbit phase out, one has to do the following 
```
t = get_TOAs('.tim')
m = get_model('.par')
delay = m.delay(t)
orbit_phase = m.components['binary_model_name'].binary_instance.orbit()
```
Howevery this should be integerated to a convenient function.

paulray:
I think it is `orbits()` instead of `orbit()`

paulray:
Also see #475 
Is that corrected?

paulray:
And to copy some comments from Slack:
* Add docstrings that explain what "phase" means (eccentric or mean anomaly)
* Add unit tests for both parameterizations (PB, PBDOT, and FBx)
* How is it getting the times? Are they cached in the model object? That seems very dangerous!  One could have one model object and multiple sets of TOAs!

luojing1211:
@paulray, the last point deserves a separate issue, it is cached in the "stand along binaries".

paulray:
Yep, I'll make an issue for that. On my flight yesterday I looked through the binary code, which I hadn't done in detail before. It is great that the orbital phase calculations are handled by their own object that can be extended! But, that caching of TOAs is scary and definitely can cause some bugs/unexpected  behavior.

paulray:
See also #111, which probably can be closed.

dlakaplan:
Would be great to also provide access to the various anomalies (mean, eccentric, true)

dlakaplan:
Absolutely - I would use this for optical analysis.

bshapiroalbert:
Has anything been done related to this yet? I think this would be really useful if possible, especially with the new epoch_averaging function. Right now if we want to plot the epoch averaged residuals from the `ecorr_average()` function against orbital phase right now we have to remake a set of TOA objects with the MJDs and frequencies that come out, which is do-able, but tedious.

paulray:
Not that I'm aware of. This issue definitely needs someone's attention.

aarchiba:
The code `kepler.py`, currently unused (#693) is capable of doing a number of conversions between anomalies.

Issue #475, opened 2019-05-06
orbital phase calculations should use barycentric times


scottransom:
I just noticed that for the GUI, the topocentric (usually0 times are used to compute binary phase.  But the ephemerides are in barycentric times, so that's what should be used.

This also applies to issue #111 

https://github.com/nanograv/PINT/blob/cda85493ab3f78ac67e42527bd617fb3323570fe/pint/pintk/pulsar.py#L132

paulray:
Indeed, that sure is wrong!

aarchiba:
Shouldn't it use times as measured at the binary barycenter? Is that even a notion in pint? 

That is, DM and transverse-velocity corrections should be made before using them to compute orbital phases.

PR #786, opened 2020-10-07
Orbital phase helper method

scottransom:
This fixes all of #475, #598, #686, and #773 (I think).

You use it like:

```
In [1]: import pint.models as model                                                                                                                      

In [2]: m = model.get_model("/home/sransom/timing/0737/0737A_latest.par")                                                                                
INFO: Parameter PBDOT's value will be scaled by 1e-12 [pint.models.parameter]
INFO: Parameter PBDOT's value will be scaled by 1e-12 [pint.models.parameter]
WARNING: Unrecognized parfile line 'NPRNT                    3' [pint.models.timing_model]

In [3]: m.is_binary  # this is also a new method                                                                                                                                     
Out[3]: True

In [4]: m.orbital_phase([56789.234, 56790.765], anom='mean')                                                                                             
Out[4]: <Quantity [1.34467392, 1.17425508] rad>

In [5]: m.orbital_phase([56789.234, 56790.765], anom='eccen')                                                                                            
Out[5]: <Quantity [1.43160206, 1.25776666] rad>

In [6]: m.orbital_phase([56789.234, 56790.765], anom='true')                                                                                             
Out[6]: <Quantity [1.51917692, 1.34253751] rad>

In [7]: m.orbital_phase(56789.234)                                                                                                                       
Out[7]: <Quantity [1.34467392] rad>
```

If you call it with a TOAs object for the times, it will barycenter them first (as it should).

Comments and suggestions welcome.


paulray:
Wow, awesome! Can it also be called with a Time() object? If so, must it have scale='tdb'?

Also, if I want orbit phase from 0-1, what is the best way?  Is it to use op.to(u.cycle)? Or op.value/(2*np.pi)?

scottransom:
The answer to your first question is tricky.  It should work with Time objects (although right now there is some units bug that is causing that not to work properly), but "tdb" isn't enough.  We really need the times to be *barycentric*, not just tdb.  And there is no good way to force that.  So I'm tempted to just use the Time objects as .value no matter how they are passed and *assume* that that is what the user is giving us.

As for getting 0-1 phases, I have been dividing by (2*np.pi*u.rad).  But I think you can do it in multiple ways.  Maybe that could be a keyword argument to get unitless "full-cycle" phases if requested?

paulray:
> It should work with Time objects (although right now there is some units bug that is causing that not to work properly), but "tdb" isn't enough.

Very good point!  But if a Time object is passed, at a minimum it should be in the 'tdb' scale, otherwise it should throw an exception. And of course the docstring should be very clear that all times passed (unless they are full TOAs) must be barycenter-corrected and in the TDB timescale.

src/pint/models/timing_model.py
```diff
@@ -304,6 +304,40 @@ def phase_funcs(self):
             pfs += p.phase_funcs_component
         return pfs
 
+    @property
+    def is_binary(self):
+        """Does the model describe a binary pulsar? (True or False)"""
+        return any(x.startswith("Binary") for x in self.components.keys())
+
+    def orbital_phase(self, barytimes, anom="mean"):
+        """Return orbital phase (in radians) at barycentric MJD times"""
```
paulray, line 313: This docstring needs to be expanded. Should define the possible choices for `anom` and should use the standard docstring style defining Parameters and Returns.

scottransom:
OK.  I've made all the changes that we discussed today:
 - Good docstrings
 - Much improved time handling (including tdb checks and the astropy.Time bug fix)
 - An optional unitless cyclic return (phases in range 0-1)
 - And a .conjunction method which will find the nearest conjunction after an input MJD

I would like to have the conjunctions as an option to plot in pintk, but I'm not sure we want to wait on that for this PR

Any other comments?

PR #788, opened 2020-10-07
Observatory fallback

### LINKED ISSUE #787: Fallback to astropy observatory if PINT doesn't know about it
Astropy site list: https://github.com/astropy/astropy-data/blob/gh-pages/coordinates/sites.json
PINTsite list: https://github.com/nanograv/PINT/blob/master/src/pint/observatory/observatories.py

The astropy list is a lot longer, although less precise.  There are situations where it would be helpful to default to the astropy list if PINT is not aware.

dlakaplan:
This fixes #787.  There is a test routine in: `test_astropyobservatory.py`. 

Usage:
```python
import pint.observatory
gbt = pint.observatory.get_observatory("gbt")
keck =  pint.observatory.get_observatory("keck")
```
where the later uses astropy.  There are a few issues that I mention in the discussion for #787: 
* there is an `astropy` bug that means I use a slightly messy try/except (it is fixed is astropy development versions).
* it might be good to use metadata to track PINT observatory definitions.  Note that this isn't a change needed to accept the PR, but might be worth discussing more broadly.


paulray:
I totally agree that adding metadata about the observatory definitions would be very good. For example, where the coordinates of the observatory come from, and the date they were last changed.

Also, I think in all your code/comment IRTF should be ITRF.

dlakaplan:
> On Oct 8, 2020, at 10:45 AM, Paul Ray <notifications@github.com> wrote:
> 
> 
> I totally agree that adding metadata about the observatory definitions would be very good. For example, where the coordinates of the observatory come from, and the date they were last changed.
> 

OK, I will make a separate issue to deal with that.

> Also, I think in all your code/comment IRTF should be ITRF.

You mean it’s not Infrared Telescope Facility?

(Will fix - thanks!)

dlakaplan:
I fixed the API a bit.  

```python
import pint.observatory                                                                                                            
# a PINT observatory
gbt = pint.observatory.Observatory.get('gbt')                                                                                      
# an astropy observatory
keck = pint.observatory.Observatory.get('keck')                                                                                    
WARNING: Observatory name 'keck' is not present in PINT observatory list; searching astropy... [pint.observatory]
# check the registry
print(pint.observatory.Observatory.names())                                                                                               
dict_keys(['gbt', 'arecibo', 'vla', 'parkes', 'jodrell', 'nancay', 'ncyobs', 'effelsberg', 'gmrt', 'wsrt', 'fast', 'mwa', 'lwa1', 'ps1', 'hobart', 'most', 'chime', 'magic', 'lst', 'virgo', 'lho', 'llo', 'geo600', 'kagra', 'algonquin', 'drao', 'acre', 'keck'])
# and 'keck' is in there.

dlakaplan:
@paulray @luojing1211 or others: this passed the CI tests, and I think the interface is as needed (see usage above). If you agree please merge.

Issue #792, opened 2020-10-14
Check for observatory name conflicts


dlakaplan:
When instantiating a new observatory, check for existing observatories with the same name.  Either give a warning or error if a conflict exists.

dlakaplan:
I thought I knew how to do this.   I put some checks in `observatory/__init__.py` in the `__new__` function:
```python
         if name in cls.names():
             log.warning("Observatory '%s' already exists; replacing..." % name)
```
but I found when testing it that when I dropped in a new observatory it was triggering the warning but didn't get registered.  I think it's because I was defining in the wrong place (just in standalone code).  What would be a use-case to define an observatory outside of the `topo_obs.py` file?

dlakaplan:
Closed by https://github.com/nanograv/PINT/pull/796

Issue #790, opened 2020-10-08
Add observatory metadata


dlakaplan:
Add info for observatory entries to document where/when the coordinates were updated.

dlakaplan:
I think this is solved in:
https://github.com/dlakaplan/PINT/tree/observatory_metadata

```python
import pint.observatory
from astropy.coordinates import EarthLocation
from astropy import units as u, constants as c

keck = pint.observatory.Observatory.get('keck')
print('keck origin: %s' % keck.origin)
gbt = pint.observatory.Observatory.get('gbt')
print('gbt origin: %s' % gbt.origin)
arecibo = pint.observatory.Observatory.get('arecibo')
print(arecibo.origin)
```
returns:
```
keck origin: astropy: 'IRAF Observatory Database'
gbt origin: This is a test
None
```

dlakaplan:
Closed by https://github.com/nanograv/PINT/pull/796

PR #796, opened 2020-10-21
Observatory metadata and replacement

dlakaplan:
This addresses #790 and #792.

Examples:
```python
import pint.observatory
from astropy.coordinates import EarthLocation

keck = pint.observatory.Observatory.get('keck')
print('keck origin: %s' % keck.origin)
gbt = pint.observatory.Observatory.get('gbt')
print('gbt origin: %s' % gbt.origin)
arecibo = pint.observatory.Observatory.get('arecibo')
print(arecibo.origin)
```
returns:
```
keck origin: astropy: 'IRAF Observatory Database'
gbt origin: This is a test
None
```

In terms of observatory replacement:
```python
pint.observatory.topo_obs.TopoObs(
    "gbt",
    tempo_code="1",
    itoa_code="GB",
    itrf_xyz=[882589.65, 4924872.32, 3943729.348],
    origin="This is a test2",
)

gbt = pint.observatory.Observatory.get('gbt')
print('gbt origin: %s' % gbt.origin)
```
returns:
```
WARNING: Observatory 'gbt' already present; overwriting... [pint.observatory]
gbt origin: This is a test2
```

A set of test routines exists in `test_observatory_metadata.py`


luojing1211:
Do we want to overwrite the current observatory object? Maybe we should add
a flag "overwrite=False" in the most time. But if one really wants to
overwrite, one can set it to be True.

-Jing

scottransom:
I think I agree with @luojing1211 that having overwriting disallowed by default is probably the best bet.  If a user really wants to do this, then they can easily "force" it with that flag.

dlakaplan:
Good suggestion.  I think this is done, although I am not entirely happy with exactly how.

With:
```python
pint.observatory.topo_obs.TopoObs(
    "gbt",
    tempo_code="1",
    itoa_code="GB",
    itrf_xyz=[882589.65, 4924872.32, 3943729.348],
    origin="This is a test2",
)

gbt = pint.observatory.Observatory.get('gbt')
print('gbt origin: %s' % gbt.origin)
```
it returns:
```
WARNING: Observatory 'gbt' already present and overwrite=False; discarding... [pint.observatory]
gbt origin: This is a test
``` 
which is the original version, so unchanged as expected.   Then:

```python
pint.observatory.topo_obs.TopoObs(
    "gbt",
    tempo_code="1",
    itoa_code="GB",
    itrf_xyz=[882589.65, 4924872.32, 3943729.348],
    origin="This is a test2",
    overwrite=True,
)

gbt = pint.observatory.Observatory.get('gbt')
print('gbt origin: %s' % gbt.origin)
```
returns:
```
WARNING: Observatory 'gbt' already present; overwriting... [pint.observatory]
gbt origin: This is a test2
```
so it has been over-written.

dlakaplan:
Is `None` the correct return value if the new observatory gets discarded?

```python
            if "overwrite" in kwargs and kwargs["overwrite"]:
                log.warning(
                    "Observatory '%s' already present; overwriting..." % name.lower()
                )

                cls._register(obs, name)
                return obs
            else:
                log.warning(
                    "Observatory '%s' already present and overwrite=False; discarding..."
                    % name.lower()
                )
                return None
```

scottransom:
@dlakaplan I would think that if someone tries to overwrite without that flag set that an exception should be raised, not just a warning issued.  It seems too fraught with possible issues to not do that (IMO).

dlakaplan:
I could see it either way.  An exception that stops execution is a big change from previous behavior (silently overwrite), so I'd want to make sure this doesn't break other things.

scottransom:
Someone who really wants to overwrite will always be able to do that by simply changing the registered information (and possibly by other ways).  But your PR is making this much more likely (or easier) to happen, right?  If so, that seems different to me, and I'm fine making the change and possibly even breaking some code with it (should be extremely easy to fix).

If getting name clashes becomes common/easy, then silently overwriting seems like a recipe for incorrect results to me.

dlakaplan:
No, I don't think the PR makes changing an observatory easier.  Previously observatories were silently overwritten.  So at least this notifies the user now that this is happening.

If an Exception makes sense I'm fine with it too, just wanted to make sure that it wouldn't break anything.

dlakaplan:
CI build passed.  @luojing1211 or @paulray or others: can you merge?

PR #800, opened 2020-11-04
Add other BIPM files from TEMPO2 and a bipm_default variable in the code

scottransom:
The BIPM files are all in the format tai2tt_bipm????.clk from TEMPO2.


scottransom:
I tried to add a test that catches an exception when you give "BIPMXXXX", for instance.  But while that fails, it fails in a cascade of 3 exceptions and pytest.raises doesn't properly notice any of them!  Very confusing....

scottransom (commented):
Paul:  did you mean to merge this into my branch?  Or did that happen automatically somehow?  Seems like it could cause a conflict in the future.

BTW, I think that this PR can be merged if you want to do that.

paulray:
Yes, a conflict appeared because David edited the CHANGELOG in master and you added your line. I just merged them to resolve the conflict.

PR #807, opened 2020-11-11

bshapiroalbert:
I have updated the F-tests in the fitter so that they are compatible with the `WidebandFitter` and subsequent `CombinedResiduals' classes. The test automatically figures out if you're using a WidebandFitter or GLS/WLSFitter and get the appropriate values of chi-squared for the actual F-test. If using the full_output option with a WidebandFitter, it will also return the time and DM chi-squared values separately (as well as the total value) and the weighted and unweighted rms of the DM residuals, and the separately computed degrees of freedom. If using a GLS/WLSFitter, output remains unchanged from previous version. Documentation and tests have been updated as well. There were so many updates since the last time I worked on this that it was easier to close the first PR of this and just open a new one with a new branch, so this is the second version of this.

The only question I have is making sure that the degrees of freedom is calculated correctly for the wideband F-tests. As per Tim Pennucci this should be 2*ntoa - n_fit_params where n_fit_params should include a +1 for absolute phase. I think this is the same as `self.resids.residual_objs["toa"].get_dof() + self.resids.residual_objs["dm"].get_dof() + 1`, where the "+1" is because this absolute phase is being added in twice, instead of once, but that's I think the only thing left to check.


bshapiroalbert:
I'm will close this PR for now and open a new one as there are conflicts that need to be resolved between these branches. I think this will be the easiest/fastest thing to do. Look forward to a new version of this PR soon!

PR #808, opened 2020-11-12
Fix dof bug

luojing1211:
This PR fixes the residual degree of freedom(DOF) bugs and modified it API from a function `get_dof()` to property `dof`.

### DIFF
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -129,9 +129,11 @@ class Residuals:
     def dof(self):
         """Return number of degrees of freedom for the model."""
         if self._combined:
-            raise RuntimeError("Please use the `.dof` in the CombinedResidual"
-                               " class. The individual residual's dof is not "
-                               "calculated correctly in the combined residuals.")
+            raise RuntimeError(
+                "Please use the `.dof` in the CombinedResidual"
+                " class. The individual residual's dof is not "
+                "calculated correctly in the combined residuals."
+            )
         dof = self.toas.ntoas
         for cp in self.model.components.values():
             if "delay" in cp.modeled_quantity or "phase" in cp.modeled_quantity:
@@ -449,9 +451,11 @@ class WidebandDMResiduals(Residuals):
     def dof(self):
         """Return number of degrees of freedom for the DM model."""
         if self._combined:
-            raise RuntimeError("Please use the `.dof` in the CombinedResidual"
-                               " class. The individual residual's dof is not "
-                               "calculated correctly in the combined residuals.")
+            raise RuntimeError(
+                "Please use the `.dof` in the CombinedResidual"
+                " class. The individual residual's dof is not "
+                "calculated correctly in the combined residuals."
+            )
         dof = len(self.dm_data)
         # only get dm type of model component
         # TODO provide a function in the timing model to get one type of component
@@ -578,7 +582,6 @@ class CombinedResiduals(object):
             res._combined = True
             self.residual_objs[res.residual_type] = res
         # Disable the individual residual's dof
-
 
     @property
     def _combined_resids(self):
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -129,7 +129,7 @@ class Residuals:
     def dof(self):
         """Return number of degrees of freedom for the model."""
         if self._is_combined:
-            raise RuntimeError(
+            raise AttributeError(
                 "Please use the `.dof` in the CombinedResidual"
                 " class. The individual residual's dof is not "
                 "calculated correctly in the combined residuals."
@@ -317,9 +317,10 @@ class Residuals:
                 except:
                     return ((self.time_resids / toa_errors.to(u.s)) ** 2.0).sum()
 
-    def get_reduced_chi2(self):
+    @property
+    def reduced_chi2(self):
         """Return the weighted reduced chi-squared for the model and toas."""
-        return self.calc_chi2() / self.dof
+        return self.chi2 / self.dof
 
     def update(self):
         """Recalculate everything in residuals class after changing model or TOAs"""
@@ -451,7 +452,7 @@ class WidebandDMResiduals(Residuals):
     def dof(self):
         """Return number of degrees of freedom for the DM model."""
         if self._is_combined:
-            raise RuntimeError(
+            raise AttributeError(
                 "Please use the `.dof` in the CombinedResidual"
                 " class. The individual residual's dof is not "
                 "calculated correctly in the combined residuals."
@@ -464,6 +465,11 @@ class WidebandDMResiduals(Residuals):
                 dof -= len(cp.free_params_component)
         dof -= 1
         return dof
+
+    @property
+    def reduced_chi2(self):
+        """Return the weighted reduced chi-squared for the model and DMs."""
+        return self.chi2 / self.dof
 
     def get_data_error(self, scaled=True):
         """Get data errors.
@@ -619,6 +625,11 @@ class CombinedResiduals(object):
         # TODO In a more general case, this assumption would not be valid.
         dof -= len(self.residual_objs["toa"].model.free_params) + 1
         return dof
+
+    @property
+    def reduced_chi2(self):
+        """Return the weighted reduced chi-squared."""
+        return self.chi2 / self.dof
 
     @property
     def data_error(self):
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -438,14 +438,6 @@ class WidebandDMResiduals(Residuals):
         return self.resids.to_value(self.unit)
 
     @property
-    def chi2(self):
-        """Compute chi-squared as needed and cache the result"""
-        if self._chi2 is None:
-            self._chi2 = self.calc_chi2()
-        assert self._chi2 is not None
-        return self._chi2
-
-    @property
     def dof(self):
         """Return number of degrees of freedom for the DM model."""
         if self._is_combined:
@@ -462,11 +454,6 @@ class WidebandDMResiduals(Residuals):
                 dof -= len(cp.free_params_component)
         dof -= 1
         return dof
-
-    @property
-    def reduced_chi2(self):
-        """Return the weighted reduced chi-squared for the model and DMs."""
-        return self.chi2 / self.dof
 
     def get_data_error(self, scaled=True):
         """Get data errors.
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -11,7 +11,8 @@ from pint.phase import Phase
 from pint.utils import weighted_mean
 from pint.models.dispersion_model import Dispersion
 
-__all__ = ["Residuals", "WidebandDMResiduals", "residual_map"]
+__all__ = ["Residuals", "WidebandDMResiduals", "residual_map",
+           "CombinedResiduals", "WidebandTOAResiduals"]
 
 
 class Residuals:
@@ -19,6 +20,14 @@ class Residuals:
 
     Parameters
     ----------
+    toas: `pint.toa.TOAs` object, optional
+        The input TOAs object. Default: None
+    model: `pint.models.TimingModel` object, optinonal
+        Input model object. Default: None
+    residual_type: str, optional
+        The type of the resiudals. Default: 'toa'
+    unit: `astropy.units.Unit` object, optional
+        The defualt unit of the residuals. Default: u.s
     subtract_mean : bool
         Controls whether mean will be subtracted from the residuals
     use_weighted_mean : bool
@@ -29,6 +38,8 @@ class Residuals:
         ``pulse_number`` column of the TOAs table to assign pulse numbers. If the
         default, None, is passed, use the pulse numbers if and only if the model has
         parameter TRACK == "-2".
+    scaled_by_FO: bool, optional
+        Controls for rescaling the final residual.
     """
 
     def __new__(
@@ -571,7 +582,15 @@ class CombinedResiduals(object):
         for res in residuals:
             res._is_combined = True
             self.residual_objs[res.residual_type] = res
-        # Disable the individual residual's dof
+
+    @property
+    def model(self):
+        """ The input model from each residual.
+        """
+        model = {}
+        for res_name, res in self.residual_objs.items():
+            model[res_name] = res.model
+        return model
 
     @property
     def _combined_resids(self):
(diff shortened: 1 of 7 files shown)


pennucci:
@luojing1211  @demorest @scottransom I'm not sure it makes sense to (1) split the dof into portions related to the TOA residuals and DM residual (2) have dof be part of residuals at all, unless inherited by the fitter class.  The dof depends not just on the model, but what is actually a free parameter, i.e., the fitter.  So dof should maybe just be associated with the fitter and child objects.

pennucci:
@luojing1211 I pulled your fix_dof_bug branch (the one I think you're trying to merge) and fitter.resids.residual_objs['toa'].dof is different from fitter.resids.residual_objs['dm'].dof; the former gives ntoa - n_fit_params and the latter gives ndm - n_dm_params - 1.  Both of those are some what confusing and probably not useful.  The top-level dof via fitter.resids.dof appears correct, though.

luojing1211:
The fitter.resids.residual_objs['toa'].dof  is from the toa residual class, which is used for most of the toa fittings. Thus the dof here is necessary. If one use the combined residual, those two attributes are not useful, but if one uses them individually, they are important. I believe the calculation is ntoa - n_fit_params(contribute to delay and phase) - 1.

aarchiba:
> The fitter.resids.residual_objs['toa'].dof is from the toa residual class, which is used for most of the toa fittings. Thus the dof here is necessary. If one uses the combined residual, those two attributes are not useful, but if one uses them individually, they are important. I believe the calculation is ntoa - n_fit_params(contribute to delay and phase) - 1.

The specific parameter described here - the `.dof` parameter from a residual that's not the result of a fit - doesn't make sense. Anything that depends on its value is going to be wrong.

luojing1211:
@aarchiba Should this method lives in the fitter, and when the fitter generates a residual, it will be passed to the residual?

aarchiba:
> @aarchiba Should this method lives in the fitter, and when the fitter generates a residual, it will be passed to the residual?

I think it's reasonable that a Residual object have a `.dof` attribute - if it's actually a standalone Residual object. "Residuals" that are only part of the result of a wideband fit are strange and don't always behave the same way - in particular they don't have a natural chi-squared or degrees-of-freedom value. The WidebandResidual object does, that's fine.

luojing1211:
The Wideband residual is designed to behave like a standalone residual class. I think this ties to a bigger question that how do we make sure the correct chi2 value and dof value passed to the residual from the fitter. But for this PR. Do you think this is something we can merge if the conflict gets fixed?

aarchiba:
> The Wideband residual is designed to behave like a standalone residual class. I think this ties to a bigger question that how do we make sure the correct chi2 value and dof value passed to the residual from the fitter. But for this PR. Do you think this is something we can merge if the conflict gets fixed?

There is a definitional problem: *is* there a meaninful number of degrees of freedom, or chi-squared, for the DM residuals alone from a wideband fit? I don't think there is. For the whole wideband residuals object, sure.

aarchiba:
I think the `.dof` attribute and the `.chi2` method (?) should be removed from the DM and TOA elements of a WidebandResidual. They can be easily implemented, and have a clear definition, on the WidebandResidual itself.

pennucci:
> > The Wideband residual is designed to behave like a standalone residual class. I think this ties to a bigger question that how do we make sure the correct chi2 value and dof value passed to the residual from the fitter. But for this PR. Do you think this is something we can merge if the conflict gets fixed?
> 
> There is a definitional problem: _is_ there a meaninful number of degrees of freedom, or chi-squared, for the DM residuals alone from a wideband fit? I don't think there is. For the whole wideband residuals object, sure.

I think this is related to what @aarchiba wrote in #807 https://github.com/nanograv/PINT/pull/807#discussion_r529035027, but there is a meaningful chi-squared value for the separate DM and TOA residuals -- but there is not a meaningful segregation of dof and, thus, the reduced chi-squared.  The likelihood is separable into "TOA data" and "DM data" pieces, and you can show that the total chi-squared is equal to the sum of the pieces.  Each piece is the usual sum of ((model-data)/uncertainty)**2; for the TOA piece, that's the timing model, TOAs, and their (scaled) uncertainties, for the DM piece, that's DMX (for now), the DM measurements, and their (scaled) uncertainties.  So having a sensible way to access both the total and separate pieces of the chi-squared is reasonable, I think.  Where they go in the API is a separate question.  The dof is not separable at least because the DMX parameters are informed by both the TOAs and DM measurements.  The issue of estimating dof in GLS problems is messy in the first place.

pennucci:
> I think the `.dof` attribute and the `.chi2` method (?) should be removed from the DM and TOA elements of a WidebandResidual. They can be easily implemented, and have a clear definition, on the WidebandResidual itself.

I agree with this.

luojing1211:
> I think the `.dof` attribute and the `.chi2` method (?) should be removed from the DM and TOA elements of a WidebandResidual. They can be easily implemented, and have a clear definition, on the WidebandResidual itself.

Since the wideband residual uses the TOA and DM residual class internally, removing `.dof` from the wideband residual means removing it from the TOA residual class, and the narrow band TOA will not have `.dof` attribute either. Is that the preferred behavior? Or should we not provide this attribute for all the residual class? What do you think @scottransom @paulray @demorest

aarchiba:
> > I think the `.dof` attribute and the `.chi2` method (?) should be removed from the DM and TOA elements of a WidebandResidual. They can be easily implemented, and have a clear definition, on the WidebandResidual itself.
> 
> Since the wideband residual uses the TOA and DM residual class internally, removing `.dof` from the wideband residual means removing it from the TOA residual class, and the narrow band TOA will not have `.dof` attribute either. Is that the preferred behavior? Or should we not provide this attribute for all the residual class? What do you think @scottransom @paulray @demorest

If you are going to use narrowband residual classes internally in the wideband residual, you are going to have to do some gymnastics. I would recommend disabling their `.dof` attribute when used inside a wideband residual (perhaps just delete the attribute, or have a subclass you use for this that replaces `.dof` with something that raises AttributeError); the wideband residual can have  its own meaningful `.dof` attribute.

pennucci:
> > > I think the `.dof` attribute and the `.chi2` method (?) should be removed from the DM and TOA elements of a WidebandResidual. They can be easily implemented, and have a clear definition, on the WidebandResidual itself.
> > 
> > 
> > Since the wideband residual uses the TOA and DM residual class internally, removing `.dof` from the wideband residual means removing it from the TOA residual class, and the narrow band TOA will not have `.dof` attribute either. Is that the preferred behavior? Or should we not provide this attribute for all the residual class? What do you think @scottransom @paulray @demorest
> 
> If you are going to use narrowband residual classes internally in the wideband residual, you are going to have to do some gymnastics. I would recommend disabling their `.dof` attribute when used inside a wideband residual (perhaps just delete the attribute, or have a subclass you use for this that replaces `.dof` with something that raises AttributeError); the wideband residual can have its own meaningful `.dof` attribute.

I also agree with this; it will probably be a pain to deal with the standard/narrowband residual class attribute.

luojing1211:
I can disable them in the wideband residual. Is there any other methods we need to disable?

aarchiba:
Looking at the documentation for Residual, `chi2_reduced` doesn't make sense, nor do `get_dof` or `get_reduced_chi2`.

luojing1211:
> Looking at the documentation for Residual, `chi2_reduced` doesn't make sense, nor do `get_dof` or `get_reduced_chi2`.

Could you be more specific? If there is anything wrong I would rather fix it before merge this PR.

aarchiba:
> > Looking at the documentation for Residual, `chi2_reduced` doesn't make sense, nor do `get_dof` or `get_reduced_chi2`.
> 
> Could you be more specific? If there is anything wrong I would rather fix it before merge this PR.

When a Residuals object is the result of combining a narrowband TOAs object and a TimingModel object, there is a well-defined number of degrees of freedom, and the methods all make sense. When a CombinedResiduals object is the result of combining wideband TOAs with a TimingModel, it also has a meaningful number of degrees of freedom, and those methods also make sense. But that WidebandResiduals object contains two Residuals objects that are not the result of narrowband fits, and those internal objects do not have a meaningful number of degrees of freedom, so the `get_dof()` method on them does not make sense, and neither do `chi2_reduced` or `get_reduced_chi2()`. For a WidebandDMResiduals object, it is never the result of a fit, so it does not have a meaningful number of degrees of freedom, and thus `get_dof()` does not make sense for it.

src/pint/residuals.py
```diff
@@ -123,6 +122,20 @@ def resids_value(self):
...
+    def dof(self):
+        """Return number of degrees of freedom for the model."""
+        dof = self.toas.ntoas
+        for cp in self.model.components.values():
+            if "delay" in cp.modeled_quantity or "phase" in cp.modeled_quantity:
+                dof -= len(cp.free_params_component)
+        # Now subtract 1 for the implicit global offset parameter
+        # Note that we should do two things eventually
+        # 1. Make the offset not be a hidden parameter
+        # 2. Have a model object return the number of free parameters instead of having to count non-frozen parameters like above
```
aarchiba, line 135: would `len(self.model.free_params)` work?

src/pint/residuals.py
```diff
@@ -436,6 +437,18 @@ def chi2(self):
         assert self._chi2 is not None
         return self._chi2
 
+    @property
+    def dof(self):
+        """Return number of degrees of freedom for the DM model."""
```
aarchiba, line 442: We can't split free parameters up into DM and non-DM free parameters - many (e.g. DMX) depend on both.

pennucci:
If I follow this correctly, then I agree.  The first place I had looked 
for dof and the reduced chi2 in PINT's wideband developments were *not* 
in the individual sub-residual objects, and they really shouldn't go 
there.  But I think CombinedResiduals should have such a method/attribute.

luojing1211:
Just disable the `.dof` in the CombinedResidual class.

src/pint/residuals.py
```diff
@@ -591,6 +611,15 @@ def chi2(self):
             chi2 += res.chi2
         return chi2
 
+    @property
+    def dof(self):
+        dof = len(self._combined_resids)
+        # It assumes that the input model are the same model, and time residual has
+        # offset in the fitting
+        # TODO In a more general case, this assumption would not be valid.
+        dof -= len(self.residual_objs["toa"].model.free_params) + 1
```
aarchiba, line 620: Shouldn't `self` have its own `.model` attribute?

tests/test_residuals.py
```diff
@@ -78,3 +79,7 @@ def test_combined_residuals(self):
         assert cb_residuals.unit["toa"] == u.s
         assert cb_residuals.unit["dm"] == u.pc / u.cm ** 3
         assert cb_chi2 == phase_res.chi2 + dm_res.chi2
+        assert cb_residuals.dof == 419
+        with pytest.raises(RuntimeError):
+            cb_residuals.residual_objs["toa"].dof
+            cb_residuals.residual_objs["dm"].dof
```
aarchiba, line 85: This line is never reached.

src/pint/residuals.py
```diff
@@ -123,6 +125,26 @@ def resids_value(self):
         else:
             return self.resids.to_value(self.unit)
 
+    @property
+    def dof(self):
+        """Return number of degrees of freedom for the model."""
+        if self._is_combined:
+            raise RuntimeError(
```
aarchiba, line 132: AttributeError would be cleaner here, though it might cause weirdness with `@property`.

src/pint/residuals.py
```diff
@@ -123,6 +125,26 @@ def resids_value(self):
...
+                "Please use the `.dof` in the CombinedResidual"
+                " class. The individual residual's dof is not "
+                "calculated correctly in the combined residuals."
+            )
+        dof = self.toas.ntoas
+        for cp in self.model.components.values():
+            if "delay" in cp.modeled_quantity or "phase" in cp.modeled_quantity:
```
aarchiba, line 139: What is `modeled_quantity` for? This ignores model components that affect DM, which do affect timing (see all our narrowband work). Do you have test cases that check the `.dof` value for phase-only residuals where the model includes DM fitting? What is the right way to compute a `.dof` for a model that includes `EQUAD`, `EFAC`, or `ECORR`?

luojing1211:
@aarchiba I think all of your comments have been addressed or answered.

src/pint/residuals.py
```diff
@@ -436,6 +445,29 @@ def chi2(self):
...
+        # only get dm type of model component
+        # TODO provide a function in the timing model to get one type of component
+        for cp in self.model.components.values():
+            if Dispersion in cp.__class__.__bases__:
+                dof -= len(cp.free_params_component)
+        dof -= 1
+        return dof
+
+    @property
+    def reduced_chi2(self):
```
aarchiba, line 467: Won't this method be inherited from the base Residuals class?

aarchiba:
> @aarchiba I think all of your comments have been addressed or answered.

Looks like there are two still outstanding and one new one?

src/pint/residuals.py
luojing1211: Is this explanation clear?

luojing1211:
Sorry, I forget to push the submit button.

src/pint/residuals.py
aarchiba: I would view that as one model with two parts. But more to the point, here and now, as a user, I would expect every Residuals object, Combined or not, to have a `.model` attribute.

luojing1211: I can make an property that takes the models from all the residuals. But it would not be one model object.  I think the assumption that all residual objects have the same model may cause problem in the future.  On the other hand, if we want to view the it as one model with two parts. I think a better way to do this is to make another residual type, that defined to have one model but used in the multiple residuals small classes.

aarchiba: I think the individual contained residuals shouldn't have `model` attributes either, but that's less important - if we want to do a joint fit, speculatively in the future, we can build a joint model, whose free parameters are assembled from the free parameters of the sub-models, with whatever linkages. But for now I think the invariant that "every Residual object you get from a model keeps track of that model in its `.model` attribute" is a good one to maintain.

src/pint/residuals.py
```diff
@@ -561,8 +580,18 @@ class CombinedResiduals(object):
     def __init__(self, residuals):
         self.residual_objs = collections.OrderedDict()
         for res in residuals:
+            res._is_combined = True
             self.residual_objs[res.residual_type] = res
 
+    @property
+    def model(self):
```
aarchiba, line 587: This doesn't work for users who expect `residual.model` to be a TimingModel they can query for, say, `free_params`. If you want to provide this differently typed object, please call it something else; if you don't want to use a single TimingModel object for combined residuals, just leave `.model` raising AttributeError.

src/pint/residuals.py
```diff
@@ -612,10 +652,56 @@ def rms_weighted(self):
...
+        return self._model
+
+    @property
+    def dof(self):
+        dof = len(self._combined_resids)
         # It assumes that the input model are the same model, and time residual has
         # offset in the fitting
         # TODO In a more general case, this assumption would not be valid.
```
aarchiba, line 705: This comment isn't relevant any more, is it?

aarchiba:
Looks good to me!

luojing1211:
Will merge this after the test is done.

PR #815, opened 2020-11-17
Test and document .tim file writing and reading.

### LINKED ISSUE #798: Want round-trip tests for .tim file writing
To check that we can write .tim files reliably we should be able to create a family of tests that load a .tim file (from a string in the test module is probably easiest) into a TOAs object t1, write that out to a temporary file (string if possible), read that back in to a new TOAs object t2, and compares t1 to t2. Cases worth considering:

- EFAC/EQUAD in the .tim file
- long double precision in the TOAs - round trip should be exactly the same floating-point number, or at most a few floating-point epsilon away
- .tim files originally loaded with the `ephem` parameter set to something weird
- .tim files containing barycentered TOAs (and a mix)
- .tim files (or TOA objects) with weird and/or excessive TOA flags
- .tim files with pulse numbers for all or some TOAs (-pn TOA flag)
- various clock correction weirdnesses (e.g. clock corrections applied/not applied to a TOAs object, an observatory with no clock corrections, TOAs before the beginning/past the end of the clock corrections)
- .tim files with INCLUDEs
- .tim files with SKIP, JUMP, commented TOAs, EMIN, EMAX, FMIN, FMAX, PHASE, START, END, TIME, PHA1, PHA2, TRACK, MODE, INFO
- .tim files in each format
- .tim files with shuffled lines

Probably good for some of these to check that the loaded TOAs object is what the original .tim intended, too, but a check-round-trip test is easy to write.

aarchiba:
Some other additional documentation has been added; also a bug to do with writing pulse numbers to .tim files has been fixed.

Closes #798

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,2 +1,8 @@
 
 .rst-content code.literal { color: unset; }
+
+/* override table no-wrap */
+.wy-table-responsive table td, .wy-table-responsive table th {
+    white-space: normal;
+}
+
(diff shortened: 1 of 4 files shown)


aarchiba:

PR #814, opened 2020-11-16
Allow fitters to respect pulse numbers

aarchiba:
Fitters now accept the `track_mode="use_pulse_numbers"` keyword argument so they can use locked pulse numbers.

Closes #812 #821  .

PR #822, opened 2020-11-20
Tidy fitparams -> free_params

### LINKED ISSUE #816: fitter.set_fitparams() cannot take the output of fitter.get_fitparams()
The output of get_fitparams is an Odict of key,val pairs, but set_fitparams appears to work only on a list/tuple of param names.  Passing the output of get_fitparams to set_params raises no warning, and fixes/freezes all parameters.  There's even a comment in set_fitparams: "# TODO, maybe reconsider for the input?"

This issue can easily raise problems for users that are making reasonable assumptions about the API.

aarchiba:
Do some renaming and API cleanup. In particular, use "free/frozen" consistently instead of the confusing "fit". Deprecate old names for functions.

Closes #816

### DIFF
--- a/docs/examples/PINT_walkthrough.md
+++ b/docs/examples/PINT_walkthrough.md
@@ -6,7 +6,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.2'
-      jupytext_version: 1.4.0
+      jupytext_version: 1.5.2
   kernelspec:
     display_name: Python 3
     language: python
--- a/docs/examples/Wideband_TOA_walkthrough.md
+++ b/docs/examples/Wideband_TOA_walkthrough.md
@@ -6,7 +6,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.2'
-      jupytext_version: 1.4.0
+      jupytext_version: 1.5.2
   kernelspec:
     display_name: Python 3
     language: python
@@ -136,7 +136,7 @@ d_matrix = fitter.get_designmatrix()
 ```python execution={"iopub.execute_input": "2020-09-10T16:29:32.318273Z", "iopub.status.busy": "2020-09-10T16:29:32.311723Z", "iopub.status.idle": "2020-09-10T16:29:32.327689Z", "shell.execute_reply": "2020-09-10T16:29:32.327089Z"}
 print("Number of TOAs:", toas.ntoas)
 print("Number of DM measurments:", len(fitter.resids.residual_objs['dm'].dm_data))
-print("Number of fit params:", len(fitter.get_fitparams()))
+print("Number of fit params:", len(fitter.get_free_params()))
 print("Shape of design matrix:", d_matrix.shape)
 ```
 
--- a/docs/examples/Wideband_TOA_walkthrough.md
+++ b/docs/examples/Wideband_TOA_walkthrough.md
@@ -65,7 +65,7 @@ fitter.resids.residual_objs
 * Time residual
 
 ```python execution={"iopub.execute_input": "2020-09-10T16:29:28.342821Z", "iopub.status.busy": "2020-09-10T16:29:28.330288Z", "iopub.status.idle": "2020-09-10T16:29:28.520180Z", "shell.execute_reply": "2020-09-10T16:29:28.519607Z"}
-time_resids = fitter.resids.residual_objs['toa'].time_resids
+time_resids = fitter.resids.residual_objs["toa"].time_resids
 plt.errorbar(
     toas.get_mjds().value,
     time_resids.to_value(u.us),
@@ -78,15 +78,15 @@ plt.xlabel("MJD")
 
 ```python execution={"iopub.execute_input": "2020-09-10T16:29:28.525251Z", "iopub.status.busy": "2020-09-10T16:29:28.524698Z", "iopub.status.idle": "2020-09-10T16:29:28.527648Z", "shell.execute_reply": "2020-09-10T16:29:28.527083Z"}
 # Time RMS
-print(fitter.resids.residual_objs['toa'].rms_weighted())
-print(fitter.resids.residual_objs['toa'].chi2)
+print(fitter.resids.residual_objs["toa"].rms_weighted())
+print(fitter.resids.residual_objs["toa"].chi2)
 ```
 
 * DM residual
 
 ```python execution={"iopub.execute_input": "2020-09-10T16:29:28.556722Z", "iopub.status.busy": "2020-09-10T16:29:28.535404Z", "iopub.status.idle": "2020-09-10T16:29:28.698341Z", "shell.execute_reply": "2020-09-10T16:29:28.697831Z"}
-dm_resids = fitter.resids.residual_objs['dm'].resids
-dm_error = fitter.resids.residual_objs['dm'].get_data_error()
+dm_resids = fitter.resids.residual_objs["dm"].resids
+dm_error = fitter.resids.residual_objs["dm"].get_data_error()
 plt.errorbar(toas.get_mjds().value, dm_resids.value, yerr=dm_error.value, fmt="x")
 plt.ylabel("pc/cm^3")
 plt.xlabel("MJD")
@@ -94,8 +94,8 @@ plt.xlabel("MJD")
 
 ```python execution={"iopub.execute_input": "2020-09-10T16:29:28.703699Z", "iopub.status.busy": "2020-09-10T16:29:28.703155Z", "iopub.status.idle": "2020-09-10T16:29:28.705717Z", "shell.execute_reply": "2020-09-10T16:29:28.706203Z"}
 # DM RMS
-print(fitter.resids.residual_objs['dm'].rms_weighted())
-print(fitter.resids.residual_objs['dm'].chi2)
+print(fitter.resids.residual_objs["dm"].rms_weighted())
+print(fitter.resids.residual_objs["dm"].chi2)
 ```
 
 #### However, in the combined residuals, one can access rms and chi2 as well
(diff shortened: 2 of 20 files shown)


aarchiba:
Depends on #814 because I wanted to use `make_fake_toas`.

swiggumj:
I haven't kept completely up to speed on this, but while I'm all for making things less awful, I worry about "deprecated" and "set_fitparams" being in the same sentence since this is what @luojing1211 suggested as the best way to determine frozen/fit parameters dynamically in a notebook setting. Am I missing something?

luojing1211:
This looks good. But it may break the degree of freedom calculation since it uses free_params.

aarchiba:
> I haven't kept completely up to speed on this, but while I'm all for making things less awful, I worry about "deprecated" and "set_fitparams" being in the same sentence since this is what @luojing1211 suggested as the best way to determine frozen/free parameters dynamically in a notebook setting. Am I missing something?

It is a bad way to determine them. "fitparams" is confusing because it does not use free/frozen; and it is on the wrong object: if you want to know which parameters are free in the model you are fitting, you can use `f.model.free_params`. If you want to set which parameters are free in that model, you can (now) do `f.model.free_params = ['F0']`.

The getter version of the `free_params` property is unchanged; and all tests pass, so the `dof` calculation is either correct or untested.

paulray:
These big API changes will break some downstream code/notebooks. So, we should definitely get them in before tagging v0.8 so things can be stable in the API for a while

aarchiba:
> These big API changes will break some downstream code/notebooks. So, we should definitely get them in before tagging v0.8 so things can be stable in the API for a while

We should also do a survey for other awful API warts so we can clean them up as early as possible - for example the `.select()` disaster on TOAs objects. Perhaps a systematic sweep through notebooks found in the wild for things people have to do in an awful way? Also going through and filing bugs, perhaps with the tag bad-API?

I am of course open to discussing these API changes - for example, I left `get_free_params` (renamed from `get_fitparams`) on Fitter objects, and maybe I shouldn't have, but it and its cousins are useful for obtaining dictionaries relevant to fitting.

scottransom:
@aarchiba Can you please avoid using terms like "awful" and "disaster"?  Much of the code in Pint has been written by students and/or postdocs without a ton of experience.  It is really disheartening to have your work described by someone else as those things.  Especially since it is *opinion*, and we are actually getting real science done now even with those "warts".

luojing1211:
The uncertain APIs are the reason why we are not claiming 1.0 yet. I think our strategy or the modern software development strategy is to put software out to the community, get feedback, and then improve it rapidly. Some APIs are guarantee to be awful, and I know this from day one. @aarchiba I am super glad you point them out. That is exactly what we want. Let us get them identified and fixed.

swiggumj:
>if you want to know which parameters are free in the model you are fitting, you can use f.model.free_params. If you want to set which parameters are free in that model, you can (now) do f.model.free_params = ['F0'].

Since `set_fitparams` is currently used in the NANOGrav notebooks, I'll need to keep careful track of this, but it looks like the updates should be extremely minor. Same with `.select()`.

aarchiba:
> @aarchiba Can you please avoid using terms like "awful" and "disaster"? Much of the code in Pint has been written by students and/or postdocs without a ton of experience. It is really disheartening to have your work described by someone else as those things. Especially since it is _opinion_, and we are actually getting real science done now even with those "warts".

You're right that I should not be so blunt. 

I am concerned that to achieve some of PINT's goals, in particular that it be a flexible library for pulsar timing tasks, it is necessary that the API be comprehensible and usable. This is a situation where code review - where our procedure of discussing PRs on the telecon, for example - could make a big difference, if we gave enough importance to API design. The current condition of the API means that I keep running into puzzling and poorly documented problems when I try to do things that seem basic. I can't be the only one. 

I am aware that I have said this before; the longer we wait to do API cleanup the more our current API gets baked into scripts and notebooks in NANOGrav and elsewhere.

scottransom:
I totally agree about the API issues.  And I'm really glad that they are being (slowly) addressed.  I was mostly just concerned about the bluntness.

We've been working on PINT for years, and for most of that time, there were only a couple people doing any coding or attending meetings.  Although there were also sporadic contributions from summer students and things like that.

It is really great that we have many people doing all of these things now -- including attending meetings and discussing issues and development.

Let's just keep all of this positive and also remember that "perfect is the enemy of the good" (especially with a very limited amount of development time available).

aarchiba:
API question: currently Fitter objects have a number of methods for extracting parameter names and values from their `.model` attribute. Is this the right place for these methods, or should they be on the model object itself?

My reasoning is that if it's useful to be able to pull out a dictionary mapping parameters to values, either the free parameters or all of them, that's something one would want to be able to do to a model without needing to construct a Fitter object. And if you have a Fitter object, it's not hard to just take its `.model` attribute. Documentation-wise, this then means that Fitter objects have only methods related to fitting. Unfortunately TimingModel objects are a little messy and have a lot of messy methods and attributes, but at least these are pretty natural.

scottransom:
Yeah, I think that sounds pretty good to me.  Although one thing *is* different about a Fitter instance than just a stand-alone model instance.  It is only in the context of the former that free/frozen really makes any sense.  Because you can load in a parfile that someone could have just edited to remove (or add) the fit flags, and the model still makes sense.

So as long as the methods are really only dealing with information about the free params to be fit, then maybe keeping them as Fitter methods is fine.  I definitely think that the timing models need better ways of accessing their params and especially their delay/phase and derivative methods.

luojing1211:
@aarchiba @scottransom I agree with you 100%. The timing model and parameter structures need to be optimized. My plan is to work on some part of this after the paper resubmission. We should have a list of those things. Is it helpful to have this list on the project board of this repo?

aarchiba:
> Yeah, I think that sounds pretty good to me. Although one thing _is_ different about a Fitter instance than just a stand-alone model instance. It is only in the context of the former that free/frozen really makes any sense. Because you can load in a parfile that someone could have just edited to remove (or add) the fit flags, and the model still makes sense.
> 
> So as long as the methods are really only dealing with information about the free params to be fit, then maybe keeping them as Fitter methods is fine. I definitely think that the timing models need better ways of accessing their params and especially their delay/phase and derivative methods.

If you would prefer to see the frozen flags moved from the timing model to the fitter, we can talk about that. But then that makes it quite inconvenient to take a model and try a different Fitter on it - you need to then pull the frozen flags from the one fitter to the other. I think it makes sense to view which parameters are frozen as part of the model. Of course someone can change the par file, or the model object, to flip the frozen/free flags around; that produces a different model, just like changing the values would.

aarchiba:
> @aarchiba @scottransom I agree with you 100%. The timing model and parameter structures need to be optimized. My plan is to work on some part of this after the paper resubmission. We should have a list of those things. Is it helpful to have this list on the project board of this repo?

This specific API change I suggested here would fit naturally into this PR, and would not be a lot of work. Broader API issues, I'm not sure what the best approach is; my inclination is to file them as bugs for now, as we have a procedure for managing bug reports and PRs. A label for API issues would make it easier to collect all API-change bug reports in one place.

scottransom:
As I mentioned in another comment, I would like to see the get/set fit/free/frozen params methods be available in the Fitter, by simply calling the relevant new methods in the timing model.  I would prefer to *not* have those be deprecated.  The reasons are that 1) a fitter is the only place where free/frozen really matters, and so that is where at least some people (like me) will expect to find it and 2) it costs us basically nothing.

aarchiba:
> As I mentioned in another comment, I would like to see the get/set fit/free/frozen params methods be available in the Fitter, by simply calling the relevant new methods in the timing model. I would prefer to _not_ have those be deprecated. The reasons are that 1) a fitter is the only place where free/frozen really matters, and so that is where at least some people (like me) will expect to find it and 2) it costs us basically nothing.

Have you tried finding anything in the PINT documentation lately? A profusion of functions hurts learnability and intelligibility of the documentation. The current methods are deprecated because the use of the word "fit" to mean "free" is confusing. As it currently stands, the methods to obtain a list of free parameters and their corresponding Parameter objects, numerical values, or uncertainties remain on the Fitter object only; I had considered moving them to the Fitter's `model` attribute but I could duplicate them there instead. "There should be one-- and preferably only one --obvious way to do it." If we move them, then the Fitter object contains no methods for manipulating Parameters in any way; all those methods will live on Fitter's `.model` attribute, as manipulating parameters, including free versus frozen, manipulates the model. Discoverability is fairly straightforward; the Fitter doc page is short and tractable, tab completion works well and does not offer anything to do with parameters, and the `model` attribute is prominently displayed.

An argument for attaching free/frozen to models rather than fitters is that one could quite reasonably switch between WLSFitter to get close to a true value quickly and MCMCFitter to explore around it, and one wants the list of which parameters are free or frozen to be preserved as one goes back and forth. This would also allow one to distinguish between models being compared with an F test or other mechanism that had an Occam factor.

scottransom:
> Have you tried finding anything in the PINT documentation lately? A profusion of functions hurts learnability and intelligibility of the documentation.

I'm not asking for a profusion of functions.  I'm asking for the *same* function to be available in two different contexts.

> The current methods are deprecated because the use of the word "fit" to mean "free" is confusing.

I completely disagree.  For any user of TEMPO/Tempo2, you put a 1 in a parfile when you want to fit the parameter.  There is no "free" or "frozen".  Check out the Tempo2 Tutorial and here:  http://tempo.sourceforge.net/ref_man_sections/freeformat.txt

> As it currently stands, the methods to obtain a list of free parameters and their corresponding Parameter objects, numerical values, or uncertainties remain on the Fitter object only

Right.  And they should be moved to the timing model class.  I'm all for that.  *But* I would like to be able to access them directly from the Fitter as well.  Fitters must have a timing model, so I don't see that as being strange at all.

> "There should be one-- and preferably only one --obvious way to do it." If we move them, then the Fitter object contains no methods for manipulating Parameters in any way; all those methods will live on Fitter's `.model` attribute, as manipulating parameters, including free versus frozen, manipulates the model.

What I'm suggesting *is* one way to do it.  I'm suggesting that the same methods get called!

> Discoverability is fairly straightforward; the Fitter doc page is short and tractable, tab completion works well and does not offer anything to do with parameters, and the `model` attribute is prominently displayed.

How is that hurt by having a few .get* and .set* params?

> An argument for attaching free/frozen to models rather than fitters is that one could quite reasonably switch between WLSFitter to get close to a true value quickly and MCMCFitter to explore around it, and one wants the list of which parameters are free or frozen to be preserved as one goes back and forth. This would also allow one to distinguish between models being compared with an F test or other mechanism that had an Occam factor.

I totally agree!  I've been agreeing with putting those methods into the timing model for days now.  I just want the user to be able to access them from the fitter *also*.  We can easily put documentation in the Fitter class that theses are just helper methods thta modify the model itself.  And the same result can be had by modifying the model.

To me, this makes cleaner code and it *helps* the user because they are able to adjust fit/free and frozen the only place they matter -- in the Fitter.  (even though, by convention, they basically need to live in the timing model)

I basically completely in agreement with most of what you are suggesting.

Issue #801, opened 2020-11-04
Parameters included in set_fitparams not added to the model should trigger a warning


swiggumj:
Currently, if parameters are listed in `fitter.set_fitparams(...)` but not explicitly added to the timing model ahead of time, those params are not included in the fit. This should warn the user and suggest a fix if fits for added parameters are indeed intended.

dlakaplan:
@swiggumj : I'll work on this.  Can you post some examples of how you've been encountering this?

tcromartie:
As requested in #813, I'm adding that this came up in reading 'STIG' (which should be 'STIGMA') from a parfile. No warnings are given, the parameter seems to just be ignored. It might be helpful to give a warning and advise that 'STIGMA' should be used, or to just interpret it automatically.

swiggumj:
Here's a quick example:
```
f0355 = pint.fitter.WLSFitter(t0355,m0355)
f0355.set_fitparams('RAJ','DECJ','F0','F1','DM','JUMP1')
f0355.fit_toas()
```
I've made a fitter object, run `set_fitparams` to determine parameters that are unfrozen, then I run the fit. Any parameters not included here, are frozen, however, if unknown parameters (including actual parameter names outside of existing model components) are added, they are simply ignored. For example, if I replaced the line above with:
```
f0355.set_fitparams('RAJ','DECJ','F0','F1','DM','JUMP1','CAPYBARA','PB')
```
The exact same thing will happen and no warnings are output about parameters listed that aren't actually fit (or in some cases, aren't recognized at all).

dlakaplan:
Thanks, Joe.

Could you provide a complete example (how you generate your f0355 and m0355 objects) along with the relevant files?   If the files are too ugly to attach here, just send to me.

Also, just for completeness could you list the right way to do this?

swiggumj:
Answered some of this offline, but for posterity...
```
source = 'J0355+28'
t0355 = toa.get_TOAs('CHIME_J0355+28_princeton.tim', usepickle=False, bipm_version='BIPM2015', ephem='DE430')
m0355 = models.get_model('J0355+28_PINT_20200929.par')
# Added to cull TOAs with large uncertainties
t0355.select(t0355.get_errors() < 1300.0*u.us)
f0355 = pint.fitter.WLSFitter(t0355,m0355)
f0355.set_fitparams('RAJ','DECJ','F0','F1','DM','JUMP1')
f0355.fit_toas()
f0355.print_summary()
```
[...]

PR #823, opened 2020-11-20
Set fitparams

dlakaplan:
I believe this fixes https://github.com/nanograv/PINT/issues/801, but I am not 100% certain because of other concurrent changes that @aarchiba has been making.

### DIFF
--- a/tests/test_fitparams.py
+++ b/tests/test_fitparams.py
@@ -1,0 +1,40 @@
+import logging
+import os
+import unittest
+
+import astropy.units as u
+
+import pint.models as models
+from pint import fitter, toa
+from pinttestdata import datadir
+
+
+class TestFitparams(unittest.TestCase):
+    """Test whether unfreezing parameters that are not in the model will cause fitting to fail"""
+
+    @classmethod
+    def setUpClass(cls):
+        # J0613 is in equatorial
+        cls.parfile = os.path.join(datadir, "NGC6440E.par")
+        cls.model = models.get_model(cls.parfile)
+
+        cls.timfile = os.path.join(datadir, "NGC6440E.tim")
+        cls.toas = toa.get_TOAs(cls.timfile, usepickle=False)
+        cls.fitter = fitter.WLSFitter(cls.toas, cls.model)
+        cls.log = logging.getLogger("TestFitparams")
+
+    def test_fit_params(self):
+        """
+        unfreeze valid parameters and fit
+        unfreeze invalid parameters and fail
+        """
+        # this should succeed since all of these are valid parameters
+        self.fitter.set_fitparams("RAJ", "DECJ", "F0", "F1", "DM")
+        self.fitter.fit_toas()
+
+        # this should fail
+        self.assertRaises(
+            AttributeError,
+            self.fitter.set_fitparams,
+            ("RAJ", "DECJ", "F0", "F1", "DM", "CAPYBARA"),
+        )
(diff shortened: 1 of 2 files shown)


aarchiba:
I think this conflicts with #822 - we should talk about the implications. In particular I think set_fitparams needs to be deprecated or directly deleted, and instead it should be possible to set the free parameters in a model. See the code and discussion on that PR.

tests/test_fitparams.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        # J0613 is in equatorial
+        cls.parfile = os.path.join(datadir, "NGC6440E.par")
+        cls.model = models.get_model(cls.parfile)
+
+        cls.timfile = os.path.join(datadir, "NGC6440E.tim")
+        cls.toas = toa.get_TOAs(cls.timfile, usepickle=False)
+        cls.fitter = fitter.WLSFitter(cls.toas, cls.model)
```
aarchiba, line 23: This test case demonstrates the problem with the API - one should be able to freeze or thaw parameters without needing to load TOAs and construct a Fitter.

scottransom: I'm confused..  You *can* freeze and thaw without a fitter instance.  .frozen is a model parameter attribute that you can just set:
```
n [1]: m.F0.frozen                                                             
Out[1]: False

In [2]: m.F0.frozen = True                                                      

In [3]: m.F0.frozen                                                             
Out[3]: True
```
Is there something wrong with that?

scottransom: But going back to my last comment, the *value* of .frozen really only matters in the context of a Fitter.

aarchiba: Indeed, Patrick's notebooks have a lot of for loops full of getattr to freeze everything and then manually unfreeze a few things one by one. It would be useful to be able to just set free_params to the list of free parameters - but you can't do that without constructing a Fitter.

scottransom: Are you just saying that you would like a free_params() and/or freeze_params() method in the timing model instances?  I think that is completely fine.

But you can do it with a one liner now without a Fitter (m=model):
`for p in m.params:  getattr(m, p).frozen = False if p in ['F0', 'F1'] else True`

scottransom: Or if you want `free_params` to be a list of the free params, a list comp does it:
`free_params = [p for p in mm.params if getattr(mm, p).frozen==False]`

So I'm just confused why you say you need a fitter for these things?

If it is just that you want the syntactic sugar of set_free_params() or somesuch in the model, as I said, I'm totally fine with that!

aarchiba: Since the `.frozen` feature of parameters lives in the model object, I don't want any manipulation of it, individually or in bulk, to have to go through a Fitter object - and `set_fitparams` does, it lives in a Fitter object. So I want to make that ability - set in bulk, which is featured in example notebooks because it's really convenient not to have to write a loop (let alone use `getattr`) - available from model objects directly. Which #822 does. I'd like to remove it from Fitter objects, which #822 also does, with a deprecation warning.

scottransom: I'd rather not have it removed from fitter objects.  I think it is useful there.  Fitter objects need a model, and they are the only places where frozen matters.

Once again, I'm completely happy to have the bulk setting and getting put into the timing model objects.  I just want the fitter to be able to access that directly.

aarchiba:
I think this is not needed any more after #822 - I think tests like the one below demonstrate what you wanted: https://github.com/nanograv/PINT/blob/master/tests/test_timing_model.py#L169

aarchiba:
Can this be closed?

scottransom:
I think so.  It is unfortunate that we had a couple people working on the same issue.  We should try to avoid this in the future.  Would be not great if we accidentally stepped on the toes of a student project or something like that.

Issue #817, opened 2020-11-18
Free Travis CI is going away


scottransom:
There is a nice discussion of this on the astropy-dev list:  https://groups.google.com/g/astropy-dev/c/45frnTEAX-U

Astropy switched from Travis CI to github's "Actions" for CI.

https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

JPGlaser:
So we have a few options. We could use GitHub's Status API to do CI on our own hardware (depending on the strain/frequency of CI requests across nanograv's projects).
https://docs.github.com/en/free-pro-team@latest/rest/guides/building-a-ci-server

Short-term there is a possible solution to **local** runs of Travis via docker images:
https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally/49019950#49019950

aarchiba:
For local CI, `tox` was intended to be a reasonable option - runs the code with a variety of dependency versions and tests and makes doc builds. It might need a little attention to bring it back to current convenient use. 

I have a desktop machine at work I could set to running CI jobs - it's not exactly high-powered but we don't necessarily need much - but I'm not sure the IT folks would be okay with it, and I'd need their help setting up Docker and making it accessible from the Web.

aarchiba:
Closed by #882

PR #825, opened 2020-11-23
Fix ELL1 bugfix for orbital_phase() #795

### LINKED ISSUE #795: Orbital phase calc fails for ELL1 models
```
m.orbital_phase(bats)

~/src/PINT/src/pint/models/stand_alone_psr_binaries/binary_orbits.py in __getattr__(self, name)
     54         try:
---> 55             return super(orbits, self).__getattribute__(name)
     56         except AttributeError:

AttributeError: 'OrbitPB' object has no attribute 'tt0'
```

Looks like ELL1 models don't have the `tt0` attribute that this is trying to use.

scottransom:
This should fix #795 
Also fixed two errors in docstrings

### DIFF
--- a/src/pint/models/stand_alone_psr_binaries/ELL1_model.py
+++ b/src/pint/models/stand_alone_psr_binaries/ELL1_model.py
@@ -46,10 +46,10 @@ class ELL1BaseModel(PSR_BINARY):
         """
         ttasc = t - TASC
         """
+        t = self.t
         if not hasattr(self.t, "unit") or self.t.unit == None:
             t = self.t * u.day
-        t = self.t
-        ttasc = (t - self.TASC).to("second")
+        ttasc = (t - self.TASC.value * u.day).to("second")
         return ttasc
 
     def a1(self):
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -361,6 +361,8 @@ class TimingModel(object):
                 bts = np.asarray(barytimes.mjd)
             else:
                 raise ValueError("barytimes as Time instance needs scale=='tdb'")
+        elif isinstance(barytimes, MJDParameter):
+            bts = np.asarray(barytimes.value)
         else:
             bts = np.asarray(barytimes)
         bbi = b.binary_instance  # shorthand
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -358,11 +358,11 @@ class TimingModel(object):
             bts = self.get_barycentric_toas(barytimes)
         elif isinstance(barytimes, time.Time):
             if barytimes.scale == "tdb":
-                bts = np.asarray(barytimes.mjd)
+                bts = np.asarray(barytimes.mjd_long)
             else:
                 raise ValueError("barytimes as Time instance needs scale=='tdb'")
         elif isinstance(barytimes, MJDParameter):
-            bts = np.asarray(barytimes.value)
+            bts = np.asarray(barytimes.value)  # .value is always a MJD long double
         else:
             bts = np.asarray(barytimes)
         bbi = b.binary_instance  # shorthand
--- a/tests/test_orbit_phase.py
+++ b/tests/test_orbit_phase.py
@@ -20,10 +20,6 @@ class TestOrbitPhase(unittest.TestCase):
         os.chdir(datadir)
         cls.pJ1855 = "B1855+09_NANOGrav_dfg+12_modified_DD.par"
         cls.mJ1855 = m.get_model(cls.pJ1855)
-        cls.pJ1855ell1 = "B1855+09_NANOGrav_12yv3.wb.gls.par"
-        cls.mJ1855ell1 = m.get_model(cls.pJ1855ell1)
-        cls.pJ0737 = "0737A_latest.par"
-        cls.mJ0737 = m.get_model(cls.pJ0737)
 
     def test_barytimes(self):
         log = logging.getLogger("test_barytimes")
@@ -72,26 +68,28 @@ class TestOrbitPhase(unittest.TestCase):
 
     def test_J1855_ell1(self):
         log = logging.getLogger("test_J1855_ell1")
-        phs1 = self.mJ1855ell1.orbital_phase(self.mJ1855ell1.TASC.value, anom="mean")
+        mJ1855ell1 = m.get_model("B1855+09_NANOGrav_12yv3.wb.gls.par")
+        phs1 = mJ1855ell1.orbital_phase(mJ1855ell1.TASC.value, anom="mean")
         assert phs1.value == 0.0, "Mean anom != 0.0 at TASC as value"
-        phs1 = self.mJ1855ell1.orbital_phase(self.mJ1855ell1.TASC, anom="mean")
+        phs1 = mJ1855ell1.orbital_phase(mJ1855ell1.TASC, anom="mean")
         assert phs1.value == 0.0, "Mean anom != 0.0 at TASC as MJDParam"
-        phs1 = self.mJ1855.orbital_phase(self.mJ1855ell1.TASC.value + 0.1, anom="mean")
-        phs2 = self.mJ1855.orbital_phase(self.mJ1855ell1.TASC.value + 0.1, anom="ecc")
+        phs1 = self.mJ1855.orbital_phase(mJ1855ell1.TASC.value + 0.1, anom="mean")
+        phs2 = self.mJ1855.orbital_phase(mJ1855ell1.TASC.value + 0.1, anom="ecc")
         assert phs2 != phs1, "Eccen anom == Mean anom"
 
     def test_J0737(self):
         log = logging.getLogger("test_J0737")
         # Find a conjunction which we have confirmed by GBT data and Shapiro delay
-        x = self.mJ0737.conjunction(55586.25)
+        mJ0737 = m.get_model("0737A_latest.par")
+        x = mJ0737.conjunction(55586.25)
         assert np.isclose(x, 55586.29643451057), "J0737 conjunction time is bad"
         # And now make sure we calculate the true anomaly for it correctly
-        nu = self.mJ0737.orbital_phase(x, anom="true").value
-        omega = self.mJ0737.components["BinaryDD"].binary_instance.omega().value
+        nu = mJ0737.orbital_phase(x, anom="true").value
+        omega = mJ0737.components["BinaryDD"].binary_instance.omega().value
         # Conjunction occurs when nu + OM == 90 deg
         assert np.isclose(
             np.degrees(np.fmod(nu + omega, 2 * np.pi)), 90.0
         ), "J0737 conjunction time gives bad true anomaly"
         # Now verify we can get 2 results from .conjunction
-        x = self.mJ0737.conjunction([55586.0, 55586.2])
+        x = mJ0737.conjunction([55586.0, 55586.2])
         assert len(x) == 2, "conjunction is not returning an array"
--- a/tests/test_orbit_phase.py
+++ b/tests/test_orbit_phase.py
@@ -45,32 +45,36 @@ class TestOrbitPhase(unittest.TestCase):
         assert np.all(phs.value >= 0), "Not all phases >= 0"
         assert np.all(phs.value <= 2 * np.pi), "Not all phases <= 2*pi"
         phs2 = self.mJ1855.orbital_phase([ts[0], ts[49]], anom="ecc")
-        assert phs2[0] == phs[0], "Eccen anom != Mean anom"
+        assert np.isclose(phs2[0], phs[0]), "Eccen anom != Mean anom"
         assert phs2[1] != phs[49], "Eccen anom == Mean anom"
         phs3 = self.mJ1855.orbital_phase([ts[0], ts[49]], anom="true")
-        assert phs3[0] == phs[0], "Eccen anom != True anom"
+        assert np.isclose(phs3[0], phs[0]), "Eccen anom != True anom"
         assert phs3[1] != phs[49], "Eccen anom == True anom"
 
     def test_J1855_zero_ecc(self):
         self.mJ1855.ECC.value = 0.0  # set the eccentricity to zero
         self.mJ1855.OM.value = 0.0  # set omega to zero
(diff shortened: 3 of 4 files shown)


src/pint/models/stand_alone_psr_binaries/ELL1_model.py
```diff
@@ -46,10 +46,10 @@ def ttasc(self):
         """
         ttasc = t - TASC
         """
+        t = self.t
         if not hasattr(self.t, "unit") or self.t.unit == None:
             t = self.t * u.day
-        t = self.t
-        ttasc = (t - self.TASC).to("second")
+        ttasc = (t - self.TASC.value * u.day).to("second")
```
aarchiba, line 52: Isn't `self.TASC.quantity` a Quantity? If that has units of days then you could drop `* u.day`, and if it doesn't then `.value*u.day` will give the wrong answer?

src/pint/models/timing_model.py
```diff
@@ -361,6 +361,8 @@ def orbital_phase(self, barytimes, anom="mean", radians=True):
                 bts = np.asarray(barytimes.mjd)
             else:
                 raise ValueError("barytimes as Time instance needs scale=='tdb'")
+        elif isinstance(barytimes, MJDParameter):
+            bts = np.asarray(barytimes.value)
```
aarchiba, line 365: Likewise, `.value` versus `.quantity.to_value(u.day)`?

tests/test_orbit_phase.py
```diff
@@ -69,6 +71,16 @@ def test_J1855_zero_ecc(self):
...
 
+    def test_J1855_ell1(self):
+        log = logging.getLogger("test_J1855_ell1")
+        phs1 = self.mJ1855ell1.orbital_phase(self.mJ1855ell1.TASC.value, anom="mean")
+        assert phs1.value == 0.0, "Mean anom != 0.0 at TASC as value"
+        phs1 = self.mJ1855ell1.orbital_phase(self.mJ1855ell1.TASC, anom="mean")
+        assert phs1.value == 0.0, "Mean anom != 0.0 at TASC as MJDParam"
+        phs1 = self.mJ1855.orbital_phase(self.mJ1855ell1.TASC.value + 0.1, anom="mean")
+        phs2 = self.mJ1855.orbital_phase(self.mJ1855ell1.TASC.value + 0.1, anom="ecc")
+        assert phs2 != phs1, "Eccen anom == Mean anom"
```
aarchiba, line 82: How different should they be? Floating-point equality is not nice.

tests/test_orbit_phase.py
```diff
@@ -20,6 +20,8 @@ def setUpClass(cls):
         os.chdir(datadir)
         cls.pJ1855 = "B1855+09_NANOGrav_dfg+12_modified_DD.par"
         cls.mJ1855 = m.get_model(cls.pJ1855)
+        cls.pJ1855ell1 = "B1855+09_NANOGrav_12yv3.wb.gls.par"
+        cls.mJ1855ell1 = m.get_model(cls.pJ1855ell1)
         cls.pJ0737 = "0737A_latest.par"
         cls.mJ0737 = m.get_model(cls.pJ0737)
         cls.timfile = "test1.tim"
```
aarchiba, line 27: Is this `timfile` and the associates `toas` object ever used? Loading `.tim` files is slow and this gets executed for each test in the class.

tests/test_orbit_phase.py
```diff
@@ -20,6 +20,8 @@ def setUpClass(cls):
         os.chdir(datadir)
         cls.pJ1855 = "B1855+09_NANOGrav_dfg+12_modified_DD.par"
         cls.mJ1855 = m.get_model(cls.pJ1855)
+        cls.pJ1855ell1 = "B1855+09_NANOGrav_12yv3.wb.gls.par"
+        cls.mJ1855ell1 = m.get_model(cls.pJ1855ell1)
```
aarchiba, line 24: Is it necessary to parse this `.par` file for every test in this class? Would it not be preferable to just parse the `.par` files for the tests where they are actually used?

tests/test_orbit_phase.py
```diff
@@ -69,6 +71,16 @@ def test_J1855_zero_ecc(self):
         phs3 = self.mJ1855.orbital_phase(self.mJ1855.T0.value + 0.1, anom="true")
         assert phs3 == phs1, "True anom != Mean anom"
 
+    def test_J1855_ell1(self):
+        log = logging.getLogger("test_J1855_ell1")
```
aarchiba, line 75: What does this line accomplish? The variable `log` appears not to ever be used.

scottransom:
BTW, I think this looks ready to go....  @paulray can you run your code that had triggered #795 now?

aarchiba (commented):
I don't see how this fixes #795 ? I mean if the tests pass it must be fixed but this does not seem to touch the `.tt0` property, which appeared to be missing in that bug report?

src/pint/models/stand_alone_psr_binaries/ELL1_model.py
paulray, line 52: Why this change? Is it to ensure you get a Quantity with units of days, rather than a `Time()`?

scottransom:
> I don't see how this fixes #795 ? I mean if the tests pass it must be fixed but this does not seem to touch the `.tt0` property, which appeared to be missing in that bug report?

That's because the problem wasn't with the .tt0 property!  It was with the format of the values that were used in the standalone binary model.  In particular, it didn't like that m.TASC was a MJDParameter.

src/pint/models/stand_alone_psr_binaries/ELL1_model.py
scottransom: That doesn't work because you can't subtract Time quantities like that form non Time quantities.  Here is an example:
```
In [7]: 50000.0*u.day - mm.TASC.quantity                                                                 
TypeError                                 Traceback (most recent call last)
<ipython-input-7-01b65069e5a2> in <module>
----> 1 50000.0*u.day - mm.TASC.quantity

~/python_venvs/py3/lib/python3.8/site-packages/astropy/time/core.py in __rsub__(self, other)
   2016     def __rsub__(self, other):
   2017         out = self.__sub__(other)
-> 2018         return -out
   2019 
   2020     def _time_comparison(self, other, op):

TypeError: bad operand type for unary -: 'Time'
```

scottransom: And self.t is a numpy array of non-Time quantities

src/pint/models/timing_model.py
scottransom: That also doesn't work.  Maybe our MJDParameter needs work:
```
In [8]: mm.TASC.quantity.to_value(u.day)                                                                 
<ipython-input-8-a2d9b63ca624> in <module>
----> 1 mm.TASC.quantity.to_value(u.day)

~/python_venvs/py3/lib/python3.8/site-packages/astropy/time/core.py in to_value(self, format, subfmt)
    983         # only, to make life easier for TimeDelta.to_value()).
    984         if format not in self.FORMATS:
--> 985             raise ValueError('format must be one of {}'
    986                              .format(list(self.FORMATS)))
    987 

ValueError: format must be one of ['jd', 'mjd', 'decimalyear', 'unix', 'unix_tai', 'cxcsec', 'gps', 'plot_date', 'stardate', 'datetime', 'ymdhms', 'iso', 'isot', 'yday', 'datetime64', 'fits', 'byear', 'jyear', 'byear_str', 'jyear_str', 'pulsar_mjd', 'mjd_long', 'pulsar_mjd_long', 'mjd_string', 'pulsar_mjd_string']
```

tests/test_orbit_phase.py
scottransom: Yeah, I don't actually know the right answer for that.  All I'm really testing there is that they are different.  Because there have been mistakes in the past where people have computed all of the anomalies for ELL1 as being the same.  So I'm not really sure what the right thing to do here (or where I would get the values to test against).

tests/test_orbit_phase.py
scottransom: It is only used in one of the tests.  I guess I could move it to that particular test.

tests/test_orbit_phase.py
scottransom: Truthfully, I have no idea.  I was just following what is done in many other test files.  Anyone know?

src/pint/models/stand_alone_psr_binaries/ELL1_model.py
scottransom: Yeah, the problem wasn't with .tt0, it was with the fact that self.t is always a numpy array, and I don't think we want that to be an array of Times (since this is a standalone module).  If self.t is not made of Times, then we can't subtract a Time from them.

tests/test_orbit_phase.py
scottransom: Sure.  I can do that.

tests/test_orbit_phase.py
scottransom: Done.

tests/test_orbit_phase.py
aarchiba: I'm pretty sure that many other files have done what you did. According to the documentation getLogger does not have side effects, so this is a pointless no-op. But if one wanted to log something during the tests one could use the `log` variable and arrange for... something? to happen. But really `pytest` keeps things like print statements together and shows them only if the test fails so I don't see the point of this. Possibly it predates the use of `pytest` and is entirely vestigial?

src/pint/models/timing_model.py
aarchiba: Maybe `MJDParameter` should look like a Quantity with units of days? possibly with an extra property to get the result as a Time object? By the time it's in a Parameter I'm not sure we need all the machinery of Time (as here)? Not really relevant for this PR, except in that a programming error led to a very confusing error message, not at all the intent of using Time and Quantity objects.

tests/test_orbit_phase.py
aarchiba: I mean 1e-3*eccentricity should probably cover it - I'm just concerned that someone might compute the "same" anomaly using slightly different order of operations and get a few ULP difference.

tests/test_orbit_phase.py
aarchiba: `pytest` has nicer facilities to support test fixtures that make it easier to ensure they are only constructed if needed, so a sweep through the test cases to modernize and simplify them would probably produce a low-effort speedup in the testing runtime. Not relevant to this PR especially though perhaps to #817 .

src/pint/models/timing_model.py
aarchiba: Actually I think this might be the result of a bare `except:` that makes any error during property access look like the property doesn't exist. Possibly here: https://github.com/nanograv/PINT/blob/6854a9ecb5cc61b7836d3a54b3e9e22691bce14c/src/pint/models/timing_model.py#L223

scottransom: I think the problem is the `.to_value(u.day)` bit.  That doesn't seem to work with even regular Time objects.

However, ```mm.TASC.quantity.to_value('mjd_long')``` works.  But it gives identical results to .value (both are numpy float128), so is there really an advantage?

```
In[31]: mm.TASC.quantity.to_value('mjd_long') == mm.TASC.value                                          
Out[31]: True
```

aarchiba: The problem was never about the numbers but about the assumption that a Quantity is in the units you think it's in. Using `x.value` will get you a number that abruptly changes if someone used `u.s` instead of `u.d`. That doesn't really arise in the same way for Time objects, and `MJDParameter` might be documented so that `.value` is always the same as `.to_value('mjd_long')`. (I think `t.mjd_long` is a shorter way to spell `t.to_value('mjd_long')` but both might be only in post-4 Astropy I know I added a number of convenience Time functions around then.)

scottransom: So I just checked, and while you can reset the units of an MJDParameter to be (for instance) `u.s`, if you use `.value` or `.quantity` on it, you still get an MJD.  I think that is because of the "pulsar_mjd" format of the Time instance.  So I think that using `.value` is identical to using `.quantity.to_value('mjd_long')`.  And you are right that (at least with my astropy), that `.mjd_long` doesn't work yet.

tests/test_orbit_phase.py
paulray: Yeah, I think that log stuff must be vestigial.  I support cleaning up the test cases, since when I make a new one I'd like to have a template to work from, but there are many different ones in our suite and I don't know which template is preferred.

aarchiba: I'd be happy to convert a test case or two to a modern pytest-based style, if I had one where I knew what was being tested so I could be sure I wasn't missing the point. Pointing to these in the docs ("how to write a test case") might not be a bad idea either.

paulray: Thanks, that would be enormously helpful! How about test_pulse_number.py. It was written to check that `Residuals` respect pulse numbers, but it could be extended to include tests to show that Fitters also handle them (as was apparently missed previously!)

aarchiba: Well, uh, funny you should say that, #814 has some modern tests I wrote in `test_pulse_number.py`: https://github.com/nanograv/PINT/pull/814/files#diff-b7b91b052c8f8a2c7e59db5eab7db1e0a88093116baf523c0bcc58d518c2f68cR48

aarchiba: I could produce a more whiz-bang version, or a simpler cleaner version, if we were going to use it as an example.

paulray: The most important target audience is a person submitting a PR that knows they need to add a unit test, but isn't sure just how to do it. So, clean and simple is my preference.

paulray: A couple things:

1. That test imports pytest, but not "unittest" and the test classes don't inherit from `unittest.TestCase`. I don't know what unittest is or if we should be using it.
1. It would be great if in the example where fancy @pytest.mark stuff is done, it has a 1-line comment explaining what that mark is doing (and maybe even a link to the docs on that topic).

scottransom: Logging lines are gone from this pull request

PR #831, opened 2020-11-26
Add paper validation example notebook

luojing1211:


docs/tutorials.rst
```diff
@@ -17,3 +17,4 @@ We don't really have any proper tutorials yet. But for the moment, we have a few
    examples/understanding_fitters.ipynb
    examples/Wideband_TOA_walkthrough.ipynb
    examples/build_model_from_scratch.ipynb
```
aarchiba, line 19: This notebook is in here twice.

docs/examples/paper_validation_example.md
```diff
@@ -0,0 +1,825 @@
+# Validation Example for PINT paper
```
aarchiba, line 1: Does this work with Jupytext? Like, does `make notebooks` work? It seems to be in a different format from all the others.

luojing1211: This is what I got,  
```
jupytext --sync paper_validation_example.md 
[jupytext] Reading paper_validation_example.md
[jupytext] Warning: 'paper_validation_example.md' is not a paired notebook
```

aarchiba: Does it construct a `.ipynb` file? That works?

luojing1211: I believe it is. It passes the notebook test. However, since there is a dependency, the doc build failed, and we have some eraf failure that fails everyone.

luojing1211: Is there a way to skip running this notebook when building the docs. On the other hand, what is the preferred format?

aarchiba: The idea is that jupytext should maintain a `.ipynb` and a `.md` version of each notebook, keeping them in sync automatically; the `.md` version contains no results and is checked into version control, and the `.ipynb` is created and run while building the docs. The instructions obviously need a little tidying up: https://pulsar-pint.readthedocs.io/en/getattr_cleanup/working-with-notebooks.html

aarchiba: One could create a separate directory for notebooks that couldn't be or shouldn't be re-run (although I would worry about them getting out of date and no one noticing) and then include them in the same way as the existing ones. There is still the awkwardness of checking a notebook with results in to git - any change and you get huge unreadable diffs. Merges are a nightmare. (I've tried this; the tool `nbdiff` does what it can but notebooks in git are asking for trouble.)

luojing1211: So when building the docs, the notebook should not run, right?

aarchiba: All existing notebooks are run when building the docs. This is desirable, as it forces us to notice when we break one of them. If you want to introduce a new category of notebook that is *not* built when we build the docs, we can discuss whether that is a good idea and if so how to set that up.

aarchiba: I think the way to do this would be to have a directory `examples-tempo` for notebooks that require TEMPO/TEMPO2 to run. One could then simply check in the `.ipynb` files, and ensure that `make examples-tempo` would successfully re-run them if TEMPO/TEMPO2 were available. One could then include them in the examples section of the docs alongside the others. This would have the advantage that github renders notebooks, so they could be viewed online easily. (The point of the markdown conversion is to ensure that we can handle modifications and re-running gracefully by not including the output; that doesn't seem relevant for this new category of notebook.)

aarchiba: Perhaps just put the `.ipynb` file in `docs/examples-built/` and then list it where appropriate in the documentation?

aarchiba:
If we want this in for the paper, if you post the `.ipynb` somewhere I can set up non-runnable notebooks and include it as part of #849 .

luojing1211:
I will add the .ipynb file this afternoon. I have problem to connect my office workstation.

luojing1211:
@aarchiba I just added the .ipynb file to this PR.  Feel free to download it.

aarchiba:
Superseded by #849

PR #834, opened 2020-11-26
Add check for STIGMA <= 0

### LINKED ISSUE #813: ELL1H issues: STIGMA crashes & lack of warnings when param is not fit
This issue came up in the context of NG 15yr notebooks. If 'STIG' is included in the parfile, STIGMA isn't fit, but no warning is given, either (the same might be true for 'VARSIGMA'). If 'STIGMA' is included in the parfile (in this case, for J1802-2124), the parfile is read, but the prefit residual plot is blank. Then, when the fit occurs, it crashes with the error "array must not contain infs or NANs." Fitting for H3/H4 works fine for this source.

aarchiba:
Fixes #813

### DIFF
--- a/src/pint/models/binary_ell1.py
+++ b/src/pint/models/binary_ell1.py
@@ -82,8 +82,7 @@ class BinaryELL1Base(PulsarBinary):
         super(BinaryELL1Base, self).setup()
 
     def validate(self):
-        """ Validate parameters
-        """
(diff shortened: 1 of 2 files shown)


aarchiba:
Should be uncontroversial and fine to merge if it passes Travis. (Passes tox at home.)

PR #838, opened 2020-11-30
Invalidate pickles based on settings

### LINKED ISSUE #837: pickles are not regenerated if ephem parameter is changed
The `usepickle` argument is meant to provide a transparent cache of TOAs objects. So if the `.tim` file changes, they should be regenerated. But if the `ephem` argument to `get_TOAs` is changed, the pickle is silently loaded and contains data processed with the old `ephem` value. The pickle should be considered invalid and either regenerated or at least raise an exception if it does not use the `ephem` value the user requested.

aarchiba:
This updates the pickle-checking code so that pickles are regenerated if they are reloaded with different settings (for example `ephem="de436"`).

Closes #837

### DIFF
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -77,10 +77,10 @@ JD_MJD = 2400000.5
 def get_TOAs(
     timfile,
     ephem=None,
-    include_bipm=True,
-    bipm_version=bipm_default,
-    include_gps=True,
-    planets=False,
+    include_bipm=None,
+    bipm_version=None,
+    include_gps=None,
+    planets=None,
     usepickle=False,
     tdb_method="default",
 ):
@@ -120,15 +120,16 @@ def get_TOAs(
         Filename or file-like object containing the TOA data.
     ephem : string
         The name of the solar system ephemeris to use; defaults to "DE421".
-    include_bipm : bool
-        Whether to apply the BIPM clock correction.
-    bipm_version : string
+    include_bipm : bool or None
+        Whether to apply the BIPM clock correction. Defaults to True.
+    bipm_version : string or None
         Which version of the BIPM tables to use for the clock correction.
-    include_gps : bool
-        Whether to include the GPS clock correction.
-    planets : bool
+    include_gps : bool or None
+        Whether to include the GPS clock correction. Defaults to True.
+    planets : bool or None
         Whether to apply Shapiro delays based on planet positions. Note that a
         long-standing TEMPO2 bug in this feature went unnoticed for years.
+        Defaults to False.
     usepickle : bool
         Whether to try to use pickle-based caching of loaded clock-corrected TOAs objects.
     tdb_method : string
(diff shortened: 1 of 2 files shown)


aarchiba:
If the user does not specify `ephem` (or whatever) as an argument to `get_TOAs`, it will accept whatever is in the pickle; but if a specific `ephem` is requested, disagreement will result in the pickle being regenerated.

src/pint/toa.py
```diff
@@ -144,12 +145,43 @@ def get_TOAs(
...
+                log.info("Pickle contains wrong ephem")
+                updatepickle = True
+            if planets is not None and t.planets != planets:
+                log.info("Pickle contains wrong planets")
+                updatepickle = True
         else:
             # Pickle either did not exist or is out of date
             updatepickle = True
-    t = TOAs(timfile)
+    if not usepickle or updatepickle:
```
scottransom, line 176: Note that we could be finer grained than this, which would help speed things up.  For instance, if a new ephem is used, all we need to do is re-compute the important table columns, not completely re-process the text .tim file.  The same applies to (I think) all of the other keyword arguments.

src/pint/toa.py
```diff
@@ -77,10 +77,10 @@
 def get_TOAs(
     timfile,
     ephem=None,
-    include_bipm=True,
-    bipm_version=bipm_default,
-    include_gps=True,
-    planets=False,
+    include_bipm=None,
```
scottransom, line 80: What is the reasoning for removing the default values from the signature and putting them into the code?  Seems like it makes the signature harder to understand.  Is this a recommended practice?

src/pint/toa.py
aarchiba: Do we have statistics on which of these operations (parsing the text file, applying clock corrections, or computing Earth positions and velocities) are slow? The later operations also depend on the result of the earlier operations, I think?

src/pint/toa.py
aarchiba: This makes it possible to load a pickle of unknown settings (as in one test case). It is custom to use "None" to indicate an unspecified default, and to put the default in the docstring, if the code wants to distinguish between unspecified and an explicit value.

src/pint/toa.py
scottransom: Yeah, you are right about some of them depending on earlier operations.  But I think we just need to be careful with that.  And I'm not sure that we have the detailed info on the timing of each step in get_TOAs() handy.  But it would be pretty easy to get.  I'm pretty sure that the posvels are the most costly.  And the text .tim file reading and parsing is costly too.  But I'd have to poke around a bit.  It is possible that @paulray might have the more information from the profiling that his student did this past summer (and which is now in the PINT paper).  I'll tag Chloe here as well in case she pays attention to this stuff still!  @champagne-cmd

aarchiba: Clock corrections are kind of a pain - un-applying one and applying a different one requires some non-trivial code. The other two are easy.

scottransom:
Is this one ready to merge, @aarchiba ?  I see that it has some of the fine-grained-ness that we talked about, which is good (and helpful even without all of it).

aarchiba:
I think it should be fine, I triggered a Travis rerun just in case but it passes tox on my machine.

scottransom:
OK.  I'll merge when Travis finishes.  This might help with the get-TOAs-and-model code I want to write in the next couple days.

Issue #820, opened 2020-11-19
Proper motion evaluation changes epoch time frame


kerrm:
I recently noticed models with a proper motion were taking much longer to evaluate phase.  Profiling reveals the slowdown is in delay calculations like the solar system geometry delay and solar system Shapiro delay, which (in my timing model), invoke AstrometryEquatorial.ssb_to_psb_xyz_ICRS.  That in turn ultimately calls AstrometryEquatorial.get_psr_coords, which uses astropy and the apply_space_motion method of SkyCoordinate to compute the position at the provided epochs.

In ssb_to_psb_xyz_ICRS, this is invoked like this:

`L_hat = self.ssb_to_psb_xyz_ICRS(epoch=tbl["tdbld"].astype(numpy.float64))`

get_psr_coords takes a kwarg epoch (default None).  If it is not already an instance of astropy.Time, as is the case with the above call, it does this:

`newepoch = Time(epoch, format="mjd")`

The default scale for time is TAI, but the obstime attribute of the pulsar SkyCoordinate has scale=='tdb'.  Calculating the deltas in apply_space_motion is then very slow, and moreover, seems like it's in error since the argument was already in TDB.  This can be checked by requesting the pulsar coordinates at POSEPOCH: doing so yields a position that's off by about 0.13 mu-as in my test.

I propose the simplest possible fix, namely adding

`newepoch = Time(epoch, scale="tdb",format="mjd")`

and adding some documentation to indicate that if the epoch is passed as float MJDs they need to be TDBs.  This speeds up the calculation substantially and results in a <2e-4 mu-as offset when checking against POSEPOCH.  (It *should* be 0, but -- see below.)

Questions/issues:
(1) Do you agree with the solution?
(2) Should other coordinate systems also expect their epoch to be TDB?  I think so.
(3) In testing this out, I found out that astropy fails a very simple sanity check, namely

`c0.apply_space_motion(dt=0*u.day).separation(c0).to(u.microarcsecond)                                                              
Out[5]: <Angle 0.00019839 uarcsec>`

Is this a problem?  If so, we might want to raise an issue.  I'm considering adding this to the PINT test suite, but it's obviously going to fail until it's fixed upstream.
(4) Any thoughts on other tests?

I'm happy to put together a PR.

aarchiba:
[...]

PR #842, opened 2020-11-30
Fix astrometry time scale

kerrm:
See #820 for some discussion.  This fixes the issue by using scale='tdb' for the astropy.Time object that is created when the various astrometry methods are called with vector-float arguments.

During implementation, a couple of other things cropped up:
1. The "new" method of evaluating proper motion was only applied to equatorial coordinates.  I updated the get_psr_coords method of AstrometryEcliptic to use the astropy machinery.
2. I simplified the base cases (no epoch or PM==0) for both classes.
3. The "obliquity" kwarg was missing in the two "dummy distance" methods in utils.py, which meant the new method of applying the evaluating the proper motion always produced a SkyCoord with the IERS2010 obliquity of the ecliptic, which in turn was causing tests to fail.  (Lucky the tests were there!)  I added that.
4. I updated test_Galactic.py to follow the new scale=='tdb' guideline and also strengthened the precision requirements on the tests by a few orders of magnitude.

### DIFF
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -267,7 +267,7 @@ class AstrometryEquatorial(Astrometry):
         Parameters
         ----------
         epoch: `astropy.time.Time` or Float, optional
-            new epoch for position.  If Float, MJD is assumed
+            new epoch for position.  If Float, MJD(TDB) is assumed
 
         Returns
         -------
@@ -279,23 +279,18 @@ class AstrometryEquatorial(Astrometry):
 
         """
         if epoch is None or (self.PMRA.value == 0.0 and self.PMDEC.value == 0.0):
-            dRA = 0.0 * u.hourangle
-            dDEC = 0.0 * u.deg
-            broadcast = 1
-            newepoch = self.POSEPOCH.quantity
             return coords.SkyCoord(
-                ra=self.RAJ.quantity + dRA,
-                dec=self.DECJ.quantity + dDEC,
-                pm_ra_cosdec=self.PMRA.quantity * broadcast,
-                pm_dec=self.PMDEC.quantity * broadcast,
-                obstime=newepoch,
-                frame=coords.ICRS,
-            )
+                ra=self.RAJ.quantity,
+                dec=self.DECJ.quantity,
+                pm_ra_cosdec=self.PMRA.quantity,
+                pm_dec=self.PMDEC.quantity,
+                obstime=self.POSEPOCH.quantity,
+                frame=coords.ICRS)
         else:
             if isinstance(epoch, Time):
                 newepoch = epoch
             else:
-                newepoch = Time(epoch, format="mjd")
+                newepoch = Time(epoch, scale="tdb", format="mjd")
             position_now = add_dummy_distance(self.get_psr_coords())
             position_then = remove_dummy_distance(
                 position_now.apply_space_motion(new_obstime=newepoch)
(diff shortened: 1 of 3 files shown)


src/pint/models/astrometry.py
```diff
@@ -535,10 +531,21 @@ def barycentric_radio_freq(self, toas):
...
+
+        Parameters
+        ----------
+        epoch: `astropy.time.Time` or Float, optional
+            new epoch for position.  If Float, MJD(TDB) is assumed
+
+        Returns
+        -------
+        position
+        PulsarEcliptic SkyCoord object optionally with proper motion applied
```
aarchiba, line 544: Should this be indented to render okay? Maybe check the output of `make docs`

tests/test_Galactic.py
```diff
@@ -76,7 +76,19 @@ def test_proper_motion(self):
...
-        assert sep < 1e-6 * u.arcsec, msg
+        assert sep < 1e-9 * u.arcsec, msg
+
+        # sanity check that evaluation at POSEPOCH returns something very close to 0
+        J0613_icrs = self.modelJ0613.coords_as_ICRS()
+        J0613_icrs_alt = self.modelJ0613.coords_as_ICRS(
+            epoch=self.modelJ0613.POSEPOCH.quantity.mjd
+        )
+        sep = J0613_icrs_alt.separation(J0613_icrs)
+        msg = (
```
aarchiba, line 87: Is this message needed? `pytest` gives pretty extensive context when failures occur, and in particular the test name (no human is calling the function, no need for a conveniently short name).

tests/test_Galactic.py
```diff
@@ -76,7 +76,19 @@ def test_proper_motion(self):
             "Applying proper motion for +100d failed with separation %.1e arcsec"
             % sep.arcsec
         )
-        assert sep < 1e-6 * u.arcsec, msg
+        assert sep < 1e-9 * u.arcsec, msg
```
aarchiba, line 79: Can these be split into separate test cases, so we can see separate failures if something goes wrong?

tests/test_Galactic.py
```diff
@@ -160,19 +169,23 @@ def test_ecliptic_to_galactic(self):
...
         msg = (
             "Ecliptic to Galactic conversion for +100d failed with separation %.1e arcsec"
             % sep.arcsec
         )
-        assert sep < 1e-6 * u.arcsec, msg
+        assert sep < 1e-9 * u.arcsec, msg
+
+
+if __name__ == "__main__":
+    unittest.main()
```
aarchiba, line 191: This isn't needed with `pytest`.

kerrm: No, but it's helpful when iterating the test in a console.

aarchiba: I just use `pytest tests/test_Galactic.py`.

aarchiba: (`pytest` generally provides nicer output and better control of what happens in tests. It also supports parametrized tests and various other nice things `unittest` doesn't, but even for `unittest` tests it is generally a nicer tool.)

tests/test_Galactic.py
kerrm: I split the "sanity check" out into its own method.  These other two were originally grouped so I'm inclined to leave them be.

src/pint/models/astrometry.py
kerrm: I overlooked it when updating the docstring, fixed in next commit.

tests/test_Galactic.py
kerrm: I would have argued yes, but now that I've split it into a separate method, context should be sufficient, so deleted!

aarchiba:
Closes #820

PR #844, opened 2020-12-01
Getitem implementation for TOAs

### LINKED ISSUE #824: .select() modifies TOAs objects and keeps a stack
Currently the only way to select a subset of TOAs from a TOAs object - and get a TOAs object as the result - is to modify the TOAs object in place with `.select()`. This keeps a stack of previous states inside the TOAs object. This is a clunky and non-pythonic API. The stack has no place here, and modifying the object in-place is inconvenient, doubly so as TOAs objects do not implement `.copy()`. Much preferable would be a mechanism like Astropy tables or numpy arrays where the same condition array would be passed as an index: `t[t.get_mjds()>58000*u.d]` would return a fresh well-behaved TOAs object. This is easily implemented with `__getitem__`, which might naturally also allow extracting columns or using integer indices or slices or various other obvious kinds of indexing that do not currently work.

aarchiba:
This allows users to use numpy-like indexing to select subsets of TOAs objects, creating new objects in the process. It supports boolean indexing, slice indexing ([::10] to pick every 10th TOA), and list indexing ([[1,3,6,9]]). `toas.select()` is marked as deprecated but nevertheless it no longer fails when handed an all-False selection.

Closes #824

### DIFF
--- a/docs/examples/time_a_pulsar.md
+++ b/docs/examples/time_a_pulsar.md
@@ -1,0 +1,138 @@
+---
+jupyter:
+  jupytext:
+    formats: ipynb,md
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: '1.2'
+      jupytext_version: 1.7.1
+  kernelspec:
+    display_name: Python 3
+    language: python
+    name: python3
+---
+
+# Time a pulsar
+
+This notebook walks through a simple pulsar timing session, as one might do with TEMPO/TEMPO2: load a `.par` file, load a `.tim` file, do a fit, plot the residuals before and after. This one also displays various additional information you might find useful, and also ignores but then plots TOAs with large uncertainties.
+
+```python
+import astropy.units as u
+import matplotlib.pyplot as plt
+
+import pint.fitter
+import pint.models
+from pint.models import get_model
+from pint.residuals import Residuals
+from pint.toa import get_TOAs
+```
+
+```python
+parfile = "NGC6440E.par"
+timfile = "NGC6440E.tim"
+```
+
+```python
+m = get_model(parfile)
+m
+```
+
+```python
+t_all = pint.toa.get_TOAs(timfile, ephem="de436")
+t_all.print_summary()
+```
+
+There are many messages here. As a rule messages marked `INFO` can safely be ignored, they are simply informational; take a look at them if something unexpected happens. Messages marked `WARNING` or `ERROR` are more serious. (These messages are emitted by the python `logger` module and can be suppressed or written to a log file if they are annoying.)
+
+
+Let's discard the data points with uncertainties $>30\,\mu\text{s}$ - uncertainty estimation is not always reliable when the signal-to-noise is low.
+
+```python
+error_ok = t_all.table["error"] <= 30 * u.us
+t = t_all[error_ok]
+t.print_summary()
+```
+
+```python
+# These are pre-fit residuals
+rs = Residuals(t, m).phase_resids
+xt = t.get_mjds()
+plt.figure()
+plt.plot(xt, rs, "x")
+plt.title("%s Pre-Fit Timing Residuals" % m.PSR.value)
+plt.xlabel("MJD")
+plt.ylabel("Residual (phase)")
+plt.grid()
+```
+
+```python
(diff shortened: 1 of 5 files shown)


aarchiba:
This does not currently support indexing with a single integer to construct a TOA object; currently there does not seem to be a way to make a TOA object out of the information in a TOAs object without assembling it by hand; even `write_TOA_line` accepts named arguments rather than TOA objects, and where it is called one can see the complexities that would be involved: when writing out TOA lines the clock corrections are un-applied and some non-trivial bookkeeping is done.

scottransom:
Love it!  Thanks for this, @aarchiba!

Any chance you can change the examples (commented out) in examples/fit_NGC6440E.py so that they use fancy indexing?

That script is always the example that I give to people to see how you might do an old-school TEMPO-like run in a program.

aarchiba:
Does that actually appear in our online docs anywhere? Is there any reason not to turn it into a notebook and add it to our tutorials? It's maybe not ideal for this purpose as it doesn't have obvious subsets to choose but there is a reason you keep sending it to people. And I don't find our tutorials as accessible as I would like.

scottransom:
Huh.  That's kind of strange...  I think that most or all of that script is/was part of `PINT_walkthrough.md`.  I just uilt the docs locally and it doesn't get converted into a notebook automatically, though.  I need to learn how to use the docs and doc building stuff more...

scottransom:
https://nanograv-pint.readthedocs.io/en/latest/examples/PINT_walkthrough.html has most of it, I think.

aarchiba:
Mmm. I think actually a version of this would go in "How do I time a pulsar?"

The way the notebooks work is you git clone, you run make notebooks, then you open the (newly created) Jupyter notebooks in a JupyterLab. You edit them, update, do whatever, hit save. Under the hood, jupytext maintains a Markdown copy of just the input side. When you're done, you use git add on the markdown files or the whole examples directory, then commit the updated markdown files.

`.py` files play no part of this process. I realize the instructions on working with notebooks are a little vague. And we have no place for pre-rendered notebooks (that can't be automatically rerun because they take forever or depend on things, but will inevitably get out of date and break without us noticing). I'm also unsure that there is an easy way to download the notebooks to work with them yourself. So there is work to do.

But I think it makes sense for me to make sure that subset selections are demonstrated somewhere findable in the tutorials/howtos. Easiest would be subsets by telescope or frequency.

scottransom:
yeah, I'm with you on all of that.  I do find that `.py` files are really useful, still, though.  Having a script you can run is often (IMO) way easier than dealing with notebooks.  Especially for real work (I like notebooks better for teaching or documenting things, though).

There are select examples in the PINT_walkthrough that I mentioned, though.  As well as in the `fit_NGC6440E.py` script.

aarchiba:
You can always download the code from a notebook as a `.py` file; if the notebook is appropriately structured you can go ahead and run it.

paulray:
Many people never use notebooks, so I like having standalone python script examples as well.

aarchiba:
> Many people never use notebooks, so I like having standalone python script examples as well.

There are a few possibilities here (which we can mix and match):

- Include pretty-printed python scripts, with outputs and download links in the documentation as examples.
- Arrange for notebooks to be accompanied by download links for `.ipynb` and `.py` files (the latter can be automatically generated). This could include "notebooks" that were just prettified scripts.
- Provide longer but readable scripts for download, with a short description in the docs.

But most of this should be another PR, for general doc improvements. I have an initial notebooky version of your example, Scott, and am trying to tidy it and make sure it demonstrates selection. Will appear soon.

aarchiba:
Notebook: https://pulsar-pint.readthedocs.io/en/getitem/examples/time_a_pulsar.html

Other doc/notebook/example enhancements see #845

aarchiba:
This is ready to merge, but might warrant discussion in the telecon; at the least someone else should look at it and see if they agree with my design decisions.

aarchiba:
New documentation:

- Residuals: https://pulsar-pint.readthedocs.io/en/wideband_improvements/api/pint.residuals.html
- Fitter: https://pulsar-pint.readthedocs.io/en/wideband_improvements/api/pint.fitter.html

Issue #845, opened 2020-12-01
Example code needs improvement.


aarchiba:
- We have a number of notebooks making up part of our documentation. Ideally it would be easy for users to download `.ipynb` and (maybe) `.py` versions of each; these can be automatically generated by `jupytext`, and are during the doc build process.
- We have some example `.py` standalone scripts. It would be nice if these were at least listed in the docs, possibly including pretty-printing (but a link to the file on github provides that and the ability to download) and/or output (including plots).
- Should we include some notebooks that cannot be rerun during the doc build process? The risk is them getting out of date, but some notebooks take a long time to run (e.g. MCMC-based notebooks) and others need external software (notably TEMPO/TEMPO2; see #831 ).

aarchiba:
Fixed in #849

PR #849, opened 2020-12-03
Make notebooks downloadable

aarchiba:
Each notebook in the documentation now has a download link on its page. Relevant to #845


aarchiba:
I would like to be able to download the notebooks as `.py` files; `jupytext` is quite capable of making such files, and even using them instead of Markdown, so one option would be to convert our example notebooks (again) so that runnable `.py` files were what lived in our repository, and the notebooks would be generated as they are now.

scottransom:
If that still produced "pretty" notebooks (and not too un-pretty .py files), I'd be highly in favor of that.

aarchiba:
The notebooks would be identical, and the `.py` files should be okay - I'd use the "percent" format, where Jupyter cell breaks are marked with `# %%`. The text-heavy Jupyter notebooks would become a bit comment-heavy, and the Markdown would naturally not be rendered, but the markdown/python files intentionally do not contain cell outputs anyway. And the "percent" format is understood by (for example) Spyder, which (I'm told) has an "execute this code cell" button. So I think it's probably a better solution overall; the only downside is that unless I'm very clever we'll make modification histories for the notebooks cumbersome to access (there will be a break where I do the conversion). 

(... possibilities for cleverness now percolating.)

aarchiba:
An example of the `.py` format is here: https://github.com/nanograv/PINT/pull/849/files#diff-a98c375a058be7b8ade21175fed010f543d20262d93cb2b3dbe515a7f26dbf55

aarchiba:
What it looks like rendered: https://pulsar-pint.readthedocs.io/en/notebook_doc_fixes/examples/Wideband_TOA_walkthrough.html 
The `.py` link points to github, so it won't work yet.

aarchiba:
Supersedes #831  - this adds the paper validation notebook, so it'd be good to merge it before the paper. Ready to merge once someone looks over it and agrees it's a good idea. (The change in notebook formats might warrant discussion on the telecon but I'm not sure anyone but me understands how we handle example notebooks anyway.)

luojing1211:
Can we merge this?

PR #853, opened 2020-12-04

bshapiroalbert:
I have updated the F-tests in the fitter so that they are compatible with the WidebandFitter and subsequent `CombinedResiduals' classes. Due to the recent updates to the `Residual` class, this basically has only changed the outputs of the `ftest()` function if a `WidebandFitter` is used, since all `Residual` objects now have the same attribute `.dof` and `.chi2`.  If using the full_output option with a WidebandFitter, it will also return the time and DM chi-squared values separately (as well as the total value) and the weighted and unweighted rms of the DM residuals. If using a GLS/WLSFitter, output remains unchanged from previous version. Documentation and tests have been updated as well. There were enough API changes since the last time I worked on this that it was easier to open a new PR.

In response to @aarchiba comments on the old PR, #807 , I have separated the tests of the `ftest` function in `test_fitter.py` to have one for a `WLSFitter` and a separate test for the `WidebandFitter`. I have also removed the output of separate DoF and chi2 for the TOA and DM residuals because, as previously discussed, they have no actual physical meaning and should not be considered separately. Comments are welcome but I believe this van be merged as soon as possible assuming it passes all the tests.


aarchiba:
Looks generally pretty good. I'd like to see some statistical tests - does the claimed false positive probability from the F test agree with the actual null-hypothesis fraction of false positives - but that's fine for another PR (which I'm happy to write).

PR #855, opened 2020-12-06
TimingModels look like dicts of parameters

aarchiba:
They support .keys(), len(), in, iteration, the usual. In particular you can now get at parameters in a loop without needing getattr:

```python
for k, v in model.items():
    v.frozen = k not in ['F0', 'F1']
```
Likewise `model[k]` works, `k in model`, and so on. This does make somewhat redundant a number of model properties and methods, but it provides a convenient and familiar way to iterate over or manipulate TimingModel objects.


aarchiba:
Ready for review. Just a quality-of-life change.

PR #859, opened 2020-12-07
Deprecate chi2_reduced

### LINKED ISSUE #857: chi2_reduced and reduced_chi2 are both defined
The Residuals class has both defined, whereas CombinedResiduals just has reduced_chi2 defined.  The get_summary function in fitter.py asks for chi2_reduced.

aarchiba:
Fixes #857


aarchiba:
This is ready to go.

Issue #713, opened 2020-05-22
the correct way to get_PSR_freq


luojing1211:
Currently get_PSR_freq(modelF0=True) only returns F0  in the model, however, should we get the F0 at TOA? F0(TOA) = F0 + dt * F1 +.... 
Secondly, this check seems duplicated with the Spindown component's validation. 

https://github.com/nanograv/PINT/blob/6fbbd924ff6b9a1c455ffd1351286a75442fe130/src/pint/residuals.py#L135-L156

paulray:
When we need precise frequency at a TOA (e.g. in zima.py), I use this:
```
F_local = m.d_phase_d_toa(ts)
```
I seem to recall testing that in residuals but it makes it really slow. It is an outstanding problem that F0 is used for frequency in some places like this. It is often a sufficient approximation, but not always. I timed a pulsar that changed in period by a factor of two in a couple years.

paulray:
See this discussion: https://github.com/nanograv/PINT/issues/415

luojing1211:
Yeah, I remember that. Should we at least provide a Taylor expansion version of F0 at TOA? Plus, the check should use the spindown validation.

paulray:
Sure, those changes would be fine with me.

scottransom:
Could he have it as an option to use the full `F_local = m.d_phase_d_toa(ts)` but use the quick normal version by default?

paulray:
That sounds fine.  For option 3, it will need to be passed a TOAs object as an argument.
I think the functionality of getting the pulsar freq is used elsewhere in PINT, so we should make them all use this one function, with its three options depending on their needs for speed and/or accuracy.

luojing1211:
Yes, for option 2, one will need a timestamp at least.
-Jing

paulray:
Copying this from a different discussion with @aarchiba :

> Ah, so there are four possible ways to handle phase -> time conversion:
> 
> - Use the full delay calculation (at what time? before or after delay corrections?)
> - Use the Taylor series but don't worry about delay/timescales
> - Use F0 from the model
> - Use some fiducial F0 that is not a fit parameter

We should make a PR to address this, after #861 is merged. For the full delay calculation, isn't the right time just the topocentric time of the TOA?  You want the observed apparent frequency of the pulsar to convert a phase offset to a time offset.

abhisrkckl:
Fixed in 
https://github.com/nanograv/PINT/blob/master/src/pint/residuals.py#L246

[...]

PR #861, opened 2020-12-07
Wideband improvements

### LINKED ISSUE #851: DMJUMP fitting is not working
In some cases, just freeing the DMJUMP parameters and iterating appears stable, in others it goes off the rails.  Perturbing the DMJUMP parameters is similar.  In all cases, the uncertainties on the DMJUMP and DMX parameters become very large and all nearly identical (out to 7 or 8 decimal places).  So something is out of whack.

aarchiba:
This is exploring whether we can test and improve PINT's handling of wideband TOAs. It is intended to add, if the appropriate flags are present, the columns `dm` and `dme` to TOAs objects, and eventually everything that works with wideband TOAs should use those.

Will hopefully fix #851 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 ### Removed
 - Removed Python 2.7 support from travis and tox testing suites and from requirements files
 - Removed "landscape" code checker since that package is no longer supported by its author
+- Removed scale_by_F0 and scaled_by_F0 as they did not do what they were intended to (PR #861)
 ### Fixed
 - Fixed bug in processing of PHASE commands in .tim file. They are now applied even if pulse numbers are not being used
 - Substantial speed increase in Residuals calculation due to removal of redundant phase calculation
@@ -35,6 +36,7 @@ modified after a fit and are displayed with a model's .par file output.
 - New observatories will no longer overwrite existing ones silently.  Will either raise ValueError or require overwrite=True
 - Large speed increase when using Ecliptic coordinates
 - Changed Residuals so that use_weighted_mean and subtract_mean are attributes set on initialization
+- WidebandDMResiduals now support access to their parts through .toa and .dm attributes (PR #861)
 
 ## [0.7.0] - 2020-05-27
 ### Changed
--- a/docs/examples/Wideband_TOA_walkthrough.py
+++ b/docs/examples/Wideband_TOA_walkthrough.py
@@ -76,7 +76,7 @@ type(fitter.resids)
 # #### If we look into the resids attribute, it has two independent Residual objects.
 
 # %%
-fitter.resids.residual_objs
+fitter.resids.toa, fitter.resids.dm
 
 # %% [markdown]
 # #### Each of them can be used independently
@@ -84,7 +84,7 @@ fitter.resids.residual_objs
 # * Time residual
 
 # %%
-time_resids = fitter.resids.residual_objs["toa"].time_resids
+time_resids = fitter.resids.toa.time_resids
 plt.errorbar(
     toas.get_mjds().value,
     time_resids.to_value(u.us),
@@ -96,23 +96,23 @@ plt.xlabel("MJD")
 
 # %%
 # Time RMS
-print(fitter.resids.residual_objs["toa"].rms_weighted())
-print(fitter.resids.residual_objs["toa"].chi2)
+print(fitter.resids.toa.rms_weighted())
+print(fitter.resids.toa.chi2)
 
 # %% [markdown]
 # * DM residual
 
 # %%
-dm_resids = fitter.resids.residual_objs["dm"].resids
-dm_error = fitter.resids.residual_objs["dm"].get_data_error()
+dm_resids = fitter.resids.dm.resids
+dm_error = fitter.resids.dm.get_data_error()
 plt.errorbar(toas.get_mjds().value, dm_resids.value, yerr=dm_error.value, fmt="x")
 plt.ylabel("pc/cm^3")
 plt.xlabel("MJD")
 
 # %%
 # DM RMS
-print(fitter.resids.residual_objs["dm"].rms_weighted())
-print(fitter.resids.residual_objs["dm"].chi2)
+print(fitter.resids.dm.rms_weighted())
+print(fitter.resids.dm.chi2)
 
 # %% [markdown]
 # #### However, in the combined residuals, one can access rms and chi2 as well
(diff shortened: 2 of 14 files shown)


aarchiba:
I added some substantial doc improvements, in particular I made sure that WidebandTOAFitter was documented at all, and I clarified how to use WidebandTOAResiduals.

aarchiba:
New documentation:

- Residuals: https://pulsar-pint.readthedocs.io/en/wideband_improvements/api/pint.residuals.html
- Fitter: https://pulsar-pint.readthedocs.io/en/wideband_improvements/api/pint.fitter.html

paulray:
The documentation of `scaled_by_F0` is wrong. That is NOT intended to say whether time or phase residuals are being computed!  It is a setting for how time residuals are computed from phases in `calc_time_resids()`. The default is to do it the way TEMPO/Tempo2 do it and convert from phase to time using model.F0. This is usually a good approximation for many pulsars. However, for some pulsars the current F can be a long way from F0 because of large frequency derivatives. In this case, the conversion should be done using `get_PSR_freq(modelF0=False)` which is MUCH slower but returns the local frequency using `d_phase_d_toa`.

aarchiba:
> The documentation of `scaled_by_F0` is wrong. That is NOT intended to say whether time or phase residuals are being computed! It is a setting for how time residuals are computed from phases in `calc_time_resids()`. The default is to do it the way TEMPO/Tempo2 do it and convert from phase to time using model.F0. This is usually a good approximation for many pulsars. However, for some pulsars the current F can be a long way from F0 because of large frequency derivatives. In this case, the conversion should be done using `get_PSR_freq(modelF0=False)` which is MUCH slower but returns the local frequency using `d_phase_d_toa`.

Good to know. But is that what the implementation does?

pennucci:
Quick questions, not sure where to put these:
--Shouldn't the last line in `test_wideband_residuals` be `assert len(r.dm.dm_data) <= len(wb_toas)`, not strictly less than?
--Why in `test_wideband_residuals_dof` is the parameter `"DMJUMP"` set free?  Shouldn't it be `"DMJUMP1"`?  As an aside, is there a way to recover the condition each JUMP/DMJUMP parameter corresponds to?  i.e., how would someone stumbling upon `DMJUMP1` know that it corresponds to the selection `-fe L-wide`?
--The assertion in the test that fails, `test_wideband_fit_dmjump`, should look for -10 not 10 (same as you have it in `test_wideband_fit_dmjump_all`).  I know it currently fails because of not all TOAs having DMs, but as a hack, if you give each TOA an arbitrary DM and a really, really large DM uncertainty, then the test passes (checking DMJUMP1 == -10 within some reasonable atol).  I know this is not what the test is after, but it's a sanity check that it should work.

aarchiba:
> --Shouldn't the last line in `test_wideband_residuals` be `assert len(r.dm.dm_data) <= len(wb_toas)`, not strictly less than?

In this case, by construction not all TOAs have DM data, and this is checking that that is reflected in the residual object.

> --Why in `test_wideband_residuals_dof` is the parameter `"DMJUMP"` set free? Shouldn't it be `"DMJUMP1"`? As an aside, is there a way to recover the condition each JUMP/DMJUMP parameter corresponds to? i.e., how would someone stumbling upon `DMJUMP1` know that it corresponds to the selection `-fe L-wide`?

Er. It should be DMJUMP1, and I should get an error for setting anything on DMJUMP. That's not good. I'm not sure how maskParameters work, but I think they have a `key_value` attribute that contains either `fe` or `L-wide`, and presumably another attribute that contains the other?

> --The assertion in the test that fails, `test_wideband_fit_dmjump`, should look for -10 not 10 (same as you have it in `test_wideband_fit_dmjump_all`). I know it currently fails because of not all TOAs having DMs, but as a hack, if you give each TOA an arbitrary DM and a really, really large DM uncertainty, then the test passes (checking DMJUMP1 == -10 within some reasonable atol). I know this is not what the test is after, but it's a sanity check that it should work.

Indeed, one of the later tests does use DMs on all TOAs and I had to set the result to -10. I should change that. (But getting the tolerances right is not trivial.)

pennucci:
Yeah, in my own tests I had made the same typo -- leaving out the 1 on DMJUMP and was none the wiser.  I was pulling my hair out trying to figure out why I wasn't getting the numbers I expected.  So yeah, I think that should throw an error.

aarchiba:
> Yeah, in my own tests I had made the same typo -- leaving out the 1 on DMJUMP and was none the wiser. I was pulling my hair out trying to figure out why I wasn't getting the numbers I expected. So yeah, I think that should throw an error.

Maybe file that as a bug? I'll try to fix it but the Parameter code is really hard to understand so it might take a while to figure that out.

pennucci:
In the process, just trying to figure out where in the code to point to.

paulray:
Ah, I see you got rid of all the occurrences of `scale_by_F0` and now there is a hardwired `m.F0` in the design matrix. 

I think the idea is that when you need a frequency to convert a phase to a time residual, you should call `get_PSR_freq()` which has a `modelF0` argument.  By default, this is True so it will just return the F0 from the model, but it already supports the mode of returning `d_phase_d_toa` to give the local frequency and #713 was planned to add another method that quickly evaluated the frequency taylor series instead of doing the full delay calculation.  Presumably supporting that will be held for a separate PR.

aarchiba:
Ah, so there are *four* possible ways to handle phase -> time conversion:

- Use the full delay calculation (at what time? before or after delay corrections?)
- Use the Taylor series but don't worry about delay/timescales
- Use F0 from the model
- Use some fiducial F0 that is not a fit parameter

Since the existing `scaled_by_F0` did none of this (it decided on whether certain quantities were in phase or time units), I took it out and decided that implementing any of those other options belonged in some future PR.

aarchiba:
Ready to merge, unless we want to delay this until we can correctly implement more sophisticated phase -> time conversion.

pennucci:
FWIW, I just used WidebandTOAFitter to fit DMJUMPs for a couple of new pulsars in the 15y data set that do not already have DMJUMPs measured / aligned templates, and it returned sensible enough values.  I was previously using 12.5-year results for testing.  Again, the uncertainties (for DMJUMPs and DMX) are completely bogus, but it looks to do what it is supposed to do.

aarchiba:
I think the DMJUMP uncertainties might be something we need to pursue in another PR.

luojing1211:
I think this is good to go. Since this has some API optimizing, we should make sure this is ready for next. The scale_F0 part should not keep up from doing the busying week analysis. So I would say merge it and let the user test it.

PR #864, opened 2020-12-08
Stop ddk sini

luojing1211:
This PR adds a failure when DDK model has sini input.


luojing1211:
I think this is ok to merge when the tests pass.

tests/test_ddk.py
```diff
@@ -97,6 +98,11 @@ def test_K96(self):
         for p in testp.keys():
             adf = self.modelJ1713.d_phase_d_param(self.toasJ1713, delay, p)
 
+    def test_SINI_input(self):
```
aarchiba, line 101: Can you add a test to see what happens when loading a par file?

luojing1211: updated the tests. It passes locally. Waiting for the online test.

tests/datafile/ddk_sini.par
```diff
@@ -0,0 +1,1333 @@
...
+PMLAMBDA            5.2671  1              0.0021
+PMBETA             -3.4428  1              0.0043
+PX                  0.8211  1              0.0258
+F0    218.8118437960826270  1  0.0000000000000988
+F1     -4.083888637248D-16  1  1.433249826455D-21
+PEPOCH        55391.000000
+START            53393.560
+FINISH           57387.663
+DM               15.917131
+DMX               6.000000
```
aarchiba, line 13: If this is just to test `SINI` you can zap all the DMX/ECORR values and have something nice and short. If `get_model` could read from StringIO, you could even just embed it in the test itself, so it was clear what was going on. (As it stands you have to write it to a temporary file; I do this in a few cases, see other PRs, but it's annoying.)

tests/test_ddk.py
```diff
@@ -97,6 +98,13 @@ def test_K96(self):
...
+        self.modelJ1713.SINI.value = 0.9
+        with pytest.raises(AttributeError):
+            self.modelJ1713.validate()
+        with pytest.raises(AttributeError):
```
aarchiba, line 105: This should be a separate test so we can see what fails, if they start failing.

src/pint/models/binary_ddk.py
```diff
@@ -99,3 +99,7 @@ def validate(self):
                 raise MissingParameter(
                     "DDK", "DDK model needs proper motion parameters."
                 )
+        if self.SINI.quantity is not None:
+            raise AttributeError(
+                "DDK model do not accept `SINI` as input. Please" " use `KIN` instead."
```
aarchiba, line 104: do not -> does not.

tests/test_ddk.py
```diff
@@ -97,6 +98,15 @@ def test_K96(self):
...
             adf = self.modelJ1713.d_phase_d_param(self.toasJ1713, delay, p)
 
+    def test_sini_from_value(self):
+        self.modelJ1713.SINI.value = 0.9
+        with pytest.raises(AttributeError):
+            self.modelJ1713.validate()
+
+    def test_sini_from_par(self):
+        with pytest.raises(AttributeError):
+            _ = mb.get_model("ddk_sini.par")
```
aarchiba, line 108: no need to assign to anything. Just call the function.

aarchiba:
Looks good to me.

paulray:
According to travis, this still needs blackening and test_model_manual.py is failing.

aarchiba:
The `test_model_manual` failure is probably because it checks that all `.par` files in our example distribution load okay, and this adds one that shouldn't load okay.

luojing1211:
Can we somehow exclude that .par file gets loading test?

aarchiba:
> Can we somehow exclude that .par file gets loading test?

You could go investigate `test_model_manual` if you wanted to address it that way, but why not use an ultra-simplified `.par` file in a triple-quoted string you can pass to `get_model`? I know it's annoying that `get_model` doesn't accept StringIO objects but you can write it in a temporary directory. This par file isn't useful for anything else, so it's not useful to have it floating around in our `datafiles` directory.

aarchiba:
#871 makes it a lot easier to use triple-quoted strings for par files and has an example of an ultra-simple par file created in this way.

luojing1211:
I will fix it when #871 is merged.

tests/test_ddk.py
```diff
@@ -97,6 +98,40 @@ def test_K96(self):
...
+            "E  0.074  1  0.001\n"
+            "T0 55388.836  1  0.0002\n"
+            "PB 67.825129  1  0.0001\n"
+            "OM 176.19845  1  0.0013\n"
+            "M2  0.283395  1  0.0104\n"
+            "KOM   83.100  1  1.800\n"
+            "SINI  0.8     1  0.562\n"
+            "K96  1"
+        )
+        par_name = os.path.join(datadir, "test_ddk_sini.par")
```
aarchiba, line 127: This writes files to the datadir. That's not a good idea. With #871 get_model can read from a StringIO object, so this is unnecessary: `mb.get_model(StringIO(temp_par_str))`

tests/test_ddk.py
```diff
@@ -97,6 +98,40 @@ def test_K96(self):
...
+        temp_par = open(par_name, "w")
```
aarchiba, line 128: `with` statements should always be used when writing to files:
```python
with open(par_name, "w") as f:
    f.write(temp_par_str)
```
This ensures they are always closed and the buffer flushed to disk even if exceptions occur.

src/pint/models/binary_ddk.py
```diff
@@ -99,3 +99,8 @@ def validate(self):
                 raise MissingParameter(
                     "DDK", "DDK model needs proper motion parameters."
                 )
+        if self.SINI.quantity is not None:
+            raise AttributeError(
```
aarchiba, line 103: This seems like a ValueError - the user has passed a bad `.par` file or tried to create an invalid model. More, because TimingModel objects use both properties and `__getattr__`, AttributeErrors behave strangely.

tests/test_ddk.py
```diff
@@ -97,6 +98,40 @@ def test_K96(self):
...
+        temp_par.write(temp_par_str)
+        temp_par.close()
+        with pytest.raises(AttributeError):
+            mb.get_model(par_name)
+        os.remove(par_name)
```
aarchiba, line 133: `pytest` provides a fixture `tmp_dir` that is a directory that will be deleted at the end of a test. If you're not inside a pytest test, 
```python
with TemporaryDirectory() as td:
    # Do things with files in td
# now td and all its contents are erased
```
In particular, if this test fails, the `os.remove` will never be executed.

tests/test_ddk.py
luojing1211: Changed to StringIO for fake parfile.

tests/test_ddk.py
luojing1211: This deleted.

src/pint/models/binary_ddk.py
luojing1211: Changed to ValueError.

tests/test_ddk.py
luojing1211: This is deleted.

tests/test_ddk.py
```diff
@@ -97,6 +99,35 @@ def test_K96(self):
         for p in testp.keys():
             adf = self.modelJ1713.d_phase_d_param(self.toasJ1713, delay, p)
 
+    def test_sini_from_value(self):
+        self.modelJ1713.SINI.value = 0.9
+        with pytest.raises(ValueError):
+            self.modelJ1713.validate()
+
+    def test_sini_from_par(self):
+        temp_par_str = (
```
aarchiba, line 108: Why not use a triple-quoted string here?

luojing1211: I consider the triple-quoted string is part of docstrings. Does this implementation has potential risk?

aarchiba: Triple quotes provide a way to include long strings of any kind without a need for `\n`, commas, numerous quotes, and so on. Using multiple short strings inside is just more awkward syntax for the same thing.

paulray: Yeah, triple quoted strings are nice, and easy to copy/paste into since they don't have all the extra cruft.

luojing1211: Just changed.

aarchiba:
Looks good to me.

PR #865, opened 2020-12-09
Improved support for orbiting observatories

kerrm:
This PR addresses a number of issues as well as re-factors some of the implementation for handling event-based data from orbiting observatories.

- It fixes the longstanding "2 mus" discrepancy  between, e.g., Fermi phases produced with PINT and those produced with the tempo2 plugin.  The origin is the "topocentric" term v_earth*r_obs/c^2.  This can be calculated directly by astropy, but it is very slow to do a coordinate transform from the inertial coordinates tabulated by spacecraft to the earth-fixed frame required by the SOFA TT to TDB routine.  Thus one implementation of TT to TDB in PINT invoked the SOFA routine with "location=None" and calculated the topocentric term externally.  Unfortunately, the astropy wrapper interpreted None as the origin in _geodetic_ coordinates, not _geocentric_ coordinates, which Paul spotted!  So the fix for us is to always pass a location (the Geocenter).  Once astropy has been fixed, we can pass None instead, which will be somewhat faster.  In the current implementation, it will be a one-line change, modulo version checking, which can be done once and only once.

- It removes the tt2tdb_method kwargs present in the Observatory hierarchy.  These were implemented to trace the "2mus" problem and are no longer needed.  Because topocentric and orbiting observatories still do their TT to TDB corrections differently, I have added an internal _get_TDBs method which topocentric or orbiting observatories implement in order to compute the correct time transfer.  The resulting code is much simpler and shorter.

- It defines a new SatelliteObs class, from which all currently supported (Fermi, NICER, NuSTAR, RXTE... hmm, I'm not sure if I remembered XMM...) observatories inherit.  Indeed, they are all the same except for NICER, which has some unique code to manage extrapolation of orbital files.  Accordingly, fermi_obs.py, nicer_obs.py, nustar_obs.py, and rxte_obs.py are deleted.  The resulting code is much simpler and shorter.  The new method to for instantiating one of these special observatories is in satellite_obs.py:

get_satellite_observatory("Fermi", ft2_file)
get_satellite_observatory("NICER", FPorbfile)
etc.

- Updated relevant scripts where necessary.

- Added new tests to verify the 2 mus error is fixed for Fermi by comparing to tempo2 explicitly, and to test topocentric phasing for NICER.  Coverage *should* be up.  I've additionally tweaked the tests to run more quickly.  Both test_fermiphase.py and test_photonphase.py now run in a total time of about 1min, down from >3.  Perhaps this will be helpful when they're run a million times for CI...


kerrm:
PS -- the lines of code changed is a dirty lie.  It's counting the new FITS file for the NICER test as +1000k, and black did it's usual changing of function signatures to many lines.  It's definitely shorter in the real world...

paulray:
The astropy fix (https://github.com/astropy/astropy/pull/11134) has now been merged. However, we can't rely on it until we set PINT's requirements to a minimum of whatever version this gets backported to.

paulray:
This is great!  Can you add a few lines to CHANGELOG?

paulray:
It is failing on the oldest supported versions, but now we are moving to astropy>=4.0, so I bet that will fix this failure.

kerrm:
The error message from travis-ci is pretty opaque.  Do you have an idea of what might be wrong?  It's just apparently failing the precision test.  So maybe it's supposed to (if those versions deliver bad results)?

matteobachetti:
@kerrm @paulray Travis is abandoning us. We need to find a new CI. Github Actions is really good, and the move is fast enough.

src/pint/event_toas.py
```diff
@@ -145,6 +152,10 @@ def load_fits_TOAs(
         Force this time system
     timeref : str, default None
         Forse this time reference
+    minmjd : float, default 0
```
aarchiba, line 155: This is really minor, it's not like we'll have satellite data from before 1858, but wouldn't -infinity be the natural default here?

src/pint/event_toas.py
```diff
@@ -183,6 +194,13 @@ def load_fits_TOAs(
     if weights is not None:
         new_kwargs["weights"] = weights
 
+    # mask out times/columns outside of mjd range
+    mjds_float = np.asarray([r[0] + r[1] for r in mjds])
+    idx = np.logical_and((mjds_float > minmjd), (mjds_float < maxmjd))
```
aarchiba, line 199: `&` would be clearer here, as well as putting them in order ((a<b) & (b<c)); sadly a<b<c doesn't work for arrays.

src/pint/event_toas.py
```diff
@@ -122,7 +122,14 @@ def _get_timeref(hdu):
...
 
 def load_fits_TOAs(
-    eventname, mission, weights=None, extension=None, timesys=None, timeref=None
+    eventname,
+    mission,
+    weights=None,
+    extension=None,
+    timesys=None,
+    timeref=None,
+    minmjd=0,
```
aarchiba, line 131: I suppose these exist to save loading time? (one could always select down the resulting TOAs) Might be worth mentioning in the docstring?

src/pint/fermi_toas.py
```diff
@@ -137,7 +158,7 @@ def load_Fermi_TOAs(
             weights = weights[idx]
 
     # limit the TOAs to ones in selected MJD range
-    mjds_float = np.array([r[0] + r[1] for r in mjds])
+    mjds_float = np.asarray([r[0] + r[1] for r in mjds])
```
aarchiba, line 161: Why this change? The input is always a list so you'll always wind up with a new array.

src/pint/observatory/special_locations.py
```diff
@@ -70,9 +69,7 @@ def __init__(
         self.bipm_version = bipm_version
         self._bipm_clock = None
 
-        super(SpecialLocation, self).__init__(
-            name, aliases=aliases, tt2tdb_mode=tt2tdb_mode
-        )
+        super(SpecialLocation, self).__init__(name, aliases=aliases)
```
aarchiba, line 72: This can just be `super().__init__(...)`.

src/pint/observatory/special_locations.py
```diff
@@ -196,14 +193,17 @@ def tempo2_code(self):
...
     def posvel(self, t, ephem):
         return objPosVel_wrt_SSB("earth", t, ephem)
 
 
-class SpacecraftObs(SpecialLocation):
-    """Observatory-derived class for a spacecraft observatory."""
+class T2SpacecraftObs(SpecialLocation):
+    """An observatory with position tabulated following Tempo2 convention.
+
+    In tempo2, it is possible to specify the GCRS position of the observatory via the -telx, -tely, and -telz flags in a TOA file.  This class is able to obtain its position in this way, i.e. by exaiming the flags in a TOA table.
```
aarchiba, line 205: It would be nice to word-wrap this.

paulray:
@matteobachetti yes, NANOGrav is setting up its own CI server that we'll be transitioning to.

src/pint/scripts/photonphase.py
```diff
@@ -212,9 +206,8 @@ def main(argv=None):
 
     # Compute model phase for each TOA
     iphss, phss = modelin.phase(ts, abs_phase=True)
-    # ensure all postive
-    negmask = phss < 0.0
-    phases = np.where(negmask, phss + 1.0, phss)
+    phases = phss.value
+    phases[phases < 0.0] += 1
```
aarchiba, line 210: is `phases` guaranteed to be between -1 and 1? `phases %= 1` should accomplish the same thing whether this is so or not.

kerrm: Yes, the PINT convention is that the fractional part is on [-0.5, 0.5), or possibly (-0.5,0.5].  I suppose mod is safer, but I hate to use it when a simple add will do!

src/pint/observatory/special_locations.py
kerrm: I thought black did this!  I'll fix it.

src/pint/fermi_toas.py
kerrm: To me, there's no downside to using asarray, even though they are indeed functionally identical here.

src/pint/event_toas.py
kerrm: Agree.  I left it as originally written but with another vote I'll change!

paulray: I wrote it originally because I didn't think & did an element-wise and.  If there is a better way than `logical_and`, that is fine with me.

src/pint/event_toas.py
kerrm: Enh...  I have a slight preference for point comparisons, and 0 (1858!) does seem OK.  I don't feel strongly about it, though.

src/pint/event_toas.py
kerrm: Yeah, most of those operators are overloaded to numpy functions, so ~, &, and | all work as expected.

You know what blew my mind the other day?  @ implements matrix multiplication on arrays (as opposed to casting to np.matrix and using *).  I had no idea...

src/pint/event_toas.py
kerrm: Yes, I moved the selection upstream to avoid instantiating TOA object we don't need -- it's very, very slow compared to an array operation.  I could add something to the docstring.

src/pint/observatory/special_locations.py
kerrm: Is that true for 2.7?

paulray: PINT only supports 3.6+

kerrm: I see this concern is OBE!

src/pint/scripts/photonphase.py
aarchiba: If it soothes your discomfort, floating-point operations are generally faster than branches and out-of-order memory access...

src/pint/event_toas.py
aarchiba: That was a many-year battle, the numpy folks trying to get that into mainline python. And good God it came in in 3.5, so we can use it in PINT.

aarchiba:
Looks good to me.

src/pint/scripts/photonphase.py
scottransom: I just did some crude timing with a big array and they are more similar than I would have thought:
```
In [22]: a = u.Quantity(np.random.standard_normal(1000000))                                                   

In [23]: %timeit "phases = a.value ; phases[phases < 0.0] += 1"                                               
8.14 ns ± 1.13 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)

In [24]: a = u.Quantity(np.random.standard_normal(1000000))                                                   

In [25]: %timeit "a %= 1"                                                                                     
9.38 ns ± 0.801 ns per loop (mean ± std. dev. of 7 runs, 100000000 loops each)
```

kerrm: Alright, mod it is!

scottransom: Yeah, I think an in-place mod is probably a bit nicer.  There are probably a few other places in PINT where that can be changed....

PR #867, opened 2020-12-09
Basic DMX Binning Code

hwahl16:
Notebook with some basic DMX binning code that will take into account the receiver and frequency ratio.

### DIFF
--- a/docs/examples/DMX_binning_example_code.ipynb
+++ b/docs/examples/DMX_binning_example_code.ipynb
@@ -1,0 +1,460 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "WARNING: AstropyDeprecationWarning: The private astropy._erfa module has been made into its own package, pyerfa, which is a dependency of astropy and can be imported directly using \"import erfa\" [astropy._erfa]\n"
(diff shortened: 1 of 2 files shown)


aarchiba:
I need to update the instructions, but notebooks need special handling in git. Specifically we use `jupytext` to strip out the outputs and store them in the `py:percent` format in git. `jupytext --set-formats ipynb,py:percent my.ipynb` should accomplish this. This way subtle output changes don't lead to horrible merge conflicts in git. You should also make the notebook visible by listing it under the tutorial examples, and check how everything looks with `make notebooks` and `make docs`.

paulray:
This doesn't look like something that is intended to be merged into PINT. I think it is an in-progress development notebook. A PR is one way to make it easy for people to look at your code and comment, but not the only way. For example you could push your development branch up to your own fork and then have people look at it and comment.

When your development has gotten to the stage that it might need to be merged in, the actual DMX binning routine should be a python script, not a notebook. It is fine (and good!) to have a notebook that demonstrates the functionality, but it should be done by calling PINT code. You don't want code in a notebook that others may need to use, since it won't be accessible to other scripts.

aarchiba:
Ah, okay. Unfortunately it's hard to read notebooks in PRs, so using jupytext to keep a `.py` version in sync would result in something easier for us to review. (Once you've told it to make the association it updates the `.py` every time you change the notebook, or vice versa.)

paulray:
Should this be closed?  I believe it is obsolete?

PR #871, opened 2020-12-09
Free params checking

### LINKED ISSUE #868: Setting free_params does not fail as it should
`model = get_model(parfile);
model.free_params = ['garbage_parameter']`,
does not fail with `ValueError: garbage_parameter is not recognized as a parameter or alias` as I think it should (parfile doesn't have H4),
and
`model = get_model(parfile);
model.free_params = ['H4']`,
does not fail with `Parameter(s) are familiar but not in the model: H4` as I think it should,
and, perhaps worst of all,
`model = get_model(parfile);
model.free_params = ['DMJUMP']`
does not fail at all!  It should fail in the same way as the first example, but the silence over the subtle typo (the parameter should be `DMJUMP1` or `DMJUMP2`, etc.) leaves the user none the wiser!

aarchiba:
Enhance checking done in `model.free_params` (and other uses of `parameter.aliases`).

Closes #868

### DIFF
--- a/src/pint/models/model_builder.py
+++ b/src/pint/models/model_builder.py
@@ -14,6 +14,7 @@ from pint.models.timing_model import (
     ignore_prefix,
     DEFAULT_ORDER,
 )
+from pint.models.parameter import maskParameter
 from pint.utils import PrefixError, interesting_lines, lines_of, split_prefixed_name
 
 __all__ = ["get_model"]
@@ -168,7 +169,7 @@ class ModelBuilder(object):
                     pre, idxstr, idxV = split_prefixed_name(p)
                     if pre in [par.prefix] + par.prefix_aliases:
                         prefixs[par.prefix].append(p)
-                except:  # FIXME: is this meant to catch KeyErrors?
+                except ValueError:  # FIXME: is this meant to catch KeyErrors?
                     continue
 
         return prefixs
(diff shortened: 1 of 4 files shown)


aarchiba:
Not sure whether it addresses everything in the bug.

aarchiba:
Okay, ready to merge and it fixes the bug.

src/pint/models/model_builder.py
```diff
@@ -283,8 +290,6 @@ def get_model(parfile):
 
     Parameters
     ----------
-    name : str
-        Name for the model.
     parfile : str
         The parfile name
```
paulray, line 294: Expand doc to say it accepts a file-like object?

aarchiba: Indeed,

src/pint/models/timing_model.py
```diff
@@ -322,11 +322,11 @@ def free_params(self):
...
+        params_true = {self.match_param_aliases(p) for p in params}
         for p in self.params:
-            getattr(self, p).frozen = p not in params
-            params.discard(p)
-        if params:
+            getattr(self, p).frozen = p not in params_true
+            params_true.discard(p)
+        if params_true:
             raise ValueError(
                 "Parameter(s) are familiar but not in the model: {}".format(params)
```
luojing1211, line 331: Is there any example for this case? Shouldn't the params_true be a subset of self.params?

aarchiba: I *think* this branch will never be reached, because aliases should not be translated for parameters that are not actually in the model. But in case someone should change the implementation of `match_param_aliases` (for example if aliases were tracked at the Component level instead of the Parameter level), I think it's good to have this here to ensure an exception is raised.

luojing1211: I agree to have a function in the component level is a good idea. But that can be a separate issue.

luojing1211:
I think this PR is good to go.

PR #874, opened 2020-12-10
Fitter error checking

aarchiba:
Verify and improve the error checking in WLSFitter (ideally others too) so that most cases of parameter degeneracy can be reported.

### DIFF
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -824,7 +824,8 @@ class PowellFitter(Fitter):
 
     def fit_toas(self, maxiter=20):
         # check that params of timing model have necessary components
-        self.model.maskPar_has_toas_check(self.toas)
+        self.model.validate()
+        self.model.validate_toas(self.toas)
         # Initial guesses are model params
         fitp = self.model.get_params_dict("free", "num")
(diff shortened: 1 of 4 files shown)


aarchiba:
Still to do: there's an xfail on DMX bins that don't contain anything.

aarchiba:
There's some numerical stability complications, so that in order to keep passing tempo2 compatibility tests I had to largely disable it for GLS-based fitters, including the Wideband. They will still detect parameters that don't do anything, and the PR also includes explicit checking of DMX bins and mask parameters. You can call `fit_toas` with `threshold=1e-14` to gain numerical stability and also be notified of degeneracies (in any Fitter).

aarchiba:
Please take a look, but I think this is ready to merge.

pennucci:
I just tried it on some data where the DMX ranges don't match to the data, and it certainly tells me so (listing all the DMX ranges...!).

aarchiba:
Automated code could catch the exception, which contains a list of the problematic DMX ranges (or mask parameters) and so could automatically freeze or remove them.

aarchiba:
Ready to go.

PR #875, opened 2020-12-13
Bump astropy requirement to 4.0 and deal with new erfa package

paulray:
Finally requires astropy>=4.0.
Also new astropy splits out erfa support into "pyerfa" package. This adds code to deal with that.


paulray:
Ready to merge

luojing1211:
Just get to know that Astropy 4.0 updates the leap second automatically. So it is good.

PR #877, opened 2020-12-14
Recent wideband API changes break fitter.get_summary() if residuals are WidebandTOAResiduals.

pennucci:
This is a somewhat inelegant, but simple solution to new wideband API changes in fitter.get_summary().  There's a check against fitter.residuals.__class__.__name__ and then two additions to the string output get changed if the residuals are WidebandTOAResiduals.  This change would be needed for upcoming processing of the 15yr dataset.

### DIFF
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -172,15 +172,15 @@ class Fitter(object):
         )
         if is_wideband:
             s += "Prefit TOA residuals Wrms = {}, Postfit TOA residuals Wrms = {}\n".format(
-                    self.resids_init.toa.rms_weighted(), self.resids.toa.rms_weighted()
-    )
+                self.resids_init.toa.rms_weighted(), self.resids.toa.rms_weighted()
+            )
             s += "Prefit DM residuals Wrms = {}, Postfit DM residuals Wrms = {}\n".format(
-                    self.resids_init.dm.rms_weighted(), self.resids.dm.rms_weighted()
-    )
+                self.resids_init.dm.rms_weighted(), self.resids.dm.rms_weighted()
+            )
         else:
             s += "Prefit residuals Wrms = {}, Postfit residuals Wrms = {}\n".format(
-                    self.resids_init.rms_weighted(), self.resids.rms_weighted()
-        )
+                self.resids_init.rms_weighted(), self.resids.rms_weighted()
+            )
         s += "Chisq = {:.3f} for {} d.o.f. for reduced Chisq of {:.3f}\n".format(
             self.resids.chi2, self.resids.dof, self.resids.reduced_chi2
         )
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -167,23 +167,13 @@ class Fitter(object):
             is_wideband = False
 
         # First, print fit quality metrics
-        s = "Fitted model using {} method with {} free parameters to {} TOAs\n".format(
-            self.method, len(self.model.free_params), self.toas.ntoas
-        )
+        s = f"Fitted model using {self.method} method with {len(self.model.free_params)} free parameters to {self.toas.ntoas} TOAs\n"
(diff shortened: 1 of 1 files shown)


luojing1211:
Is the PR pass your local tox tests? Since travis-ci takes too long to finish the tests, we will look into the local tox test this week.

pennucci:
I'm running tox now.  Just pushed the blackened version of the change.

pennucci:
tox 'passed':
`=== 648 passed, 20 skipped, 13 xfailed, 15 xpassed, 156 warnings in 1541.21s (0:25:41) ===`
and ended with
`congratulations :)`

luojing1211:
This PR is a small fix and passes the local tox test, should be able to merge. If there is anyone else who wants to double-check, please feel free to do so.

src/pint/fitter.py
```diff
@@ -160,13 +160,27 @@ def get_summary(self, nodmx=False):
         import uncertainties.umath as um
         from uncertainties import ufloat
 
+        # Check if Wideband or not
+        if self.resids.__class__.__name__ == "WidebandTOAResiduals":
```
aarchiba, line 164: Because the Fitter constructs the Residuals, only the WidebandTOAFitter will ever construct residuals that need this branch. A more natural approach would be to use the inheritance tree and the fact that methods in WidebandTOAFitter override the versions in Fitter. That would require breaking this method into smaller methods - one for each hunk where the wideband fitter differs from the Fitter. I'm not sure that this solution is a very good one.

src/pint/fitter.py
```diff
@@ -160,13 +160,27 @@ def get_summary(self, nodmx=False):
...
+
         # First, print fit quality metrics
         s = "Fitted model using {} method with {} free parameters to {} TOAs\n".format(
             self.method, len(self.model.free_params), self.toas.ntoas
         )
-        s += "Prefit residuals Wrms = {}, Postfit residuals Wrms = {}\n".format(
-            self.resids_init.rms_weighted(), self.resids.rms_weighted()
-        )
+        if is_wideband:
+            s += "Prefit TOA residuals Wrms = {}, Postfit TOA residuals Wrms = {}\n".format(
```
aarchiba, line 174: You can use f-strings instead of `.format()`:
```python
s += f"Prefit TOA residuals Wrms = {self.resids_init.toa.rms_weighted()}"
```

src/pint/fitter.py
```diff
@@ -348,13 +362,22 @@ def get_summary(self, nodmx=False):
...
-                        ecc.nominal_value,
-                        self.resids.rms_weighted(),
-                        self.toas.ntoas,
-                        outstring=True,
-                    )
+                    if is_wideband:
+                        s += pint.utils.ELL1_check(
+                            self.model.A1.quantity,
+                            ecc.nominal_value,
+                            self.resids.toa.rms_weighted(),
```
aarchiba, line 369: Does the `ELL1` condition really not depend on the DM residuals? It makes sense but I just wanted to check.

src/pint/fitter.py
pennucci: Ok; then we'll go with this simpler fix for now.

src/pint/fitter.py
pennucci: Well, not in a direct way.  The DM data will influence the TOA residuals (typically the TOA RMS will be larger), but I don't think the baseline ELL1_check needs to be changed.

src/pint/fitter.py
pennucci: Done!

pennucci:
Forgot about tox; doing that now.

pennucci:
tox:
`=== 652 passed, 20 skipped, 13 xfailed, 15 xpassed, 152 warnings in 1446.48s (0:24:06) ===`
and
`congratulations :)`
Should be good?

PR #889, opened 2020-12-15
Added ability to read TOAs using commands from timing model

scottransom:
This allows you to call get_TOAs() with a timing model as an optional keyword param so that it uses the EPHEM, CLOCK, and PLANET_SHAPIRO that the model uses.

There is also a convenience function that sits in models.model_builder that loads both the model and the TOAs in one shot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,8 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - Fixed bug in solar wind model that prevented fitting
 - Fix pintempo script so it will respect JUMPs in the TOA file.
 ### Added
-- Added a get_model_and_toas() function in model_builder to read both, including model-based commands affecting the TOAs
-- Added ability to load TOAs including relevant commands (e.g. EPHEM, CLOCK, PLANET_SHAPIRO) from a timing model in get_TOAs()
+- Added a get_model_and_toas() function in model_builder to read both, including model-based commands affecting the TOAs (PR #889)
+- Added ability to load TOAs including relevant commands (e.g. EPHEM, CLOCK, PLANET_SHAPIRO) from a timing model in get_TOAs() (PR #889)
 - Added metadata to observatory definition, to keep track of the data origin
 - Added other bipm???? files from TEMPO2
 - Added ability to find observatories in [astropy](https://github.com/astropy/astropy-data/blob/gh-pages/coordinates/sites.json) if not present in PINT
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - Fixed bug in solar wind model that prevented fitting
 - Fix pintempo script so it will respect JUMPs in the TOA file.
 ### Added
+- Added a get_model_and_toas() function in model_builder to read both, including model-based commands affecting the TOAs
+- Added ability to load TOAs including relevant commands (e.g. EPHEM, CLOCK, PLANET_SHAPIRO) from a timing model in get_TOAs()
 - Added metadata to observatory definition, to keep track of the data origin
 - Added other bipm???? files from TEMPO2
 - Added ability to find observatories in [astropy](https://github.com/astropy/astropy-data/blob/gh-pages/coordinates/sites.json) if not present in PINT
--- a/docs/examples/fit_NGC6440E.py
+++ b/docs/examples/fit_NGC6440E.py
@@ -1,9 +1,10 @@
 #! /usr/bin/env python
 """Demonstrate use of pint in a script."""
 from __future__ import print_function, division
+import pint.toa
 import pint.fitter
 import pint.residuals
-import pint.models.model_builder as mb
+import pint.models as mm
 
 # import matplotlib
 # matplotlib.use('TKAgg')
@@ -17,11 +18,11 @@ parfile = os.path.join(datadir, "NGC6440E.par")
 timfile = os.path.join(datadir, "NGC6440E.tim")
 
 # Read the timing model and the TOAs
-m, t = mb.get_model_and_toas(parfile, timfile)
+m, t = mm.get_model_and_toas(parfile, timfile)
 
 # If we wanted to do things separately we could do
 # Define the timing model
-# m = mb.get_model(parfile)
+# m = mm.get_model(parfile)
 # Read in the TOAs, overriding some things from the model
 # t = pint.toa.get_TOAs(timfile, model=m)
 
--- a/docs/examples/fit_NGC6440E.py
+++ b/docs/examples/fit_NGC6440E.py
@@ -1,8 +1,6 @@
 #! /usr/bin/env python
 """Demonstrate use of pint in a script."""
 from __future__ import print_function, division
-import pint.toa
-import pint.models
 import pint.fitter
 import pint.residuals
 import pint.models.model_builder as mb
@@ -18,11 +16,14 @@ datadir = os.path.dirname(os.path.abspath(str(__file__)))
 parfile = os.path.join(datadir, "NGC6440E.par")
 timfile = os.path.join(datadir, "NGC6440E.tim")
 
-# Define the timing model
-m = mb.get_model(parfile)
+# Read the timing model and the TOAs
+m, t = mb.get_model_and_toas(parfile, timfile)
 
-# Read in the TOAs
-t = pint.toa.get_TOAs(timfile)
+# If we wanted to do things separately we could do
+# Define the timing model
+# m = mb.get_model(parfile)
+# Read in the TOAs, overriding some things from the model
+# t = pint.toa.get_TOAs(timfile, model=m)
 
 # Examples of how to select some subsets of TOAs
 # These can be un-done using t.unselect()
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -42,14 +42,12 @@ ignore_params = set(
     [
         "START",
         "FINISH",
-        "CLK",
         "EPHVER",
         "UNITS",
         "TIMEEPH",
         "T2CMETHOD",
         "DILATEFREQ",
         "NTOA",
-        "CLOCK",
         "TRES",
         "TZRMJD",
         "TZRFRQ",
@@ -200,6 +198,10 @@ class TimingModel(object):
         )
         self.add_param_from_top(
             strParameter(name="EPHEM", description="Ephemeris to use"), ""
+        )
+        self.add_param_from_top(
+            strParameter(name="CLOCK", description="Timescale to use", aliases=["CLK"]),
+            "",
         )
         self.add_param_from_top(
             strParameter(name="UNITS", description="Units (TDB assumed)"), ""
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -155,7 +155,10 @@ def get_TOAs(
             ephem = model["EPHEM"].value
             log.info(f"Using EPHEM = {ephem} from the given model")
         if include_bipm is None and model["CLOCK"].value is not None:
-            if "BIPM" in model["CLOCK"].value:
+            if model["CLOCK"].value == "TT(TAI)":
+                include_bipm = False
+                log.info("Using CLOCK = TT(TAI), so setting include_bipm = False")
+            elif "BIPM" in model["CLOCK"].value:
                 clk = model["CLOCK"].value.strip(")").split("(")
                 if len(clk) == 2:
                     ctype, cvers = clk
@@ -1470,7 +1473,7 @@ class TOAs(object):
                 raise ValueError("Some TOAs have 'clkcorr' flag and some do not!")
         # An array of all the time corrections, one for each TOA
         log.info(
-            "Applying clock corrections (include_GPS = {0}, include_BIPM = {1})".format(
+            "Applying clock corrections (include_gps = {0}, include_bipm = {1})".format(
                 include_gps, include_bipm
             )
         )
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -169,7 +169,7 @@ def get_TOAs(
                             )
         if planets is None and model["PLANET_SHAPIRO"].value:
             planets = True
-            log.info(f"Using PLANET_SHAPIRO = True from the given model")
+            log.info("Using PLANET_SHAPIRO = True from the given model")
 
     updatepickle = False
     recalc = False
@@ -222,7 +222,7 @@ def get_TOAs(
         if t.ephem is not None:
             # If you read a .tim file using TOAs(), the ephem is None
             # and so no recalculation is needed, just calculation!
-            log.info(f"Ephem changed, recalculation needed")
+            log.info("Ephem changed, recalculation needed")
         recalc = True
         updatepickle = True
     if recalc or "tdb" not in t.table.colnames:
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -83,6 +83,7 @@ def get_TOAs(
     bipm_version=None,
     include_gps=None,
     planets=None,
+    model=None,
     usepickle=False,
     tdb_method="default",
 ):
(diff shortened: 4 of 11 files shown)


aarchiba:
Can you add tests that `.as_parfile` produces parfiles that contain the "new" parameters? Round-trips make good tests too.

scottransom:
Hmmm.....  So the CLOCK line *does* get put into as_parfile() now.  However, if it isn't a BIMP**** (which is all we use internally, I think!), then what gets written out is wrong.  Not sure how we want to handle this.... @luojing1211 @paulray ?

aarchiba:
> Hmmm..... So the CLOCK line _does_ get put into as_parfile() now. However, if it isn't a BIMP**** (which is all we use internally, I think!), then what gets written out is wrong. Not sure how we want to handle this.... @luojing1211 @paulray ?

What goes wrong with CLOCK? Don't we just write back out what we read in?

aarchiba:
A CHANGELOG.md entry would be good.

docs/examples/fit_NGC6440E.py
```diff
@@ -18,11 +16,14 @@
 parfile = os.path.join(datadir, "NGC6440E.par")
 timfile = os.path.join(datadir, "NGC6440E.tim")
 
-# Define the timing model
-m = mb.get_model(parfile)
+# Read the timing model and the TOAs
+m, t = mb.get_model_and_toas(parfile, timfile)
```
aarchiba, line 20: It's worth changing the notebooks too, to work this way. If you don't ever use notebooks, you can just make the changes to the `.py` files, they're what's in git anyway, then just doing `make notebooks`.

src/pint/toa.py
```diff
@@ -143,6 +150,27 @@ def get_TOAs(
         Completed TOAs object representing the data.
 
     """
+    if model:
+        # If the keyword args are set, override what is in the model
+        if ephem is None and model["EPHEM"].value is not None:
+            ephem = model["EPHEM"].value
+            log.info(f"Using EPHEM = {ephem} from the given model")
+        if include_bipm is None and model["CLOCK"].value is not None:
+            if "BIPM" in model["CLOCK"].value:
```
aarchiba, line 159: Is it worth emitting a warning if something went wrong - if `CLOCK` is set but we don't know what to do with it?

tests/test_toa_reader.py
```diff
@@ -52,3 +77,27 @@ def test_jump_3(self):
...
+    def test_ephem(self):
+        assert self.x.ephem == "DE421"
+
+    def test_planets(self):
+        assert self.x.planets is False
+
+    def test_clock(self):
+        assert self.x.clock_corr_info["bipm_version"] == bipm_default
+
+    def test_model_override1(self):
```
aarchiba, line 90: What does the "override" here mean? It would be good to have tests to check that values specified as keyword arguments to `get_TOAs` override values obtained from the model, but that's not what this checks.

tests/test_toa_reader.py
```diff
@@ -52,3 +77,27 @@ def test_jump_3(self):
...
+        parstr = StringIO(simplepar)
+        m = get_model(parstr)
+        y = toa.get_TOAs("test1.tim", model=m)
+        assert y.ephem == "DE436"
+        assert y.planets is True
+        assert y.clock_corr_info["bipm_version"] == "BIPM2017"
+
+    def test_model_override2(self):
```
aarchiba, line 98: Leaving these functions inside TestTOAReader means that the setup function, which loads some TOAs, is done for each of them. Just unindenting them so they're bare test functions will get them run without the class-based setup. (Classes are only needed for `unittest`.)

docs/examples/fit_NGC6440E.py
scottransom: yeah, I agree.  Will try to get that in tomorrow.

src/pint/toa.py
scottransom: Yeah, that's a good idea.  I'll do that.  Right now, if a CLOCK line is not TT(BPIMXXXX) we basically don't do anything with it.

tests/test_toa_reader.py
scottransom: The model is overriding the default behavior of the TOA reading code.  But I can also check the overriding of the override!

tests/test_toa_reader.py
scottransom: Ah.  Didn't know that.  Cool.

scottransom:
> What goes wrong with CLOCK? Don't we just write back out what we read in?

Yeah, that's what we do.  But it isn't what we use internally.  So if we read in a file with UTC(NIST) (which we ignore, since by default we use BIPM2015) and then do a fit, the resulting post-fit parfile *should* have TT(BIPM2015), but it actually gets UTC(NIST) again.

aarchiba:
> > What goes wrong with CLOCK? Don't we just write back out what we read in?
> 
> Yeah, that's what we do. But it isn't what we use internally. So if we read in a file with UTC(NIST) (which we ignore, since by default we use BIPM2015) and then do a fit, the resulting post-fit parfile _should_ have TT(BIPM2015), but it actually gets UTC(NIST) again.

This could tie in to the question of what we do with unfamiliar CLOCK settings. Or maybe more useful, when a fitter makes a new model, it could set CLOCK (and other things) to what was actually used on the TOAs, just like it sets START and FINISH.

docs/examples/fit_NGC6440E.py
scottransom: Do we want to use the new function in all of the notebooks?  Or just tutorial-style ones?

aarchiba: Might as well switch over completely unless it's doing something like load TOAs with no model in sight.

scottransom:
I think that I've handled all of the issues.  There is still the issue (probably not to be resolved in this PR) about CLOCK getting written out with the wrong value if a fit is done.

aarchiba:
Maybe file a bug for the CLOCK thing?

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - Fixed bug in solar wind model that prevented fitting
 - Fix pintempo script so it will respect JUMPs in the TOA file.
 ### Added
+- Added a get_model_and_toas() function in model_builder to read both, including model-based commands affecting the TOAs
```
aarchiba, line 22: It's nice to have "PR #889" next to these

src/pint/toa.py
```diff
@@ -126,12 +127,18 @@ def get_TOAs(
...
         Whether to include the GPS clock correction. Defaults to True.
     planets : bool or None
         Whether to apply Shapiro delays based on planet positions. Note that a
         long-standing TEMPO2 bug in this feature went unnoticed for years.
         Defaults to False.
+    model : A valid PINT timing model or None
+        If a valid timing model is passed, model commands (such as BIPM version,
+        planet shapiro delay, and solar system ephemeris) that affect TOA loading
+        are applied.
+        Defaults to False.
```
aarchiba, line 141: Maybe drop "Defaults to" here; definitely not False.

aarchiba:
I'm okay to merge this with these two changes.

luojing1211:
Are we going to handle the TT(TAI), which means no BIMP correction in this PR?

src/pint/toa.py
```diff
@@ -143,6 +150,27 @@ def get_TOAs(
         Completed TOAs object representing the data.
 
     """
+    if model:
+        # If the keyword args are set, override what is in the model
+        if ephem is None and model["EPHEM"].value is not None:
```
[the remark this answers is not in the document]
scottransom: For get_TOAs(), we use the default EPHEM (DE421) if nothing is specified in the kwargs.  If a model is specified, we use the EPHEM from the model, *unless* the kwarg `ephem` is set, then that takes precedence over everything.

src/pint/models/timing_model.py
```diff
@@ -201,6 +199,10 @@ def __init__(self, name="", components=[]):
         self.add_param_from_top(
             strParameter(name="EPHEM", description="Ephemeris to use"), ""
         )
+        self.add_param_from_top(
+            strParameter(name="CLOCK", description="Timescale to use", aliases=["CLK"]),
```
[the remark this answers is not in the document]
scottransom: By default, if we don't specify any kind of clock, PINT will TT(BIPM), with the default BIPM being BIPM2015 (as of right now).    If we explicitly set `include_bipm` to False, do we get TT(TAI)?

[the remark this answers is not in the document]
luojing1211: Yes. UTC -> TT is implmented as TT =  UTC + constant + (bimp_correction if using bipm)

scottransom:
@luojing1211 As long as you have `include_bipm` set to False, you do not get BIPM corrections.  However, you also do not get a warning that you are not using the `CLOCK` specified in a model if that is present.  If we still want that warning, let me know.

src/pint/models/timing_model.py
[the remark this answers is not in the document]
scottransom: Are you asking if we should specify the allowed values for CLOCK?  If so, where would we put that?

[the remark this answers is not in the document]
luojing1211: The acceptable value should include `TT(TAI)`, which turns off the BIPM correction. It should be near line 177

scottransom:
Is it worth putting:
```
from pint.toa import get_TOAs
```
in the top level `pint.__init__.py` file?

That would allow:
```
> import pint
> m, t = pint.get_model_and_toas("test.par", "test.tim")
```
rather than having to know that you need to import those from `pint.models` or `pint.toa`.

src/pint/models/timing_model.py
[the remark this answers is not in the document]
scottransom: Yes, as I mentioned in another comment, as log as `include_bipm` is False, that works and you get TT(TAI).  So I think it is OK.

aarchiba:
> Is it worth putting:
> 
> ```
> from pint.models.model_builder import get_model, get_model_and_toas
> from pint.toa import get_TOAs
> ```
> 
> in the top level `pint.__init__.py` file?
> 
> That would allow:
> 
> ```
> > import pint
> > m, t = pint.get_model_and_toas("test.par", "test.tim")
> ```
> 
> rather than having to know that you need to import those from `pint.models` or `pint.toa`.

There's a risk of circular import disaster. I'm not completely sure whether `pint` gets imported on the way to importing `pint.whatever` but `from module import thing` doesn't work unless `module` can be completely loaded before the current one. 

Having things appear in the wrong places can also make the documentation more confusing - for example in the documentation you find TimingModel under `pint.models.timing_model.TimingModel` but the examples reference it as `pint.models.TimingModel` - if you go to the documentation for `pint.models`, it's not there.

aarchiba:
> @luojing1211 As long as you have `include_bipm` set to False, you do not get BIPM corrections. However, you also do not get a warning that you are not using the `CLOCK` specified in a model if that is present. If we still want that warning, let me know.

Can you add a test case with `CLOCK TT(TAI)` and check that the right thing happens?

scottransom:
> > Is it worth putting:
> > ```
> > from pint.models.model_builder import get_model, get_model_and_toas
> > from pint.toa import get_TOAs
> > ```
> > 
> > 
> > in the top level `pint.__init__.py` file?
> > That would allow:
> > ```
> > > import pint
> > > m, t = pint.get_model_and_toas("test.par", "test.tim")
> > ```
> > 
> > 
> > rather than having to know that you need to import those from `pint.models` or `pint.toa`.
> 
> There's a risk of circular import disaster. I'm not completely sure whether `pint` gets imported on the way to importing `pint.whatever` but `from module import thing` doesn't work unless `module` can be completely loaded before the current one.
> 
> Having things appear in the wrong places can also make the documentation more confusing - for example in the documentation you find TimingModel under `pint.models.timing_model.TimingModel` but the examples reference it as `pint.models.TimingModel` - if you go to the documentation for `pint.models`, it's not there.

Currently `get_model()` and `get_model_and_toas()` are both defined in `pint.models.model_builder` which is a bit buried.  However, `get_model()` was also imported into `pint.models`, so I did the same thing for `get_model_and_toas()`.  I was just wondering whether it would be better to move them to the top of the stack since they are likely to be used a lot by our users.

If there is a real problem with circular import issues, then we should figure out if we should keep them available from `pint.models`.  Or maybe there is a better place to put them overall?

paulray:
I would love to have `pint.get_model_and_toas()`

scottransom:
> > @luojing1211 As long as you have `include_bipm` set to False, you do not get BIPM corrections. However, you also do not get a warning that you are not using the `CLOCK` specified in a model if that is present. If we still want that warning, let me know.
> 
> Can you add a test case with `CLOCK TT(TAI)` and check that the right thing happens?

It doesn't work correctly now.  It overrides TT(TAI) with TT(BIPM2015).  I'll fix that an make a test case.

scottransom:
> I would love to have `pint.get_model_and_toas()`

I think I would too.  I'll test it to see if it causes circular reference issues.  Maybe I could remove `get_model()` and `get_model_and_toas()` from `pint.models` then.

paulray:
I you remove `pint.models.get_model()` it will break a LOT of scripts!

scottransom:
There are definitely circular reference issues if I put those functions in `pint.__init__.py`....

scottransom:
OK.  I think this is probably good to go now.

swiggumj:
@scottransom I might want to touch base with you tomorrow to get a high level overview of how this works so I can implement appropriate changes in timing_analysis once this is merged.

scottransom:
@swiggumj  Sure thing.  But as long as we are planning on the parfiles having all the important stuff (including good CLOCK and EPHEM lines, PLANET_SHAPIRO = Y and CORRECT_TROPOSPHERE = Y all you should have to do is use:

```
```
And everything should just work.  Note that that means that the ephem and clock stuff that are in the config files would *not* change or override things.  If you wanted that, you would want to split things into a `get_model()` call and a `get_TOAs()` call where the model is passed, but you override the CLOCK and EPHEM options.

aarchiba:
> @swiggumj Sure thing. But as long as we are planning on the parfiles having all the important stuff (including good CLOCK and EPHEM lines, PLANET_SHAPIRO = Y and CORRECT_TROPOSPHERE = Y all you should have to do is use:
> 
> ```
> from pint.models import get_model_and_toas
> m, t = get_model_and_toas(parfile, timfile)
> ```
> 
> And everything should just work. Note that that means that the ephem and clock stuff that are in the config files would _not_ change or override things. If you wanted that, you would want to split things into a `get_model()` call and a `get_TOAs()` call where the model is passed, but you override the CLOCK and EPHEM options.

Any reason not to just add the optional arguments to `get_model_and_toas`? All it has to do is pass them downward to `get_TOAs`.

scottransom:
> > And everything should just work. Note that that means that the ephem and clock stuff that are in the config files would _not_ change or override things. If you wanted that, you would want to split things into a `get_model()` call and a `get_TOAs()` call where the model is passed, but you override the CLOCK and EPHEM options.
> 
> Any reason not to just add the optional arguments to `get_model_and_toas`? All it has to do is pass them downward to `get_TOAs`.

Yeah, I was thinking about exactly that this morning.  Is it possible just to put `**kwargs` and just pass them as-is (and then to mention to check the get_TOAs() docs for what is valid?).  Or do we need to explicitly list the options?

scottransom:
I just checked that and it works just fine (and barfs if you pass a keyword arg that isn't in get_TOAs().  So would you be OK with that?  I kinda like it (and we don't need to copy arguments from get_TOAs()).

aarchiba:
Using `**kwargs` does work, and it's very common in matplotlib, but as you may have noticed it makes the documentation harder to read and it defeats tab completion. So when it's just a few options I prefer to specify them explicitly. (This is another advantage of using `None` as the default value - you don't end up with `BIPM2015` baked into callers like this.) See for example Astropy's `download_files_in_parallel` which hands a bunch of its options down to `download_file`: https://docs.astropy.org/en/stable/api/astropy.utils.data.download_files_in_parallel.html#astropy.utils.data.download_files_in_parallel

scottransom:
Hmm.  Those are good points.  Not sure if I think they outweigh the code duplication, but for this particular function, which will probably be used by people on the command line, maybe tab-completion wins out.  I'll do that.

scottransom:
Let's merge this!

Issue #879, opened 2020-12-14
JUMP tel does not recognize alternative representations


aarchiba:
`JUMP tel @ 0` does not match TOAs at the barycenter; you have to use `JUMP tel barycenter 0`.

PR #890, opened 2020-12-15
Fix noise models so scaled uncertainties make sense

aarchiba:
Adds doc improvements and changes to noise models and mask parameters.

Closes #888
Closes #880

PR #891, opened 2020-12-16
Fix mask parameter flags

luojing1211:
This PR fixes the maskParameter key, key_value problems.
Here is the change:
* The flag type of keys are required to have leading '-' 
* The key_value are parsed to the right type. If not one of the pre-defined types, it will be an one element list of string
* The telescope key accepts aliases.


src/pint/models/parameter.py
```diff
@@ -1348,26 +1349,35 @@ def __init__(
...
+                key_info = self.key_identifier[key.lower()]
+                if len(key_value) != key_info[1]:
+                    errmsg = f"key {key} takes {key_info[1]} element(s)."
+                    raise ValueError(errmsg)
+                key_value_parser = key_info[0]
+            else:
+                if not key.startswith("-"):
+                    raise ValueError(
+                        "A key/flag(excpet for 'mjd', 'freq', "
+                        "'name', 'tel') needs a leading '-'."
```
paulray, line 1376: Perhaps "A JUMP keyed to a TOA flag requires a leading '-'. Legal JUMP keywords that don't require a leading '-' are MJD, FREQ, NAME, TEL."

paulray:
We should check whether Tempo2 allows JUMPs on TOA flags without the leading "-", since we might want to retain backwards compatibility?

aarchiba:
This will have conflicts with #890 where I made some minor changes, particularly to the docs, while I was trying to get my test cases working.

src/pint/models/parameter.py
```diff
@@ -1581,19 +1593,20 @@ def select_toa_mask(self, toas):
         # get the table columns
         # TODO Right now it is only supports mjd, freq, tel, and flagkeys,
         # We need to consider some more complicated situation
+        # THis will have a problem if the key has '-' in it.
         key = self.key.replace("-", "")
         tbl = toas.table
         if (
-            key.lower() not in column_match.keys()
+            self.key.lower() not in column_match.keys()
```
aarchiba, line 1600: `in` on dictionaries does the same thing as on their keys; you can leave out the `.keys()` here.

tests/test_mask_parameter.py
```diff
@@ -0,0 +1,111 @@
...
+
+from pint.models.model_builder import get_model
+from pint.models.parameter import maskParameter
+from pint.toa import get_TOAs
+from pinttestdata import datadir
+
+import copy
+
+
+class TestParameters(unittest.TestCase):
```
aarchiba, line 19: Please use the simpler `pytest` syntax: just make each test a function. For the ones that need those TOAs loaded, you can set up a fixture. The current arrangement re-loads the TOAs for every test whether they use it or not. If you're not modifying the TOAs, you can make the fixture module-level and then the TOAs will be loaded once and then used for all tests.

tests/test_mask_parameter.py
```diff
@@ -0,0 +1,111 @@
...
+        assert mp_special_obs1.key_value == ["barycenter"]
+        mp_special_obs2 = maskParameter("test2", key="tel", key_value=0)
+        assert mp_special_obs2.key_value == ["geocenter"]
+        with pytest.raises(ValueError):
+            mp_wrong_keyvalue = maskParameter(
+                "test2", key="tel", key_value=["gbt", "ao"]
+            )
+
+    def test_name_mask(self):
+        # Not sure about the use case for name mask
```
aarchiba, line 77: `name` is often the input file the TOAs were generated from. One could reasonably select all the TOAs that were generated from a single file.

tests/test_mask_parameter.py
```diff
@@ -0,0 +1,111 @@
...
+        os.chdir(datadir)
+        self.toas = get_TOAs("J1713+0747_NANOGrav_11yv0_short.tim")
+
+    def test_mjd_mask(self):
+        mp = maskParameter("test1", key="mjd", key_value=[54000, 54100])
+        assert mp.key == "mjd"
+        assert mp.key_value == [54000, 54100]
+        assert mp.value == None
+        select_toas = mp.select_toa_mask(self.toas)
+        assert np.all(self.toas.table["mjd_float"][select_toas] <= 54100)
```
aarchiba, line 31: Can you check equality - that the set selected by `select_toa_mask` is the same as the set of TOAs with the right frequencies? You might find a utility function useful here, since that's what one wants to check for almost all these selections.

src/pint/models/parameter.py
```diff
@@ -1581,19 +1593,20 @@ def select_toa_mask(self, toas):
         # get the table columns
         # TODO Right now it is only supports mjd, freq, tel, and flagkeys,
         # We need to consider some more complicated situation
+        # THis will have a problem if the key has '-' in it.
```
aarchiba, line 1596: What does this mean about how keys should  be specified? Is `JUMP -tel` allowed as an alternative form of `JUMP tel`? Should it be?

aarchiba:
Closes #879

src/pint/models/parameter.py
luojing1211: I am a little bit hesitant to put the word "JUMP" here. Since this is in the maskParameter, which is not only for JUMPs. Is this for keys general enough?

paulray: Oh, I see. Well I'm just trying to make it as clear as possible for some user who loads a par file and wants to be clearly told why it is failing

paulray: Do all maskParameters respect MJD, FREQ, etc.?

luojing1211: I am not sure if anyone used ECORR EFAC like this. Those are quite general, so I believe one can.

aarchiba: In PINT, yes, all EFAC/EQUAD/JUMP/etc. use the same kinds of selection criteria.

src/pint/models/parameter.py
```diff
@@ -1581,19 +1593,20 @@ def select_toa_mask(self, toas):
         # get the table columns
         # TODO Right now it is only supports mjd, freq, tel, and flagkeys,
         # We need to consider some more complicated situation
+        # THis will have a problem if the key has '-' in the middle of the key.
```
aarchiba, line 1596: This is a bug! It's easy enough to remove it only at the beginning:
```python
if self.key.startswith("-"):
    key = self.key[1:]
else:
    key = self.key
```
for example.

tests/test_mask_parameter.py
luojing1211: Can you explain more about this? I am confused about the set of TOAs with the right frequencies.

tests/test_mask_parameter.py
luojing1211: Ok, got it, can make some more test on it.

tests/test_mask_parameter.py
aarchiba: So `select_toa_mask` returns a list of TOA indices that match the mask. Right now you check that all the TOAs returned match the mask criterion - but that would pass if it returned no TOAs. Why not use `np.where` to get a list of TOAs that match the criterion, and compare it to the list returned by `select_toa_mask`? They should be exactly equal.

src/pint/models/parameter.py
luojing1211: This is fixed in the update

tests/test_mask_parameter.py
luojing1211: This is fixed in the update.

tests/test_mask_parameter.py
luojing1211: The return of `select_toa_mask` returns the indices of selection, so I compare the indices between the result and `np.where()` return instead.

src/pint/models/parameter.py
luojing1211: Changed.

src/pint/models/parameter.py
aarchiba: I'm not sure what you mean?

tests/test_mask_parameter.py
luojing1211: Test added.

src/pint/models/parameter.py
```diff
@@ -1581,19 +1594,24 @@ def select_toa_mask(self, toas):
         # get the table columns
         # TODO Right now it is only supports mjd, freq, tel, and flagkeys,
         # We need to consider some more complicated situation
-        key = self.key.replace("-", "")
+        # THis will have a problem if the key has '-' in the middle of the key.
```
aarchiba, line 1597: This comment can go now.

aarchiba:
Looks good to me.

luojing1211:
I think this is good to merge. @scottransom Can you merge this?

PR #892, opened 2020-12-16
Implement model.sun_angle(toas)

aarchiba:
This adds a method to any timing model that contains Astrometry that allows it to compute the sun angle at each TOA.

This could be used to implement solar wind models, but for now just gives a convenient sun angle.

### DIFF
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -64,7 +64,7 @@ class Astrometry(DelayComponent):
         return self.coords_as_ECL(epoch=epoch).cartesian.xyz.transpose()
 
     def sun_angle(self, toas):
-        """Compute the pulsar-Earth-Sun angle.
+        """Compute the pulsar-observatory-Sun angle.
 
         This angle takes into account the motion of the Sun around the solar system barycenter.
         """
@@ -73,7 +73,7 @@ class Astrometry(DelayComponent):
         obs_vec = tbl["ssb_obs_pos"].quantity
         # FIXME: sun pos may not have been computed
         sun_vec = tbl["obs_sun_pos"].quantity
-        psr_vec = self._parent.ssb_to_psb_xyz_ICRS(epoch=tbl["tdb"])
+        psr_vec = self.ssb_to_psb_xyz_ICRS(epoch=tbl["tdb"])
         osv = sun_vec - obs_vec
         osv /= ((osv ** 2).sum(axis=1) ** 0.5)[:, None]
         cos = (osv * psr_vec).sum(axis=1)
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -66,6 +66,9 @@ class Astrometry(DelayComponent):
     def sun_angle(self, toas):
         """Compute the pulsar-observatory-Sun angle.
 
+        This is the angle between the center of the Sun and the direction to
+        the pulsar, as seen from the observatory (for each TOA).
+
         This angle takes into account the motion of the Sun around the solar system barycenter.
         """
         tbl = toas.table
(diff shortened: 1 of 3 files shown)


src/pint/models/astrometry.py
```diff
@@ -69,6 +63,22 @@ def ssb_to_psb_xyz_ECL(self, epoch=None):
         # TODO: would it be better for this to return a 6-vector (pos, vel)?
         return self.coords_as_ECL(epoch=epoch).cartesian.xyz.transpose()
 
+    def sun_angle(self, toas):
+        """Compute the pulsar-Earth-Sun angle.
```
paulray, line 67: Technically this is the pulsar-observatory-Sun angle. Some of our observatories are not on the Earth.
Also, maybe add a clarification to the documentation saying that this is "the angle from the center of the Sun to the pulsar direction, as seen from the observatory"

src/pint/models/astrometry.py
```diff
@@ -69,6 +63,22 @@ def ssb_to_psb_xyz_ECL(self, epoch=None):
...
+
+        This angle takes into account the motion of the Sun around the solar system barycenter.
+        """
+        tbl = toas.table
+
+        obs_vec = tbl["ssb_obs_pos"].quantity
+        # FIXME: sun pos may not have been computed
+        sun_vec = tbl["obs_sun_pos"].quantity
+        psr_vec = self._parent.ssb_to_psb_xyz_ICRS(epoch=tbl["tdb"])
```
luojing1211, line 76: `ssb_to_psb_xyz_ICRS` is in the class. Any reason to use it from the `._parent`?

src/pint/models/astrometry.py
```diff
@@ -80,15 +90,14 @@ def solar_system_geometric_delay(self, toas, acc_delay=None):
...
-        L_hat = self.ssb_to_psb_xyz_ICRS(epoch=tbl["tdbld"].astype(numpy.float64))
-        re_dot_L = numpy.sum(tbl["ssb_obs_pos"] * L_hat, axis=1)
+        L_hat = self.ssb_to_psb_xyz_ICRS(epoch=tbl["tdbld"].astype(np.float64))
+        re_dot_L = np.sum(tbl["ssb_obs_pos"] * L_hat, axis=1)
         delay = -re_dot_L.to(ls).value
-        if self.PX.value != 0.0 and numpy.count_nonzero(tbl["ssb_obs_pos"]) > 0:
+        if self.PX.value != 0.0 and np.count_nonzero(tbl["ssb_obs_pos"]) > 0:
             L = (1.0 / self.PX.value) * u.kpc
-            # TODO: numpy.sum currently loses units in some cases...
+            # TODO: np.sum currently loses units in some cases...
```
paulray, line 98: Is this still true? You can't safely use units with np.sum?

src/pint/models/solar_wind_dispersion.py
```diff
@@ -23,42 +23,41 @@ class SolarWindDispersion(Dispersion):
...
+                description="Solar Wind density at 1 AU",
             )
         )
         self.add_param(
             floatParameter(
-                name="SWM", value=0.0, units="", description="Solar Wind Model"
+                name="SWM",
+                value=0.0,
+                units="",
+                description="Solar Wind Model (0 is the only option)",
```
paulray, line 47: How about "Solar Wind Model Selection (0 = Edwards et al. 2006, more to come)"

src/pint/models/astrometry.py
aarchiba: I'm pretty sure astropy fixed this with 4.0. (line is only here because I cleaned up numpy -> np)

src/pint/models/solar_wind_dispersion.py
luojing1211: Is `SWM` a new type of solar wind model other than the spherical one? If that is the case, I would say make it a new component. Of if it is an alias, we should just add it to the parameter alias.

src/pint/models/astrometry.py
aarchiba: Also I rewrite this whole function in the actual solar wind PR #894

PR #895, opened 2020-12-16
Catch TypeError when SINI is missing (i.e., None)

rossjjennings:
With ELL1 binary models, SINI is often `None`, which doesn't get caught by the `try` block that's supposed to catch missing SINI values in `fitter.get_summary()` since it produces a `TypeError` rather than a `ValueError`, creating a mysterious error message for whoever called `get_summary()`. This fixes that by catching `TypeError` in addition to `ValueError`.


aarchiba:
Looks good to me.

Issue #893, opened 2020-12-16
sun_angle and solar wind need better testing


aarchiba:
PR #892 introduces a sun_angle calculation but the testing is basic; it should be compared to some known and trusted values, that specifically include the sun's offset from the solar system barycenter. Likewise we should have some way to test solar wind models built on that.

paulray:
It looks like the definition of the scale factor is off between Tempo2 and PINT by a factor of 2. When I use `zima` to create TOAs with SOLARN0 10 and then fit them with Tempo2, it gets a value of 5.0.  But the shape and location of the cusp is fine, so I think the angle calculations are correct.

paulray:
Spoke too soon. There is a small difference in the solar angle computed by PINT and Tempo2.

![sunanglediff](https://user-images.githubusercontent.com/1771428/102406668-eb59df00-3fb8-11eb-8afc-79affad0d82f.png)

paulray:
Something is funky with the solar wind and PINT.  I create a set of TOAs like this:
`zima sw.par sw.tim --ntoa 365 --duration 365 --error 0.05 --plot`

Then I try to fit with PINT and it does not get 0 residuals, but has a strange wiggle.

This is the par file (rename to sw.par):
[sw.txt](https://github.com/nanograv/PINT/files/5715826/sw.txt)

Making this more difficult for me is that something in all the new changes has broken `pintempo` and `pintk` so they are giving nonsense.

scottransom:
I'm going to try and fix the pintk (and maybe pintempo) issues today.

aarchiba:
Be aware of #906 which fixed subtle bugs with solar angle

paulray:
Possibly relevant paper: https://arxiv.org/abs/2012.11726

aarchiba:
I had a student who looked at solar wind modelling, and indeed it's much messier than our usual assumptions. But surely at least we can know where the Sun is and what this model predicts...

PR #898, opened 2020-12-17
Solar wind/sun angle improvements

aarchiba:
This is meant to address #893 . One improvement(?) is that users can request a `sun_angle` computation using the barycenter instead of the center of the actual Sun.


aarchiba:
The calculation is now correct, or at least, agrees with TEMPO. Still no really good tests; see #893 .

aarchiba:
Ready to merge if tests can come later.

paulray:
Cool, I'll merge and run my TEMPO comparisons

PR #900, opened 2020-12-17
Fitter now sets informational parameters

### LINKED ISSUE #841: .tim files cannot record full information required to load them repeatably
The full TOAs object obtained from a `.tim` file depends on the value of `ephem` passed to `get_TOAs`; similarly several clock correction parameters. NANOGrav folks would like to be able to encapsulate these options in some canonical way, in the interest of replicability. The relevant options could be recorded as TOA flags, if those TOA flags were respected upon loading of `.tim` files. One would have to think through what to do with files for which, say, `ephem` was specified only for some TOAs, or for which conflicting values of `ephem` were specified.

### LINKED ISSUE #886: Backwards-compatibility for writing tempo parfiles label:"feature request"
What I'm thinking of is something that complements `model.as_parfile()`, which might be `model.as_tempo_parfile()`.  Or if a parfile-writing utility exists in PINT, then something like `write_parfile(model, format='tempo')`.  This would require adding the `CLK`, `TIMEEPH`, `T2CMETHOD`, and (probably) `INFO -f` lines, and if the fitter is `WidebandTOAFitter` (or, perhaps, if there is `DMJUMP` in the model), adding `DMDATA 1`.  There might be one or two other necessary lines to add (for tempo, not tempo2) that I've forgotten.  The would greatly facilitate sanity cross-checks with tempo (especially for wideband fitting) where appropriate.

### LINKED ISSUE #896: output parfiles can contain bogus CLOCK values
Now that CLOCK is recognized as a valid parameter, it will get written out (for example, `.as_parfile()`).  However, PINT only internally uses (I believe) TT(TAI) if we specify `include_bipm` = False or a TT(BIPMXXXX) if we specify the BIPM version (where XXXX is a year).

However, if we fit using TT(BIPMXXXX), the CLOCK parameter value does not get updated to TT(BIPMXXXX)

This is related to #886

aarchiba:
In an effort to record useful information and be compatible with
tempo/tempo2, Fitter objects now record properties of the TOAs and fit
result in the models they produce. These then naturally get written out
to par files.

Closes #896 
Closes #886 
Closes #841


src/pint/models/timing_model.py
```diff
@@ -228,6 +221,62 @@ def __init__(self, name="", components=[]):
...
+            strParameter(
+                name="TIMEEPH",
+                description="Time ephemeris to use for TDB conversion; for PINT, always FB90",
+            ),
+            "",
+        )
+        self.add_param_from_top(
+            strParameter(
+                name="T2CMETHOD",
+                description="Something to do with time conversion; for PINT, always TEMPO",
```
pennucci, line 241: "Method for transforming from terrestrial to celestial frame (IAU2000B/TEMPO)."

src/pint/models/timing_model.py
```diff
@@ -228,6 +221,62 @@ def __init__(self, name="", components=[]):
         self.add_param_from_top(
             MJDParameter(name="FINISH", description="End MJD for fitting"), ""
         )
+        self.add_param_from_top(
+            strParameter(
+                name="INFO",
+                description="Tells TEMPO to write some extra information about frontend/backend combinations; -f is recommended",
```
pennucci, line 227: The argument to INFO could be any TOA flag, -f is what we use for fe/be combos.  But you could just specify -fe or even -gof or -snr.

src/pint/models/timing_model.py
```diff
@@ -228,6 +221,62 @@ def __init__(self, name="", components=[]):
...
+                name="T2CMETHOD",
+                description="Something to do with time conversion; for PINT, always TEMPO",
+            ),
+            "",
+        )
+        self.add_param_from_top(
+            boolParameter(
+                name="DILATEFREQ",
+                value=False,
+                description="Tells TEMPO2 o do something with frequencies; PINT always acts like it is N",
```
pennucci, line 249: "Whether or not to apply gravitational redshift and time dila-tion to observing frequency (Y/N)."

pennucci: In tempo2.

src/pint/models/timing_model.py
paulray: Is PINT really TEMPO? Or is it IAU2000B?

aarchiba: Ah, thanks. And I don't know how PINT does it, exactly, but I agree IAU2000B sounds more plausible.

src/pint/models/timing_model.py
aarchiba: INFO will not be written to the output par file unless it is set, either in the input par file or on the model object, to some value. So the NANOGrav scripts will have to put INFO -f.

src/pint/models/timing_model.py
aarchiba: Aha thanks!

aarchiba:
All tests pass, I think this can be merged.

PR #901, opened 2020-12-17
Add neptune

paulray:
Add neptune position to TOAs table and to Solar System Shapiro calculation.


src/pint/toa.py
```diff
@@ -1638,7 +1638,7 @@ def compute_posvels(self, ephem=None, planets=False):
             if c in self.table.colnames:
                 log.info("Column {0} already exists. Removing...".format(c))
                 self.table.remove_column(c)
-        for p in ("jupiter", "saturn", "venus", "uranus"):
+        for p in ("jupiter", "saturn", "venus", "uranus", "neptune"):
```
aarchiba, line 1641: Can this tuple go up top in a variable so it doesn't have to be repeated all these times?

aarchiba:
Sounds like the Enterprise folks want all the planets?

paulray:
There will be a storage and time cost to adding all the planets, but I guess it will only be paid if PLANET_SHAPIRO is true?

aarchiba:
Sounds like they're not very interested.

PR #904, opened 2020-12-17
WIP: find and resolve DMJUMP uncertainty anomalies

aarchiba:
@pennucci reports some fits involving `DMJUMP` parameters wind up with improbably large uncertainties on them. The intention here is to resolve these.

### DIFF
--- a/tests/test_wideband_dm_data.py
+++ b/tests/test_wideband_dm_data.py
@@ -14,21 +14,26 @@ from numpy.testing import assert_allclose
 
 from pint.models import get_model
 from pint.residuals import Residuals, WidebandTOAResiduals
-from pint.toa import get_TOAs
+from pint.toa import get_TOAs, make_fake_toas
 from pint.fitter import WidebandTOAFitter
 
 os.chdir(datadir)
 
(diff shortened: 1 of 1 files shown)


aarchiba:
Unfortunately to date I haven't been able to reproduce them - see test case.

pennucci:
I checked the test case, and I see it passes.  But if you use the `tim_all` TOAs and the `par,` it 'fails', even though the `assert_allclose` is still reasonably close.  Still, in the examples I've seen, I wonder if it has something to do with DMX and/or regular JUMPs.

pennucci:
#933 is in process of fixing this in part.

luojing1211:
Since the DMJUMP is fixed, can we close this PR?

aarchiba:
Do we now have test cases that verify that it is fixed? Have we solved all the problems?

luojing1211:
We have some test case about the DMJUMP derivative, which leads to this problem.

aarchiba:
I think this may be fixed by #1047 ? In any case we can revive it if a bug is filed.

Issue #689, opened 2020-05-06
Need a function to combine two TOAs objects


paulray:
Say I have two `TOAs` objects, perhaps read from two different .tim files.
If I want to do a joint fit, I need to combine them into one object (or we need to make `Fitter` objects support multiple `TOAs` objects). At the moment, it is not clear how to do this safely, given that `TOAs` objects have properties like solar system ephemerides and clock corrections applied, etc. that may not be easily combinable. 

Currently, I'm just putting them in two .tim files then making an all.tim with include statements and reading them in again. But, it would be good to have a smart TOAs combining function that made sure they were compatible and merged the TOA tables.

paulray:
Joe Swiggum also asked about doing this.

paulray:
Interestingly I just found that inside mcmc_fitter.py, there is a `concat_toas()` function. This could form the basis of this, but probably needs some confirmation that the resulting TOAs object is all consistent.

PR #908, opened 2020-12-18
Added ability to merge TOAs instances

scottransom:
This merges a list of compatible TOAs instances.  It implements #689


src/pint/toa.py
```diff
@@ -1917,3 +1920,73 @@ def read_toa_file(self, filename, process_includes=True, top=True):
...
+
+    Returns
+    -------
+    A new TOAs instance with all the combined and grouped TOAs
+    """
+    # Check each TOA object for consistency
+    ephems = [tt.ephem for tt in TOAs_list]
+    if len(set(ephems)) > 1:
+        log.warning(f"EPHEMs are inconsistent in merge_TOAs(): {ephems}\nNot merging.")
+        return None
```
aarchiba, line 1946: I think this is a clear case for an exception - the function is unable to do the specific thing you called it to do.

src/pint/toa.py
```diff
@@ -1917,3 +1920,73 @@ def read_toa_file(self, filename, process_includes=True, top=True):
...
+    num_cols = [len(tt.table.columns) for tt in TOAs_list]
+    if len(set(num_cols)) > 1:
+        log.warning(
+            f"Numbers of table columns are inconsistent in merge_TOAs(): {num_cols}\nNot merging."
+        )
+        return None
+    # Use a copy of the first TOAs instance as the base for the joined object
+    nt = copy.deepcopy(TOAs_list[0])
+    nt.filename = [tt.filename for tt in TOAs_list]
```
aarchiba, line 1979: Can this merge TOAs objects that have lists of filenames?

tests/test_toa_reader.py
```diff
@@ -132,3 +132,17 @@ def test_tttai():
...
+
+
+def test_toa_merge():
+    filenames = ["NGC6440E.tim", "testtimes.tim", "parkes.toa"]
+    toas = [toa.get_TOAs(ff) for ff in filenames]
+    nt = toa.merge_TOAs(toas)
+    assert len(nt.observatories) == 3
+    assert nt.table.meta["filename"] == filenames
+    assert nt.ntoas == 78
+    # now test a failure if ephems are different
```
aarchiba, line 144: This could be made less dependent on details of the tim files by checking the length against the sum of the lengths.

tests/test_toa_reader.py
```diff
@@ -132,3 +132,17 @@ def test_tttai():
...
+    toas[0].ephem = "DE436"
+    nt = toa.merge_TOAs(toas)
+    assert nt is None
+
```
aarchiba, line 148: It works be good to have tests that check whether this refuses illegal merges, and what happens if you successively merge multiple files one by one. You could also do a more exactly comparison between this and a TOAs merged using INCLUDE statements.

src/pint/toa.py
scottransom: Fair enough.  TypeError?

src/pint/toa.py
scottransom: Good point.  I think that will be rare, but I'll implement that today.

tests/test_toa_reader.py
scottransom: I'm checking one type of the illegal merges.  But the syntax for all the others is exactly the same, so I'm not sure it is worth testing them all.  Consecutive merges now work and are tested.  Comparing closely with INCLUDEs is non-trivial given the way that commands are handled.  But INCLUDEs re handled better now (filenames of included files go in the .filename list) and they merge just fine (and that is tested).

scottransom:
I think I've handled all @aarchiba 's comments.  If others are OK, it is probably ready to be merged.

PR #899, opened 2020-12-17
add_DMX_range function (also small model.compare update)

almcewen0:
**WORK IN PROGRESS**

NOTE: this pull request has ben moved to https://github.com/nanograv/PINT/pull/910.

Created a wrapper function to add DMX components to a dispersion model. Avoids overwriting DMX indices. 

Also, I accidentally committed a few very minor changes to model.compare. I was going to remove them from this, but since they are really minor (and likely to be merged right away anyway), I thought I'd just include them here.

PR #910, opened 2020-12-18
Adding add_DMX_range and remove_DMX_range functions (from MR #899)

almcewen0:
The commit stream got all messed up on the previous MR (https://github.com/nanograv/PINT/pull/899), so I reset and opened a new one. The old MR is not closed for the time being for comments, but can be removed when this is merged.

### DIFF
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -418,17 +418,19 @@ class DispersionDMX(Dispersion):
         index : float, int, list, np.ndarray
             Number or list/array of numbers corresponding to DMX indices to be removed from model. 
         """
-        
-        if isinstance(index,int) or isinstance(index,float):
-            indices=[index]
-        elif not isinstance(index,list) or not isinstance(index,np.ndarray):
-            raise TypeError(f"index must be a float, int, list, or array - not {type(index)}")
+
+        if isinstance(index, int) or isinstance(index, float):
+            indices = [index]
+        elif not isinstance(index, list) or not isinstance(index, np.ndarray):
+            raise TypeError(
+                f"index must be a float, int, list, or array - not {type(index)}"
+            )
         for index in indices:
             index_rf = f"{int(index):04d}"
             for prefix in ["DMX_", "DMXR1_", "DMXR2_"]:
                 self.remove_param(prefix + index_rf)
         self.validate()
-    
+
     def get_indices(self):
         """Returns an array of integers corresponding to DMX parameters. 
 
@@ -437,10 +439,10 @@ class DispersionDMX(Dispersion):
         inds : np.ndarray
         Array of DMX indices in model.
         """
-        inds=[]
+        inds = []
         for p in self.params:
-            if 'DMX_' in p:
-                inds.append(int(p.split('_')[-1]))
+            if "DMX_" in p:
+                inds.append(int(p.split("_")[-1]))
         return np.array(inds)
 
     def setup(self):
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -319,42 +319,78 @@ class DispersionDMX(Dispersion):
                 description="Dispersion measure",
             )
         )
+
+        self.add_DMX_range(None, None, dmx=0, frozen=False, index=1)
+
+        self.dm_value_funcs += [self.dmx_dm]
+        self.set_special_params(["DMX_0001", "DMXR1_0001", "DMXR2_0001"])
+        self.delay_funcs_component += [self.DMX_dispersion_delay]
+
+    def add_DMX_range(self, mjd_start, mjd_end, index=None, dmx=0, frozen=True):
+        #### Setting up the DMX title convention. If index is None, want to increment the current max DMX index by 1.
+        if index is None:
+            index = self.get_prefix_mapping_component("DMX_").max() + 1
+        i = f"{int(index):04d}"
+
+        if mjd_end is not None and mjd_start is not None:
+            if mjd_end < mjd_start:
+                raise ValueError("Starting MJD is greater than ending MJD.")
+        elif mjd_start != mjd_end:
+            raise ValueError("Only one MJD bound is set.")
+
+        #### Check to see if the user-input index is already assigned to a different DMX component.
+        if int(index) in self.get_prefix_mapping_component("DMX_"):
+            raise ValueError(
+                "Index '%s' is already in use in this model. Please choose another."
+                % index
+            )
+
+        #### Add three DMX parameters (start MJD, end MJD, DMX value)
         self.add_param(
             prefixParameter(
-                name="DMX_0001",
+                name="DMX_" + i,
                 units="pc cm^-3",
-                value=0.0,
+                value=dmx,
                 unit_template=lambda x: "pc cm^-3",
                 description="Dispersion measure variation",
                 description_template=lambda x: "Dispersion measure",
-                paramter_type="float",
+                parameter_type="float",
+                frozen=frozen,
             )
         )
         self.add_param(
             prefixParameter(
-                name="DMXR1_0001",
+                name="DMXR1_" + i,
                 units="MJD",
                 unit_template=lambda x: "MJD",
                 description="Beginning of DMX interval",
                 description_template=lambda x: "Beginning of DMX interval",
                 parameter_type="MJD",
                 time_scale="utc",
+                value=mjd_start,
             )
         )
         self.add_param(
             prefixParameter(
-                name="DMXR2_0001",
+                name="DMXR2_" + i,
                 units="MJD",
                 unit_template=lambda x: "MJD",
                 description="End of DMX interval",
                 description_template=lambda x: "End of DMX interval",
                 parameter_type="MJD",
                 time_scale="utc",
+                value=mjd_end,
             )
         )
-        self.dm_value_funcs += [self.dmx_dm]
-        self.set_special_params(["DMX_0001", "DMXR1_0001", "DMXR2_0001"])
-        self.delay_funcs_component += [self.DMX_dispersion_delay]
+        self.setup()
+        self.validate()
+
+    def remove_DMX_range(self, index):
+        """Removes all three DMX parameters associated with a given index."""
+        index = "{:04d}".format(index)
+        for prefix in ["DMX_", "DMXR1_", "DMXR2_"]:
+            self.remove_param(prefix + index)
+        self.validate()
 
     def setup(self):
         super(DispersionDMX, self).setup()
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -428,6 +428,20 @@ class DispersionDMX(Dispersion):
             for prefix in ["DMX_", "DMXR1_", "DMXR2_"]:
                 self.remove_param(prefix + index_rf)
         self.validate()
+    
+    def get_indices(self):
+        """Returns an array of integers corresponding to DMX parameters. 
+
+        Returns
+        -------
+        inds : np.ndarray
+        Array of DMX indices in model.
+        """
+        inds=[]
+        for p in self.params:
+            if 'DMX_' in p:
+                inds.append(int(p.split('_')[-1]))
+        return np.array(inds)
 
     def setup(self):
         super(DispersionDMX, self).setup()
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -333,8 +333,8 @@ class DispersionDMX(Dispersion):
 
         #### Setting up the DMX title convention. If index is None, want to increment the current max DMX index by 1.
         if index is None:
-            dct = self.get_prefix_mapping_component('DMX_') 
-            index = np.max(list(dct.keys()))+1 
+            dct = self.get_prefix_mapping_component("DMX_")
+            index = np.max(list(dct.keys())) + 1
         i = f"{int(index):04d}"
 
         if mjd_end is not None and mjd_start is not None:
@@ -343,14 +343,12 @@ class DispersionDMX(Dispersion):
         elif mjd_start != mjd_end:
             raise ValueError("Only one MJD bound is set.")
 
-        #### Check to see if the user-input index is already assigned to a different DMX component.
         if int(index) in self.get_prefix_mapping_component("DMX_"):
             raise ValueError(
                 "Index '%s' is already in use in this model. Please choose another."
                 % index
             )
 
-        #### Add three DMX parameters (start MJD, end MJD, DMX value)
         self.add_param(
             prefixParameter(
                 name="DMX_" + i,
@@ -393,7 +391,7 @@ class DispersionDMX(Dispersion):
 
     def remove_DMX_range(self, index):
         """Removes all three DMX parameters associated with a given index."""
-        index = "{:04d}".format(index)
+        index = f"{int(index):04d}"
         for prefix in ["DMX_", "DMXR1_", "DMXR2_"]:
             self.remove_param(prefix + index)
         self.validate()
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -419,7 +419,7 @@ class DispersionDMX(Dispersion):
             Number or list/array of numbers corresponding to DMX indices to be removed from model. 
         """
 
-        if isinstance(index, int) or isinstance(index, float):
+        if isinstance(index, int) or isinstance(index, float) or isinstance(index,np.int64):
             indices = [index]
         elif not isinstance(index, list) or not isinstance(index, np.ndarray):
             raise TypeError(
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -410,7 +410,7 @@ class DispersionDMX(Dispersion):
         return index
 
     def remove_DMX_range(self, index):
-        """Removes all DMX parameters associated with a given index/list of indices."""
+        """Removes all DMX parameters associated with a given index/list of indices.
 
         Parameters
         ----------
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -327,6 +327,10 @@ class DispersionDMX(Dispersion):
         self.delay_funcs_component += [self.DMX_dispersion_delay]
 
     def add_DMX_range(self, mjd_start, mjd_end, index=None, dmx=0, frozen=True):
+        """ Add DMX range to a dispersion model with specified start/end MJDs and DMX. 
+            returns: index [int]
+        """
+
         #### Setting up the DMX title convention. If index is None, want to increment the current max DMX index by 1.
         if index is None:
             index = self.get_prefix_mapping_component("DMX_").max() + 1
@@ -384,6 +388,7 @@ class DispersionDMX(Dispersion):
         )
         self.setup()
         self.validate()
+        return index
 
     def remove_DMX_range(self, index):
         """Removes all three DMX parameters associated with a given index."""
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -333,7 +333,8 @@ class DispersionDMX(Dispersion):
 
         #### Setting up the DMX title convention. If index is None, want to increment the current max DMX index by 1.
         if index is None:
-            index = self.get_prefix_mapping_component("DMX_").max() + 1
+            dct = self.get_prefix_mapping_component('DMX_') 
+            index = np.max(list(dct.keys()))+1 
         i = f"{int(index):04d}"
 
         if mjd_end is not None and mjd_start is not None:
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -328,7 +328,27 @@ class DispersionDMX(Dispersion):
 
     def add_DMX_range(self, mjd_start, mjd_end, index=None, dmx=0, frozen=True):
         """ Add DMX range to a dispersion model with specified start/end MJDs and DMX. 
-            returns: index [int]
+
+        Parameters
+        ----------
+
+        mjd_start : float
+            MJD for beginning of DMX event.
+        mjd_end : float
+            MJD for end of DMX event.
+        index : int, None
+            Integer label for DMX event. If None, will increment largest used index by 1. 
+        dmx : float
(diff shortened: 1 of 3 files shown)


paulray:
Should this be merged?

aarchiba:
I think there are a number of unaddressed comments on the previous version.

almcewen0:
> I think there are a number of unaddressed comments on the previous version.

I think I have actually addressed them all now - forgot to reply to your other comments. just did, though.

src/pint/models/dispersion_model.py
```diff
@@ -319,42 +319,83 @@ def __init__(self):
...
         )
+
+        self.add_DMX_range(None, None, dmx=0, frozen=False, index=1)
+
+        self.dm_value_funcs += [self.dmx_dm]
+        self.set_special_params(["DMX_0001", "DMXR1_0001", "DMXR2_0001"])
+        self.delay_funcs_component += [self.DMX_dispersion_delay]
+
+    def add_DMX_range(self, mjd_start, mjd_end, index=None, dmx=0, frozen=True):
+        """ Add DMX range to a dispersion model with specified start/end MJDs and DMX.
```
aarchiba, line 330: This is not the format we use for docstrings. Use `make docs` to check how Sphinx renders what you have written. We use `napoleon`'s NumPy-style docstrings: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy

src/pint/models/timing_model.py
```diff
@@ -1706,17 +1706,23 @@ def compare(self, othermodel, nodmx=True, threshold_sigma=3.0, verbosity="max"):
                             newstr += " {:28f}".format(otherpar.value)
                         if otherpar.value != par.value:
                             sys.stdout.flush()
-                            log.warning(
-                                "Parameter %s not fit, but has changed between these models"
-                                % par.name
-                            )
+                            if par.name == "START" or par.name == "FINISH":
```
aarchiba, line 1709: Are these the only parameters that change without being fit? Fitter objects now set a number of parameters based on the results of the fit (e.g. EPHEM, NTOA, CHI2)

src/pint/models/timing_model.py
```diff
@@ -2223,7 +2229,7 @@ def remove_param(self, param):
             param_name = param.name
         if param_name not in self.params:
             raise ValueError(
-                "Tried to remove parameter {} but it is not listed: {}".formmat(
+                "Tried to remove parameter {} but it is not listed: {}".format(
```
aarchiba, line 2232: If you're changing this might as well use f-strings.

src/pint/models/dispersion_model.py
```diff
@@ -319,42 +319,83 @@ def __init__(self):
...
+        """ Add DMX range to a dispersion model with specified start/end MJDs and DMX. 
+            returns: index [int]
+        """
+
+        #### Setting up the DMX title convention. If index is None, want to increment the current max DMX index by 1.
+        if index is None:
+            index = self.get_prefix_mapping_component("DMX_").max() + 1
+        i = f"{int(index):04d}"
+
+        if mjd_end is not None and mjd_start is not None:
```
aarchiba, line 339: Do you want to allow users to add extra None/None DMX ranges?

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,142 @@
...
+        dm_mod.add_DMX_range(58200, 58300, 3, 1, frozen=False)
+
+
+def test_remove_nonexistent_index():
+    """ Check for error when a unused DMX index is removed. """
+    dm_mod = DispersionDMX()
+    with pytest.raises(ValueError):
+        dm_mod.remove_DMX_range(3)
+    with pytest.raises(ValueError):
+        dm_mod.add_DMX_range(None, 58000, 3, 1, frozen=False)
```
aarchiba, line 39: What is this testing? Why does it raise a ValueError?

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,142 @@
...
+
+def test_duplicate_index():
+    """ Check for error when a duplicate DMX index is used. """
+    dm_mod = DispersionDMX()
+    dm_mod.add_DMX_range(58000, 58100, 3, 1, frozen=False)
+    with pytest.raises(ValueError):
+        dm_mod.add_DMX_range(58200, 58300, 3, 1, frozen=False)
+
+
+def test_remove_nonexistent_index():
```
aarchiba, line 33: This function name is repeated so this test is never run.

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,142 @@
...
+def test_only_one_MJD():
+    """ Check for error when one of the MJDs is None and the other isn't. """
+    dm_mod = DispersionDMX()
+    with pytest.raises(ValueError):
+        dm_mod.add_DMX_range(None, 58000, 3, 1, frozen=False)
+    with pytest.raises(ValueError):
+        dm_mod.add_DMX_range(58000, None, 3, 1, frozen=False)
+
+
+def test_duplicate_index():
```
aarchiba, line 25: This function name is also repeated.

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def test_remove_nonexistent_index():
+    """ Check for error when a unused DMX index is removed. """
+    dm_mod = DispersionDMX()
+    with pytest.raises(ValueError):
+        dm_mod.remove_DMX_range(3)
+
+
+def test_add_DMX():
```
aarchiba, line 59: Perhaps also test dynamic index allocation? That is, don't specify the index and ensure that everything works?

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,142 @@
...
+    """ Check that the DMX model contains the DMX events after they are added. """
+    dm_mod = DispersionDMX()
+    index = 3
+    dmx = 1.0
+    mjd_start = 58000.0
+    mjd_end = 58100.0
+    dm_mod.add_DMX_range(mjd_start, mjd_end, index, dmx, frozen=False)
+    assert len(dm_mod.params) == 7
```
aarchiba, line 67: This test is going to break if `dm_mod` ever changes the default number of parameters. Better to record the number before and confirm that three have been added.

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,142 @@
...
+def test_remove_DMX():
+    """Check the DMX model no longer has the DMX components after they are removed. """
+    dm_mod = DispersionDMX()
+    index = 3
+    dmx = 1.0
+    mjd_start = 58000.0
+    mjd_end = 58100.0
+    dm_mod.add_DMX_range(mjd_start, mjd_end, index, dmx, frozen=False)
+    dm_mod.remove_DMX_range(index)
+    for pn in zip(["DMX_", "DMXR1_", "DMXR2_"]):
```
aarchiba, line 88: What is this `zip` supposed to do?

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,142 @@
...
+        nm = str(pn) + str("{:04d}".format(index))
+        assert nm not in dm_mod.params
+
+
+def test_remove_DMX():
```
aarchiba, line 93: This test has a repeated name and thus hides the previous one.

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,142 @@
...
+    PEPOCH 57000
+    DM 10 0
+    """
+    index = 3
+    dmx = 1.0
+    mjd_start = 58000.0
+    mjd_end = 58100.0
+
+    model = get_model(
+        io.StringIO(
```
aarchiba, line 116: Why not just add all this into the `par_base` string?

src/pint/models/dispersion_model.py
```diff
@@ -319,42 +319,83 @@ def __init__(self):
...
+
+    def add_DMX_range(self, mjd_start, mjd_end, index=None, dmx=0, frozen=True):
+        """ Add DMX range to a dispersion model with specified start/end MJDs and DMX. 
+            returns: index [int]
+        """
+
+        #### Setting up the DMX title convention. If index is None, want to increment the current max DMX index by 1.
+        if index is None:
+            index = self.get_prefix_mapping_component("DMX_").max() + 1
+        i = f"{int(index):04d}"
```
aarchiba, line 337: It would be good to have tests of what happens if you don't provide `index` or if you provide strange values (negative, larger than 9999, floating-point...)

src/pint/models/dispersion_model.py
```diff
@@ -319,42 +319,83 @@ def __init__(self):
...
+
+        if mjd_end is not None and mjd_start is not None:
+            if mjd_end < mjd_start:
+                raise ValueError("Starting MJD is greater than ending MJD.")
+        elif mjd_start != mjd_end:
+            raise ValueError("Only one MJD bound is set.")
+
+        #### Check to see if the user-input index is already assigned to a different DMX component.
```
aarchiba, line 345: I think the `if` statement conveys the same information as this comment.

src/pint/models/dispersion_model.py
```diff
@@ -319,42 +319,83 @@ def __init__(self):
...
+        if int(index) in self.get_prefix_mapping_component("DMX_"):
+            raise ValueError(
+                "Index '%s' is already in use in this model. Please choose another."
+                % index
+            )
+
+        #### Add three DMX parameters (start MJD, end MJD, DMX value)
```
aarchiba, line 352: This information is also visible in the code immediately below the comment.

src/pint/models/dispersion_model.py
```diff
@@ -319,42 +319,83 @@ def __init__(self):
...
                 time_scale="utc",
+                value=mjd_end,
             )
         )
-        self.dm_value_funcs += [self.dmx_dm]
-        self.set_special_params(["DMX_0001", "DMXR1_0001", "DMXR2_0001"])
-        self.delay_funcs_component += [self.DMX_dispersion_delay]
+        self.setup()
+        self.validate()
+        return index
```
aarchiba, line 391: Does this return the generated index if none was provided?

src/pint/models/dispersion_model.py
```diff
@@ -319,42 +319,83 @@ def __init__(self):
...
+
+    def remove_DMX_range(self, index):
+        """Removes all three DMX parameters associated with a given index."""
+        index = "{:04d}".format(index)
```
aarchiba, line 395: You can use f-strings for this.

src/pint/models/dispersion_model.py
```diff
@@ -319,42 +319,83 @@ def __init__(self):
...
+        for prefix in ["DMX_", "DMXR1_", "DMXR2_"]:
```
aarchiba, line 396: It would be good to have a test to verify what happens if a nonexistent index is provided - is the TimingModel damaged?

paulray:
@almcewen0 can you merge nanograv/master in with this so that the tests run under GitHub actions and not travis-ci?

src/pint/models/timing_model.py
almcewen0: No, not necessarily. Those parameters are just likely to change AND be relatively uninteresting between the two par files, while things like EPHEM, NTOA, and CHI2 might be good to highlight with the warning statements as opposed to info statements.

That said, it is quite easy to add more parameters to this exclude list, so if there are specific parameters that should be similarly "demoted," I can do that.

src/pint/models/dispersion_model.py
almcewen0: we needed this to allow the initial instance of DMX (the "fake" one for the model builder), which is made with None MJDs.

tests/test_dmxrange_add_sub.py
almcewen0: removed

tests/test_dmxrange_add_sub.py
almcewen0: removed

src/pint/models/timing_model.py
almcewen0: done.

tests/test_dmxrange_add_sub.py
almcewen0: done.

tests/test_dmxrange_add_sub.py
almcewen0: done

tests/test_dmxrange_add_sub.py
almcewen0: unsure! got rid of it. (:

tests/test_dmxrange_add_sub.py
almcewen0: fixed.

tests/test_dmxrange_add_sub.py
almcewen0: done.

src/pint/models/dispersion_model.py
almcewen0: okay, so as it stands, using a negative value raises a PrefixError, numbers larger than 9999 pass (so you can have DMX_99999 or whatever), and floating points are just int cast, so they round down. what behavior is preferable?

src/pint/models/dispersion_model.py
almcewen0: removed

src/pint/models/dispersion_model.py
almcewen0: removed

src/pint/models/dispersion_model.py
almcewen0: yep!

src/pint/models/dispersion_model.py
almcewen0: done

tests/test_dmxrange_add_sub.py
almcewen0: i'm a little confused with this test file, it looks like maybe there was a C&P error or something... that second "with" is supposed to test if we add a None MJD. I just moved it to another test.

src/pint/models/dispersion_model.py
almcewen0: added.

almcewen0: added.

src/pint/models/dispersion_model.py
aarchiba: Does DMX_99999 actually work in the other parts of the code? Like, is there a search specific to four-digit post-fixes? If they work everywhere, it's probably fine (though they will break alphabetic sorting), but if they give other kinds of trouble better to reject them. Floating point is probably fine. Where are these tests in the test suite?

src/pint/models/dispersion_model.py
aarchiba: Would it be better to reject None entirely and make the bogus one by hand?

src/pint/models/dispersion_model.py
aarchiba: Which test verifies this?

tests/test_dmxrange_add_sub.py
aarchiba: It seems to still be here?

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,140 @@
...
+    comp = getattr(dm_mod, nm)
+    assert comp.value == mjd_start
+    nm = "DMXR2_" + "{:04d}".format(index)
+    comp = getattr(dm_mod, nm)
+    assert comp.value == mjd_end
+    dm_mod.remove_DMX_range(index)
+    for pn in ["DMX_", "DMXR1_", "DMXR2_"]:
+        nm = str(pn) + str("{:04d}".format(index))
+        assert nm not in dm_mod.params
+    init_len = str(len(model.params))
```
aarchiba, line 137: Why are you converting to a string here?

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,140 @@
...
+    assert comp.value == dmx
+    nm = "DMXR1_" + "{:04d}".format(index)
+    comp = getattr(dm_mod, nm)
+    assert comp.value == mjd_start
+    nm = "DMXR2_" + "{:04d}".format(index)
+    comp = getattr(dm_mod, nm)
+    assert comp.value == mjd_end
+    dm_mod.remove_DMX_range(index)
+    for pn in ["DMX_", "DMXR1_", "DMXR2_"]:
+        nm = str(pn) + str("{:04d}".format(index))
```
aarchiba, line 135: What is this `str` for? Isn't the content already a string?

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,140 @@
...
+    index = 3
+    dmx = 1.0
+    mjd_start = 58000.0
+    mjd_end = 58100.0
+
+    model = get_model(io.StringIO(par_base))
+
+    dm_mod = model.components["DispersionDMX"]
+    dm_mod.add_DMX_range(mjd_start, mjd_end, index, dmx, frozen=False)
+    nm = "DMX_" + "{:04d}".format(index)
```
aarchiba, line 124: All these `.format` would be clearer as f-strings.

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,140 @@
...
+    index2 = None
+    dmx2 = 1.0
+    mjd_start2 = 58200.0
+    mjd_end2 = 58300.0
+    dm_mod.add_DMX_range(mjd_start1, mjd_end1, index1, dmx1, frozen=False)
+    dct = dm_mod.get_prefix_mapping_component("DMX_")
+    ind = np.max(list(dct.keys())) + 1
+    dm_mod.add_DMX_range(mjd_start2, mjd_end2, index2, dmx2, frozen=False)
+    dct = dm_mod.get_prefix_mapping_component("DMX_")
+    assert np.max(list(dct.keys())) == ind
```
aarchiba, line 64: This test requires a specific algorithm for choosing the dynamic indices. Better to check that the returned index corresponds to a real parameter and was not previously in the set of indices.

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,140 @@
...
+    nm = "DMXR2_" + "{:04d}".format(index)
+    comp = getattr(dm_mod, nm)
+    assert comp.value == mjd_end
+    dm_mod.remove_DMX_range(index)
+    for pn in ["DMX_", "DMXR1_", "DMXR2_"]:
+        nm = str(pn) + str("{:04d}".format(index))
+        assert nm not in dm_mod.params
+    init_len = str(len(model.params))
+    with pytest.raises(ValueError):
+        dm_mod.remove_DMX_range(3)
```
aarchiba, line 139: Why is this a 3 and not `index`? It would be fine to replace all the `index` with 3 (and get rid of a lot of string formatting) or use `index` throughout, but the mix is asking for trouble.

tests/test_dmxrange_add_sub.py
```diff
@@ -0,0 +1,140 @@
...
+    dm_mod = DispersionDMX()
+    with pytest.raises(ValueError):
+        dm_mod.add_DMX_range(None, 58000, 3, 1, frozen=False)
+    with pytest.raises(ValueError):
+        dm_mod.add_DMX_range(58000, None, 3, 1, frozen=False)
+
+
+def test_dynamic_index():
+    dm_mod = DispersionDMX()
+    index1 = 3
```
aarchiba, line 51: If you only use these variables once it's probably clearer to put the values in the right places directly (here and in other tests). It'll make them much shorter and easier to read.

src/pint/models/dispersion_model.py
aarchiba: I don't see where the test is that checks the TimingModel or Component is undamaged? (Like, other parameters not deleted for example.)

src/pint/models/dispersion_model.py
almcewen0: okay, i think i fixed them. let me know if they are still no good.

tests/test_dmxrange_add_sub.py
almcewen0: ah, i think it got overwritten somehow. but i think it's good now

src/pint/models/dispersion_model.py
almcewen0: i added a test for negative index and floating point, but how can i check that everything else works with DMX_99999? i don't really know what all interacts with the DMX parameters.

src/pint/models/dispersion_model.py
almcewen0: should be test_dynamic_index.

src/pint/models/dispersion_model.py
almcewen0: at the end of test_model_usage, i just added a quick loop to check this.

tests/test_dmxrange_add_sub.py
almcewen0: yea, i was worried that it would just point to the list and i haven't imported deepcopy, so i thought that a string would work. but i've now removed this to improve the test.

tests/test_dmxrange_add_sub.py
almcewen0: just a habit, i suppose. but i removed it.

tests/test_dmxrange_add_sub.py
almcewen0: fixed.

tests/test_dmxrange_add_sub.py
almcewen0: okay i changed this a bit (i was forgetting that i had set it up to return the index of the new DMX range when one is added). let me know if it seems good enough now.

tests/test_dmxrange_add_sub.py
almcewen0: whoops, that is just a typo. should be fixed.

tests/test_dmxrange_add_sub.py
almcewen0: i thought it might be easier to be able to tell which variable are which in the call to add/remove the ranges. but i can change it if this doesn't make sense.

paulray:
Ready to merge?

almcewen0:
> Ready to merge?

should be!

Issue #797, opened 2020-10-21
Warning methods


dlakaplan:
It seems that PINT uses:
`log.warning()` for warnings, but this doesn't raise any sort of exception and is therefore hard to examine from any but a user-output perspective (i.e., complicated for test routines).  Instead we could use `warnings.warn()` which is a subclass of `Exception` that doesn't stop the program.

If we could swap the standard, then the `warnings.warn()` could handle the output side as well:
you can get warnings issued with `warnings.warn()` to be passed to the logger automatically, but not the other way around https://docs.python.org/3/library/logging.html#integration-with-the-warnings-module

dlakaplan:
@scottransom @luojing1211 @paulray What are your thoughts?

aarchiba:
From https://docs.python.org/3/howto/logging.html : 

> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
> 
> logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

PR #912, opened 2020-12-20
Allow no toa for the noise model parameter

luojing1211:
The noise design matrix falls when ECORR does not have TOAs. Since the noise parameters are not fitted ECORR's validate_toa() never gets ran. This PR fix the failing and add the warning print for the noise model's validate_toas.

### DIFF
--- a/src/pint/models/noise_model.py
+++ b/src/pint/models/noise_model.py
@@ -10,6 +10,7 @@ from pint.models.parameter import floatParameter, maskParameter
 from pint.models.timing_model import Component
 from collections import OrderedDict
 
+
 class NoiseComponent(Component):
     def __init__(self,):
         super(NoiseComponent, self).__init__()
--- a/src/pint/models/noise_model.py
+++ b/src/pint/models/noise_model.py
@@ -44,7 +44,7 @@ class NoiseComponent(Component):
                 )
             if not suppress_warning:
                 log.warning(msg)
-            return no_toa_param
+        return no_toa_param
 
 
 class ScaleToaError(NoiseComponent):
--- a/src/pint/models/noise_model.py
+++ b/src/pint/models/noise_model.py
@@ -8,7 +8,7 @@ from astropy import log
 
 from pint.models.parameter import floatParameter, maskParameter
 from pint.models.timing_model import Component
-
+from collections import OrderedDict
 
 class NoiseComponent(Component):
     def __init__(self,):
@@ -26,25 +26,6 @@ class NoiseComponent(Component):
 
     def validate(self,):
         super(NoiseComponent, self).validate()
-
-    def validate_toas(self, toas, suppress_warning=False):
-        no_toa_param = []
-        for par_name in self.get_params_of_type('maskParameter'):
-            par = getattr(self, par_name)
-            mask = par.select_toa_mask(toas)
-            if len(mask) == 0:
-                no_toa_param.append(par_name)
-
-        if no_toa_param != []:
-            if len(no_toa_param) == 1:
-                msg = f"Noise Parameter {no_toa_param[0]} does not correspond to any TOAs"
-            elif len(no_toa_param) > 1:
-                msg = (
-                    f"Parameters {' '.join(no_toa_param)} do not correspond to any TOAs"
-                )
-            if not suppress_warning:
-                log.warning(msg)
-        return no_toa_param
 
 
 class ScaleToaError(NoiseComponent):
@@ -397,27 +378,23 @@ class EcorrNoise(NoiseComponent):
         The weights used are the square of the ECORR values.
 
         """
-        no_toa_params = self.validate_toas(toas, suppress_warning=True)
         tbl = toas.table
         t = (tbl["tdbld"].quantity * u.day).to(u.s).value
         ecorrs = self.get_ecorrs()
-        umats = []
+        umats = OrderedDict()
         for ec in ecorrs:
-            if ec.name in no_toa_params:
-                continue
             mask = ec.select_toa_mask(toas)
-            umats.append(create_quantization_matrix(t[mask]))
-        nc = sum(u.shape[1] for u in umats)
+            if mask != []:
+                umats[ec.name] = (mask, create_quantization_matrix(t[mask]))
+        nc = sum(u[1].shape[1] for u in umats.values())
         umat = np.zeros((len(t), nc))
         weight = np.zeros(nc)
         nctot = 0
-        for ct, ec in enumerate(ecorrs):
-            if ec.name in no_toa_params:
-                continue
-            mask = ec.select_toa_mask(toas)
-            nn = umats[ct].shape[1]
-            umat[mask, nctot : nn + nctot] = umats[ct]
-            weight[nctot : nn + nctot] = ec.quantity.to(u.s).value ** 2
+        for ec, umt in umats.items():
+            nn = umt[1].shape[1]
+            mask = umt[0]
+            umat[mask, nctot : nn + nctot] = umt[1]
+            weight[nctot : nn + nctot] = getattr(self, ec).quantity.to(u.s).value ** 2
             nctot += nn
         return (umat, weight)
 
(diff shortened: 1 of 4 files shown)


src/pint/models/noise_model.py
```diff
@@ -27,6 +27,25 @@ def __init__(self,):
     def validate(self,):
         super(NoiseComponent, self).validate()
 
+    def validate_toas(self, toas, suppress_warning=False):
```
aarchiba, line 30: This does something different from what `validate_toas` does in every other case: all other versions of the function raise an exception if any free parameter has no TOAs. This one doesn't care if the parameters are free, and doesn't raise an exception. If you still want a function that does what this one does, please call it something else.

src/pint/models/noise_model.py
```diff
@@ -378,18 +397,23 @@ def ecorr_basis_weight_pair(self, toas):
...
         """
+        no_toa_params = self.validate_toas(toas, suppress_warning=True)
         tbl = toas.table
         t = (tbl["tdbld"].quantity * u.day).to(u.s).value
         ecorrs = self.get_ecorrs()
         umats = []
         for ec in ecorrs:
+            if ec.name in no_toa_params:
+                continue
             mask = ec.select_toa_mask(toas)
```
aarchiba, line 408: Can't you just skip the ones that don't select any TOAs? And record them here if you need a list of the ones to skip, later?

src/pint/models/noise_model.py
```diff
@@ -378,18 +397,23 @@ def ecorr_basis_weight_pair(self, toas):
...
             umats.append(create_quantization_matrix(t[mask]))
         nc = sum(u.shape[1] for u in umats)
         umat = np.zeros((len(t), nc))
         weight = np.zeros(nc)
         nctot = 0
         for ct, ec in enumerate(ecorrs):
+            if ec.name in no_toa_params:
+                continue
             mask = ec.select_toa_mask(toas)
```
aarchiba, line 417: Select again gives you the right tool to skip the ones that aren't needed. And it's more efficient than searching through all the maskParameters in a separate loop.

src/pint/models/noise_model.py
luojing1211: The original valiate_toa() will never raise any exception since the noise parameters are not get fitted currently. I understand the point to have only one behaviour for this type of functions. However, I think the noise model should warn the use about no TOA found, which is the goal of this function. I can give a different name, but in the end, if we want to warning the user about this, we will have to run the "new_name" function every time the fitting starts. Is there any risk to expend the behaviours  of validate_toa() functions

src/pint/models/noise_model.py
luojing1211: OK.

src/pint/models/noise_model.py
aarchiba: On every other kind of object, `validate_toa` raises exceptions to indicate anything missing. Having a function of the same name do something different on this kind of object is confusing. Particularly as this is called through inheritance and by the timing model. All the other versions do nothing to warn of parameters with no TOAs unless they are being fit. Certainly they don't return anything. If you wanted all the `validate_toa` versions to warn about no-TOA-but-frozen parameters, and this one to begin raising exceptions if noise parameters became fittable, then we could have consistent behaviour. But this function is not necessary, as I say below, to fix the bug in question.

luojing1211: Ok. I can remove this function here and I do think it is reasonable to give a warning when a frozen parameter has no toas. I got a lot of questions about the error massage caused by lacking of TOAs in the noise_design_matrix(). It is misleading.

src/pint/models/timing_model.py
```diff
@@ -1987,17 +1991,29 @@ def validate_toas(self, toas):
...
+        # not warning parameters.
         for c in self.components.values():
             try:
                 c.validate_toas(toas)
             except MissingTOAs as e:
                 bad_parameters += e.parameter_names
+
+        # first print out the warnings
+        if warning_parameters:
+            log.warning(missingTOAs_msg(warning_parameters))
```
aarchiba, line 2016: If you use `warnings.warn` here you can use `pytest.warns` below to verify that warnings are emitted.

tests/test_gls_fitter.py
```diff
@@ -77,3 +78,20 @@ def test_gls_compare(self):
...
+        mcorr2 = maskParameter(
+            name="ECORR", key="-f", key_value="fake_backend2", value=1, units=u.us
+        )
+        self.m.add_param_from_top(mcorr1, "EcorrNoise", setup=True)
+        self.m.add_param_from_top(mcorr2, "EcorrNoise", setup=True)
+        assert "ECORR5" in self.m.params
+        assert "ECORR6" in self.m.params
+        fitter = GLSFitter(self.t, self.m)
+        chi2 = fitter.fit_toas()
+        assert fitter.resids.dof == 3914
```
aarchiba, line 96: Can this test be done with a much smaller synthetic data set? GLSFitter tests can be very slow.

src/pint/models/timing_model.py
luojing1211: I was thinking about it, but in the end, I followed the tradition. Just want to know, why does astropy implement its own warning? Is there any reason we should stick with astropy warning?

tests/test_gls_fitter.py
luojing1211: I can have one step to delete part of the TOAs, still real data, but no need to add more  .tim files.

src/pint/models/timing_model.py
scottransom: We've had this discussion before, I think.  I think it is just that the astropy logging is much more than warnings.  It has multiple levels (including INFO and DEBUG -- which we should use more often) and has colors and other nice capabilities.  It isn't like it is a new dependency or anything, and we use it all over in PINT.  So I don't think it is worth worrying about.  I'm not sure that the `pytest.warns` is a big enough improvement to cause us to switch.  But I'm happy to hear other arguments.

aarchiba: Astropy itself has largely made the switch over to using `warnings`; the `warnings` mechanism can easily be logged using the Astropy logging mechanism. If we use `warnings` we can write tests that verify that we emit the important warnings, and our users can choose what to do with them - for example, use `-Werror` to promote them into exceptions if they want to write code that doesn't emit any. There is discussion of this in #797 ; in one comment there I added:

From https://docs.python.org/3/howto/logging.html :

>    warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning
>
>    logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

tests/test_gls_fitter.py
aarchiba: `make_fake_toas` is an easy way to make synthetic data without adding any `.tim` files.

paulray:
Can this be merged?

aarchiba:
I think it's important that this fix not add another very slow test to the test suite, and currently it loads and fits (using the GLSFitter) a large set of real TOAs. PR #990 is my own fix for the problem, implemented when I didn't realize this solved the same problem.

paulray:
I agree that we should make sure not to add very slow tests.  That should be fixed, then we have to decide if this and #990 are mutually exclusive.

aarchiba:
This and #990 fix exactly the same problem, as far as I can tell. The tests here don't verify that a warning is emitted and do take much longer to run.

luojing1211:
If #990 fix this problem better. I can close this one merge the other one.

aarchiba:
This one also warns about EFACs and EQUADs with no TOAs, and it does that during validation, which #990 doesn't. I can put that feature in #990, along with tests, if that is how we want to handle this.

luojing1211:
If it does not take a long to add it. Feel free to do it.

aarchiba:
I'm not sure why this test didn't pick it up but in expanding the tests for #990 I got errors from EFAC/EQUAD with no TOAs as well - astropy Quantity arrays don't like *= with an empty mask.

aarchiba:
Superseded by #990

PR #914, opened 2020-12-20
Fix 908

### LINKED ISSUE #839: pickles are not regenerated if INCLUDEd files change
Pickles are meant to serve as transparent caches of loaded and corrected TOAs objects, and as a result they check the modification time of the `.tim` file against that of the `.pickle.gz` file. But they do not go on to check `.tim` files added by `INCLUDE` statements, so that one could modify one of these `INCLUDE`d files and not see the changes reflected in the loaded `TOAs` object because the pickle was incorrectly re-used.

### LINKED ISSUE #840: pickle invalidation is based on time not content
NANOGrav is discussing the possibility of passing pickles around to get better efficiency and replicability (they encode the `ephem` setting). This may result in pickles that are older than the `.tim` files that still do not need to be regenerated. One could add a hash of the input file(s) to TOAs objects (but note #839 ) so that the contents of their `.tim` files could quickly be checked for changes.

### LINKED ISSUE #913: travis-ci tests are failing
There are now a bunch of failing tests on the master branch. We need to diagnose and fix. That should be the last thing needed to tag 0.8

aarchiba:
PR #908 broke CI in two different ways; this fixes them. It has the side advantage of using hashes to invalidate pickles rather than modification dates.

Closes #913 
Closes #840 
Closes #839


paulray:
Can I merge this?

aarchiba:
Take a quick look at the code but as far as I can tell, yes.

src/pint/toa.py
```diff
@@ -1277,8 +1288,12 @@ def pickle(self, filename=None):
         self.pintversion = pint.__version__
         if filename is not None:
             pickle.dump(self, open(filename, "wb"))
-        elif (self.filename is not None) and (type(self.filename) is not list):
-            pickle.dump(self, gzip.open(self.filename + ".pickle.gz", "wb"))
+        elif self.filename is not None:
+            if isinstance(self.filename, str):
+                filename = self.filename
+            else:
+                filename = self.filename[0]
```
scottransom, line 1295: This is problematic, IMO, and it changes the intent of what I wrote.  In this case, if there have been merges of .tim files, the pickle file will automatically get the name of the first file.  I don't think that should happen, though.  Because that is a huge assumption that that is what you would want.  If there is more than one filename (including from an INCLUDE), I think we should have to specify the name of the pickle file.  I know that that was not the way it was before, but that's because we didn't keep track of INCLUDE filenames (which we should have).

tests/datafile/parkes.toa
```diff
@@ -1,9 +1,9 @@
 # These are Parkes format and have been changed to also be Parkes observatory
```
scottransom, line 1: I added that comment about changing the observatory to Parkes.  The comment needs updated as it is no longer correct!

src/pint/toa.py
```diff
@@ -78,6 +78,12 @@
 JD_MJD = 2400000.5
 
 
+def _compute_hash(filename):
+    h = hashlib.sha256()
+    h.update(open(filename, "rb").read())
```
scottransom, line 83: Doesn't this mean that every .tim file will be opened and read at least twice now?  Can't we do this as part of the first read?

scottransom (commented):
Thanks for the fixes, Anne!  And sorry about the breakages!

src/pint/toa.py
aarchiba: So as it stands, pickles are not created for merged files at all, only for files loaded from a single .tim file - which have multiple filenames now, if they have INCLUDEs. And when you have a file with INCLUDEs, the pickle gets named after the top-level file. If you are working with merged files, they have to be loaded one by one and then merged; the loading process is accelerated by pickling each of the files to be merged individually.

So your concern only applies if the user explicitly calls `.pickle` themselves.

src/pint/toa.py
aarchiba: They will be opened and read twice. Reading a binary file without parsing is not a very expensive operation. Trying to build the hash as they are read is a pain in the neck, because the code needs to be innterlaced with the parsing code, and because hashes operate on binary data while parsing operates on text (so one has to deal with Unicode conversions).

scottransom: I was thinking about doing it under the hood at the beginning of the parsing -- for instance reading the file as a whole as you are doing for the hashing, but then maybe using that as a StringIO object for the rest of the reading?

src/pint/toa.py
scottransom: I'm concerned with both with merged files that a user calls pickle on (which I think they should be required to specifically give a name) and also with the pickling happening for files with INCLUDEs.  Naming after the top-level file is what we had been doing, but I'm not sure that that is what we should continue to do (there are arguments for it and also against, I think).  Personally, I think the user should be required to specify the filename if their TOAs object came from more than one file no matter how that happened.

src/pint/toa.py
aarchiba: There's still the annoying text/binary thing. And the cost of reading a binary file the size of our .tim files is very low. It's also going to be something of a pain to keep track of all the INCLUDEs and their hashes.

scottransom: Yup.  I don't disagree.  I don't feel too strongly and I do think that using hashes somehow is a big improvement.

src/pint/toa.py
aarchiba: So I saw this in action with the `timing_analysis` from the NANOGrav Gitlab: they had to create a `.tim` file with INCLUDEs, and turning on auto-pickling made reloading `.tim` files fast. The problem was that someone might modify one of the INCLUDEd files and the cache wouldn't be invalidated - but the hashes fix that. So the pickles capture all the TOAs referred to by the top-level file, and the automatic name makes it easy to use transparent caching. 

Manual pickling is a different story; of course users can pickle any old python object to any old file and unpickle it by hand later. The point of having pickle support code here, in PINT, is to make transparent caching work. And it can, just fine, for `.tim` files with INCLUDEs. It even works for merged TOAs objects by caching their components.

PR #916, opened 2020-12-20
Allow TT(BIPM)

### LINKED ISSUE #911: TT(BIPM) does not recognised by get_model_and_toas()
The models.get_model_and_toas() does not support the par file input TT(BIPM), which I think is for the latest version in TEMPO. It has BIPM in the string but without the version/year, this will search the BIPM file named"bipm" instead.

aarchiba:
Currently this relies on `tai2tt_bipm.clk` being a symlink to the most recent `tai2tt_bipm2019.clk` file. This should probably be cleaned up when the clock correction code is cleaned up.

Closes #911


scottransom:
This looks like it has the changes from #914 in it as well (although maybe that is just a github thing given that they were both in progress at the same time).

I notice that my comments on #914 went in too late.  But please check them out.  I think the pickling of multiple files should change.

aarchiba:
Yes, I branched this off #914 because I wanted it to be able to pass tests; that's not a problem, merge-wise. I answered some of your comments on #914, and we can continue this discussion there.

PR #918, opened 2020-12-20
Ensure that merged files don't get auto-pickled

aarchiba:
This expands on #914

### DIFF
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -955,6 +955,8 @@ class TOAs(object):
         self.ephem = None
         self.clock_corr_info = {}
         self.obliquity = None
+        self.merged = False
+        self.hashes = {}
 
         if (toalist is not None) and (toafile is not None):
             raise ValueError("Cannot initialize TOAs from both file and list.")
@@ -1288,6 +1290,8 @@ class TOAs(object):
         self.pintversion = pint.__version__
         if filename is not None:
             pickle.dump(self, open(filename, "wb"))
+        elif self.merged:
+            raise ValueError("Merged TOAs need specified file names")
         elif self.filename is not None:
             if isinstance(self.filename, str):
                 filename = self.filename
@@ -1809,6 +1813,7 @@ class TOAs(object):
         self.ephem = tmp.ephem
         self.planets = tmp.planets
         self.hashes = tmp.hashes
+        self.merged = tmp.merged
 
     def read_toa_file(self, filename, process_includes=True, top=True):
         """Read TOAs from the given filename.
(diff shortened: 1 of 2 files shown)


aarchiba:
The way we're handling pickling and unpickling is a shambles, and I'd sort of like to take it apart and start over. For that matter the whole TOAs object could use some revamping, with as little API change as possible, so that they are constructed in a more coherent way and so that one never has them containing lists of TOA objects.

scottransom:
For this PR, are you proposing that `.merged` gets set when an INCLUDE has been processed?  Or only after explicitly using `merge_TOAs()`?

I don't disagree with your comments above.  But a lot of the issues are because we have decided to support the timfile standards of TEMPO and TEMPO2.  And those things include commands that affect the timing model, the toas in the file, and also multiple levels of INCLUDEs.  Those are all a PITA, but are used all over the comunity.

aarchiba:
My idea is that a `.tim` file with INCLUDEs can still be pickled without confusion, producing a pickle that is named according to the top-level file. Files that have been merged at the python level, as you say, there is no natural top-level file name for them. So pickles need to be explicitly named. And anyway, their components can be pickled.

aarchiba:
The handling of pickles is a different question. I realize we're stuck with the TEMPO/TEMPO2-compatible `.tim` files, and that that produces a few headaches. But the question of how we handle pickles and construction of the TOAs objects is largely internal to PINT, and it's something we can do better. For example, I don't think TOAs objects should know anything about pickles. The fact that they try to unpickle themselves means we have to unpickle a new TOAs object then copy over all the things we think are relevant; we should just unpickle them directly, but you can't do that from inside a constructor. Likewise constructing TOAs from lists shouldn't be in the constructor. Let get_TOAs and friends handle all the searching for pickles and whatnot.

aarchiba:
To be clear, I'm not proposing doing any of this in this PR, I just think some cleanup would help make all this code a lot more manageable.

scottransom:
I guess I'm OK with auto-pickles using the top-level file name if there has been an INCLUDE -- especially since that is what we have been doing already!

scottransom:
I'm still not sure about my question above though:  when are you planning to set the `.merged` attribute?  After just a "manual" call of `merge_TOAs()` or also when an INCLUDE has been processed?  I don't see it getting set anywhere right now.

aarchiba:
I used `.merged` because it was already getting set (only) when manually merging TOAs objects.

paulray:
OK to merge this PR?

scottransom:
Yeah, I think so.  I'll do it now.

PR #919, opened 2020-12-21
Support fake wideband TOAs

aarchiba:


paulray:
Merging since fake WB TOAs are very useful!

PR #922, opened 2020-12-21
Add component build example

luojing1211:
This PR adds component building examples to the documentation. This PR will be a part of version 0.8 and the paper appendix.


src/pint/residuals.py
```diff
@@ -242,26 +242,15 @@ def get_PSR_freq(self, modelF0=True):
...
-                    "more than one PSR frequency parameter found; "
-                    + "should be only one from %s" % F0names
-                )
-            return F0 * u.Hz
+            # TODO this function will be re-write and move to timing model soon.
+            # The following is a temproary patch.
+            if 'Spindown' in self.model.components:
+                return self.model.F0.quantity
+            elif 'P0' in self.model.params:
+                return 1.0/self.model.P0.quantity
```
paulray, line 250: Maybe add `.to(u.Hz)`?

paulray:
I don't think we use markdown notebooks anymore. I think they are supposed to be .py now. I'm not sure exactly how to implement this, though.

luojing1211:
You are right. Let me see if I can change it.

paulray:
Maybe just this? `jupytext --to py How_to_build_a_timing_model_component.md`

paulray:
Maybe add a sentence at the beginning of the notebook like this:

"""
PINT's design makes it easy to add a new, custom timing model component to meet specific needs. This notebook demonstrates how to write your own timing model component with the minimal requirements so that PINT can recognize and use it in fits. Here, we implement a new spindown class that is parameterized by P0, P1, instead of the usual Taylor series in frequency.
"""

src/pint/residuals.py
luojing1211: udated.

PR #925, opened 2020-12-22
update paper validate notebook for version 0.8

luojing1211:
This PR updates the paper validate notebook for PINT version 0.8


luojing1211:
The tests failed at the wideband check. But I think we can merge this PR. Since it is not related to any of our test.

aarchiba:
For others wanting to review the validation notebook, it can be seen here: https://github.com/luojing1211/PINT/blob/update_paper_validation_example/docs/examples-rendered/paper_validation_example.ipynb

- Is there a URL or other easy way to get the NANOGrav 11-year data set that is required?
- Maybe mention a PINT version? Either PINT 0.8 (with a link) or a mention of the earliest version this is likely to work on? And a link to the Github?
- Can you easily (`glob.glob`?) provide a list of pulsars they could run this script on? That is, values they could swap in for `J1600-3053`?
- Maybe a comment in the notebook about the T2CMETHOD when the warning is emitted?
- Why does this fit improve on the NANOGrav 11-year par file?
- "Write the TEMPO postfit residuals to a new .par file" - er, model, surely?
- Under "TEMPO2 run" there's a confusing note about T2CMETHOD
- You don't actually need to create a temporary par file any more, you can modify the string and read from a StringIO; this would also allow the user to actually substitute in another pulsar, as you suggest is possible above.
- What's the story on the ECL parameter for tempo2? Is that a PINT requirement or a TEMPO2 requirement? Is it still a requirement, with the tempo2 compatibility PR that went in?
- The list under "compare the parameter" is slightly malformed - the last entry starting "If" is supposed to be part of the previous line.
- Why is `NE_SW` unrecognized? Why is there a warning about `NE_SW2`? If this is to be expected they should perhaps be explained in this notebook.

luojing1211:
This notebook, right now, is meant to provide the code for the PINT paper. I agree it is not perfect and did not use the newer features, but serves the purpose that the referee has a way to reproduce the result in the paper. We can definitely improve it after the resubmission of the paper. But here is my answer.

> * Is there a URL or other easy way to get the NANOGrav 11-year data set that is required?
The test data is included in the example directory, one does not require to download the 11-year data to run this script. 
> * Maybe mention a PINT version? Either PINT 0.8 (with a link) or a mention of the earliest version this is likely to work on? And a link to the Github?
There is a cell in the notebook that displays the version of PINT, tempo and tempo2.
> * Can you easily (`glob.glob`?) provide a list of pulsars they could run this script on? That is, values they could swap in for `J1600-3053`?
> * Maybe a comment in the notebook about the T2CMETHOD when the warning is emitted?
I can add that.
> * Why does this fit improve on the NANOGrav 11-year par file?
I don't think I ever mention that this fit improves 11-year par file.
> * "Write the TEMPO postfit residuals to a new .par file" - er, model, surely?
I can fix this.
> * Under "TEMPO2 run" there's a confusing note about T2CMETHOD
I can fix this.
> * You don't actually need to create a temporary par file any more, you can modify the string and read from a StringIO; this would also allow the user to actually substitute in another pulsar, as you suggest is possible above.
Can we improve this after the paper resubmission?
> * What's the story on the ECL parameter for tempo2? Is that a PINT requirement or a TEMPO2 requirement? Is it still a requirement, with the tempo2 compatibility PR that went in?
ECL is hard coded in the TEMPO2 code. Can add this.
> * The list under "compare the parameter" is slightly malformed - the last entry starting "If" is supposed to be part of the previous line.
I can fix it. 
> * Why is `NE_SW` unrecognized? Why is there a warning about `NE_SW2`? If this is to be expected they should perhaps be explained in this notebook.
Can we fix this after paper resubmission? I don't know why PINT did not recognize the solar wind, but it is set to be zero.

aarchiba:
> This notebook, right now, is meant to provide the code for the PINT paper. I agree it is not perfect and did not use the newer features, but serves the purpose that the referee has a way to reproduce the result in the paper. We can definitely improve it after the resubmission of the paper. But here is my answer.

I know it's meant to satisfy what the referee wants. I'm trying to offer suggestions to make it better at doing that.

> > * Is there a URL or other easy way to get the NANOGrav 11-year data set that is required?
> 
> The test data is included in the example directory, one does not require to download the 11-year data to run this script.

This is worth saying. What's more, is there a particular directory one must run this notebook from? I know we don't provide a way for users to find the PINT example datafiles, so I guess this must be run from its location in a checked-out PINT source distribution?

> > * Maybe mention a PINT version? Either PINT 0.8 (with a link) or a mention of the earliest version this is likely to work on? And a link to the Github?
> 
> There is a cell in the notebook that displays the version of PINT, tempo and tempo2. I can add the earliest version.

The goal here is for readers to know what they need to run this notebook.

> > * Can you easily (`glob.glob`?) provide a list of pulsars they could run this script on? That is, values they could swap in for `J1600-3053`?
> 
> Not sure how one can swap the pulsar with out changing the data set path.

You have a note that suggests users can just change the source name. If that's not true, don't say so. If they can easily swap in a different pulsar, that's worth saying too, and listing the ones they could easily drop in.

> > * Why does this fit improve on the NANOGrav 11-year par file?
> 
> I don't think I ever mentioned that this fit improves 11-year par file.

It does, though, the chi-squared is lower. Is this to be expected? Is the fit actually supposed to accomplish anything? A little more explanation would help.

> > * You don't actually need to create a temporary par file any more, you can modify the string and read from a StringIO; this would also allow the user to actually substitute in another pulsar, as you suggest is possible above.
> 
> Can we improve this after the paper resubmission?

This relates to whether we can or can not substitute in other pulsars.

> > * Why is `NE_SW` unrecognized? Why is there a warning about `NE_SW2`? If this is to be expected they should perhaps be explained in this notebook.
> 
> Can we fix this after paper resubmission? I don't know why PINT did not recognize the solar wind, but it is set to be zero.

This may be a PINT bug/problem we should investigate.

Is it also worth saying something about the timing model? At least, is it a binary pulsar? What aspects of PINT is this notebook validating?

aarchiba:
This looks much better! A few comments still:

- You seem to give the same github URL for both tempo_utils and tempo2_utils, is this a mistake?
- When you explain that PINT only uses the IAU2000B precession-nutation model but the NANOGrav data uses the TEMPO model, what does that mean? Do we expect disagreement between PINT and TEMPO? PINT and TEMPO2? as a result?
- I think the reason to do fitting in this notebook is so that the timing model will soak up some of the differences between TEMPO/TEMPO2 and PINT? Would it be a better comparison with no fitting? This might be worth explaining.

luojing1211:
> This looks much better! A few comments still:
> 
> * You seem to give the same github URL for both tempo_utils and tempo2_utils, is this a mistake?

That is not a mistake, it is the same package, named tempo_utils but contains both Tempo and Tempo2 utilts.  I just want to be clear that we need both. 

> * When you explain that PINT only uses the IAU2000B precession-nutation model but the NANOGrav data uses the TEMPO model, what does that mean? Do we expect disagreement between PINT and TEMPO? PINT and TEMPO2? as a result?

Yes, the tempo prefit residual's shape is coming from the difference of precession-nutation model. Can add a little bit more.

> * I think the reason to do fitting in this notebook is so that the timing model will soak up some of the differences between TEMPO/TEMPO2 and PINT? Would it be a better comparison with no fitting? This might be worth explaining.

The pre-fit residual is the comparison result of with no fitting, since we are using the same .par file.

paulray:
Ready to merge?

luojing1211:
Waiting for @aarchiba's approval.

aarchiba:
> > This looks much better! A few comments still:
> > 
> > * You seem to give the same github URL for both tempo_utils and tempo2_utils, is this a mistake?
> 
> That is not a mistake, it is the same package, named tempo_utils but contains both Tempo and Tempo2 utilts. I just want to be clear that we need both.

Ah, this is confusing - does `tempo_utils` include TEMPO and TEMPO2 both? Because currently the notebook doesn't provide any other indication for where to obtain those.

> > * I think the reason to do fitting in this notebook is so that the timing model will soak up some of the differences between TEMPO/TEMPO2 and PINT? Would it be a better comparison with no fitting? This might be worth explaining.
> 
> The pre-fit residual is the comparison result of with no fitting, since we are using the same .par file.

I still think the notebook could use some explanation of why fitting is being done and why we are doing the comparisons that we do.

luojing1211:
@paulray  I am adding more explanations. This will be updated shortly

luojing1211:
@aarchiba  Is this good to go?

aarchiba:
Thanks, those explanations at the top help a lot. I think this is good to go.

Issue #870, opened 2020-12-09
Slow tests


paulray:
When using `make test` it now reports the slowest 20 tests.  This is important since our CI resources are limited and it is taking a really long time for travis to run.  So, we need to do some work to speed up the tests where possible.
```
=============================================================== slowest 20 durations ================================================================
171.51s call     tests/test_fitter.py::test_ftest_nb
70.75s call     tests/test_fitter.py::test_ftest_wb
69.66s call     tests/test_gls_fitter.py::TestGLS::test_gls_fitter
68.22s call     tests/test_gls_fitter.py::TestGLS::test_gls_compare
65.44s call     tests/test_parameters.py::TestParameters::test_START_FINISH_in_par
51.22s call     tests/test_utils.py::test_dmxparse
46.12s call     tests/test_ell1h.py::TestELL1H::test_J0613_H4
45.03s call     tests/test_event_optimize.py::TestEventOptimize::test_result
43.73s call     tests/test_ell1h.py::TestELL1H::test_J0613_STIG
34.35s call     tests/test_photonphase.py::test_result
33.35s call     tests/test_gls_fitter.py::TestGLS::test_whitening
33.27s call     tests/test_ecorr_average.py::TestEcorrAverage::test_ecorr_average
29.81s call     tests/test_fbx.py::TestFBX::test_derivative
22.55s call     tests/test_widebandTOA_fitting.py::TestWidebandTOAFitter::test_fitting_no_full_cov
22.24s call     tests/test_widebandTOA_fitting.py::TestWidebandTOAFitter::test_fitting_full_cov
22.04s setup    tests/test_ell1h.py::TestELL1H::test_J0613_H4
21.86s call     tests/test_B1855_9yrs.py::TestB1855::test_derivative
20.85s call     tests/test_wls_fitter.py::Testwls::test_wlf_fitter
20.38s setup    tests/test_ecorr_average.py::TestEcorrAverage::test_ecorr_average
18.49s setup    tests/test_fbx.py::TestFBX::test_B1953_binary_delay
```
@bshapiroalbert can you see about the test_ftest ones? Those are the two slowest tests. If they can be reworked to use much smaller numbers of TOAs that would be great.

bshapiroalbert:
Hey @paulray, I will see if I can make those a bit faster by the end of this week, sorry they are so slow. I think I just grabbed a par and tim file that were already in accessible in the PINT repo and didn't look too much into the speed or number of TOAs.

[...]

Issue #869, opened 2020-12-09
Test suite takes too long


aarchiba:
Currently the PINT test suite takes many minutes to run. While it is possible to run individual test files with `pytest tests/test_thing.py` and individual tests with `pytest tests/test_thing.py::test_thing_works`, running all the tests takes so long it disrupts development and discourages running the whole test suite.

There are tools for improving this:

- Change the default setting for `hypothesis`
- Work with simplified or reduced data, particularly with reduced sets of TOAs; `make_fake_toas` can be used to generate small test sets
- Use `pytest --durations=N` to identify slow tests
- Rewrite tests to test more specific features (unit tests)
- Mark slow tests as slow and set up options so that it is easy to run the fast ones
- Arrange for fixtures (e.g. loaded par files) to be re-used where this does not interfere with testing
- Change (suitable) tests to run with `usepickle=True` so that loaded TOAs are cached
- Identify what tests actually test, especially slow ones, and discard redundant ones
- Reduce the range of parameters used by `pytest.mark.parametrized` where this is excessive

aarchiba:
This is particularly important as Travis has become unusable (and will disappear completely: #817 ) so users need to run our own test suites.

aarchiba:
#871 provides infrastructure and demonstrates how to use an ultra-simplified par file in a triple-quoted string.

aarchiba:
Duplicate of #870

PR #927, opened 2020-12-22
Narrowband F-test test speed up

bshapiroalbert:
I have amended the narrowband F-test test to use significantly fewer TOAs which as greatly reduced the time of the test from ~170 seconds as mentioned in #870 and #869 to ~6 seconds on my local machine. Speeding up the WB tests is trickier since there are not many WB tim and par files to work with right now, though if a stripped down one could be made that would make things faster I think.


aarchiba:
Perhaps consider `make_fake_toas`? One of the PRs allows it to make wideband fake TOAs.

paulray:
I'll merge now to speed up testing. You can make a new PR when you want to speed up the WB testing.
Make sure to pull nanograv/master since I've fixed the GitHub actions testing now.

Issue #691, opened 2020-05-08
pintk - right click to delete doesn't work


clp3ef:
In its original form, pintk allowed a user to right click to delete a particular TOA. This function is no longer usable and currently commented out of the code. This function could be re-implemented, allowing quick and easy deletion of particular TOAs. (Deleting TOAs is currently allowed by selecting TOAs using left click-and-drag and pressing 'd'. Whether right click should act the same as pressing 'd', right click-and-drag should delete all selected TOAs, or right click should act only on single TOAs, is up for discussion)

PR #928, opened 2020-12-22
Pintk right click fix

champagne-cmd:
This addresses issue #691 and reinstates the right-click to delete functionality in `pintk`. Right-clicking on a single `TOA` deletes it (multiple `TOAs` can be deleted at once with the key-press option, but if the right-click to delete should be able to do the same please let me know). There appears to be a bug when deleting `TOAs` after adding a jump, but I think this is an issue with the jump code rather than the delete functionality, so I will address this in another pull request soon.


src/pint/pintk/plk.py
```diff
@@ -1143,20 +1143,50 @@ def stationaryClick(self, event):
...
+                    self.jumped = jumped_copy
+                    # check if point to be deleted is jumped
+                    if all_jumped[ind] == True:
+                        log.warn(
+                            "cannot delete jumped toas. Delete interfering jumps before deleting toas."
+                        )
+                        return None
+                    # create boolean array to readjust all_toas without point to be deleted
+                    toas_to_delete = np.zeros(self.psr.all_toas.ntoas, dtype=bool)
+                    toas_to_delete[ind] = True
```
paulray, line 1168: @aarchiba you recently improved TOAs to change the way selection worked. One of the main places where the old `.select()` was used was in pintk. Can you take a look at this and see if what Chloe is doing here is the current "preferred" way?

aarchiba: I'm quite confused by this code because it doesn't seem to use either the old or the new method of selecting TOAs - it manipulates the table directly, then goes rooting around in the stack. The new method would just be `new_toas = self.psr.all_toas[~toas_to_delete]` but I'm not really clear on what's going on with the jumps and the "selected". It looks like maybe the "selected" process is separate from the question of deleting TOAs?

paulray: Right, since Chloe is going to be working on pintk, she has the opportunity to update how it handles selecting TOAs. It does selections so you can highlight a subset of TOAs and fit, or JUMP, or add/subtract phase turns. So, if you two can discuss the best way to handle that, that would be great.

paulray:
@champagne-cmd if you merge nanograv/master into this PR, the tests will stop failing. I fixed the GitHub actions.

src/pint/pintk/plk.py
champagne-cmd: @aarchiba perhaps I can email you after Christmas to discuss? I will play around with this some more also, I basically copied the code from the keypress delete, so both probably need to be modified.

paulray:
I'm going to merge this.  Then, we can have a broader discussion of how pintk should handle its sets of TOAs as people can delete them, fit subsets, and all that kind of stuff.

paulray:
Oops, it is Draft so I can't merge it.

champagne-cmd:
> Oops, it is Draft so I can't merge it.

I just made it ready for review @paulray if you'd like to merge.

PR #933, opened 2020-12-29
Fix dmjump derivative

luojing1211:
Fix the bug in the DMjump fitting. The DMJUMP fitting was not working due to the dmjump component registers the wrong dmjump derivative functions.

### DIFF
--- a/tests/test_wideband_dm_data.py
+++ b/tests/test_wideband_dm_data.py
@@ -121,6 +121,16 @@ class TestDMData:
             dm_jump_map[dmj.key_value[0]] = dmj
         for be in all_backends:
             assert all(dm_jump_value[toa_backends == be] == -dm_jump_map[be].quantity)
+        for dmj_param in dm_jump_params:
+            # test derivative function in the dmjump component
+            d_dm_d_dmjump = self.model.d_dm_d_dmjump(self.toas, dmj_param.name)
+            be = dmj_param.key_value
+            assert all(d_dm_d_dmjump[toa_backends == be] == -1.0 * u.Unit(''))
+            d_delay_d_dmjump = self.model.d_delay_d_dmjump(self.toas, dmj_param.name)
+            assert all(d_delay_d_dmjump == 0.0 * (u.s / dmj_param.units))
+            # Test the registered functions
+            assert self.model.delay_deriv_funcs[dmj_param.name] == [self.model.d_delay_d_dmjump]
+            assert self.model.dm_derivs[dmj_param.name] == [self.model.d_dm_d_dmjump]
 
         r = WidebandTOAResiduals(
             self.toas, self.model, dm_resid_args=dict(subtract_mean=False)
--- a/tests/test_wideband_dm_data.py
+++ b/tests/test_wideband_dm_data.py
@@ -121,18 +121,6 @@ class TestDMData:
             dm_jump_map[dmj.key_value[0]] = dmj
         for be in all_backends:
             assert all(dm_jump_value[toa_backends == be] == -dm_jump_map[be].quantity)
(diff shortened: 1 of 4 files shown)


pennucci:
Copy/pasted from slack: "I've looked at a number of pulsars, including 1614, 2043, 1903, 1937, 1713.  The DMJUMP uncertainties are comparable to the posterior widths found in the enterprise analyses (from the earlier 12.5-year wideband analysis), even though the DMX uncertainties increase by a factor of a few (3-5ish), but that's to be expected, I think, given the additional free params that have covariance.  The DMJUMP parameter values sometimes change a bit, but within reason/error, and the whole thing is stable under fit iteration."  So this looks good to me.  Merge it?

scottransom:
This seems like an appropriate time to bring back up the idea of having all of our derivative functions tested (maybe even semi-automatically?) by one of the autodifferentiation packages available for Python, now?  I'm worried that we have other subtle derivative bugs in the code, and right now, tests are done (at least sometimes) with fixed differential sizes.

luojing1211:
I agree with you Scott. I think we should merge this PR for the NG 15 year
data analysis and open a new PR for more reliable derivative tests.
-Jing

aarchiba:
Doesn't the paper claim we can use numerical differentiation to verify derivatives? In any case I think `numdifftools` could work to verify our derivatives.

aarchiba:
I think having a test here would be valuable. For one thing I don't actually know what the derivative should look like - I don't fully understand what all the different d_thing_d_otherthing functions are supposed to do. So a test here would both define what the derivatives should look like and check that this function does so correctly.

luojing1211:
I can add some tests for this PR. For the systematically derivative tests I think we need a separate PR.

luojing1211:
I added a small test for the dmjump fix. It is not a full systematic derivative test. But dmjump derivative is basically -1 or 0. I am not sure if this addresses @aarchiba 's question. We should have a discussion about how to do it systematically in the separate issue or  PR

tests/test_wideband_dm_data.py
```diff
@@ -121,6 +121,18 @@ def test_dm_jumps(self):
             dm_jump_map[dmj.key_value[0]] = dmj
         for be in all_backends:
             assert all(dm_jump_value[toa_backends == be] == -dm_jump_map[be].quantity)
+        for dmj_param in dm_jump_params:
+            # test derivative function in the dmjump component
+            d_dm_d_dmjump = self.model.d_dm_d_dmjump(self.toas, dmj_param.name)
+            be = dmj_param.key_value
+            assert all(d_dm_d_dmjump[toa_backends == be] == -1.0 * u.Unit(""))
```
aarchiba, line 128: Should this test that all the others are zero? Also, can you put the derivative test in a separate test method with a descriptive name so we can tell what has failed, if it does?

luojing1211: Just changed.

tests/test_wideband_dm_data.py
```diff
@@ -140,6 +140,38 @@ def test_dm_jumps(self):
...
+        # First get the toas for jump
+        toa_backends, valid_flags = self.toas.get_flag_value("fe")
+        toa_backends = np.array(toa_backends)
+        all_backends = list(set(toa_backends))
+        dm_jump_value = self.model.jump_dm(self.toas)
+        dm_jump_params = [
+            getattr(self.model, x)
+            for x in self.model.params
+            if (x.startswith("DMJUMP"))
+        ]
```
aarchiba, line 154: This shared code can all go in the setup method and be saved in `self`.

aarchiba:
Docs failure is unrelated. Fix due shortly from #936

PR #934, opened 2020-12-29
Script for timings

champagne-cmd:
Script for replicating timing tables in the PINT paper. The plan is to turn the python file currently in this pull request into a notebook to make it easier to work with.


profiling/paper_timing_tables/timing_comparison.py
```diff
@@ -0,0 +1,437 @@
...
+    if m.PLANET_SHAPIRO.value:
+        use_planets = True
+    model_ephem = "DE421"
+    if m.EPHEM is not None:
+        model_ephem = m.EPHEM.value
+
+    # time import statements in pintempo script
+    print("timing imports...")
+    total = 0
+    for i in range(MAXIT):
```
paulray, line 247: This will not work. Python imports are aggressively cached, so these are basically no-ops.  To measure the import speed, you need to make a script with just those import statements and run it as a separate process.

profiling/paper_timing_tables/timing_comparison.py
```diff
@@ -0,0 +1,358 @@
...
+            shell=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
+        end = timeit.default_timer()
+        total = total + (end - start)
+    t2time_arr.append(total / MAXIT)  # average time
+
+
+if __name__ == "__main__":
```
aarchiba, line 94: Can you move almost all of this to a function, so that this script can run as a module? Then the `if __name__ == "__main__":` would just use `argparse` to set things up and call it. Or if you want to make a notebook, that should be a fairly easy conversion.

profiling/paper_timing_tables/timing_comparison.py
```diff
@@ -0,0 +1,358 @@
...
+    """
+    # Generate simple, fake TOAs for the timing runs
```
aarchiba, line 96: What is this doing here?

profiling/paper_timing_tables/timing_comparison.py
```diff
@@ -0,0 +1,358 @@
...
+        overwrite=True,
+    )
+    """
+    # explore more complex model
+    # use J1910+1256 with the following parameter additions to the par file (to ensure GLS fit with Tempo2):
+    # - TNRedAmp -14.227505410948254
+    # - TNRedGam 4.91353
+    # - TNRedC 45
+
+    # add needed params for GLS fitting
```
aarchiba, line 245: Why are you using `subprocess` and `echo` and modifying these files? Better to modify them with python - for example, open them with append mode and just write the new line. What happens if this is run multiple times? Are these files created afresh every time this script is run?

profiling/paper_timing_tables/timing_comparison.py
```diff
@@ -0,0 +1,358 @@
...
+    gls = ""
+    if fitter == "gls":
+        gls = " --gls"
+    usepickle = ""
+    if pickle:
+        usepickle = " --usepickle"
+    total = 0
+    for i in range(MAXIT):
+        start = timeit.default_timer()
+        subprocess.call(
```
aarchiba, line 48: If you use `check_call` you will get an exception instead of wrong numbers if something goes wrong while running the subprocesses.

profiling/paper_timing_tables/timing_comparison.py
```diff
@@ -0,0 +1,358 @@
...
+    )
+    ascii.write(
+        complex_comparison,
+        "complex_tables.pdf",
+        Writer=ascii.Latex,
+        latexdict={"tabletype": "table*"},
+        overwrite=True,
+    )
+
+    # remove added params from par file for future use
```
aarchiba, line 353: This will not get run if an exception - including `KeyboardInterrupt` - arises during execution.

I recommend using `with TemporaryDirectory as d:` and creating and using all files in that directory - this will be deleted when you exit the `with`  block.

profiling/paper_timing_tables/timing_comparison.py
champagne-cmd: Thank you for all the helpful tips @aarchiba , sorry it took me so long to reply! I will be deleting the files at the conclusion of the run, I failed to delete them all in this current version but I am working on implementing better functionality right now.

profiling/paper_timing_tables/timing_comparison.py
```diff
@@ -0,0 +1,338 @@
...
+        print("Making fake TOAs...")
+        for num in ntoas_simple:
+            call = (
+                "zima --startMJD 53478 --duration 700 --freq 1400 2000 --ntoa "
+                + str(num)
+                + " "
+                + par_simple
+                + " "
+                + tempdir
+                + "/NGC6440E_fake"
```
aarchiba, line 106: It's a good idea to use `os.path.join()` to join paths rather than concatenating with a '/'; there are weird corner cases (and Windows) where `join()` works but that doesn't. It may be worth a `from os.path import join` if you use it a lot.

profiling/paper_timing_tables/timing_comparison.py
```diff
@@ -0,0 +1,338 @@
...
+                + str(num)
+                + ".tim"
+            )
+
+            subprocess.check_call(
+                call, shell=True, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL
```
aarchiba, line 112: Do you actually need the shell? It's better (faster, more efficient, more secure, more reliable) to pass a list of command-line arguments and leave out the shell. (`subprocess.check_call(['ls', '-l', tempdir])`)

profiling/paper_timing_tables/timing_comparison.py
```diff
@@ -0,0 +1,338 @@
...
+            model_ephem = m.EPHEM.value
+
+        for tim in timfiles:
+            tim = tempdir + "/" + tim  # add temporary directory path to file
+            # no pickle time of get_TOAs
+            print("timing get_TOAs w/o pickling...")
+            total = 0
+            for i in range(MAXIT):
+                start = timeit.default_timer()
+                toa.get_TOAs(
```
aarchiba, line 196: You can use the (new) `model` argument to `get_TOAs` to avoid having to pull the ephemeris and planets setting and so forth out of the model yourself.

profiling/paper_timing_tables/timing_comparison.py
```diff
@@ -0,0 +1,338 @@
+"""
+A script for comparing the timing of PINT and TEMPO/Tempo2 for standard fitting 
+cases. Recreates Tables 7, 8, and 9 from PINT paper.
+Requires: 
+    - TEMPO
+    - TEMPO2
```
aarchiba, line 6: Do you want to mention that it needs to be run from the same location as its data files? (There are hacks you can use based on `__file__` to look in the same directory as the `.py` file but I don't think it's worth it.)

champagne-cmd:
@aarchiba I decided not to use the temporary directory for this notebook since having the datafiles already created would speed up the process for future runs if the user wished to use this script multiple times. There are checks for the zima files so that the user doesn't have to wait for that process if the files are already there, and they can also simply skip that cell in the notebook. Let me know what you think.

paulray:
I'm merging this to get it into the main branch. If further changes are needed we can make a new PR

PR #936, opened 2021-01-05
build docs in CI

### LINKED ISSUE #935: CI does not run make docs
Tox only runs make docs if python 3.7 is available, which it isn't on our github-action. 

Ideally, a fix would not only run make docs to see if it worked but make available the generated docs to people reviewing the PR - we see a lot of changes with broken docstrings, and it's not easy to inspect new docs in un-merged PRs (I have a hack set up with readthedocs but it doesn't scale).

aarchiba:
Build and publish documentation as part of CI checks.

Fixes #935

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -45,9 +45,6 @@ jobs:
             python: '3.7'
             tox_env: 'oldestdeps'
           - os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


aarchiba:
Ready to go; making the docs as an artifact might be unnecessary (and can easily be removed).

PR #941, opened 2021-01-07
Small changes and added tests for template code

kerrm:
This is code I think only I use, but I needed to make these changes and I'm tired of keeping a forked branch locally!  Very small changes.  Summary of changes:

-- bug fix when handling pickled object
-- can specify line color in light curve fit plotting

EDIT: added some test coverage as stubs to motivate development!


paulray:
Note that there is no test coverage of any of this, I think.  It would be great to add some unit tests.

kerrm:
Yeah... :). I should work something up.  I've already tested the
derivatives and such so it wouldn't be too bad to add at least that to the
test suite, which will also serve to at least touch this code.

paulray:
Should I merge?

tests/test_templates.py
```diff
@@ -0,0 +1,71 @@
...
+
+class TestPrimitives(unittest.TestCase):
+    """ Test instantiation and various properties of LCPrimitive objects.
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        pass
+
+    def test_basic_functionality(self):
```
aarchiba, line 23: There's no need to create a class; we're using pytest so you can just create this as a function.

kerrm:
I'd say go ahead.  Can always pick up additional changes in a future PR!

tests/test_templates.py
```diff
@@ -0,0 +1,71 @@
...
+        g1 = lcprimitives.LCGaussian(p=[0.01, 0.5])
+        # defined as 1/sigma/(2*pi)^0.5 * (-0.5*(x-x0)^2) WRAPPED
+        # because this is a narrow gaussian, it's essentially the same as
+        # an unwrapped version
+        expected_val = gauss(0.50, 0.5, 0.01)
+        assert abs(g1(0.50) - expected_val) < 1e-6
+        expected_val = gauss(0.48, 0.5, 0.01)
+        assert abs(g1(0.48) - expected_val) < 1e-6
+
+        # explicitly test wrapping
```
aarchiba, line 39: It's nicer to create a separate fest function for testing separate properties like this, so bugs can be more easily localized.

kerrm:
Is there a quick guide for this?  I just copy and paste previous tests and
have no real idea what I'm doing...  I mean, it's clear enough what is
being tested, but not what the pytest structure is buying.

aarchiba:
> Is there a quick guide for this? I just copy and paste previous tests and have no real idea what I'm doing... I mean, it's clear enough what is being tested, but not what the pytest structure is buying.

So `pytest` requires very little structure - just write a function whose name starts with `test_` and that raises an exception or an assertion failure if the test fails. `unittest` required all this elaborate class stuff, and it still works under `pytest`, but it adds boilerplate to no good purpose.

If you want fixtures, that is, code that's run to set up some stuff the same way for each of a bunch of tests, you can do that with classes and their setUp method or you can just write a pytest fixture function that does the setup, then pass its name as an argument. The pytest docs are pretty good: https://docs.pytest.org/en/stable/

kerrm:
Perfect -- that sentence, really, is exactly the context I wanted!  I'll
revise this when I add more tests in the next PR.

tests/test_templates.py
```diff
@@ -0,0 +1,157 @@
...
+
+    assert str(lct).strip() == expected_val.strip()
+
+
+def test_template_simulation():
+    lct = lctemplate.get_gauss2()
+    lct.random(100)
+
+
+def test_simple_fit():
```
aarchiba, line 100: Can you split this into binned and unbinned tests? That way if one breaks we know whether the other also breaks. You can just duplicate the setup code or you can use a pytest fixture (and if the setup is slow you can use a module-level fixture so it only gets run once).

tests/test_templates.py
```diff
@@ -0,0 +1,157 @@
...
+delta   : 0.4800 +\- 0.0000
+Delta   : 0.0200 +\- 0.0000
+"""
+
+    assert str(lct).strip() == expected_val.strip()
+
+
+def test_template_simulation():
+    lct = lctemplate.get_gauss2()
+    lct.random(100)
```
aarchiba, line 97: Is this just generating a random sample drawn from the template? Are there some basic properties you could test, even just whether it's the right length and between 0 and 1?

tests/test_templates.py
```diff
@@ -0,0 +1,157 @@
...
+    # check that that template correctly evaluates weighted sum of comps
+    expected_val = (
+        0.25 * gauss(0.49, 0.5, 0.01)
+        + 0.35 * gauss(0.49, 0.48, 0.01)
+        + (1 - 0.25 - 0.35)
+    )
+    assert abs(lct(0.49) - expected_val) < 1e-6
+
+
+def test_template_string_representation():
```
aarchiba, line 64: These kinds of test can be annoying as they fail if the formatting changes in the slightest or a number changes in the last place. A multiline regular expression might be an easy way to add a little flexibility? Just replace "---------" with "-+" and things like that? On the other hand at least it's easy to know whether a failure is significant.

tests/test_templates.py
```diff
@@ -0,0 +1,157 @@
...
+    # defined as 1/sigma/(2*pi)^0.5 * (-0.5*(x-x0)^2) WRAPPED
+    # because this is a narrow gaussian, it's essentially the same as
+    # an unwrapped version
+    expected_val = gauss(0.50, 0.5, 0.01)
+    assert abs(g1(0.50) - expected_val) < 1e-6
+    expected_val = gauss(0.48, 0.5, 0.01)
+    assert abs(g1(0.48) - expected_val) < 1e-6
+
+
+def test_prim_gauss_wrapping():
```
aarchiba, line 30: Should the number of wraps be based on the width of the Gaussian?

tests/test_templates.py
```diff
@@ -0,0 +1,157 @@
...
+
+from pinttestdata import datadir
+from pint.templates import lcfitters, lcprimitives, lctemplate
+
+
+def gauss(x, x0, s):
+    return 1.0 / s / (2 * np.pi) ** 0.5 * np.exp(-0.5 * (x - x0) ** 2 / s ** 2)
+
+
+def test_prim_gauss_definition():
```
aarchiba, line 13: Have you considered using von Mises functions rather than wrapped Gaussians? They're similar, approximately equal if narrow, and the von Mises distribution satisfies a similar maximum entropy property to the Gaussian. (Also pdf and cdf are implemented in scipy.stats.vonmises.)

tests/test_templates.py
```diff
@@ -0,0 +1,157 @@
+import logging
+
+import numpy as np
+
+from pinttestdata import datadir
+from pint.templates import lcfitters, lcprimitives, lctemplate
+
+
+def gauss(x, x0, s):
```
aarchiba, line 9: Should this be using the bin-averaged gaussian rather than the sampled gaussian? This is available from scipy.stats.norm.cdf. The sampled version gets erratic when the width gets small compared to a sample.

tests/test_templates.py
kerrm: Yes, it's implemented.

tests/test_templates.py
kerrm: Agree it's annoying, but this code hasn't changed in many years.  The data are frozen so in fact a number shouldn't change in the last place!  Frankly it's not worth the time to make it more robust -- I'm likely to be the one breaking it, in which case I'll fix the test!

tests/test_templates.py
kerrm: It would make the derivatives pretty expensive.

kerrm: Moreover, in the internal caching code, I believe it does evaluate the integrated model + derivatives, albeit using something like a Simpson's rule, iirc.  But this test is just of the point evaluation of the profile, anyway.

PR #945, opened 2021-01-12
Correct documentation.

### LINKED ISSUE #943: Incorrect docs
https://github.com/nanograv/PINT/blob/e8ae8bf5067f69ed87cbbcb4a1ef475ce80b9cba/src/pint/observatory/__init__.py#L219-L220

I think the actual allowed values are 'default' and 'ephemeris', not 'astropy'.  @kerrm isn't that right?

aarchiba:
Closes #943

### DIFF
--- a/src/pint/observatory/__init__.py
+++ b/src/pint/observatory/__init__.py
@@ -213,11 +213,11 @@ class Observatory(object):
         # TOA metadata which may be necessary in some cases.
         raise NotImplementedError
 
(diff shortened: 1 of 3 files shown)


paulray:
I see that `grp=None` was removed from several lines for get_TDBs and get_gcrs. The grp argument is used by satellite_obs.py, as recently noted by @kerrm .  So, I'm not sure if those should be deleted or not.

aarchiba:
Ah, sorry, I got everything covered by the test suite but I guess those aren't.

PR #946, opened 2021-01-12
Future cleanup

### LINKED ISSUE #940: Remove future and six
Now that we are Python 3 only, we should probably remove all of the `from __future__` and `six` imports from the codebase.

aarchiba:
Closes #940


aarchiba:
Codecov is infuriating as always. Decreased coverage is because I removed the `__future__` line, which was covered.

paulray:
Thanks for this!  Nice to see all that cleanup!
So classes should not inherit from `object` anymore?

aarchiba:
> Thanks for this! Nice to see all that cleanup!
> So classes should not inherit from `object` anymore?

Classes always inherit from object, unless they inherit from something else. In Python 2, only "new-style classes" inherited from objects, and you had to explicitly say so (`class Thing(object):`) but in python 3 it's no longer needed. (Old-style classes were dropped - it's part of why an incompatible change from 2 to 3 was needed.)

paulray:
Ah, thanks! I had assumed that Python 3 adopted new-style classes so I figured they should have the object inheritance. Now I get it.

aarchiba:
If everyone's happy, this is ready to go; the coverage failure isn't going away and isn't important.

paulray:
I broke black when I fixed a conflict with master. Can you blacken and push?

paulray:
The bad file is src/pint/templates/lcprimitives.py

aarchiba:
Done.

PR #947, opened 2021-01-12
Updated degeneracy check in WidebandTOAFitter

pennucci:
This was probably just missed in the original pass through when DegeneracyWarning was created; the same check is performed in GLSFitter.


aarchiba:
Is there a quick and easy test case? That would help with coverage and also ensure nobody breaks this without noticing.

pennucci:
I had assumed that there were other tests associated with the development of DegeneracyWarning, and this one part of the code was just missed when updating GLSFitter and WidebandTOAFitter.  It passes all current tests, I think; do you think it needs a separate test?

aarchiba:
I mean, you changed a linear algebra exception into a no-exception case, so if it were being tested your modification would produce a test failure. So probably no test case with `norm==0` is never run; coverage confirms that the warning message is never executed. (I *think* I put in test cases for the narrowband case that you might be able to copy and modify, but I suspect I didn't have an easy way offhand to introduce a no-effect parameter for wideband TOAs, which is how I missed this check.)

aarchiba:
You might should also get a warning about a degenerate parameter "combination" out of the SVD if threshold is set to a reasonable value, but I'm not sure that is well tested either - the SVD stuff gave me headaches in the GLS fitter.

paulray:
Can this be merged?

pennucci:
Anne might be waiting for me to put a test in; I can't do this now.

pennucci:
@aarchiba, I'm still pretty new to tests, but I just had a look in test_fitter_error_checking.py and it looks like you wrote a test for this back in mid-December.  That test doesn't get executed, I suppose, because when I test it by hand (on the master branch), it fails, but when I test it with this branch, it passes (since a DegeneracyWarning is raised now).

pennucci:
The test is, obviously, test_jump_everything_wideband().

aarchiba:
That test passes for me on master.

pennucci:
Argh, this was some pebkac that I can't reproduce.  Yes, it does pass.  However, where does the 'first' DegeneracyWarning get tested, namely the one that I mirrored in this PR, which gets raised in the event of a null column vector in the design matrix of GLSFitter?  The easy example test cases I can think of get picked up in validate_toas, maybe model.validate.

aarchiba:
> Argh, this was some pebkac that I can't reproduce. Yes, it does pass. However, where does the 'first' DegeneracyWarning get tested, namely the one that I mirrored in this PR, which gets raised in the event of a null column vector in the design matrix of GLSFitter? The easy example test cases I can think of get picked up in validate_toas, maybe model.validate.

It's possible that this isn't tested any more, that I wrote a test that triggered it but then improved `validate` so that that test case is caught elsewhere. But I think I was able to hit that by choosing infinite-frequency TOAs and fitting for DM, or TOAs at the barycenter and fitting for position. Degenerate linear combinations are trickier but not essential here.

aarchiba:
In fact I think I may have forgotten to put this check in because I was thinking in terms of infinite-frequency (or all-the-same-frequency) TOAs and fitting for DM, and obviously that won't work for wideband TOAs. But the position fitting for barycentered TOAs should trigger irrelevant-parameter warnings for both kinds of TOA.

aarchiba:
Incidentally python >= 3.8 has the `breakpoint()` command, which will drop you into the python debugger when executed, so you can inspect the call stack and figure out where that line is being called from.

pennucci:
Good call, that's easy.  I just wrote a test for both fitters.  And yeah, the lines are printed when the errors are raised.  I had to match against the slightly different warning string, since the other DegeneracyWarning also is raised.

tests/test_fitter_error_checking.py
```diff
@@ -134,6 +134,22 @@ def test_jump_everything_wideband():
...
+def test_null_vector():
+    model = get_model(io.StringIO("\n".join([par_base])))
+    model.free_params = ["ELONG", "ELAT"]
+    toas = make_fake_toas(58000, 58900, 10, model, obs="barycenter", freq=1400.0, dm=15.0, dm_error=1e-4*u.pc*u.cm**-3)
+    fitter = pint.fitter.GLSFitter(toas, model)
+    with pytest.warns(pint.fitter.DegeneracyWarning, match=r".*degeneracy.*following parameter.*ELONG\b"):
+        fitter.fit_toas(threshold=1e-14)
+    for p in fitter.model.free_params:
+        assert not np.isnan(fitter.model[p].value)
+    fitter = pint.fitter.WidebandTOAFitter(toas, model)
```
aarchiba, line 146: Could you put the GLSFitter and the WidebandTOAFitter in separate tests? That way if the tests fail we'll know which one, or if it's both.

aarchiba: You could actually parameterize this test over WidebandTOAFitter, GLSFitter, and also WLSFitter with `@pytest.mark.parametrize()` to have one piece of code test all three.

pennucci:
I had a feeling you might prefer separate test functions.  But yeah, pytest.mark.parameterize is the tool to use.  I'm happy to be learning about this stuff...!  However, iirc, WLSFitter doesn't raise DegeneracyError in quite this way, as it doesn't have a step where it normalizes the design matrix.  I'll have a look tomorrow about these things.

pennucci:
@aarchiba I parameterized the test, but only over GLS and WB fitters, since there's no analogous normalization of the design matrix in WLSFitter.  If that dev happens, it can simply be added to this test parameterization.  Also, previously the CI failed on black because I didn't blacken the code.  But I just tried, and black didn't do anything, so hopefully it doesn't fail with the new code.

aarchiba:
> @aarchiba I parameterized the test, but only over GLS and WB fitters, since there's no analogous normalization of the design matrix in WLSFitter. If that dev happens, it can simply be added to this test parameterization.

There is, actually, and no warning is emitted: https://github.com/nanograv/PINT/blob/7a56f3f77f2a512cd57b014b810d4bfba454964f/src/pint/fitter.py#L941

I may have left it out because the SVD threshold will pick it up; this is possible but more complicated (and broke a bunch of test cases) for GLS-based fitters. (They use a less-numerically-stable calculation based on the normal equations, which squares the condition number, so condition number problems happen much sooner, but fixing them breaks TEMPO/TEMPO2 compatibility.) I don't know whether it's worth adding the more specific zero-column check.

pennucci:
Ah, I missed those lines, as I was looking for the same for loop as in the other fitters.  Would you like me to try and add some code to that effect, or should we leave this for now?

aarchiba:
Test failure is unrelated. Looks good to me.

PR #948, opened 2021-01-14
WIP: Added chi2 setters, ensure fitter.resids.chi2 attributes set correctly after GLS/Wideband fit

pennucci:
This also replaces the calls the minimize_func in GLS and Wideband fitters with the more transparent calls to update pars and resids; this avoids a redundant call to calc_chi2 in some places.  The WIP status is mostly because this doesn't address why the baseclass Residual calc_chi2 function doesn't return the right value; I think this is something common to the zero-iteration calculations in both GLS fitters.


src/pint/residuals.py
```diff
@@ -679,6 +691,13 @@ def data_error(self):
             errors.append((rs.residual_type, rs.get_data_error()))
         return collections.OrderedDict(errors)
 
+    def calc_chi2(self):
```
aarchiba, line 694: I think this should call `calc_chi2` on the components - that is, it should not use the cache. The point of `calc_chi2` is to force recomputation, and this implementation doesn't.

Alternatively, you could have CombinedResiduals not cache anything, just compute the sum of the chi2 attributes of the parts every time it is requested. This accomplishes the same thing (and is what is there now).

src/pint/fitter.py
```diff
@@ -1527,5 +1532,9 @@ def fit_toas(self, maxiter=1, threshold=0, full_cov=False):
                 self.resids.noise_resids = noise_resids
 
         self.update_model(chi2)
+        self.resids.chi2 = chi2
+
+        # A cheat for now:
+        self.resids.toa.chi2 = self.resids.chi2 - self.resids.dm.chi2
```
aarchiba, line 1538: You could just use `chi2-self.resids.dm.chi2` here and avoid needing a setter in CombinedResiduals.

src/pint/fitter.py
```diff
@@ -1181,7 +1181,9 @@ def fit_toas(self, maxiter=1, threshold=0, full_cov=False):
                 fitpv[pn] = np.longdouble((pv + dpv) / fitp[pn].units)
                 # NOTE We need some way to use the parameter limits.
                 fitperrs[pn] = errs[uind]
-            self.minimize_func(list(fitpv.values()), *list(fitp.keys()))
+            newparams = dict(zip(list(fitp.keys()), list(fitpv.values())))
```
aarchiba, line 1184: Why this change?

aarchiba:
I think it would be useful to have tests - at a minimum check that the `.chi2` attribute agrees with the value returned from the fitter (including for wideband fitting), possibly also (marked with @pytest.mark.xfail) a test confirming that it gives the same result as `calc_chi2`. Ideally we could add some tests that compute a chi-squared against a known analytical value, or check the statistic has the right distribution, but those tests are more complicated and annoying.

src/pint/fitter.py
pennucci: You're right, given the other comment.

src/pint/residuals.py
pennucci: Yeah, you're right -- the alternatively part.  I was messing with something more involved for this calc_chi2 (calling WidebandTOAFitter in some cases), but turns out it didn't do what I wanted.  So I can revert.

src/pint/fitter.py
pennucci: It's more transparent, but more importantly, within minimize_func, the Residuals get re-instantiated, so self.resids._chi2 gets reset to None, so the return statement of that function winds up calling calc_chi2, which is not necessary (especially given this dev) and time consuming.

aarchiba: Does this behave any differently? It seems like both call `set_params` followed by `update_resids`? (One could check using `breakpoint()`)

aarchiba:
Actually looking at the code we probably *do* want to recompute chi-squared once the fitter is done - the GLSFitter returns a linearized version of the chi-squared, which might or might not be very close to the one with everything evaluated at the returned parameter values.

pennucci:
> I think it would be useful to have tests - at a minimum check that the `.chi2` attribute agrees with the value returned from the fitter (including for wideband fitting), possibly also (marked with @pytest.mark.xfail) a test confirming that it gives the same result as `calc_chi2`. Ideally we could add some tests that compute a chi-squared against a known analytical value, or check the statistic has the right distribution, but those tests are more complicated and annoying.

Absolutely; I just can't write those right now...

src/pint/fitter.py
pennucci: Yes it does; update_resids resets the residuals in the same way, however there's no querying for self.resids.chi2 at the end, which is the return statement of minimize_func (since it's purpose is to return a chi2, which we don't need here anyway) and which calls calc_chi2 because self.resids.chi2 at that point is None.  This fix makes the updating of resids and pars transparent, and then a few lines later it explicitly sets self.resids.chi2 to the end result of the fit, it doesn't rely on calc_chi2 at all.

pennucci:
I'm closing this, as we resolved to not have a chi2 setter, at least for now, and the other changes (i.e., w.r.t. minimize_func) were absorbed by an earlier merged PR #949.  Can be re-opened if need be.

PR #950, opened 2021-01-14
Allow renaming of .tim files

aarchiba:
If `.tim` files are renamed, pickles should be smart enough to invalidate themselves, if necessary, even if the original files are unchanged.


aarchiba:
Ready to merge if everyone likes it.

Issue #810, opened 2020-11-13
Weighted mean of post-fit residuals is not using scaled uncertainties; affecting chi2 calculation.


pennucci:
When I calculated the weighted mean of post-fit residuals it is ~identically zero when **not** applying EFAC/EQUADs, but presumably these factors should be used when fitting the mean phase, and thus the weighted mean should be zero when using the scaled TOA uncertainties as weights.  I found this using WidebandTOAFitter, but I tried with GLSFitter and it's the same, so I think it's a general issue.  This also affects the chi-squared calculation, since it uses the post-fit residuals; while the chi-squared calculation doesn't require zero-mean, I think this is implied in any set of post-fit timing residuals.

demorest:
For GLS fits it's slightly more complicated.  chi2 requires the full noise model not just residuals and uncertainties (either scaled or not).  This is returned by the fitter, and I think it's best to use that value rather than duplicating the calculation elsewhere.  This means at least for chi2 purposes it doesn't matter where the zero-point of the residuals is.

aarchiba:
This is indeed a bug where mean subtraction (when using a weighted mean) uses `get_errors` and not `get_data_error`. Fixed as part of a PR I'm working on, which should make mean subtraction/whitened residuals more readily available and consistent.

PR #953, opened 2021-01-18
Fix GLS and wideband chi-squared

aarchiba:
Computation of the GLS and wideband chi-squared were wrong before in red-noise cases. Also the GLS computation incorrectly froze all parameters.

### DIFF
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -776,7 +776,7 @@ class WidebandTOAResiduals(CombinedResiduals):
         m = copy.deepcopy(self.model)
         m.free_params = []
         f = WidebandTOAFitter(
-            self.toas, m, additional_args=dict(toa=dict(track_mode="use_pulse_numbers"))
+            self.toas, m, additional_args=dict(toa=dict(track_mode=self.toa.track_mode))
         )
         try:
             return f.fit_toas(maxiter=1, full_cov=full_cov)
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -347,7 +347,7 @@ class Residuals:
             from pint.fitter import GLSFitter
 
             m = copy.deepcopy(self.model)
-            m.free_parameters = []
+            m.free_params = []
             f = GLSFitter(self.toas, m, residuals=self)
             try:
                 return f.fit_toas(maxiter=1, full_cov=full_cov)
@@ -729,6 +729,7 @@ class WidebandTOAResiduals(CombinedResiduals):
             self.toas, self.model, residual_type="toa", **toa_resid_args
         )
         dm_resid = Residuals(self.toas, self.model, residual_type="dm", **dm_resid_args)
+        self._chi2 = None
 
         super().__init__([toa_resid, dm_resid])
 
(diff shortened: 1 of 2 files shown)


luojing1211:
This looks good. Is it a good idea to separate these fitting functions from the fitter, so one doesn't need to build a fitter in the residual? Of course, it would be a separate PR.

pennucci:
It does look good.  I'm still curious about the choice to not cache the chi2 value at the end of fit_toas, but I guess I can understand why.

luojing1211:
It matches how TOAResidual's Chi2. If one only uses it in the fitter, caching chi2 is not a problem since the fitter rebuilds the residual class after every fit. However, if one modifies the model in the residual class, `Residusal.model`, the cached chi2 will be wrong.

aarchiba:
> It does look good. I'm still curious about the choice to not cache the chi2 value at the end of fit_toas, but I guess I can understand why.

That chi2 value is computed based on pre-fit values of the orbit. It will be valid, for example, even if the output SINI > 1. It will not be the same value one would obtain if one ran calc_chi2 on the residuals object.

aarchiba:
> This looks good. Is it a good idea to separate these fitting functions from the fitter, so one doesn't need to build a fitter in the residual? Of course, it would be a separate PR.

I don't think the Fitter introduces any appreciable extra machinery. And it is conceptually clearer here: you really do need to fit for the noise instance to get a useful chi-squared, and the way you fit for things in PINT is using a Fitter.

aarchiba:
> > It does look good. I'm still curious about the choice to not cache the chi2 value at the end of fit_toas, but I guess I can understand why.
> 
> That chi2 value is computed based on pre-fit values of the orbit. It will be valid, for example, even if the output SINI > 1. It will not be the same value one would obtain if one ran calc_chi2 on the residuals object. This is demonstrated in one of the test cases.

There's an argument to be made that `fit_toas` should not return its `chi2` value at all, because unless all parameters are frozen it is not the chi-squared value corresponding to any specific set of model parameters.

paulray:
> There's an argument to be made that fit_toas should not return its chi2 value at all, because unless all parameters are frozen it is not the chi-squared value corresponding to any specific set of model parameters.

I haven't been following this discussion, but I don't quite understand what this means. Can you explain?  In any chi2 minimization fitter, don't you end up with a set of final parameters and a chi2 value that corresponds to the data-model for that set of parameters?

There is a mention above about computing chi2 from pre-fit values. Why would that be done? Don't you normally want chi2 after a fit (unless you are trying to find delta-chi2 from the fitting process)?

scottransom:
I agree with @paulray.  I'm not understanding that.  Heck, we are using chi^2 as the function to be minimized by the fitter.  So it is definitely meaningful in some way.  The main thing is that we would like it to mean what most people *think* chi^2 usually means.

@aarchiba Are you specifically talking about returning a chi^2 pre-fit when an actual fit is not done?  I totally think it is valid to ask what the chi^2 is of any model with any set of parameters (frozen or not, it doesn't matter if you aren't fitting).  But *post-fit*, we should really be able to know the true iterated/converged post-fit chi^2 and # of DOF.  And that should (IMO) totally be returned by the fitter after fitting.

aarchiba:
> > There's an argument to be made that fit_toas should not return its chi2 value at all, because unless all parameters are frozen it is not the chi-squared value corresponding to any specific set of model parameters.
> 
> I haven't been following this discussion, but I don't quite understand what this means. Can you explain? In any chi2 minimization fitter, don't you end up with a set of final parameters and a chi2 value that corresponds to the data-model for that set of parameters?
> 
> There is a mention above about computing chi2 from pre-fit values. Why would that be done? Don't you normally want chi2 after a fit (unless you are trying to find delta-chi2 from the fitting process)?

No. Well, in any reasonable fitting algorithm you would, but we do not use a fitting algorithm. PINT (and I think tempo/tempo2) does a "fit" by computing residuals and derivatives at the pre-fit model parameters, then treating that as a linear least-squares fitting problem to compute the needed change in the model parameters. It then adjusts the model in this way and returns the new parameters. It does not ever evaluate the model at these new parameter values (which are often invalid); the chi-squared value it returns is an estimate of what the chi-squared should be if the linear least-squares problem were exact. 

The new residuals object evaluates the model at the new parameter values (often obtaining nothing but NaN), and if `.calc_chi2` is run, computes the chi-squared value at those model parameters.

aarchiba:
> I agree with @paulray. I'm not understanding that. Heck, we are using chi^2 as the function to be minimized by the fitter. So it is definitely meaningful in some way. The main thing is that we would like it to mean what most people _think_ chi^2 usually means.
> 
> @aarchiba Are you specifically talking about returning a chi^2 pre-fit when an actual fit is not done? I totally think it is valid to ask what the chi^2 is of any model with any set of parameters (frozen or not, it doesn't matter if you aren't fitting). But _post-fit_, we should really be able to know the true iterated/converged post-fit chi^2 and # of DOF. And that should (IMO) totally be returned by the fitter after fitting.

The issue right now is that we need a chi-squared value with the correct statistics in order to support the F test. For models with uncorrelated errors, this just means we need a weighted-mean-subtracted set of residuals, or to subtract the weighted mean in the chi-squared computation. For models with correlated errors, to get reasonable statistics we need to operate on the whitened residuals, which requires a fitting procedure (since our residuals object is not whitened - as I understand it whitening is subtracting a best-fit noise instance, analogous to subtracting the weighted mean).

pennucci:
I think I understand that but assuming convergence (maybe not a good assumption, but let's call that "post-fit"), then the values returned by either fit_toas or calc_chi2 should give you the right statistics I think.  As per my original comment regarding "pre-fit" calculations -- all I'm getting at is the simple intuition that the chi2 value is related to the evaluation of the likelihood, so given the data, some model parameters, including a noise model, you should be able to get a value for the likelihood without a fit.  I understand that the GLS tricks we use make this hand-wavy, but that's what I was getting at.  And w.r.t. the weighted mean, see #810; a number of us have noticed in v0 timing that post-fit (whitened) residuals have a non-zero mean.

Issue #952, opened 2021-01-18
Turn tracking should be automatic by default


aarchiba:
Currently PINT is able to use `track_mode="use_pulse_numbers"` to lock in phase wraps, but it is cumbersome to specify this. Much easier would be a mode that used pulse numbers if they were available but not if they weren't; this would be a sensible default.

luojing1211:
Can we close this?

PR #954, opened 2021-01-18
Use pulse numbers if present

aarchiba:
Currently pulse numbers are ignored unless `track_mode="use_pulse_numbers"` is specified. With this PR, TOAs with pulse numbers take advantage of them.


aarchiba:
Closes #952

PR #955, opened 2021-01-20
Fix an "AttributeError: can't set attribute" error relating to the dof property

fjankowsk:
Hello,

This pull request fixes an "AttributeError: can't set attribute" error that I got when using the MCMCFitter under python 3.8.4.

The `dof` attribute is an instance property, and there is no setter defined for it, so it should be immutable. Perhaps a private variable `__dof` or similar could be used like is done for `chi2`; otherwise, the degrees of freedom will be recalculated on every call (as it is now, from what I can see).

I got the same error with PINT release 0.8.1.

OK?

Cheers.
Fabian


aarchiba:
Could you add a very short test case to this? This bug would have been caught if anywhere in our test suite the Residuals `update` method were called. Maybe just a test that creates a residuals object, queries the `dof`, changes the model (just add something small to a parameter), calls `update`, and then checks the `dof` again (shouldn't have changed)? Should be easy to use 10 TOAs worth of fake data and some handy par file.

paulray:
Let's merge this and make an issue to add a test case for this problem.

PR #963, opened 2021-01-28
Make ELL1H model accept the H3=0 and H4=0

luojing1211:
@bshapiroalbert Can you try this branch?

### DIFF
--- a/src/pint/models/binary_ell1.py
+++ b/src/pint/models/binary_ell1.py
@@ -265,8 +265,8 @@ class BinaryELL1H(BinaryELL1Base):
     def validate(self):
         """Parameter validation."""
         super(BinaryELL1H, self).validate()
-        if self.H3.quantity is None:
-            raise MissingParameter("ELL1H", "H3", "'H3' is required for ELL1H model")
+        # if self.H3.quantity is None:
+        #     raise MissingParameter("ELL1H", "H3", "'H3' is required for ELL1H model")
         if self.SINI.quantity is not None:
             warn("'SINI' will not be used in ELL1H model. ")
         if self.M2.quantity is not None:
--- a/src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
+++ b/src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
@@ -54,7 +54,10 @@ class ELL1Hmodel(ELL1BaseModel):
 
     def delayS(self):
         if set(self.fit_params) == set(["H3", "H4"]):
-            stigma = self.H4 / self.H3
+            if self.H3 == 0.0:
+                stigma = 0.0
+            else:
+                stigma = self.H4 / self.H3
         elif set(self.fit_params) == set(["H3", "STIGMA"]):
             stigma = self.STIGMA
         elif set(self.fit_params) == set(["H3"]):
@@ -167,11 +170,16 @@ class ELL1Hmodel(ELL1BaseModel):
             return 0, basis_func
 
     def d_STIGMA_d_H4(self):
+        if self.H3 == 0.0:
+            raise ValueError("To fit for H4, H3 should be significant(not 0.0).")
         return 1.0 / self.H3
 
     def d_STIGMA_d_H3(self):
         if set(self.fit_params) == set(["H3", "H4"]):
-            d_stigma_d_H3 = -self.H4 / self.H3 / self.H3
+            if self.H3 == 0.0:
+                d_stigma_d_H3 = 0.0
+            else:
+                d_stigma_d_H3 = -self.H4 / self.H3 / self.H3
         elif set(self.fit_params) == set(["H3", "STIGMA"]):
             d_stigma_d_H3 = 0.0
         elif set(self.fit_params) == set(["H3"]):
@@ -388,7 +396,10 @@ class ELL1Hmodel(ELL1BaseModel):
 
     def d_delayS_d_par(self, par):
         if set(self.fit_params) == set(["H3", "H4"]):
-            stigma = self.H4 / self.H3
+            if self.H3 == 0:
+                stigma = 0.0
+            else:
+                stigma = self.H4 / self.H3
         elif set(self.fit_params) == set(["H3", "STIGMA"]):
             stigma = self.STIGMA
         elif set(self.fit_params) == set(["H3"]):
--- a/src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
+++ b/src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
@@ -58,7 +58,7 @@ class ELL1Hmodel(ELL1BaseModel):
                 if self.H4 == 0.0:
                     stigma = 0.0
                 else:
-                    raise ValueError("To use H4, H3 needs to be significant(H3 >= H4).")
+                    raise ValueError("To use H4, H3 needs to be significant(H3 != 0).")
             else:
                 stigma = self.H4 / self.H3
         elif set(self.fit_params) == set(["H3", "STIGMA"]):
--- a/src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
+++ b/src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
@@ -171,7 +171,7 @@ class ELL1Hmodel(ELL1BaseModel):
 
     def d_STIGMA_d_H4(self):
         if self.H3 == 0.0:
-            raise ValueError("To fit for H4, H3 should be significant(not 0.0).")
+            return 0.0
         return 1.0 / self.H3
 
     def d_STIGMA_d_H3(self):
--- a/src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
+++ b/src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
@@ -55,7 +55,10 @@ class ELL1Hmodel(ELL1BaseModel):
     def delayS(self):
         if set(self.fit_params) == set(["H3", "H4"]):
             if self.H3 == 0.0:
-                stigma = 0.0
+                if self.H4 == 0.0:
+                    stigma = 0.0
+                else:
+                    ValueError("To use H4, H3 needs to be significant(H3 >= H4).")
             else:
                 stigma = self.H4 / self.H3
         elif set(self.fit_params) == set(["H3", "STIGMA"]):
@@ -229,7 +232,10 @@ class ELL1Hmodel(ELL1BaseModel):
 
     def d_ELL1H_fourier_harms_d_par(
         self, selected_harms, phi, stigma, par, factor_out_power=0
-    ):
+    ):if self.H4 == 0.0:
+                    stigma = 0.0
+                else:
+                    ValueError("To use H4, H3 needs to be significant(H3 >= H4).")
         """This is a overall derivative  function.  """
         # Find the right derivative  function for fourier components
         df_name = "d_fourier_component_d_" + par.lower()
@@ -397,7 +403,10 @@ class ELL1Hmodel(ELL1BaseModel):
     def d_delayS_d_par(self, par):
         if set(self.fit_params) == set(["H3", "H4"]):
             if self.H3 == 0:
-                stigma = 0.0
+                if self.H4 == 0.0:
+                    stigma = 0.0
+                else:
+                    ValueError("To use H4, H3 needs to be significant(H3 >= H4).")
             else:
                 stigma = self.H4 / self.H3
         elif set(self.fit_params) == set(["H3", "STIGMA"]):
--- a/src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
+++ b/src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
@@ -232,10 +232,7 @@ class ELL1Hmodel(ELL1BaseModel):
 
     def d_ELL1H_fourier_harms_d_par(
         self, selected_harms, phi, stigma, par, factor_out_power=0
-    ):if self.H4 == 0.0:
-                    stigma = 0.0
-                else:
-                    ValueError("To use H4, H3 needs to be significant(H3 >= H4).")
+    ):
         """This is a overall derivative  function.  """
         # Find the right derivative  function for fourier components
         df_name = "d_fourier_component_d_" + par.lower()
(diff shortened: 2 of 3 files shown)


bshapiroalbert:
Thanks for putting this quick fix in there @luojing1211! I just ran one quick test, but it seems to work as expected (see test output below with respect to the H3, H4 line). Thanks!

Screen Shot 2021-01-28 at 11 35 31 AM

luojing1211:
Does that mean, both H3 and H4 are significant for this pulsar?

bshapiroalbert:
Yes, for this pulsar that is correct. It's saying that removing just H4 makes a significant change, as does removing H3 and H4, so they are thus important and should be added to the model. This is expected for this pulsar (J1832) I think. But before when removing H3 and H4 (which in the `ftest()` function just sets them to zero and freezes them), it would just return an error, so it's showing that this is apparently working now :)

src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
```diff
@@ -54,7 +54,10 @@ def __init__(self):
 
     def delayS(self):
         if set(self.fit_params) == set(["H3", "H4"]):
-            stigma = self.H4 / self.H3
+            if self.H3 == 0.0:
```
aarchiba, line 57: Should this raise an exception or set stigma to `np.inf` if H4 is not zero?

src/pint/models/binary_ell1.py
```diff
@@ -265,8 +265,8 @@ def setup(self):
     def validate(self):
         """Parameter validation."""
         super(BinaryELL1H, self).validate()
-        if self.H3.quantity is None:
-            raise MissingParameter("ELL1H", "H3", "'H3' is required for ELL1H model")
+        # if self.H3.quantity is None:
```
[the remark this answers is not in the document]
luojing1211: Yes. The default value for H3 is zero in the stand-along-binary

aarchiba:
This should have (very simple, with perhaps 10 fake TOAs) test cases with:

- H3, H4 not set
- H4 set to zero, H3 zero
- H4 not zero, H3 zero

There should probably also be some test cases to verify that fitting does not behave unexpectedly when the initial values have H3=0 or unset, for H4 zero or not, and the two parameters with various combinations of frozen and unfrozen.

These tests can probably be set up with `pytest.mark.parametrize` or a `pytest.fixture` to avoid repeating setup code.

luojing1211:
Yes. That is why this is still a WIP PR

tests/test_ell1h.py
```diff
@@ -12,111 +13,239 @@
...
+      TASC 54889.991808565  1 0.000000012
+      EPS1 0.0000025554  1 0.0000002783
+      EPS2 0.0000036160  1 0.0000000847
+      H3 2.7507208E-7  1       1.5114416E-7
+      H4 2.0262048E-7  1       1.1276173E-7
+      """
+
+
+@pytest.fixture
+def simple_par_str():
```
aarchiba, line 45: What is this fixture for? Can't they just access the variable simple_par?

tests/test_ell1h.py
```diff
@@ -12,111 +13,239 @@
...
+    return simple_par
+
+
+@pytest.fixture
+def toasJ0613():
+    toas = toa.get_TOAs("J0613-0200_NANOGrav_9yv1.tim", ephem="DE421", planets=False)
+    return toas
+
+
+@pytest.fixture
```
aarchiba, line 55: If you're confident these will not be modified by any test you can use `@pytest.fixture(scope="module")` to arrange for the TOAs to be loaded only once rather than for every test that uses them.

tests/test_ell1h.py
```diff
@@ -12,111 +13,239 @@
...
+def simple_par_str():
+    return simple_par
+
+
+@pytest.fixture
+def toasJ0613():
+    toas = toa.get_TOAs("J0613-0200_NANOGrav_9yv1.tim", ephem="DE421", planets=False)
```
aarchiba, line 51: No need to assign this to a variable, you can just return it.

tests/test_ell1h.py
```diff
@@ -12,111 +13,239 @@
...
+        diff = np.abs(p.value - op.value)
+        sigma = diff / op.uncertainty_value
+        # Fit th
+        assert sigma < 0.7, "refit %s is %lf sigma different from original value" % (
+            pn,
+            sigma,
+        )
+
+
+def test_no_H3_H4(toasJ0613, simple_par_str):
```
aarchiba, line 173: the fixture `simple_par_str` is never used and can be omitted here

tests/test_ell1h.py
```diff
@@ -12,111 +13,239 @@
...
+        "H4 2.0262048E-7  1       1.1276173E-7", "H4 0  1  0"
+    )
+    H3H4_zero = H4_zero.replace("H3 2.7507208E-7  1       1.5114416E-7", "H3 0  1  0")
+    H3H4_zero_model = model.get_model(StringIO(H3H4_zero))
+    assert H3H4_zero_model.H3.value == 0.0
+    assert "H3" in H3H4_zero_model.free_params
+    assert H3H4_zero_model.H4.value == 0.0
+    assert "H4" in H3H4_zero_model.free_params
+    f = ff.WLSFitter(test_toas, H3H4_zero_model)
+    # Fitting H4 with H3 == 0, will give a ValueError
```
aarchiba, line 232: Wait, shouldn't this work? Like, shouldn't  one be able to start with H3==H4==0 and fit for both? Presumably the fit value for H3 will be nonzero

scottransom: Yeah, I agree with @aarchiba here.  We should be able to start a fit with H3 == H4 == 0 and get a reasonable result for a system where there is a small amount of Shapiro signal.  Users will *definitely* try this.

aarchiba: It will probably require special-casing the derivatives, maybe just returning zero for derivatives with respect to H4 if H3 is zero. But that should work (fitting will require more than one step, almost certainly).

tests/test_ell1h.py
```diff
@@ -12,111 +13,223 @@
...
+            assert np.nanmax(relative_diff) < tol, msg
+        else:
+            continue
+
+
+def test_J0613_H4(toasJ0613, modelJ0613):
+    log = logging.getLogger("TestJ0613.fit_tests")
+    f = ff.GLSFitter(toasJ0613, modelJ0613)
+    f.fit_toas()
+    f.model.free_params = ("H3", "H4")
```
aarchiba, line 138: This line does not appear to do anything since the `fit_toas` call is before this?

tests/test_ell1h.py
```diff
@@ -12,111 +13,223 @@
...
+            log.debug(
+                "derivative relative diff for %s, %lf"
+                % ("d_delay_d_" + p, np.nanmax(relative_diff).value)
+            )
+            assert np.nanmax(relative_diff) < tol, msg
+        else:
+            continue
+
+
+def test_J0613_H4(toasJ0613, modelJ0613):
```
aarchiba, line 134: Can you give some indication - is this a case where H3 and H4 are both nonzero?

tests/test_ell1h.py
```diff
@@ -12,111 +13,223 @@
...
+        assert sigma < 0.7, "refit %s is %lf sigma different from original value" % (
+            pn,
+            sigma,
+        )
+
+
+def test_no_H3_H4(toasJ0613):
+    """ Test no H3 and H4 in model.
+    """
+    no_H3_H4 = simple_par.replace("H4 2.0262048E-7  1       1.1276173E-7", "")
```
aarchiba, line 169: This will silently stop working if the value or uncertainty change in any digit, or the spacing changes. Perhaps a regular expression to replace r"^H4 .*$" with nothing would be more reliable?

tests/test_ell1h.py
```diff
@@ -12,111 +13,223 @@
...
+    print(no_H3_H4)
+    no_H3_H4_model = model.get_model(StringIO(no_H3_H4))
+    assert no_H3_H4_model.H3.value == None
+    assert no_H3_H4_model.H4.value == None
+    test_toas = toasJ0613[::20]
+    f = ff.WLSFitter(test_toas, no_H3_H4_model)
+    f.fit_toas()
+
+
+def test_H3_H4_pairs(toasJ0613):
```
aarchiba, line 180: Er, this test only seems to test one model? Which of H3, H4 are zero here? Nicest if the function name makes that clear (nobody ever needs to call these functions so long names are not a problem).

tests/test_ell1h.py
```diff
@@ -12,111 +13,223 @@
...
+def zero_H4(toasJ0613):
+    H4_zero = simple_par.replace("H4 2.0262048E-7  1       1.1276173E-7", "H4 0  1  0")
+    H4_zero_model = model.get_model(StringIO(H4_zero))
+    assert H4_zero_model.H4.value == 0.0
+    assert H4_zero_model.H3.value != 0.0
+    f = ff.WLSFitter(test_toas, H4_zero_model)
+    f.fit_toas()
+
+
+def zero_H3(toasJ0613):
```
aarchiba, line 200: Shouldn't this test raise an exception? Isn't the model invalid if H4!=0 but H3==0?

tests/test_ell1h.py
luojing1211: Since the simple_par is a defined string in the beginning. I don't expect, it gets changed.

tests/test_ell1h.py
luojing1211: Fitting H3 and H4 changes A1 and EPS1 3 to 4 sigma and caused a failure on the parameter change. The value of H3 and H4 are derived from sini and M2.

aarchiba: Er, what I meant was: setting the free parameters *after* the fitting instead of before doesn't do anything and seems like a mistake?

tests/test_ell1h.py
```diff
@@ -12,111 +13,204 @@
...
+        fitting result will not be checked here.
+    """
+    simple_model = model.get_model(StringIO(simple_par))
+
+    test_toas = toasJ0613[::20]
+    f = ff.WLSFitter(test_toas, simple_model)
+    f.fit_toas()
+
+
+def zero_H4(toasJ0613):
```
aarchiba, line 171: This needs to be called `test_zero_H4`  for pytest to find it; you can use `pytest -v tests/test_ell1h.py` to have it print out the names of the tests it is actually running.

luojing1211: Good catch.

src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
```diff
@@ -54,7 +54,13 @@ def __init__(self):
...
+                if self.H4 == 0.0:
+                    stigma = 0.0
+                else:
+                    raise ValueError("To use H4, H3 needs to be significant(H3 >= H4).")
```
aarchiba, line 61: Wait, is the condition really H3 >= H4, as this claims, or is it just H3 != 0? The condition described in the exception message does not seem to be the condition the code tests for.

tests/test_ell1h.py
```diff
@@ -12,111 +13,202 @@
...
+    assert H3_zero_model.H4.value != 0.0
+    H3_zero_model.H4.frozen = False
+    test_toas = toasJ0613[::20]
+    with pytest.raises(ValueError):
+        f = ff.WLSFitter(test_toas, H3_zero_model)
+
+
+def test_zero_H3_H4_fit_H3_H4(toasJ0613):
+    H4_zero = simple_par.replace("H4 2.0262048E-7  1       1.1276173E-7", "H4 0  1  0")
+    H3H4_zero = H4_zero.replace("H3 2.7507208E-7  1       1.5114416E-7", "H3 0  1  0")
```
aarchiba, line 194: These lines appear to be unnecessary.

tests/test_ell1h.py
```diff
@@ -12,111 +13,202 @@
...
+def test_zero_H3_H4_fit_H3(toasJ0613):
+    H3H4_zero2_model = model.get_model(StringIO(simple_par))
+    H3H4_zero2_model.H3.value = 0.0
+    H3H4_zero2_model.H3.frozen = False
+    H3H4_zero2_model.H4.value = 0.0
+    H3H4_zero2_model.H4.frozen = True
+    test_toas = toasJ0613[::20]
+    f = ff.WLSFitter(test_toas, H3H4_zero2_model)
+    # This should work
+    f.fit_toas()
```
aarchiba, line 214: Does fitting for H3 actually work? If all the derivatives with respect to H3 are zero (looks like they might be?) it might look like a completely degenerate parameter and the fitter would not change it at all, so fitting from this starting point would never work even if it didn't raise an exception.

aarchiba:
This seems to have a few unaddressed comments and has been untouched for a while.

tests/test_ell1h.py
luojing1211: Added a derivative test that d_delay_d_H3 > 0 for that set of TOAs.

aarchiba: Er, could you point to the code? The only derivative test I see is the one above that checks the claimed derivative against a numerical derivative for the special case of H4 and STIGMA both zero.

tests/test_ell1h.py
```diff
@@ -12,111 +13,203 @@
...
+
+def test_zero_H3_H4_fit_H3_H4(toasJ0613):
+    H3H4_zero_model = model.get_model(StringIO(simple_par))
+    H3H4_zero_model.H3.value = 0.0
+    H3H4_zero_model.H4.value = 0.0
+    H3H4_zero_model.H3.frozen = False
+    H3H4_zero_model.H4.frozen = False
+    # Make sure H3's derivative does not return zero.
+    test_toas = toasJ0613[::20]
+    d_h3 = H3H4_zero_model.d_delay_d_param(test_toas, "H3")
```
luojing1211, line 200: Here is the H3 derivative test.

aarchiba: Wait why do we expect the mean of the derivative to be positive?

aarchiba:
The MacOS failure will go away if you fold in the changes from master - a recent PR fixed the hypothesis settings so this should no longer occur.

luojing1211:
@aarchiba Can we merge this?

Issue #964, opened 2021-02-02
numpy 1.20.0 causes failures


paulray:
As @champagne-cmd discovered, numpy 1.20.0 is out and our CI tests started using it. This causes some problems.

`numpy.genfromtext()` changed how it works and that causes `test_jump.py` to fail because it returns a list of arrays instead of just an array.

Also, we get lots of deprecation warnings from using `np.int` and `np.float` which they now recommend against.

Here are the full numpy 1.20 release notes:
https://numpy.org/devdocs/release/1.20.0-notes.html

paulray:
Here are examples of some of the new warnings and deprecations. I think we should just fix these. There seems to be no downside.

tests/test_ddk.py:63
  /home/runner/work/PINT/PINT/tests/test_ddk.py:63: SyntaxWarning: "is" with a literal. Did you mean "=="?
    if type(par).__name__ is "boolParameter":

tests/test_utils.py:312
  /home/runner/work/PINT/PINT/tests/test_utils.py:312: DeprecationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    np.int,

tests/test_utils.py:314
  /home/runner/work/PINT/PINT/tests/test_utils.py:314: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    np.float,

paulray:
I pushed a fix that solves the crash in the test, but @champagne-cmd would you mind going through and making a PR that fixes those deprecated uses of np.float and np.int?

champagne-cmd:
Yes I will get on that!

champagne-cmd:
@paulray is there preference for using `np.int64` or `np.int32` or `int`? Using the built-in `int` will have its precision vary based on the operating system being run on, whereas `np.int64` and `np.int32` specify their respective precision and don't vary based on os.

[...]

PR #965, opened 2021-02-03
Numpy deprecations fix

champagne-cmd:
Addresses issue #964 by swapping out deprecated Numpy data types with the appropriate built-in data types. `np.int` was replaced with `np.int64` (not deprecated) in case we wanted 64 bits of precision despite the os being run on, but I can go back and change these simply to `int` if this is unnecessary.


aarchiba:
I think we actually do want 64 bits - things like turn counts could exceed 4 billion in plausible situations. I'm not sure we work on any platform where the default integer size is 32 but it's probably good to specify anyway.

PR #966, opened 2021-02-03
Don't try to parse .tim files as if they were pickles.

paulray:
This fixes the bug I mentioned on the telecon.  It was trying to read a .tim file as if it was a pickle. Now it tries to unpickle only files that end in .pickle or .pickle.gz.

### DIFF
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -303,7 +303,7 @@ def load_pickle(toafilename, picklefilename=None):
         If no pickle is found.
     """
     picklefilenames = (
-        [toafilename + ext for ext in (".pickle.gz", ".pickle", "")]
+        [toafilename + ext for ext in (".pickle.gz", ".pickle")]
         if picklefilename is None
         else [picklefilename]
     )
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -303,7 +303,7 @@ def load_pickle(toafilename, picklefilename=None):
         If no pickle is found.
     """
     picklefilenames = (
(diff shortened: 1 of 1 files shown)


paulray:
For reference, the error looked like this.
get_TOAs() tried to do this:
```
        try:
            t = load_pickle(timfile, picklefilename=picklefilename)
        except IOError:
            # Pickle either did not exist or is out of date
            updatepickle = True
```
which triggered a call  like this, which failed, but not with IOError as expected.
```
In [7]: pint.toa.load_pickle("NGC6440E_fake1k.tim")                                                                   
<ipython-input-7-a9df473300e5> in <module>
----> 1 pint.toa.load_pickle("NGC6440E_fake1k.tim")

~/src/PINT/src/pint/toa.py in load_pickle(toafilename, picklefilename)
    318         try:
    319             with open(fn, "rb") as f:
--> 320                 lf = pickle.load(f)
    321         except (IOError, pickle.UnpicklingError):
    322             pass

ValueError: could not convert string to float: ORMAT 1

```

aarchiba:
Please just catch the `ValueError`. One might have a pickle file called pretty much anything, for example `.txt` or `.html`. if some program got confused downloading it, or `.aside` if that's the user's naming convention. If we just try unpickling everything, you can access these files, while if you remove `""` as a possible addition to the name one must rename files to be able to read them. 

The line would just be `except (IOError, pickle.UnpicklingError, ValueError):`

paulray:
Sure, I can do that.

PR #967, opened 2021-02-03
Timing notebook fixes

paulray:
Modified timing comparison notebook to use AASTex deluxetable environment and print times to 3 digit precision.
Also moved the table write to a separate cell.


paulray:
@champagne-cmd this PR deletes timing_comparison.py, since I think that is now obsolete. Is that right?

champagne-cmd (approved):
Yes it is fine to delete `timing_comparison.py` @paulray , this looks good to me!

PR #969, opened 2021-02-10
Minor update to F-tests

bshapiroalbert:
This updates changes the `NITS` to an optional user input parameter to the `fitter.ftest()` function instead of hardcoding `NITS=1`. I have also updated the test for the wideband F-test so that it uses just a small number of fake TOAs and runs substantially faster because of that.


src/pint/fitter.py
```diff
@@ -569,7 +569,7 @@ def get_correlation_matrix(self, with_phase=False, pretty_print=False, prec=3):
             )
             raise AttributeError
 
-    def ftest(self, parameter, component, remove=False, full_output=False):
+    def ftest(self, parameter, component, remove=False, full_output=False, NITS=1):
```
aarchiba, line 572: This parameter is called `maxiter` in `fit_toas`; it would be nice if PINT were consistent with itself.

bshapiroalbert: Done!

aarchiba:
Looks good to me.

PR #970, opened 2021-02-10
Ensure output par files have appropriate T2CMETHOD

aarchiba:
This makes sure that if T2CMETHOD was specified in a par file, it is specified on output, but that it is always IAU2000B since that's all PINT supports. Likewise for DILATEFREQ and TIMEEPH.


aarchiba:
Ready to go unless someone thinks of other settings we should set to ensure PINT-like behaviour.

Issue #951, opened 2021-01-17
compare_model() fails when newly added parameter does not have uncertainty


luojing1211:
compare_model() compares the uncertainty difference between pre and post-fit model. However if the profit model has newly added parameter without any uncertainty, comparison will fail.

pennucci:
We definitely would like this fixed for further v0 timing.  I made a local hack with a try/except clause, printing 0.0 instead of complaining about no error, but it's definitely not a fix.  @almcewen0, have you had a chance to look at this?

almcewen0:
my apologies, last week became suddenly busy and I lost track of this. will work on it today.

pennucci:
No worries, thanks!

almcewen0:
okay, so i made a new branch in my repo called "compare_fix" and I think that I have addressed the problem. if someone would like to test it out, that would be great.

pennucci:
@almcewen0 Perhaps you can open a WIP PR?

almcewen0:
https://github.com/nanograv/PINT/pull/972

pennucci:
Thanks!

aarchiba:
I think this is fixed between #972 and #997 - and is covered by a test case in #997. Can it be closed?

almcewen0:
yes, i believe it can.

PR #972, opened 2021-02-11
Compare fix

almcewen0:
Following the issue that had been raised, I included another check for missing uncertainty on the models.

### DIFF
--- a/tests/test_compare.py
+++ b/tests/test_compare.py
@@ -5,6 +5,7 @@ import astropy.units as u
 import pint
 import pint.models as mod
 import os
+import io
 from copy import deepcopy as cp
 from pinttestdata import datadir
 
@@ -103,3 +104,58 @@ class TestCompare(unittest.TestCase):
                 if not accumulate_changes:
                     modelcp = cp(model)
         assert True, "Failure in uncertainty changing test"
+
+    def test_missing_uncertainties(self):
+        # Removes uncertainties from both models and attempts to use compare.
+
+        par_base1 = """
+            PSR J1234+5612
+            RAJ 14:34:01.00 
+            DECJ 56:14:00.00
+            F0 1 
+            PEPOCH 57000
+            DM 10 
+            DMEPOCH 57000
+            DM1 2
+            DMX     0.0
+            DMX_0001   1.0 
+            DMXR1_0001     58000.0
+            DMXR2_0001     58000.0
+            """
+
+        par_base2 = """
+            PSR J1234+5612
+            RAJ 14:34:01.00 
+            DECJ 56:14:00.00
+            F0 1  
+            PEPOCH 58000 
+            DM 10
+            DMEPOCH 57000
+            DM1 2 
+            DMX     0.0
+            DMX_0001   1.0 
+            DMXR1_0001     58000.0
+            DMXR2_0001     58000.0
+            """
+
+        model_1 = mod.get_model(io.StringIO(par_base1))
+        model_2 = mod.get_model(io.StringIO(par_base2))
+
+        for pn in model_1.params_ordered[1:]:
+            param1 = getattr(model_1, pn)
+            param2 = getattr(model_2, pn)
+            if (
+                param1 is None
+                or param2 is None
+                or param1.uncertainty is None
+                or param2.uncertainty is None
+            ):
+                continue
+            param1.frozen = False
+            param2.frozen = False
+            param1.uncertainty = 0.1 * param1.quantity
+            model_1.compare(model_2)
+            model_2.compare(model_1)
+            model_1 = mod.get_model(io.StringIO(par_base1))
+            model_2 = mod.get_model(io.StringIO(par_base2))
+        assert True, "Failure in missing uncertainty test"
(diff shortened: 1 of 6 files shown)


pennucci:
This is basically the same change as (but better than) the local hack I made to get it to work...so it looks good to me.

aarchiba:
Could you add a series of tests that cover all four cases? Old/new parameter with/without uncertainty? Even just making sure nothing crashes would be good, but you could use regular expressions to check the format is what you expect if you wanted to make it more thorough.

aarchiba:
How does this cope with RAJ/DECJ with/without uncertainties? Are they printed as decimal degrees or something?

aarchiba:
Closes #951

almcewen0:
> How does this cope with RAJ/DECJ with/without uncertainties? Are they printed as decimal degrees or something?

good question. i actually just found out while working on some of the tests that the comparison for position is not happening. so i was going to try and come up with a good way to do this, and my basic idea is to calculate the separation between the two positions using SkyCoord.separation(), and compare this to the uncertainty region for position. i'm not totally certain on how all of that shakes out mathematically, though - the uncertainty on RA is listed as "hourangle_second," so do i need to include a factor of 15 degrees/cos(dec)? should i just use the scalar separation of the two positions and compare this to some joint RA/Dec uncertainty? in cases where one of the components has uncertainty and the other does not, this could be troublesome. 

alternatively, if there is a better way to do this within astropy, that could be handy. but i haven't found anything obvious.

aarchiba:
I *think* that position uncertainties are expressed as uncertainties in the position numbers, not in terms of actual angle on the sky - but I'm not sure what the convention is. There is definitely a lack of clarity on the meaning of proper motion coming from the same cos(dec) or cos(latitude) factor. It should be possible to figure out what PINT does by digging through the Astrometry timing models, and the comparison should be compatible with the PINT implementation. Whether PINT's implementation means what our users think it does is a different question.

almcewen0:
i made the fix that i believe solves the problem, and i also added a test for missing uncertainties. i think we should be good, pending further comments!

almcewen0:
@swiggumj mentioned on the PINT call that there are a few other small changes that should be made to model.compare() about verbosity levels. would it be preferable to make those changes here, or in a separate PR? @luojing1211 @paulray

paulray:
I think this should be merged ASAP

tests/test_compare.py
```diff
@@ -103,3 +104,58 @@ def test_uncertaintychange(self):
...
+            ):
+                continue
+            param1.frozen = False
+            param2.frozen = False
+            param1.uncertainty = 0.1 * param1.quantity
+            model_1.compare(model_2)
+            model_2.compare(model_1)
+            model_1 = mod.get_model(io.StringIO(par_base1))
+            model_2 = mod.get_model(io.StringIO(par_base2))
+        assert True, "Failure in missing uncertainty test"
```
aarchiba, line 161: No need for this

tests/test_compare.py
```diff
@@ -103,3 +104,58 @@ def test_uncertaintychange(self):
...
+                or param1.uncertainty is None
+                or param2.uncertainty is None
+            ):
+                continue
+            param1.frozen = False
+            param2.frozen = False
+            param1.uncertainty = 0.1 * param1.quantity
+            model_1.compare(model_2)
+            model_2.compare(model_1)
+            model_1 = mod.get_model(io.StringIO(par_base1))
```
aarchiba, line 159: Why not just put these lines at the top of the loop instead of repeating them?

tests/test_compare.py
```diff
@@ -103,3 +104,58 @@ def test_uncertaintychange(self):
...
+            DM 10 
+            DMEPOCH 57000
+            DM1 2
+            DMX     0.0
+            DMX_0001   1.0 
+            DMXR1_0001     58000.0
+            DMXR2_0001     58000.0
+            """
+
+        par_base2 = """
```
aarchiba, line 126: Is this actually any different from par_base1? If so it might be good to have a comment to indicate how. If not why not just reuse par_base1?

aarchiba (commented):
Also there appears to be a bug - if DMEPOCH differs, change_posepoch is called? https://github.com/nanograv/PINT/pull/972/files#diff-7cbfb3877d78c1fbb69c8d3e37bf61c30f2bf40de4934ad13675f1064686b003L1615

tests/test_compare.py
almcewen0: removed

tests/test_compare.py
almcewen0: not sure I understand. I want to reset the changes in each iteration of the loop so that I can only adjust one parameter at a time. but I need at least one model in advance for the loop, so i just made both there.

tests/test_compare.py
almcewen0: nope, good point. i removed par_base2

paulray:
Can we merge now?  This is needed for 15yr

tests/test_compare.py
aarchiba: It looks like you could just create the pair of models again each time through the loop? That way they're available the first time, and newly created every time after? And the code to create them is not duplicated?

almcewen0: but i need the parameter list from the model for the loop.

PR #975, opened 2021-02-11
Fitting reliability improvement

### LINKED ISSUE #958: Strange behavior from Fitter
In this example, I take the 15yr solution for J0740+6620 and freeze all the FD parameters at 0.0.  I then fit the solution and then iterate adding one FD parameter at a time and refitting.  Each time you add a parameter, the chi2 should decrease. It does this for each parameter addition until after the first fit that includes FD5, the chi2 is 13584.93.  The problem is that a second iteration of the fitter with no further changes to the free parameters makes ch2 jump up to 13585.40, then another iteration makes it jump to 13585.41, at which point it stabilizes around that value with further iterations. No minimum chi2 fitter should find a higher value on a second iteration, right? 

So, there are two components to this issue:
(1) Running a fitter iteration should never increase chi2 since the parameter values that give the lower chi2 are still valid and could be kept.
(2) Calling `fit_toas()` should probably return a converged fit by default, unless the user really knows that they want to only run one iteration and possibly get an unconverged fit.

Is any of this behavior due to GLSFitter instead of WLSFitter?  I haven't tried that yet.

Screen Shot 2021-01-22 at 9 12 22 AM
Screen Shot 2021-01-22 at 9 12 41 AM

aarchiba:
This PR introduces new Fitters that can detect convergence and can avoid taking too-long steps into NaN-land. Incidentally contains some speedups and fixes that seemed necessary along the way but that could be spun out to their own PRs.

Accreted other changes:

* The solar system geometric delay is only calculated for the non-barycentered TOAs
* The solar wind geometry is not calculated if `NE_SW` is zero (as it is for all NANOGrav par files)
* The TOA covariance matrix includes a diagonal component even if the ScaleTOAError component is not included
* The DM covariance matrix function has fixed units
* There was a bug in detecting pulse numbers in the table
* There was a bug when adding a DM column
* Removed usage of `import pint.residuals as pr`
* Avoided multiple Residuals being constructed during Fitter construction
* Fitters now have a `.is_wideband` attribute that is True or False as appropriate

Closes #958 #978 #977

### DIFF
--- a/docs/explanation.rst
+++ b/docs/explanation.rst
@@ -325,7 +325,7 @@ the residuals produced when the model is applied to a set of TOAs. The result
 of this process is a set of best-fit model parameters, uncertainties on (and
 correlations between) these, and residuals from this best-fit model. This is
 carried out by constructing a :class:`pint.fitter.Fitter` object from
-a :class:`pint.toas.TOAs` object and
+a :class:`pint.toa.TOAs` object and
 a :class:`pint.models.timing_model.TimingModel` object and then running the
 :func:`pint.fitter.Fitter.fit_toas` method; there are several example notebooks
 that demonstrate this. Nevertheless there are some subtleties to how fitting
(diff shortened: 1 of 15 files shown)


aarchiba:
grr `cached_property` is too new

aarchiba:
Some test failures, mostly disagreement with tempo2, from the fixes for #978 .

aarchiba:
Now supports WLS, GLS, and Wideband variants. Still needs correct uncertainties, covariance matrices, noise_resids, and various other stuff that is constructed along with a fit. Also documentation! But potential users are encouraged to throw problem par/tim files at it to see if it manages to converge.

aarchiba:
The new fitters set a `converged` property to True or False as appropriate at the end of fitting. This should be used, if available, in places like the F test and example notebooks and whatnot.

aarchiba:
I apologize for the very many commits - I find myself debugging using the notebook server and using github to transfer the code across. I can do a `git rebase -i` and squash the history down into something more comprehensible, and probably should as part of rebasing off the master branch if that becomes necessary.

aarchiba:
Notes from working on v0.9 pulsars:

* Downhill is slower but not agonizingly so when the fit was working okay.
* Downhill doesn't help when the problem is degeneracy, but at least it reports it.
* LM stops immediately with lambda=0 if degeneracy leads to no progress.
* LM's "gradient descent" does not always actually descend; sometimes it's uphill. Need to work through the math to figure out if an LM-like algorithm can work here.
* Some pulsars end up with very slow but steady downhill progress and beat maxiter without ever being able to take a full step, with Downhill or LM (as-is). Found this happened with a negative M2 in one case.

aarchiba:
So the LMFitter doesn't work very reliably, and the Downhill fitters don't solve all our problems, and their code could be cleaned up substantially, but given the urgent need, and the fact that they both improve on PINT's existing Fitters, I think this could be merged now and further work on the Fitters could come in new PRs.

aarchiba:
Let me draw your attention to the documentation explaining our fitting algorithms: https://nanograv-pint--975.org.readthedocs.build/en/975/explanation.html#fitting

These include some docstrings as well but they're a little rudimentary.

aarchiba:
Wow that CI output is just spectacularly useless - all the ostensible test failures are completely blank. On all the architectures. From line 50 to line 8051.

paulray:
That is bizarre!

aarchiba:
There is room for discussion on the convergence criterion - right now it only considers it converged if a full-size step was taken and it either improved the chi-squared by less than 1e-2 (any more and we're still getting somewhere) or worsened it by less than 1e-5 (a converged solution might produce a minor worsening with a full-size step - how much worsening is tolerable? I was getting problems where a full-size step worsened things by more than 1e-6 in spite of being converged).

src/pint/toa.py
```diff
@@ -907,7 +907,7 @@ def get_freq_array(bfv, ntoas):
     ts.table["error"] = error
     if dm is not None:
         for f in ts.table["flags"]:
-            f["pp_dm"] = dm
+            f["pp_dm"] = dm.to_value(u.pc / u.cm ** 3)
```
paulray, line 910: Note that PINT's `__init__.py` defines:
```
# DM unit (pc cm^-3)                                                                                                                         
dmu = u.def_unit("dmu", u.pc * u.cm ** -3)
```

src/pint/models/astrometry.py
```diff
@@ -111,16 +111,22 @@ def solar_system_geometric_delay(self, toas, acc_delay=None):
...
-        delay = -re_dot_L.to(ls).value
-        if self.PX.value != 0.0 and np.count_nonzero(tbl["ssb_obs_pos"]) > 0:
-            L = (1.0 / self.PX.value) * u.kpc
-            # TODO: np.sum currently loses units in some cases...
-            re_sqr = (
-                np.sum(tbl["ssb_obs_pos"] ** 2, axis=1) * tbl["ssb_obs_pos"].unit ** 2
-            )
-            delay += (0.5 * (re_sqr / L) * (1.0 - re_dot_L ** 2 / re_sqr)).to(ls).value
+        delay = np.zeros(len(toas))
+        c = np.logical_and.reduce(tbl["ssb_obs_pos"] != 0, axis=1)
```
paulray, line 115: I think this check is basically "Are all TOAs barycentric?"  Maybe add that as a comment?

src/pint/toa.py
aarchiba: ooh, thanks!

src/pint/models/astrometry.py
aarchiba: Good point. It's a little more than that - even if some TOAs are not barycentric, the actual calculations are only done for the non-barycentric subset. Which is more reason for a comment.

aarchiba:
Checking convergence in the old Fitters is not trivial, I'd rather leave that for another PR.

aarchiba:
I think this is ready to merge, per our discussions in the telecon today.

docs/explanation.rst
```diff
@@ -316,6 +316,189 @@ are meant to be omitted from reading or have their time adjusted. We recommend
...
+These issues particularly arise with poorly constrained parameters like M2 or
+SINI. Users experienced with pulsar timing are frequently all too familiar with
+this phenomenon and have a collection of tricks for evading it.
+
+PINT contains a slightly more sophisticated algorithm, implemented in
+:class:`pint.fitter.DownhillFitter`, that takes more careful steps:
+
+1. Evaluate the model and its derivatives at the starting point :math:`x`, producing a set of residuals :math:`\delta y` and a Jacobian `M`.
+2. Compute :math:`\delta x` to minimize :math:`\left| M\delta x - \delta y \right|_C`, where :math:`\left| \cdot \right|_C` is the squared amplitude of a vector with respect to the data uncertainties/covariance :math:`C`.
+3. Set :math:`\lambda` to 1.
```
paulray, line 492: This renders wrong. I think it should be 4. then 5. for the next line.

paulray:
Concur. I just noted one tiny doc typo.

docs/explanation.rst
aarchiba: Oops. Quite right!

PR #980, opened 2021-02-18
Fixed default argument for get_highest_density_range()

### LINKED ISSUE #979: pint.toa.get_highest_density_range() not functioning with default arguments
If no arguments are passed to the get_highest_density_range() method; the default argument of ndays=7 is a unitless quantity meaning np.searchsorted tries to add a dimensionless quantity to an mjd quantity. Producing a UnitConversionError by default. Passing ndays as a quantity with unit 'days' rectifies this issue, so should probably do this by default. 

Possible fix is included below. Extra lines of the function

`def get_highest_density_range(self, ndays=7*u.d):`  <- Changed default argument to "7\*u.d"
`     """Print the range of mjds (default 7 days) with the most toas"""`
`        sorted_mjds = np.sort(self.get_mjds())`
`        s = np.searchsorted(sorted_mjds, sorted_mjds + ndays)` #<- Without change UnitConversionError occurs when computing sorted_mjds+ndays

poneill129:
Default argument required unit: days

closes #979

### DIFF
--- a/tests/test_density.py
+++ b/tests/test_density.py
@@ -7,9 +7,11 @@ from pint.models import get_model
 from pylab import *
 from copy import deepcopy
 import io
+import pytest
 
 
-def test_get_highest_density_range(ndays=7 * u.d):
+@pytest.mark.parametrize("ndays", [7 * u.d, 20 * u.d])
+def test_get_highest_density_range(ndays):
     par_base = """
     PSR J1234+5678
     F0 1 0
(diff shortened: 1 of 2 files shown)


aarchiba:
This looks good! But since it went uncaught, that means we aren't testing this function at all. Would you mind writing a test or two? I'd suggest putting it in `tests/tes_toa.py`, but don't follow the old-style testing code you find there. I'd recommend using pytest style, like this test: https://github.com/nanograv/PINT/blob/941fca8b36d32fd0ba5cd0886ee01febb320ec52/tests/test_fitter_error_checking.py#L63 You can use `merge_TOAs` and `make_fake_toas` to make a set of TOAs with a known densest week and check that that's what the function finds.

aarchiba:
This looks good! But since it went uncaught, that means we aren't testing this function at all. Would you mind writing a test or two? I'd suggest putting it in `tests/tes_toa.py`, but don't follow the old-style testing code you find there. I'd recommend using pytest style, like this test: https://github.com/nanograv/PINT/blob/941fca8b36d32fd0ba5cd0886ee01febb320ec52/tests/test_fitter_error_checking.py#L63 You can use `merge_TOAs` and `make_fake_toas` to make a set of TOAs with a known densest week and check that that's what the function finds.

tests/test_density.py
```diff
@@ -0,0 +1,37 @@
...
+from astropy.time import Time
+import matplotlib.pyplot as plt
+import pint.toa as toa
+from pint.models import get_model
+from pylab import *
+from copy import deepcopy
+import io
+
+
+def test_get_highest_density_range(ndays=7 * u.d):
```
aarchiba, line 12: Since we don't call test functions, there's not much point giving them arguments - or at least, we use arguments for different things (fixtures and parametrization if you're curious). So maybe just set this on the first line. Or you could run this function with twice with two different values of `ndays` using `@pytest.mark.parametrize()` to make two different tests.

tests/test_density.py
```diff
@@ -0,0 +1,37 @@
...
+    SOLARN0 0
+    """
+    model = get_model(io.StringIO(par_base))
+    toas_1 = toa.make_fake_toas(57000, 58000, 1000, model, obs="@")
+    toas_2 = toa.make_fake_toas(57500, 57507, 100, model, obs="@")
+    merged = toa.merge_TOAs([toas_1, toas_2])
+    x1, x2 = merged.get_highest_density_range()
+    x3, x4 = merged.get_highest_density_range(ndays)
+
+    assert x1.value <= 57500 + 1e-5
```
aarchiba, line 29: It's nicer to use `abs(x1.value - 57500) <= 1e-5`.

aarchiba:
Looks good to me!

PR #981, opened 2021-02-22
Fix bug in sign of pulse number test

aarchiba:
Currently PINT's test for available pulse numbers in Residuals is accidentally negated. This removes the inappropriate `not`.


paulray:
No brainer.  Merging...;

PR #983, opened 2021-02-23
Fix a couple of small bugs for extrapolation checking.

kerrm:
There was a typo in the interpolation check.  additionally, changed the behavior of the spline interpolators to allow extrapolation.  (All of the machinery for capping the maximum extrapolated interval to something sane, like 2 mins, is still in place.  This change applies simply to the spline object.)

There is a specific use case for this: Fermi FT2 files have 30s intervals, typically, and the spline knots are at the "START".  An event falling in the last interval (30s) would technically be "extrapolated", and this was causing the spline object to barf.

### DIFF
--- a/src/pint/observatory/satellite_obs.py
+++ b/src/pint/observatory/satellite_obs.py
@@ -285,14 +285,15 @@ class SatelliteObs(SpecialLocation):
     def __init__(self, name, ft2name, maxextrap=2, overwrite=False):
         self.FT2 = load_orbit(name, ft2name)
 
-        # Now build the interpolator here:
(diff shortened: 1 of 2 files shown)


aarchiba:
Looks good to me. I don't suppose there are easy test cases? We won't have a Fermi orbit file on hand.

kerrm:
There are a couple of small orbit files in tests/datafiles, currently used
in test_satobs.py and test_fermiphase.py.  I was going to replace it with
one that tickled the bug I found, but then I realized I could just
explicitly add in a time that (would have) triggered the  problem.

PR #987, opened 2021-02-25
If read model from parfile, put the parfile name in the .name attribute

### LINKED ISSUE #984: Unnecessary info/warnings reported by compare_models
There are INFO/WARNING statements currently output by `compare_models` which are either too verbose/redundant, logging level is too high, or the language used doesn't quite make sense. I've attached a screenshot of one example, which can be reproduced on the psr/b1855+09/jks branch in timing_analysis. By default, I don't think START/FINISH/NTOA/CHI2 should be reported, but they could be info statements if verbosity is increased. If kept, changing the language would be a good idea, e.g.:
```
INFO: Parameter START has changed between these models
```
For T/F (or Y/N) parameters like `CORRECT_TROPOSPHERE` and `PLANET_SHAPIRO`, the language should change and info statements here would be better than warnings. I'd suggest something like:
```
INFO: Parameter CORRECT_TROPOSPHERE has changed (turned ON in model 2)
```

Screen Shot 2021-02-24 at 9 37 26 AM

scottransom:
This helps fix #984 as the parfile name will be in the model `.name` attribute

### DIFF
--- a/src/pint/models/model_builder.py
+++ b/src/pint/models/model_builder.py
@@ -302,7 +302,12 @@ def get_model(parfile):
     except AttributeError:
         contents = None
     if contents is None:
-        return ModelBuilder(parfile).timing_model
+        # parfile is a filename and can be handled by ModelBuilder
+        mbmodel = ModelBuilder(parfile)
+        model = mbmodel.timing_model
+        if model.name=='':
+            model.name = parfile
+        return model
     else:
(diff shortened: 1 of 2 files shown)


scottransom:
If someone can take a look at this, I think it is ready to merge...  It is a tiny change (that is tested!) that only updates an attribute that was already there.

paulray (approved):

PR #990, opened 2021-03-01
Fix bug when ECORR has no TOAs

aarchiba:
Currently if an ECORR selects no TOAs, there is a crash while trying to build the noise design matrix. This PR fixes that (and just emits a warning).

Incidentially includes an improved `hypothesis` configuration.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,24 @@
+import os
+import hypothesis
+
+# Tell Hypothesis that we might be running slow tests, to print the seed blob
+# so we can easily reproduce failures from CI, and derive a fuzzing profile
+# to try many more inputs when we detect a scheduled build or when specifically
+# requested using the HYPOTHESIS_PROFILE=fuzz environment variable or
+# `pytest --hypothesis-profile=fuzz ...` argument.
+
+hypothesis.settings.register_profile(
+    "ci", deadline=None, print_blob=True, derandomize=True
+)
+hypothesis.settings.register_profile(
+    "fuzzing", deadline=None, print_blob=True, max_examples=1000
+)
+default = (
+    "fuzzing"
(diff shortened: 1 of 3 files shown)


luojing1211:
This should be fixed in the PR #912, but it has not been merged yet.

paulray:
Why has #912 sat so long without being merged? We should not let that happen.  Unless they are WIP, PRs should get merged promptly once comments are addressed.

aarchiba:
Yes, there is at least one unaddressed comment on #912 that I think is important, and has been for quite a while now.

aarchiba:
Ready to merge if it passes the test suite and a reviewer thinks it's okay.

aarchiba:
The test failure is a `hypothesis` timing problem.

paulray:
> The test failure is a `hypothesis` timing problem.

Yeah, that has been happening to macos a lot. Why is this test so fragile?

aarchiba:
<sigh /> We *should* be setting hypothesis up with different settings on CI, and probably for interactive use as well. For the life of me I can't remember how to do this, but Astropy I think sets some stuff in `conftest.py`. (Specifically, `hypothesis` fails tests that take too long, and on CI machines how long they take depends on load. One can turn off the "deadlines", in general or specifically on CI machines, which I think we should do; we can also set up "profiles" so the hypothesis tests run faster normally but if you're working on the code they test you can easily run more intensive versions.)

aarchiba:
This should prevent those hypothesis deadline failures.

paulray:
Great, once this passes, I can merge it.

aarchiba:
@paulray Would you mind checking on this one? It'll bring in the `hypothesis` setup too.

Issue #995, opened 2021-03-06
CI does not respect settings in `conftest.py`


aarchiba:
Our `conftest.py` has a number of settings, most notably configuring hypothesis to stop caring about how long tests take. But our CI is not respecting this. I suspect tox may be ignoring the file?

abhisrkckl:
Is this issue still valid?

PR #996, opened 2021-03-06
WIP: Testing conftest.py in CI

aarchiba:
It seems (#995 ) that our CI is not respecting settings in `conftest.py`. This is an attempt to fix that, and also to set up suitable options for `hypothesis`.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -26,7 +26,7 @@ default = (
     )
     else "ci"
 )  # noqa: E501
-raise ValueError(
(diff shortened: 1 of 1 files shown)


aarchiba:
I'm puzzled. `conftest.py` seems to be being picked up by CI, and on my home machine it disables deadlines. But we're still getting deadline failures from CI?

aarchiba:
Found the problem - some tests had their own hypothesis settings.

Issue #976, opened 2021-02-11
PINT does not easily provide whitened residuals


aarchiba:
If one has ECORR or red noise, there are two notions of residual - one that is the plain residuals, after subtraction of the timing model, and one that has been "whitened" by subtraction of the basis functions corresponding to those noise components. One could quite reasonably want to plot either, but PINT provides only the former. GLSFitter and WidebandTOAFitter do add a `.noise_resids` dictionary to their `resids` object, but it is not trivial to get whitened residuals from this (and it's not at all clear how this behaves for WidebandTOAFitter). Perhaps a `.whitened` attribute or property on appropriate Residuals objects? Or Fitters could offer, if relevant, a `.whitened_resids` attribute alongside their `.resids`?

Convenient access to whitened residuals would be valuable for, e.g., identifying outliers for manual TOA excision, or for visually inspecting fits for plausibility.

aarchiba:
https://gitlab.nanograv.org/nano-time/timing_analysis/blob/15yr/src/timing_analysis/utils.py#L20 is a demonstration (though wrong) of how to do it

PR #1000, opened 2021-03-07
WIP: make whitened residuals available

### LINKED ISSUE #810: Weighted mean of post-fit residuals is not using scaled uncertainties; affecting chi2 calculation.

aarchiba:
Go over the calculations of chi-squared and noise residuals so that whitened residuals are routinely available on Residuals objects.

Closes #810 #976

### DIFF
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -1410,25 +1410,7 @@ class DownhillGLSFitter(DownhillFitter):
         """
         self.threshold = threshold
         self.full_cov = full_cov
-        r = super().fit_toas(maxiter=maxiter, **kwargs)
-        # FIXME: set up noise residuals et cetera
-        # Compute the noise realizations if possible
-        ntmpar = len(self.model.free_params)
-        if not self.full_cov:
-            noise_dims = self.model.noise_model_dimensions(self.toas)
-            noise_resids = {}
-            for comp in noise_dims.keys():
-                p0 = noise_dims[comp][0] + ntmpar
-                p1 = p0 + noise_dims[comp][1]
-                noise_resids[comp] = (
-                    np.dot(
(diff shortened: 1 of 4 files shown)


aarchiba:
I think the existing code has a bug in how we do rank-reduced fits - `phiinv` needs to be rescaled by `norm**2`; that fix is in here, but I can pull it out and file a standalone PR if I'm right. Likewise I can put in a standalone fix for #810.

aarchiba:
> I think the existing code has a bug in how we do rank-reduced fits - `phiinv` needs to be rescaled by `norm**2`; that fix is in here, but I can pull it out and file a standalone PR if I'm right. Likewise I can put in a standalone fix for #810.

In fact it's not a problem because we don't normalize columns corresponding to noise basis functions (or the offset), though I have observed this causing problems. I had to copy this fix over to #1013 to avoid test failures due to numerical issues on the actual NGC6440E data set.

paulray:
Any chance of getting this merged? I'd like to be able to get whitened residuals easily.

aarchiba:
This is a bit of a mess. A more direct approach would be to make `whitened_residuals` a property on residuals objects that used the red noise object that fitters attach to construct whitened residuals. This was trying to clean up the fitters too, and I think now that the code has diverged it might be better to start simpler.

abhisrkckl:
This was implemented in #1671.
So I am closing this.

PR #1001, opened 2021-03-07
Use scaled errors for weighted averaging

### LINKED ISSUE #810: Weighted mean of post-fit residuals is not using scaled uncertainties; affecting chi2 calculation.

aarchiba:
This does the pretty obviously right thing with residuals - the mean subtraction uses a weighted mean that includes EQUADs - but that leads to a failure of a comparison with TEMPO2. Unsure what to do with that.

Closes #810

### DIFF
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -328,11 +328,11 @@ class Residuals:
         else:
             # Errs for weighted sum.  Units don't matter since they will
             # cancel out in the weighted sum.
-            if np.any(self.toas.get_errors() == 0):
+            if np.any(self.get_data_error() == 0):
                 raise ValueError(
                     "Some TOA errors are zero - cannot calculate residuals"
                 )
-            w = 1.0 / (self.toas.get_errors().value ** 2)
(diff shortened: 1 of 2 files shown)


aarchiba:
So we are faced with two options:

1. abandon tempo2 compatibility and use scaled errors for mean subtraction of residuals, or
2. simply declare that mean subtraction uses un-scaled errors and you can subtract your own mean if you want.

scottransom:
Is there some way that we can give an option for the user that they can get the "other" thing if our default is not what they want?  Or is this too tied up in the residuals/fit process that it would make that difficult?

PR #1004, opened 2021-03-10
Added MeerKAT as an observatory

scottransom:
This uses the TEMPO2 MeerKAT clock correction files (which can be obtained here:  https://archive-gw-1.kat.ac.za/public/tfr/mk2utc.clk )

### DIFF
--- a/src/pint/observatory/observatories.py
+++ b/src/pint/observatory/observatories.py
@@ -31,6 +31,9 @@ TopoObs(
     "meerkat",
     tempo_code="m",
     itoa_code="MK",
+    clock_fmt="tempo2",
+    clock_dir="TEMPO2",
+    clock_file="mk2utc.clk",
     itrf_xyz=[5109360.133, 2006852.586, -3238948.127],
 )
 TopoObs(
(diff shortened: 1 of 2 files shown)


scottransom:
yeah, I think that is correct about the GPS correction (and it looks like that is happening already).  I can add a test, but aren't we not using tests that require TEMPO2?

aarchiba:
We have a few tests that compare our results to stored results previously output by tempo2 - so a simple test would be to have tempo2 barycenter a TOA or two from MeerKAT, store those in a file, then barycenter them ourselves and compare.

scottransom:
This should be ready to go now, I think.

paulray:
Looks good, but your test needs blackening. Note that in VS Code you can set it to run blacken on save. You can also add a pre-commit hook to check blackening before you push.

scottransom:
I actually don't want VS code to blacken on save....  just when I'm working on PINT!  And I had *just* blacked it, but forgot and tweaked the "TEMPO2" line and screwed it up!  D'oh!  PEBKAC (and stupid black rules).

paulray:
Note that "editor.formatOnSave" is a per-project setting. It lives in my .vscode/settings.json in the PINT tree. So it won't try to blacken my other projects.

scottransom:
Cool.  I just put that editor.formatOnSave setting on for PINT alone.  That's cool!  I didn't know you could do that.  Thanks, @paulray !

PR #1013, opened 2021-03-22
Raise on reaching maxiter for downhill fitters

aarchiba:
Currently downhill fitters raise an exception if they can't take any more steps, but not if they reach the limit `maxiter` without converging. In the latter case they just quietly set `converged` to False, but if users don't check that (they don't, for example NANOGrav's F test harness) they just assume convergence. Raising an exception - a unique type, easy to catch - is a pretty definite signal that the fitter has not converged. But it leaves the best-fit solution so far as the current state of the fitter, so the user can resume where they left off.

It might be useful to have a separate exception type for maxiter and more dire convergence failures, as maxiter failures are resumable but other kinds are not; this could easily be added.

This introduces a slight change to the math: design matrix column normalization is extended to include the offset and noise parameters, which requires rescaling `phiinv`, because in a real example data set the scaling of the offset was causing numerical problems.


aarchiba:
Ready to review.

luojing1211:
Do you have any tests for the StepProblem error. Or is it hard to find a case to trigger the problem?

aarchiba:
> Do you have any tests for the StepProblem error. Or is it hard to find a case to trigger the problem?

It's quite hard to trigger - I have seen it in some of the NANOGrav data sets, but usually when M2/SINI is added to a data set that just can't detect them, and I'm not sure I can replicate that with one of ours in any kind of efficient way. (It takes ten model evaluations to trigger, so the test could be pretty slow.)

PR #1015, opened 2021-03-23
Control hypothesis profile

aarchiba:
Arrange settings so we no longer get annoying hypothesis test failures due to timeouts; also use randomness when interactive but not on CI.


aarchiba:
Ready for review.

Issue #1008, opened 2021-03-16
PINT is not as compatible with TEMPO as it could be


aarchiba:
It would be valuable if PINT emitted `.par` and `.tim` files that TEMPO interpreted in as similar a way to PINT as it is able. A number of modifications on the TEMPO side have resolved some issues but Joe Swiggum and David Nice have identified the following points where PINT could do better:

- `.tim` files should start with `MODE 1` as well as `FORMAT 1` (to tell TEMPO to use TOA uncertainties?)
- TEMPO expects noise parameters with different names (for example `EFAC` -> `T2EFAC`, I think)
- TEMPO is not as flexible about observatory names as PINT/TEMPO2 - specifically (I think) it only recognizes `ao` not `arecibo`

If we find others or want to clarify TEMPO's needs, we can put that here.

pennucci:
For the record, MODE 1 --> yeah, tells tempo to use the TOA uncertainties in the fit and FORMAT 1 --> indicates that TOAs are in the modern tempo2 format <file> <freq> <TOA> <TOA uncertainty> <site> <flags> without fixed-space allocation for each.  Having these would be nice conveniences to avoid the need for a timing_analysis post-processing conversion step/

aarchiba:
Thanks! I think it should be possible to do without a conversion step, although because of ephemeris and other differences we'll probably find that solutions that have converged for PINT haven't fully for tempo and tempo2. But they should be close.

aarchiba:
Some notes from today's PINT meeting:

- I think that, even for TEMPO, putting MODE 1 in the par file (which we do, or can do), has the same effect as putting MODE 1 in the tim file, except that putting it in the tim file makes various programs emit warnings because it's the old way of doing things.
- PINT does, and TEMPO does, support different values of ECL, so emitting an ECL statement is a good idea and doesn't bother TEMPO2; if users want a TEMPO2-compatible par file they should use ECL IERS2003, which may just be a different coordinate system or may have some built-in inaccuracies, I'm not sure.
[...]

PR #1017, opened 2021-03-24
TEMPO/TEMPO2 compatibility

aarchiba:
The goal of this PR is to improve compatibility of written par/tim files with tempo/tempo2. In so far as possible normal PINT par/tim files should be intelligible to these programs, but it should be reasonably easy to ask PINT to produce more-compatible par/tim files and it should still be able to read them. This will be accomplished partly by tweaking various settings to PINT produces reasonably compatible output by default, but also by allowing translation of parameter names to tempo/tempo2-friendlier versions (that are, ideally, recognized as aliases upon reading).

This should produce TEMPO-friendly par and tim files, if the model is appropriate, with:
```python
to.alias_translation = pint.toa.tempo_aliases
mo.use_aliases(reset_to_default=True, alias_translation=pint.toa.tempo_aliases)
to.write_tim_file("file.tim")
with open('file.par', 'w') as f:
    f.write(mo.as_parfile())
```

Should help with #1008


src/pint/toa.py
```diff
@@ -73,6 +73,16 @@
 
 all_planets = ("jupiter", "saturn", "venus", "uranus", "neptune", "earth")
 
+tempo_aliases = {
+    "arecibo": "AO",
+    "jodrell": "JB",
+    "A1DOT": "XDOT",
+    "STIGMA": "VARSIGMA",
+    "EFAC": "T2EFAC",
+    "EQUAD": "T2EQUAD",
```
scottransom, line 82: This is from memory, but I'm pretty sure this is correct, but T2EFAC and T2EQUAD aren't really aliases for EFAC and EQUAD, right?  Their order of ops is different.  I'll see if I can't find the documentation for that....

scottransom: Hmmm.  After looking at the TEMPO and T2 codes, maybe I'm wrong and they both are implemented the same.  I swear this was a problem sometime in the past, but maybe it was an isolated incident or something.

swiggumj:
I was scrolling through various issues that came up and wondered if this PR would also handle the fact that TEMPO expects DMDATA and NHARM to be integers. Can we simply make that so in PINT? And from a 15yr (timing_analysis) standpoint, to facilitate PINT/Tempo2 comparisons, should we specify IERS2003 as an early check (I can open a quick MR if so) for timing models?

aarchiba:
> I was scrolling through various issues that came up and wondered if this PR would also handle the fact that TEMPO expects DMDATA and NHARM to be integers. Can we simply make that so in PINT? And from a 15yr (timing_analysis) standpoint, to facilitate PINT/Tempo2 comparisons, should we specify IERS2003 as an early check (I can open a quick MR if so) for timing models?

Currently PINT has no notion of an integer-valued parameter, so it's not as easy as changing their type. It might be possible to arrange for PINT to output parameter values that happen to be integers without the `.0`, which ought to resolve that problem.

I think - if I understand correctly - there is no particular negative to all NANOGrav timing solutions using `ECL IERS2003`; this will make the coordinates mean the same thing to all three programs and should not introduce any inaccuracy. This doesn't need PINT support, but would drop nicely into the same part of `timing_analysis` that makes sure we use `EPHEM DE440`. Like that, re-fitting would be necessary, so it would only take effect as we re-fit and check new par files into git.

aarchiba:
Currently this produces par files PINT cannot read correctly as a result of #1019 .

aarchiba:
We do have booleans, and I think that on read we accept 1/0 or Y/N; I'm not sure what we write. The Parameter types are... complicated, but adding an integer type might not be too bad. (PINT would want to accept expressions like 5.0 since it has been writing these.)

davidnice:
Re: DMDATA being an integer, I fixed up Tempo a couple weeks ago so it can handle non-integer values.

Re: NHARM.  This is weird in Tempo right now.  Tempo reads it in as an integer to a temporary variable.  It prints an error if it is greater than 4.  It then does nothing else with it, i.e., it does not store it in a variable that gets used elsewhere.  Very strange!  Perhaps it is a legacy of an incomplete attempt to upgrade the ELL1H model (for which I believe NHARM is a control parameter).  It is simple enough to allow Tempo to treat it as a floating number (in fact, I will do that now), but really the parameter is just ignored.

davidnice:
I think it is OK to use IERS2003 obliquity of the ecliptic.  This isn't the most up-to-date standard value, but it is hard-wired in Tempo2 (at least the last time I checked), so it would increase compatibility across packages if we use it.  It doesn't affect scientific results in any material way since its effect is only to perturb positions in ecliptic coordinates.

demorest:
Making tempo input more robust is good, but I think it's mildly confusing for PINT to output floating point values for things that are actually integers or boolean, so I think that should still be fixed up.  Regarding NHARM, I forgot about that!  The tempo ELL1H model only implements the four-harmonic case (H3 and/or H4 params) which is a subset of what is available in tempo2.  That test is just there as a "safety feature" so that people don't get tripped up by this with no warning.

aarchiba:
Should DMDATA not be a boolean? This does mean PINT will print out Y/N.

aarchiba:
Now we have `intParameter` values, which accept floating-point values that are also integers, and output integers without decimal places.

aarchiba:
> Making tempo input more robust is good, but I think it's mildly confusing for PINT to output floating point values for things that are actually integers or boolean, so I think that should still be fixed up. Regarding NHARM, I forgot about that! The tempo ELL1H model only implements the four-harmonic case (H3 and/or H4 params) which is a subset of what is available in tempo2. That test is just there as a "safety feature" so that people don't get tripped up by this with no warning.

By default. PINT outputs booleans as Y/N. It is not clear whether TEMPO can safely read these, since it prefers 1/0 for booleans.

aarchiba:
Note to self: 

more integer paramters - `NTOA`

booleans (but are booleans safe?):  `CORRECT_TROPOSPHERE`, 'DMDATA', 'PLANET_SHAPIRO'

demorest:
tempo only accepts 1/0 for booleans.  But it looks like the only boolean params it reads are DMDATA and DMXFIX.  Would not be hard to modify tempo to also take Y/N for these.

tempo2 accepts both 1/0 and Y/N on input, but writes Y/N in output par files.

davidnice:
I have updated Tempo to allow flexible inputs of boolean parameter flags:
  -- "True" values are registered for strings starting with 'T' or 'Y' or 't' or 'y'
  -- "True" values are registered for numbers greater than zero, integer or real
  -- "False" values are registered for strings starting with 'F' or 'N' or 'f' or 'n'
  -- "False" values are registered for numbers zero and less

Tempo now uses this for:
-- DMXFIX
-- DMDATA
-- JUMP 
-- START
-- FINISH

Other boolean flags allowed by Tempo2/PINT flags (e.g., PLANET_SHAPIRO) are ignored by Tempo.

aarchiba:
> I have updated Tempo to allow flexible inputs of boolean parameter flags:
> -- "True" values are registered for strings starting with 'T' or 'Y' or 't' or 'y'
> -- "True" values are registered for numbers greater than zero, integer or real
> -- "False" values are registered for strings starting with 'F' or 'N' or 'f' or 'n'
> -- "False" values are registered for numbers zero and less
> 
> Tempo now uses this for:
> -- DMXFIX
> -- DMDATA
> -- JUMP
> -- START
> -- FINISH
> 
> Other boolean flags allowed by Tempo2/PINT flags (e.g., PLANET_SHAPIRO) are ignored by Tempo.

Okay, I think we should be compatible with current TEMPO and TEMPO2 if files are written with `alias_translation=tempo_aliases`.`

aarchiba:
Now supports remembering which alias was used on reading for par files. It would be good to add remembering which alias to use for observatories, though that's complicated because the input might use more than one alternative (so how do we pick which?). This should also affect observatory aliases in par files (TZRSITE), though handling that gracefully is not so easy.

aarchiba:
This now allows you to set TOAs objects and on TimingModel objects into modes where they write themselves in TEMPO/TEMPO2-compatible formats. Otherwise TimingModel objects try to retain the way they were written in the input par files. TOAs objects don't, yet.

Issue #1011, opened 2021-03-19
PINT has no way of working with commented-out TOAs


aarchiba:
libstempo can recognize a TOA line that is just `C ` followed by a normal TOA line as a commented-out TOA, and it can load it in and keep track of it and write it back out. PINT simply ignores that line, and TOA selections simply discard the removed TOAs. Nevertheless it is a common practice to retain excised TOAs with comment markers, and it would be good if PINT provided some level of support for that. Perhaps allowing users to request, while reading, inclusion of commented-out TOAs, which would be marked with a special flag? This would result in them being commented out on write, presumably, and we could provide an easy function to select only the non-commented TOAs as a subset, for example to create a new TOAs object for use in fitting.

swiggumj:
Just noticed you'd already opened the issue, so I'll copy the text from mine here (though it's not too different) and close the other...

In the past, NANOGrav data releases include tim files with excised TOAs included, but commented out for various reasons (those commented out include -cut flags with values that explain why they were cut). This provides more transparency into the TOA excision process, but also doesn't require any special handling to reproduce exact results if someone were to run them through TEMPO/Tempo2 immediately.

Doing this in PINT right now is difficult because TOAs that are cut are simply gone (a new table is created, which is a subset of the initial one when TOAs are selected out). Another way to handle this might be to have a specific flag that signals to PINT that TOAs with that flag should not be included in fits, and when using toa.write_TOA_file, they should be commented out. @aarchiba suggested creating this issue for further discussion since implementation might take a bit of care.

swiggumj:
If we do go the route of using a designated flag, may I suggest `-cut` for consistency with NANOGrav's data releases?

aarchiba:
So, how could this work on the PINT side?

[...]

PR #1020, opened 2021-03-26
Allow flags to comment TOAs when writing to file

scottransom:
This is for discussion with #1011, but it seems useful and has tests.
Any TOA with an `ignore` or `cut` flag will be commented on write.

### DIFF
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -1711,11 +1711,6 @@ class TOAs:
         if format.upper() in ("TEMPO2", "1"):
             outf.write("FORMAT 1\n")
 
-        # Warn the user if any of the TOAs have -ignore or -cut flags that
-        # they will be commented in the output file
-        if isinstance(commentflag, str) and len(self.get_flag_value(commentflag)[1]):
-            log.warning(f"TOAs with '-{commentflag}' flags will be commented")
-
         # Add pulse numbers to flags temporarily if there is a pulse number column
         # FIXME: everywhere else the pulse number column is called pulse_number not pn
         pnChange = False
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -1731,10 +1731,9 @@ class TOAs:
             obs_obj = Observatory.get(obs)
 
             flags = flags.copy()
+            toatime_out = toatime
             if "clkcorr" in flags:
-                toatime_out = toatime - time.TimeDelta(flags["clkcorr"])
-            else:
-                toatime_out = toatime
+                toatime_out -= time.TimeDelta(flags["clkcorr"])
             out_str = (
                 "C " if isinstance(commentflag, str) and (commentflag in flags) else ""
             )
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -1678,7 +1678,7 @@ class TOAs:
         self.compute_TDBs()
         self.compute_posvels(self.ephem, self.planets)
 
-    def write_TOA_file(self, filename, name="unk", format="tempo2"):
+    def write_TOA_file(self, filename, name="unk", format="tempo2", comments=True):
         """Write this object to a ``.tim`` file.
 
         This function writes the contents of this object to a (single) ``.tim``
@@ -1696,6 +1696,9 @@ class TOAs:
         format : str
             Format specifier for file ('TEMPO' or 'Princeton') or ('Tempo2' or '1');
             note that not all features may be supported in 'TEMPO' mode.
+        comments : boolean
+            If True, and there are "-ignore" or "-cut" flags, the TOAs will be
+            commented in the output file.  If False, all TOAs are uncommented.
         """
         try:
             # FIXME: file must be closed even if an exception occurs!
@@ -1737,7 +1740,9 @@ class TOAs:
                 toatime_out = toatime - time.TimeDelta(flags["clkcorr"])
             else:
                 toatime_out = toatime
-            out_str = "C " if ("ignore" in flags or "cut" in flags) else ""
+            out_str = (
+                "C " if (comments and ("ignore" in flags or "cut" in flags)) else ""
+            )
             out_str += format_toa_line(
                 toatime_out,
                 toaerr,
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -1678,7 +1678,7 @@ class TOAs:
         self.compute_TDBs()
         self.compute_posvels(self.ephem, self.planets)
 
-    def write_TOA_file(self, filename, name="unk", format="tempo2", commentflag="cut"):
+    def write_TOA_file(self, filename, name="unk", format="tempo2", commentflag=None):
         """Write this object to a ``.tim`` file.
 
         This function writes the contents of this object to a (single) ``.tim``
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -1728,11 +1728,12 @@ class TOAs:
             obs_obj = Observatory.get(obs)
 
             flags = flags.copy()
-            if "clkcorr" in flags.keys():
+            if "clkcorr" in flags:
                 toatime_out = toatime - time.TimeDelta(flags["clkcorr"])
             else:
                 toatime_out = toatime
-            out_str = format_toa_line(
+            out_str = "C " if ("ignore" in flags or "cut" in flags) else ""
+            out_str += format_toa_line(
                 toatime_out,
                 toaerr,
                 freq,
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -1708,6 +1708,11 @@ class TOAs:
         if format.upper() in ("TEMPO2", "1"):
             outf.write("FORMAT 1\n")
 
+        # Warn the user if any of the TOAs have -ignore or -cut flags that
+        # they will be commented in the output file
+        if len(self.get_flag_value("ignore")[1]) or len(self.get_flag_value("cut")[1]):
+            log.warning("TOAs with '-cut' or '-ignore' flags will be commented")
+
         # Add pulse numbers to flags temporarily if there is a pulse number column
         # FIXME: everywhere else the pulse number column is called pulse_number not pn
         pnChange = False
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -1678,7 +1678,7 @@ class TOAs:
         self.compute_TDBs()
         self.compute_posvels(self.ephem, self.planets)
 
-    def write_TOA_file(self, filename, name="unk", format="tempo2", comments=True):
+    def write_TOA_file(self, filename, name="unk", format="tempo2", commentflag="cut"):
         """Write this object to a ``.tim`` file.
 
         This function writes the contents of this object to a (single) ``.tim``
@@ -1696,9 +1696,9 @@ class TOAs:
         format : str
             Format specifier for file ('TEMPO' or 'Princeton') or ('Tempo2' or '1');
             note that not all features may be supported in 'TEMPO' mode.
-        comments : boolean
-            If True, and there are "-ignore" or "-cut" flags, the TOAs will be
-            commented in the output file.  If False, all TOAs are uncommented.
+        commentflag : str or None
+            If a string, and that string is a TOA flag, that TOA will be commented
+            in the output file.  If None (or non-string), no TOAs will be commented.
         """
         try:
             # FIXME: file must be closed even if an exception occurs!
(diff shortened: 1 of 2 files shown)


tests/test_toa_writer.py
```diff
@@ -42,10 +42,14 @@ def test_roundtrip_topo_toa_Tempo2format(self):
...
         t1 = toa.TOA(t1time, obs="gbt", freq=0.0)
-        ts = toa.get_TOAs_list([t1], ephem="DE421")
+        t2 = toa.TOA(t1time, obs="gbt", freq=0.0)
+        ts = toa.get_TOAs_list([t1, t2], ephem="DE421")
+        assert ts.ntoas == 2
+        ts.table[1]["flags"]["ignore"] = True  # comment TOA
         ts.write_TOA_file("testtopo.tim", format="Tempo2")
         ts2 = toa.get_TOAs("testtopo.tim")
         print(ts.table, ts2.table)
+        assert ts2.ntoas == 1  # one should be commented
```
aarchiba, line 52: Can you make this a separate test, so we can tell which things are broken and which are not?

src/pint/toa.py
```diff
@@ -1728,11 +1728,12 @@ def write_TOA_file(self, filename, name="unk", format="tempo2"):
             obs_obj = Observatory.get(obs)
 
             flags = flags.copy()
-            if "clkcorr" in flags.keys():
+            if "clkcorr" in flags:
                 toatime_out = toatime - time.TimeDelta(flags["clkcorr"])
             else:
                 toatime_out = toatime
-            out_str = format_toa_line(
+            out_str = "C " if ("ignore" in flags or "cut" in flags) else ""
```
aarchiba, line 1735: Can you make the "ignore" flag an optional argument to `write_TOA_flag`? So that users don't find their TOAs disappearing unexpectedly, and so that users *can* write files that include `-cut` TOAs that aren't commented out?

scottransom: How will people find their TOAs disappearing unexpectedly given that 1) they have to explicitly add these flags (or they would have had to do something non-standard to the .tim file (i.e. uncommenting them) before reading them in) and 2) they don't disappear, they are just commented?  The whole point of this change is to make a simple way that we can select against TOAs that we don't want to fit or whatever, and which we can easily flag to be commented *if* we re-write the TOAs.

I'm not seeing an advantage to an optional argument (which users have to remember) -- especially since if "-cut" means that we've cut something, we really don't want it included!  If you want to make a note about a TOA, use a different flag.

scottransom: If we want to make an *additional* way of commenting TOAs via an optional argument which provides a mask array, I suppose that is fine.  The utility and ability to use flags for this type of thing is exactly why TOAs have flags now.

tests/test_toa_writer.py
scottransom: Done

src/pint/toa.py
aarchiba: My thought was that if a user has a tim file in which some TOAs have a `-cut` flag that they did not add, they may be surprised that those TOAs are commented out when the TOAs are written out, and they will have disappeared when the user tries to read the tim file back in. 

I also anticipate wanting to write out a tim file with `-cut` flags indicating reasons for cutting TOAs but with those TOAs not actually commented out, so that this tim file can easily be used to evaluate our excision criteria. So I'd like to be able to turn off the commenting feature.

scottransom: This was my point at the beginning of the discussion in #1011.  We are trying to use comments and/or the `-cut` flag to do things which they (IMO) shouldn't be doing.  NANOGrav started using the `-cut` flag to describe *why* we cut (thereby commenting out) a TOA.  But the bottom line is that `-cut` == commented as defined by NANOGrav.

So for your first point, the user shouldn't have TOAs that have a cut flag and are un-commented unless they (or someone else) modified them.   But even if they do, they still need to do the non-standard thing of writing out a new TOA file *and* be ignorant as to what the flags contain and mean in order to be "surprised".

If you (or someone else) wants to evaluate *possible* reasons to cut TOAs and test them (i.e. your 2nd point), then you should be using a different flag since all strings are valid.  But `-cut` means that we have cut the TOA!  Please don't change  the definition of `-cut`.

scottransom:
I added a warning (should it be INFO instead?) that TOAs with `-ignore` and/or `-cut` flags will be commented.  And I also added an optional argument for commenting to be turned off.  I think the default should be to comment, though, given that those flags explicitly mean that we are ignoring those TOAs.

aarchiba:
Does anyone actually use `-ignore` or is it something you're introducing?

scottransom:
Introducing.  As an example of a way to add a flag that means that we want to ignore a TOA and comment it out.  Could be something else.  But, for example, I think `pintk` should have a way to select TOAs for commenting or ignoring (perhaps temporarily;  and adding a flag like this could do it).  I didn't use `-comment`, because that flag sounds like you could be giving a standard comment *about* a TOA.

aarchiba:
I ask because the original purpose for this was to allow `-cut` to work, so we are presumably going to make that produce comments. Do we need to introduce a second flag to do the same thing?

scottransom:
We certainly don't need to.  But for interactive use, I don't think `-cut` is very intuitive.  I used `-ignore` as an example of an alias that might be a bit more intuitive.

swiggumj:
Previous NANOGrav releases have used `-cut` for this purpose and in timing analysis config files, we have an `ignore` block. I guess I don't mind one way or the other, but I suggested `-cut` from a backwards-compatibility standpoint.

scottransom:
Yeah, I think we definitely want `-cut` for this.  The question is simply if we want any other aliases for commenting TOAs using flags.  I personally think `-ignore` makes sense for interactive work.  But I don't feel strongly.

scottransom:
I don't know of any others, besides what TEMPO2 has defined (and which are described in the T2 Users Manual):
```
-r (receiver)
-b (backend)
-i/-info (info)
-dm
-p (phase)
-t (telescope)
```
And IPTA DR1 paper (Verbiest et al) mentioned using a `-simul` flag and a `-dmo` flag (simultaneous observations and DM offset, respectively).

paulray:
Looks like you are missing `-pn` for pulse numbers

src/pint/toa.py
```diff
@@ -1708,6 +1711,11 @@ def write_TOA_file(self, filename, name="unk", format="tempo2"):
         if format.upper() in ("TEMPO2", "1"):
             outf.write("FORMAT 1\n")
 
+        # Warn the user if any of the TOAs have -ignore or -cut flags that
+        # they will be commented in the output file
+        if isinstance(commentflag, str) and len(self.get_flag_value(commentflag)[1]):
+            log.warning(f"TOAs with '-{commentflag}' flags will be commented")
```
aarchiba, line 1717: We already have too many messages. If you are worried about the user being surprised by commented TOAs, make the default None; if you are not worried, leave the warning out.

src/pint/toa.py
```diff
@@ -1728,11 +1736,14 @@ def write_TOA_file(self, filename, name="unk", format="tempo2"):
...
 
             flags = flags.copy()
-            if "clkcorr" in flags.keys():
+            if "clkcorr" in flags:
                 toatime_out = toatime - time.TimeDelta(flags["clkcorr"])
             else:
                 toatime_out = toatime
-            out_str = format_toa_line(
+            out_str = (
+                "C " if isinstance(commentflag, str) and (commentflag in flags) else ""
```
aarchiba, line 1744: None is hashable, so you can just leave this as `commentflag in flags` and rely on the keys all being actual flags.

tests/test_toa_writer.py
```diff
@@ -49,6 +49,25 @@ def test_roundtrip_topo_toa_Tempo2format(self):
...
         assert np.abs(ts.table["tdb"][0] - ts2.table["tdb"][0]) < 1.0e-15 * u.d
 
+    def test_commenting_toas(self):
+        # Create a barycentric TOA
+        t1time = Time(58534.0, 0.0928602471130208, format="mjd", scale="utc")
+        t1 = toa.TOA(t1time, obs="gbt", freq=0.0)
+        t2 = toa.TOA(t1time, obs="gbt", freq=0.0)
+        ts = toa.get_TOAs_list([t1, t2], ephem="DE421")
+        assert ts.ntoas == 2
+        ts.table[1]["flags"]["ignore"] = True  # flag one TOA
```
aarchiba, line 59: Maybe also check the default behaviour? That is, if the default is "cut", maybe check that that happens when you pass no argument?

tests/test_toa_writer.py
```diff
@@ -49,6 +49,25 @@ def test_roundtrip_topo_toa_Tempo2format(self):
...
+        ts.write_TOA_file("testtopo.tim", format="Tempo2")
+        ts2 = toa.get_TOAs("testtopo.tim")
+        assert ts2.ntoas == 2  # none should be commented
+        ts.write_TOA_file("testtopo.tim", format="Tempo2", commentflag=None)
+        ts3 = toa.get_TOAs("testtopo.tim")
+        assert ts3.ntoas == 2  # none should be commented
+        ts.write_TOA_file("testtopo.tim", format="Tempo2", commentflag="ignore")
+        ts4 = toa.get_TOAs("testtopo.tim")
+        assert ts4.ntoas == 1  # one should be commented
+        print(ts.table, ts2.table, ts3.table, ts4.table)
```
aarchiba, line 69: This will never do anything - `pytest` only prints the output if an exception arose, in which case this line is never executed.

src/pint/toa.py
scottransom: I'm fine leaving the warning out.  I put it in there specifically because you said that you were worried that people might be surprised by having TOAs commented.

src/pint/toa.py
scottransom: What I'm doing is stricter than that.  I'm requiring that `commentflag` must be a string.  If it is anything else, including a number, and list, an array, or a `None`, then it won't write a comment.

tests/test_toa_writer.py
scottransom: I'm fine doing that, but checking for `ignore` tests for the default behavior (since `ignore` is a string like `cut`, just not the default -- and I don't think we need to test that Python's optional arguments work) as well as the user choosing something besides the default string.  Seems like a waste of a (small amount) of test time.

tests/test_toa_writer.py
scottransom: Those print() functions were in each of the class methods from before.  I was just carrying them along.  (and note that they *are* executed and actually print stuff if you use the `-s` option for pytest (which I use a lot for debugging tests).

tests/test_toa_writer.py
aarchiba: We had a bug earlier this year where the test suite never executed a method with the default value of its argument, so that default value had the wrong units and we never noticed. Tests also detect when the public API changes, and the default value is part of that public API, so it's a good idea to check that.

src/pint/toa.py
aarchiba: Indeed; my preferred solution is to default to not commenting any TOAs. But I don't think a warning is a good idea.

tests/test_toa_writer.py
```diff
@@ -45,34 +43,54 @@ def test_roundtrip_topo_toa_Tempo2format(self):
...
         assert np.abs(ts.table["tdb"][0] - ts2.table["tdb"][0]) < 1.0e-15 * u.d
 
+    def test_commenting_toas(self):
+        # Create a barycentric TOA
+        t1time = Time(58534.0, 0.0928602471130208, format="mjd", scale="utc")
+        t1 = toa.TOA(t1time, obs="gbt", freq=0.0)
+        t2 = toa.TOA(t1time, obs="gbt", freq=0.0)
+        ts = toa.get_TOAs_list([t1, t2], ephem="DE421")
+        assert ts.ntoas == 2
+        ts.write_TOA_file("testtopo.tim", format="Tempo2")
```
aarchiba, line 56: Does this leave a stray `testtopo.tim` lying around? Can you use a temporary directory or a StringIO?

tests/test_toa_writer.py
```diff
@@ -45,34 +43,54 @@ def test_roundtrip_topo_toa_Tempo2format(self):
...
+        ts2 = toa.get_TOAs("testtopo.tim")
+        assert ts2.ntoas == 2  # none should be commented
+        ts.table[0]["flags"]["cut"] = "do not like"  # default comment flag
```
aarchiba, line 59: What happens when we use flags with spaces in them? Does PINT handle this correctly? (What would correct handling look like - there isn't a defined form of quoting is there?)

tests/test_toa_writer.py
scottransom: Yes, it leaves a stray testtopo.tim lying around (and the other tests in there also leave files).  I can clean that up, although that is really beyond the scope of what the pull request is trying to do.

scottransom: Done.  And changed all into pytest format.

tests/test_toa_writer.py
scottransom: Yeah, I don't even know what that should look like.  Especially if a flag (or the flag's value) with spaces is in a .tim file.

scottransom: I changed the value to have s/ /_/ in this case.

scottransom:
This should be good to go now.

PR #1023, opened 2021-04-05
F test returns 1 if chi2 increases

aarchiba:
The F test is used to compare a model with some parameters frozen to a model with those parameters free. Normally the chi-squared should be smaller for the model with free parameters, but if the fitting fails or if there is no improvement and numerics are inaccurate you can obtain a spurious increase. Previously we returned an F test false positive probability of False, which is interpreted as zero in a floating-point context, and so the NANOGrav `timing_analysis` tools treated this the same as a massive decrease; now we return 1, as in, the probability of obtaining more decrease than this is 1. We also emit a warning.

### DIFF
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1321,8 +1321,7 @@ def ELL1_check(A1, E, TRES, NTOA, outstring=True):
 
 
 def shklovskii_factor(pmtot, D):
-    """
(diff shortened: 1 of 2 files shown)


aarchiba:
Coverage failure is meaningless.

aarchiba:
Also removed some unneeded xfails.

PR #1025, opened 2021-04-06
change_dmepoch works even if DMEPOCH not set

### LINKED ISSUE #1024: TimingModel.compare() fails if DMEPOCH is not set
There's no reason that the comparison function can't make the same assumption the model itself does (DMEPOCH==PEPOCH, I think).

aarchiba:
This allows `TimingModel.compare` to work even when the second model doesn't have DMEPOCH set; this comes up in NANOGrav timing_analysis summaries.

Fixes #1024


luojing1211:
I think this is good, but it needs to resolve the conflict.

PR #1026, opened 2021-04-07
Rewrite model builder

luojing1211:
The old model builder is confusing and has some bugs. This PR reorganizes the model builder.
The main algorithm is the same to keep the same, comparing the overlap of the parameters in parfile and in the component. 
================ Update ==================
This PR has been modified in the following aspects:
* Add parfile parsing function, for parfile/string to a dictionary 
* Add ModelBuilder._pintify_parfile function that translates the parfile information to pint recognized parameter names and unknown parameters 
* Add ModelBuilder._setup_model function that adds the value to the model. 
* More tests 
* several bug fix 

================ Road blocks ===============
* The doc building fails due to #1127


aarchiba:
`get_model_new` is a rewrite of `get_model`, written directly without any object-orientation. It hasn't been updated since we adopted it, so it doesn't work for all our par files any more, but it might be a good place to start for this?

src/pint/models/model_builder.py
```diff
@@ -26,276 +26,296 @@ class UnknownBinaryModel(ValueError):
...
+    def __call__(self, parfile):
+        pass
+
+    @lazyproperty
+    def param_component_map(self):
+        """ Return the parameter to component map.
+        """
+        p2c_map = {}
+        for k, cp in self.components.items():
+            for p in cp.params:
```
aarchiba, line 44: Should this include aliases?

src/pint/models/model_builder.py
```diff
@@ -26,276 +26,296 @@ class UnknownBinaryModel(ValueError):
...
         A dictionary with all the parfile parameters with values in string
-
         """
         param = {}
         repeat_par = {}
-        pfile = open(parfile, "r")
-        for l in [pl.strip() for pl in pfile.readlines()]:
+        param_inpar = []
+        for l in interesting_lines(lines_of(parfile), comments=("#", "C ")):
             # Skip blank lines
```
aarchiba, line 114: `interesting_lines` does this.

src/pint/models/model_builder.py
```diff
@@ -26,276 +26,296 @@ class UnknownBinaryModel(ValueError):
...
             if not l:
                 continue
-                # Skip commented lines
+            # Skip commented lines
```
aarchiba, line 117: And this.

src/pint/models/model_builder.py
```diff
@@ -26,276 +26,296 @@ class UnknownBinaryModel(ValueError):
...
+        Note
+        ----
+        The selection algorithm:
+        1. Look at the BINARY in the par file and catche the indicated binary model
+        2. Translate para file parameters to the pint parameter name
+        3. Get all the components which holds the parameters
+        4. Check if there is any conflict in the component(i.e., mulitple
+           components in one category)
+          4.1 Find the overlap between component special special parameter and
+              in parfile component.
```
aarchiba, line 214: I'm not sure I understand this resolution mechanism. This algorithm needs to be something we can explain to users - what Components end up in their model.

src/pint/models/model_builder.py
```diff
@@ -26,276 +26,296 @@ class UnknownBinaryModel(ValueError):
...
-        if not self.param_unrecognized == {}:
-            for ctrlp in self.control_params:
-                if ctrlp in self.param_unrecognized:
-                    info[ctrlp] = self.param_unrecognized[ctrlp]
+        result_componets = []
+        param_inpar = self.parse_parfile(parfile)
+        # 1. iteration read parfile with a no component timing_model to get
+        # the overall control parameters. This will get us the binary model name
+        # build the base fo the timing model
+        self.timing_model.read_parfile(parfile)
```
aarchiba, line 221: Can't you just look up BINARY in `param_inpar`?

src/pint/models/model_builder.py
```diff
@@ -26,276 +26,296 @@ class UnknownBinaryModel(ValueError):
...
-                        if p.startswith(ctrlp):
-                            info[ctrlp] = self.param_unrecognized[ctrlp]
-        return info
+                    possible_components[cp] = [k]
+        # Get conflicting components based on categories
+        # conflict_components = {category: [component1, component2, ...]}
+        conflict_components = {}
+        for k in possible_components.keys():
+            cate = self.component_category_map[k]
+            if cate in conflict_components.keys():
```
aarchiba, line 256: `collections.defaultdict(list)` does this for you.

src/pint/models/model_builder.py
```diff
@@ -26,276 +26,296 @@ class UnknownBinaryModel(ValueError):
...
+            # Check if there are more than one components have the same rank
+            # If the highest rank has the same rank with the second highest,
+            # the conflict is not resolved by speical parameters.
+            if rank[1][1] != rank[0][1]: # has an unique highest case.
+                result_componets.append(self.components[rank[0][0]])
+                resolved.append(ca)
+            else:
+                # First no overlap case, which should mean no special parameters
+                # in the component, will do a search depends on the overlap of
+                # all parameters.
```
aarchiba, line 288: I think - if I understand this correctly - this is counting how many parameters overlap between what's in the par file and what's in the component? This is not an easy rule for users to understand. I suggest raising an exception for any ambiguous cases where a distinctive parameter doesn't force the choice. (e.g. NE_SW is only in solar_wind_dispersion so it is required.)

aarchiba:
One of the reasons I wrote `get_model_new` instead of replacing `get_model` was that that allowed me to compare the two and confirm that they produced the same timing model for all our par files as `get_model` did. Some of those tests are still around. This kind of comparison is quite important, as if we suddenly start interpreting par files differently that will impact our users. You could keep a copy of the old model builder in a test file if you want to really clean up but still do this comparison.

aarchiba:
There are some relevant tests in https://github.com/nanograv/PINT/blob/master/tests/test_model_manual.py

src/pint/models/model_builder.py
luojing1211: Good idea.

src/pint/models/model_builder.py
luojing1211: Will change it.

src/pint/models/model_builder.py
luojing1211: Yes, Counting the overlap is the idea. I am very hesitant to add any case-specific checks here in this general code. We can have a separate function to do the job of hard fail.

aarchiba: I don't think counting overlap is a good idea at all. I think it will surprise users when their model suddenly switches because they added a parameter that pushes them over the threshold. My idea was that for every model there should be at least one parameter not present in any alternatives, and that these parameters should force the selection. (For example, if you have NE_SW, you clearly need a solar wind dispersion model. No special casing here because that parameter only occurs in that model.) If you have two alternatives for which one has a subset of the parameters of the other, the one with the smaller parameter set should be chosen unless there is a parameter from the larger. Are there any cases where models cannot be distinguished in this way?

luojing1211: The component_special_parameter is meant to tell the model builder that choose one component vs the other one. The current setup is that it will check the over_lap of component_special_parameters first, if the conflict still exists, it will check the overlap of all parameters. I can remove the latter check and raise a conflict error.

src/pint/models/model_builder.py
luojing1211: Changed.

src/pint/models/model_builder.py
luojing1211: Done

src/pint/models/model_builder.py
luojing1211: I did update the code here. I think it is much simpler.

luojing1211:
@aarchiba Can you take a look at the current development? I am going to fix the last failure and add more tests.

src/pint/models/model_builder.py
```diff
@@ -24,295 +25,402 @@
...
+                    continue
+        if result is None:
+            raise UnknownBinaryModel("Pulsar system/Binary model component {}"
+                                     " is not provided.".format(system_name))
+        else:
+            return result
 
-    def sort_components(self, category_order=DEFAULT_ORDER):
-        """Sort the components into order.
+    def alias_2_pint_param(self, alias):
```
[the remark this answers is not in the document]
aarchiba: Isn't this covered by the parameter map code above?

src/pint/models/model_builder.py
```diff
@@ -24,295 +25,402 @@
...
+        """ Translate the alias to a PINT parameter name.
+        """
+        pint_par = self.param_alias_map.get(alias, None)
+        # If it is not in the map, double check if it is a repeatable par.
```
aarchiba, line 194: This is confusing - isn't JUMP a repeatable par? It doesn't have prefixes. Does it use the above option or the below?

src/pint/models/model_builder.py
```diff
@@ -24,295 +25,402 @@
...
+        components: list
+            A list of the component names.
+        params: list
+            A list of matching parameter names.
+        primary_key: str
+            The primary compare key(e.g., `params` or `component_special_params`)
+
+        Return
+        ------
+        A sorted tuple list based on how many overlap parameters.
```
aarchiba, line 231: I do not think that this is an intelligible criterion for users. Requiring them to count how many parameters match different models to know how their par file will be interpreted is not acceptable. This does not appear to ever be used anyway, so please remove it.

src/pint/models/model_builder.py
```diff
@@ -24,295 +25,402 @@
...
     """
     try:
         contents = parfile.read()
     except AttributeError:
         contents = None
     if contents is None:
         # parfile is a filename and can be handled by ModelBuilder
-        mbmodel = ModelBuilder(parfile)
-        model = mbmodel.timing_model
+        mbmodel = ModelBuilder()
```
aarchiba, line 410: Shouldn't there be one ModelBuilder object created for the entire program? This will make the lazyproperties work. Just set `model_builder = None` at the module level, then check if it's None and create one to replace it if necessary here.

src/pint/models/model_builder.py
```diff
@@ -24,295 +25,402 @@
...
+        model = mbmodel(parfile)[0]
```
aarchiba, line 411: This is confusing. Why `[0]`? What state is `model` in - I guess it is supposed to have the right components but not any values yet? Can you at least have the method that constructs this half-cooked model called something that makes that clear?

src/pint/models/model_builder.py
```diff
@@ -24,295 +25,402 @@
...
         model.name = parfile
+        model.read_parfile(parfile)
```
aarchiba, line 413: What does `read_parfile` do? Is it a function users should ever call? Shouldn't it live inside ModelBuilder?

src/pint/models/model_builder.py
```diff
@@ -24,295 +25,402 @@
...
         return model
     else:
         with tempfile.TemporaryDirectory() as td:
             fn = os.path.join(td, "temp.par")
             with open(fn, "wt") as f:
                 f.write(contents)
-            return ModelBuilder(fn).timing_model
+            mb = ModelBuilder()
```
aarchiba, line 420: If you're going to copy and paste this code from above please use the same variable names.

src/pint/models/model_builder.py
```diff
@@ -355,11 +462,11 @@ def get_model_and_toas(
         Filename to use for caching loaded file. Defaults to adding ``.pickle.gz`` to the
         filename of the timfile, if there is one and only one. If no filename is available,
         or multiple filenames are provided, a specific filename must be provided.
-
     Returns
     -------
     A tuple with (model instance, TOAs instance)
     """
+    pass
```
aarchiba, line 469: What is this `pass` for?

src/pint/models/parameter.py
```diff
@@ -1045,7 +1058,7 @@ class prefixParameter:
         Example parameter class template for quantity and value setter
     long_double : bool, optional
         Set float type quantity and value in numpy long doubles.
-    time_scale : str, optional
+    time_scale : str, optionalParameter(
```
aarchiba, line 1061: This looks wrong.

src/pint/models/timing_model.py
```diff
@@ -895,8 +905,8 @@ def remove_component(self, component):
         cp, co_order, host, cp_type = self.map_component(component)
         host.remove(cp)
 
-    def _locate_param_host(self, components, param):
-        """Search for the parameter host component.
+    def _locate_param_host(self, param):
+        """Search for the parameter host component in the timing model.
```
aarchiba, line 909: Please clarify this. Perhaps "Find the component containing the given parameter." Does this expect the string parameter name or an actual parameter object?

tests/test_model_builder.py
```diff
@@ -0,0 +1,109 @@
...
+ECORR -f L-wide_PUPPI   0.31843
+JUMP -fe L-wide      -0.000009449  1       0.000009439
+"""
+
+# @pytest.fixture
+# def tmp_dir(tmpdir):
+#     yield str(tmpdir)
+def test_model_builder_class():
+    mb = ModelBuilder()
+    categore = mb.category_component_map
```
aarchiba, line 87: categore -> category (not sure what this is supposed to be?)

tests/test_model_builder.py
```diff
@@ -0,0 +1,109 @@
...
+    assert len(mb.param_component_map['PX']) == len(categore['astrometry'])
+    assert len(mb.component_category_map) == len(mb.components)
+    assert len(mb.param_alias_map) == len(mb.param_component_map)
+    print(mb.repeatable_param)
```
aarchiba, line 91: Do not print.

tests/test_model_builder.py
```diff
@@ -0,0 +1,109 @@
...
+    assert len(mb.repeatable_param) == 24
```
aarchiba, line 92: Why 24? Won't this test be a spurious failure if someone adds a new mask parameter?

tests/test_model_builder.py
```diff
@@ -0,0 +1,109 @@
...
+    #assert len(mb.param_component_map) == 10
+
+def test_model_par():
+    mb = ModelBuilder()
+    param_inpar, repeat = mb.parse_parfile(io.StringIO(test_par1))
+    assert len(param_inpar) == 60
+    assert len(repeat) == 4
+    comps, conflict, unknown_param = mb.choose_model(param_inpar)
```
aarchiba, line 100: Shouldn't you check conflict and unknown_param?

tests/test_model_builder.py
```diff
@@ -0,0 +1,109 @@
...
+    mb = ModelBuilder()
+    categore = mb.category_component_map
+    assert len(mb.param_component_map['PX']) == len(categore['astrometry'])
+    assert len(mb.component_category_map) == len(mb.components)
+    assert len(mb.param_alias_map) == len(mb.param_component_map)
+    print(mb.repeatable_param)
+    assert len(mb.repeatable_param) == 24
+    #assert len(mb.param_component_map) == 10
+
+def test_model_par():
```
aarchiba, line 95: The test name should specify what is being tested - that the model par loads correctly? Or just without exceptions? This is some internal test of ModelBuilder, isn't it?

tests/test_model_builder.py
```diff
@@ -0,0 +1,109 @@
...
+    mb = ModelBuilder()
+    param_inpar, repeat = mb.parse_parfile(io.StringIO(test_par1))
+    assert len(param_inpar) == 60
+    assert len(repeat) == 4
+    comps, conflict, unknown_param = mb.choose_model(param_inpar)
+    # print(comps)
+    tm, unknown = mb(io.StringIO(test_par1))
```
aarchiba, line 102: Can you check anything about tm and unknown?

tests/test_model_builder.py
```diff
@@ -0,0 +1,109 @@
...
+    # print(conflict)
+    print(unknown)
+    # print(repeat)
+    assert len(comps) == 14
+
+
+    #tm = mb(test_par1)
```
aarchiba, line 109: Do you ever test that aliases are recognized?

tests/test_model_builder.py
```diff
@@ -0,0 +1,109 @@
...
+    assert len(repeat) == 4
+    comps, conflict, unknown_param = mb.choose_model(param_inpar)
+    # print(comps)
+    tm, unknown = mb(io.StringIO(test_par1))
+    # print(conflict)
+    print(unknown)
+    # print(repeat)
+    assert len(comps) == 14
+
+
```
aarchiba, line 108: Do you ever test conflict resolution?

tests/test_model_builder.py
```diff
@@ -0,0 +1,109 @@
+"""Test model builder using variance of parfile input"""
+
+from glob import glob
```
aarchiba, line 3: Are these imports all used?

src/pint/models/parameter.py
```diff
@@ -109,9 +109,11 @@ def __init__(
         get_value=lambda x: x,
         prior=priors.Prior(priors.UniformUnboundedRV()),
         set_uncertainty=fortran_float,
+        repeatable=False
```
aarchiba, line 112: Why do you need a `repeatable` attribute? Isn't it the case that only `maskParameter`s are repeatable? You could just set that in `__init__` - False for the  base class and True in `maskParameter.__init__`.

tests/test_model_builder.py
```diff
@@ -0,0 +1,109 @@
...
+    assert len(param_inpar) == 60
+    assert len(repeat) == 4
+    comps, conflict, unknown_param = mb.choose_model(param_inpar)
+    # print(comps)
+    tm, unknown = mb(io.StringIO(test_par1))
+    # print(conflict)
+    print(unknown)
+    # print(repeat)
+    assert len(comps) == 14
+
```
aarchiba, line 107: Do you test that all par files in our datafiles directory can be loaded without exceptions?

aarchiba:
The documentation for all this is visible at: https://nanograv-pint--1026.org.readthedocs.build/en/1026/api/pint.models.model_builder.html#pint.models.model_builder.get_model 

This is not a useful bit of documentation for `get_model`. You should be able to copy the existing docstring and start from that.

src/pint/models/model_builder.py
```diff
@@ -24,295 +25,402 @@
...
+            cp_ca[k] = cp.category
+        return cp_ca
+
+    @lazyproperty
+    def component_unique_params(self):
+        """ Return the unique parameter names in each componens. Note this only
+        retself.timing_model = TimingModel()urns the pint defined parameter name, not including the aliases
+        """
+        component_special_params = defaultdict(list)
+        for param, cps in param_component_map.items():
```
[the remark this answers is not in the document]
luojing1211: It not used in the current code. But I think it would be a helpful property for the user.

src/pint/models/model_builder.py
```diff
@@ -24,295 +25,402 @@
...
+    def param_alias_map(self):
+        """ Return the aliases map of all parameters
+        """
+        alias = {}
+        for k, cp in self.components.items():
+            for p in cp.params:
+                par = getattr(cp, p)
+                alias[p] = p
+                for als in par.aliases:
+                    alias[als] = p
```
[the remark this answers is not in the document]
luojing1211: This will not work, since we have components are inherited from the other ones. Some aliases will occur multiple times, but they represent the same parameter. Maybe we can keep track of the value.

src/pint/models/model_builder.py
```diff
@@ -24,295 +25,402 @@
...
+    def search_pulsar_system_components(self, system_name):
+        """ Search the component for the pulsar system, mostly binaries.
+        """
+        all_systems = self.category_component_map['pulsar_system']
+        result = None
+        # Search the system name first
+        print(system_name)
+        if system_name in all_systems:
+            result = self.components[system_name]
+        else: # search for the pulsar system aliases
```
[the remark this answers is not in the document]
luojing1211: It maps the binary name to the binary class name (e.g., DD ---> BinaryDD)

src/pint/models/model_builder.py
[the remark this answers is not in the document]
luojing1211: The parameter--> Aliases map does not have the higher index prefix and mask parameter. The gets recognize parameters like T2EQUAD2 and convert it to EQUAD2

src/pint/models/model_builder.py
luojing1211: This is for the case like "JUMP3", which is not in the aliases map. but should be recognized.

src/pint/models/model_builder.py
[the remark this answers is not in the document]
aarchiba: Will the user have access to ModelBuilder? It is currently not published in `__all__` or documented.

src/pint/models/model_builder.py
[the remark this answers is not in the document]
aarchiba: A new alias that points to a different parameter name is an error; a new occurrence of an alias that points to the same parameter name is fine.

aarchiba:
I think the cleanest way to make one global model builder object (so that the lazy properties have the effect of caching for longer than the duration of one `get_model` call) would be to have a `get_model_builder()` which returns the value of `_model_builder`, constructing and assigning a new one if that variable is None.

src/pint/models/model_builder.py
luojing1211: read_parfile reads in the parameter values into timing model. I think it is useful to put it in the timing model. The user can build a timing model manually and then reads the value in.

src/pint/models/model_builder.py
luojing1211: This is changed.

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+        def __init__(self):
+            super(SimpleModel2, self).__init__()
+            self.add_param(
+                floatParameter(name="TESTPARAMF0", aliases=["F0"], value=0.0, unit="s")
+            )
+
+    with pytest.raises(ConflictAliasError):
+        mb2 = ModelBuilder()
+    del Component.component_types["SimpleModel2"]
```
aarchiba, line 222: This should be in a try/finally to be sure it gets zapped even if an exception occurs (for example if the test fails)

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+    """
+    mb = ModelBuilder()
+    category = mb.category_component_map
+    assert len(mb.param_component_map["PX"]) == len(category["astrometry"])
+    assert len(mb.component_category_map) == len(mb.components)
+    assert len(mb.param_alias_map) == len(mb.param_component_map)
+    # test for new components
+    assert "SimpleModel" in mb.components
+    simple_comp = mb.components["SimpleModel"]
+    simple_comp.add_param(
```
aarchiba, line 156: Does this modify the ModelBuilder's internal data structures? What is it testing?

luojing1211: Since PINT uses the metaclass when the user defines a new component the"Component" class would get updated but not the model builder. So we need to have a new model builder.

src/pint/models/pulsar_binary.py
```diff
@@ -211,11 +211,14 @@ def update_binary_object(self, toas=None, acc_delay=None):
                 aliase = []
 
             if hasattr(self, par) or list(set(aliase).intersection(self.params)) != []:
-                try:
-                    pint_bin_name = self.match_param_aliases(par)
-                except ValueError:
+                pint_bin_name = self.match_param_aliases(par)
```
aarchiba, line 214: Shouldn't `match_param_aliases` raise an exception if nothing is found? That would simplify this code and avoid having to check every time the function is called.

src/pint/models/pulsar_binary.py
```diff
@@ -325,6 +328,7 @@ def change_binary_epoch(self, new_epoch):
                 getattr(self, k).quantity
                 for k in self.get_prefix_mapping("FB").values()
             ]
+            print("binary", fbterms)
```
aarchiba, line 331: Do not print here. If this is a debugging statement use `log.debug`.

src/pint/models/timing_model.py
```diff
@@ -406,12 +414,21 @@ def free_params(self, params):
...
     def match_param_aliases(self, alias):
         """Return the parameter corresponding to this alias."""
-        for p in self.params:
+        # Search the top level first.
+        for p in self.top_level_params:
             if p == alias:
                 return p
             if alias in getattr(self, p).aliases:
                 return p
-        raise ValueError("{} is not recognized as a parameter or alias".format(alias))
```
aarchiba, line 414: Raising an exception is much more pythonic and avoids the headaches caused by returning None. If you want to make it easier to catch you can create a custom subclass of ValueError.

src/pint/models/timing_model.py
aarchiba: Does this accept aliases?

src/pint/models/timing_model.py
```diff
@@ -2265,7 +2296,7 @@ def get_prefix_mapping(self, prefix):
             par = getattr(self, parname)
             if par.is_prefix and par.prefix == prefix:
                 mapping[par.index] = parname
-        return mapping
+        return OrderedDict(sorted(mapping.items()))
```
aarchiba, line 2299: Why an OrderedDict?

src/pint/models/timing_model.py
```diff
@@ -2409,13 +2440,66 @@ def get_prefix_mapping_component(self, prefix):
...
+                par = getattr(self, prefix_pname)
+                if par.is_prefix:
+                    raise ValueError(
+                        f"Found a similar prefixed parameter '{prefix_pname}'"
+                        f" But parameter {par.prefix}{idx} need to be added"
+                        f" to the model."
+                    )
+        return pname
+
+        # if pname is None: # Try the prefix parameter
```
aarchiba, line 2482: Why  is this still here?

src/pint/models/timing_model.py
```diff
@@ -2409,13 +2440,66 @@ def get_prefix_mapping_component(self, prefix):
...
+            if pname is not None:
+                par = getattr(self, pname)
+                if par.is_prefix:
+                    raise ValueError(
+                        "Prefix {} maps to mulitple parameters"
+                        ". Please specify the index as well.".format(alias)
+                    )
+            else:
+                # Not a prefix, not an alias
+                return None
```
aarchiba, line 2468: Please raise an exception if not found, as you do below.

tests/test_gls_fitter.py
```diff
@@ -10,6 +10,7 @@
 from pint import toa
 from pint.fitter import GLSFitter
 from pinttestdata import datadir
+import matplotlib.pyplot as plt
```
aarchiba, line 13: Does this force our test suite to be dependent on matplotlib? Does this still work okay if tests are running headless? Is there any need for this import?

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+    ComponentConflict,
+)
+from pint.models.timing_model import PhaseComponent, Component
+from pint.models.parameter import floatParameter
+from pint.utils import split_prefixed_name, PrefixError
+
+
+test_par1 = """
+PSR              B1855+09
+#RAJ             18:57:36.3932884         1  0.00002602730280675029
```
aarchiba, line 22: Why leave these commented lines? Please make the par file as simple and to the point as possible.

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+ECORR -f L-wide_PUPPI   0.31843
+JUMP -fe L-wide      -0.000009449  1       0.000009439
+"""
+
+
+class SimpleModel(PhaseComponent):
+    """Very simple test model component
+    """
+
+    register = True
```
aarchiba, line 96: Does this permanently register itself in PINT's class hierarchy when the test module is imported? This can result in users seeing this bogus component if they run the test suite from within their interpreter.

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+def test_model_builder_class():
+    """Test if model builder collected components information correctly
+    """
+    mb = ModelBuilder()
+    category = mb.category_component_map
+    assert len(mb.param_component_map["PX"]) == len(category["astrometry"])
+    assert len(mb.component_category_map) == len(mb.components)
+    assert len(mb.param_alias_map) == len(mb.param_component_map)
```
aarchiba, line 152: The keys of these two maps should be identical, shouldn't they?

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+    return simple_model
+
+
+def test_model_builder_class():
+    """Test if model builder collected components information correctly
+    """
+    mb = ModelBuilder()
+    category = mb.category_component_map
+    assert len(mb.param_component_map["PX"]) == len(category["astrometry"])
+    assert len(mb.component_category_map) == len(mb.components)
```
aarchiba, line 151: The keys of the left one should be the same, in some order, as the elements of the list on the right?

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+    )
+    return simple_model
+
+
+def test_model_builder_class():
+    """Test if model builder collected components information correctly
+    """
+    mb = ModelBuilder()
+    category = mb.category_component_map
+    assert len(mb.param_component_map["PX"]) == len(category["astrometry"])
```
aarchiba, line 150: This will break if someone adds an astrometry model that doesn't include PX, and that breakage will not indicate that there's anything wrong with ModelBuilder.

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+
+def test_aliases_mapping():
+    """Test if aliases gets mapped correclty
+    """
+    mb = ModelBuilder()
+    # all alases should be mapped to the components
+    assert len(mb.param_alias_map) == len(mb.param_component_map)
+
+    # Test if the param_alias_map is passed by pointer
+    _ = mb._add_alias_to_map("TESTAX", "TESTAXX", mb.param_alias_map)
```
aarchiba, line 169: If you don't use the return value, just don't assign to anything. `_ = f()` -> `f()`

aarchiba: Shouldn't this fail because there is no parameter `TESTAX`? What is this code supposed to be testing? Whether python dictionary lookups work?

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+    assert "TESTAX" in mb.param_alias_map
+    # Test existing entry
+    _ = mb._add_alias_to_map("F0", "F0", mb.param_alias_map)
```
aarchiba, line 172: Should `F0` be in the alias map or not, initially?

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+    assert mb.param_alias_map["F0"] == "F0"
+    # Test repeatable_params with differnt indices.
+    for rp in mb.repeatable_param:
```
aarchiba, line 175: Can you make the repeatable params test a separate test? I'm not quite sure I understand what repeatable parameters are.

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+def test_conflict_alias():
+    """Test if model builder detects the alais conflict.
+    """
+    mb = ModelBuilder()
+    # Test conflict parameter alias name
+    with pytest.raises(ConflictAliasError):
+        _ = mb._add_alias_to_map("F0", "F1", mb.param_alias_map)
+    # Define conflict alais from component class
```
aarchiba, line 206: Please put this in a separate test.

tests/test_model_builder.py
```diff
@@ -0,0 +1,294 @@
...
+    # Test subset
+    superset = mb._is_subset_component(sub_set_model)
+    assert superset == "Spindown"
+
+
+def test_model_from_par():
+    """Test Get model from test par file.
+    """
+    mb = ModelBuilder()
+    param_inpar, repeat = mb.parse_parfile(io.StringIO(test_par1))
```
aarchiba, line 266: Can you put `test_par1` inline here, so we can see why 14 is the right number of components?

tests/test_model_builder.py
luojing1211: This is removed.

tests/test_model_builder.py
luojing1211: This is removed.

src/pint/models/parameter.py
```diff
@@ -114,9 +114,11 @@ def __init__(
         prior=priors.Prior(priors.UniformUnboundedRV()),
         set_uncertainty=fortran_float,
         use_alias=None,
+        repeatable=False,
```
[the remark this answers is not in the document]
luojing1211: Having this attribute gives us two benefits: 
1. Much easy to identify if a parameter is repeatable. This is a much more intuitive way to do than checking the types.  It makes the ModelBuilder.repeatable_param() much more generic.  
2. They will allow repeating parameters for other types, not just the Masked parameters. 
This is something I want to use in the parameter update.

tests/test_model_builder.py
luojing1211: _add_alias_to_map() is a private function that puts the aliases in the map and it is a part of the aliases_map property This function tests if adding a new alias works.

tests/test_model_builder.py
luojing1211: It should be. Here it is testing Adding an exact same alias -> parameter entry as an existing one to the map should not fail.

src/pint/models/parameter.py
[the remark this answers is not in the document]
aarchiba: In terms of checking whether parameters are repeatable, providing a .repeatable property, established by inheritance, is a very convenient way of letting the type determine whether a parameter can be repeated.

For a parameter that isn't a mask or prefix parameter, what can it mean that it is repeatable? What meaning could we assign to having `PB 0.9` and `PB 0.8` in the same par file? Surely in order to be usable - that is, to have a usable name within PINT - any such would need to be of a new type, which could then report whether it is repeatable using the method above, rather than adding to the complexity of constructing every Parameter object.

src/pint/models/timing_model.py
```diff
@@ -406,12 +415,21 @@ def free_params(self, params):
...
+        # Search the top level first.
+        for p in self.top_level_params:
             if p == alias:
                 return p
             if alias in getattr(self, p).aliases:
                 return p
-        raise ValueError("{} is not recognized as a parameter or alias".format(alias))
+        # if not in the top level, check parameters.
+        for cp in self.components.values():
+            pint_par = cp.match_param_aliases(alias)
```
aarchiba, line 426: Can `cp.match_param_aliases` be modified to raise an exception if there is no match?

src/pint/models/timing_model.py
```diff
@@ -2534,6 +2619,176 @@ def __init__(self):
...
 
 
+class AllComponents:
+    """ A class for the components pool.
+
+    This object stores and manages the instances of component classes with class
+    attribute .register = True. This includes the PINT built-in components and
+    user defined components. It is designed for helping model building and
+    parameter seraching, not for directly data analysis. Thus, the instances do
+    not have any valid parameter values. Runing `.validate()` function will fail.
```
aarchiba, line 2629: Please explain why this class stores these half-formed invalid instances. I believe it is because classes do not know about their parameters until an instance is created, and this object needs to obtain that information?

src/pint/models/timing_model.py
```diff
@@ -2565,3 +2820,15 @@ def __str__(self):
         if self.msg is not None:
             result += "\n  " + self.msg
         return result
+
+
+class ConflictAliasError(ValueError):
```
aarchiba, line 2825: Should this subclass TimingModelError?

src/pint/models/timing_model.py
```diff
@@ -2565,3 +2820,15 @@ def __str__(self):
...
+    """If the same alias is used for different parameters."""
+
+    pass
+
+
+class UnknownBinaryModel(ValueError):
```
aarchiba, line 2831: Should this subclass TimingModelError?

tests/test_model_manual.py
```diff
@@ -9,8 +9,13 @@
 from pint.models.astrometry import AstrometryEquatorial
 from pint.models.dispersion_model import DispersionDM, DispersionDMX
 from pint.models.spindown import Spindown
-from pint.models.model_builder import UnknownBinaryModel, get_model, get_model_new
-from pint.models.timing_model import MissingParameter, TimingModel, Component
+from pint.models.model_builder import get_model, get_model_new
```
aarchiba, line 12: Since this replaces both the existing `get_model` and the broken `get_model_new`, please delete `get_model_new`. This will necessitate better tests, but testing your new code against broken code isn't helpful.

src/pint/models/model_builder.py
```diff
@@ -2,317 +2,366 @@
...
+                )
+                if v.category == "pulsar_system":
+                    # The pulsar system will be selected by parameter BINARY
+                    continue
+                else:
+                    raise ComponentConflict(m)
 
+    def _get_component_param_overlap(self, component):
+        """Check if one component's parameters are overlaped with another
+        component.
```
aarchiba, line 97: The first line of a docstring needs to be on one line.

src/pint/models/model_builder.py
```diff
@@ -2,317 +2,366 @@
...
+
+        Parameters
+        ----------
+        components: component object
```
aarchiba, line 101: Please use the actual fully qualified class name so Sphinx can make a link. Also this is not the name of the argument in the function signature. (Parameter names are part of a function's public API in python.)

src/pint/models/model_builder.py
```diff
@@ -2,317 +2,366 @@
...
+            The superset component name, or None
+        """
+        overlap = self._get_component_param_overlap(component)
+        for k, v in overlap.items():
+            if v[1] == 0:
+                return k
+        return None
+
+    def parse_parfile(self, parfile):
+        """Preprocess the par file.
```
aarchiba, line 148: Blank line after the first line in a docstring.

src/pint/models/model_builder.py
```diff
@@ -2,317 +2,366 @@
...
 
-        Parameters
-        ----------
-        category_order: list, optional
-           The order for the order sensitive component categories.
+        Parameter
+        ---------
+        param_inpar: str
+            Dictionary of the unique parameters in .par file with the key is the
+        parfile line. Function `.parse_parfile` returns this dictionary.
```
aarchiba, line 187: Please use Sphinx's syntax to refer to a function so that the documentation has a live link.

src/pint/models/model_builder.py
```diff
@@ -2,317 +2,366 @@
...
+            cf_cps = conflict_components.get(ps_cp, None)
+            if cf_cps is not None:  # Had conflict, but resolved.
+                for cf_cp in cf_cps:
+                    del conflict_components[cf_cp]
+                del conflict_components[ps_cp]
+        return selected_components, conflict_components, unrec_param
+
+    def _add_indexed_params(self, timing_model, indexed_params):
+        """Add the parameters with unknown number/indexed in parfile (maskParameter/
+        prefixParameter) to timing model.
```
aarchiba, line 271: First line of a docstring should be on one line. If too long, write a one-line summary and put this in the first paragraph.

aarchiba:
https://nanograv-pint--1026.org.readthedocs.build/en/1026/api/pint.models.model_builder.html The documentation is ill-formatted and hard to read. A module docstring would also help greatly in finding the documentation.

src/pint/models/parameter.py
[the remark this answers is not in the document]
luojing1211: I will change .repeatable to a property.

src/pint/models/pulsar_binary.py
luojing1211: This part will have to be addressed in #1042 first. It will have a conflict here, once it gets merged.

aarchiba:
Please include tests that verify that:

- All valid par files included with PINT can be loaded successfully
- Parameters specified by alias work, both when they are only part of an existing component and when the aliased name is the one that makes a new/different component necessary. (This is the bug that we are waiting for this PR to fix.)
- Parfiles that might be ambiguous about which components should be included get the correct components

aarchiba:
The documentation should probably link to the reference doc on timing models, even if #1065 isn't merged yet, so users can easily find which parameters work.

src/pint/models/timing_model.py
```diff
@@ -2565,3 +2913,21 @@ def __str__(self):
         if self.msg is not None:
             result += "\n  " + self.msg
         return result
+
+
+class ConflictAliasError(TimingModelError):
```
aarchiba, line 2918: There's no need for "Error" in the name, it's an exception. And "AliasConflict" would be clearer than "ConflictAlias".

tests/test_all_component_and_model_builder.py
```diff
@@ -0,0 +1,296 @@
...
+@pytest.fixture
+def simple_model_overlap(simple_model):
+    simple_model.add_param(floatParameter(name="F0", aliases=[], value=0.0, unit="1/s"))
+    return simple_model
+
+
+@pytest.fixture
+def simple_model_alias_overlap():
+    simple_model = SimpleModel()
+    simple_model.add_param(
```
aarchiba, line 63: Shouldn't this raise an exception? Trying to add any parameter that shares a name or alias with an existing parameter?

tests/test_all_component_and_model_builder.py
```diff
@@ -0,0 +1,296 @@
...
+    )
+    return simple_model
+
+
+def test_model_builder_class():
+    """Test if AllComponents collected components information correctly
+    """
+    mb = AllComponents()
+    category = mb.category_component_map
+    assert len(mb.param_component_map["PX"]) == len(category["astrometry"])
```
aarchiba, line 77: Can you test more than just the length? Should all the components, or all the names, be the same?

tests/test_all_component_and_model_builder.py
```diff
@@ -0,0 +1,296 @@
...
+    T2EFAC -f 430_ASP   1.147
+    T2EQUAD -f L-wide_PUPPI   0.25518
+    T2EQUAD -f 430_ASP   0.01410
+    ECORR -f 430_PUPPI   0.00601
+    ECORR -f L-wide_PUPPI   0.31843
+    JUMP -fe L-wide      -0.000009449  1       0.000009439
+    """
+    mb = ModelBuilder()
+    param_inpar, repeat = mb.parse_parfile(io.StringIO(test_par1))
+    assert len(param_inpar) == 60
```
aarchiba, line 269: Can you make this 60 more meaningful? Perhaps the number of nonempty lines in the par file? That is, please change the test so it verifies that the length of param_inpar is equal to something like `len(l for l in test_par1.split("\n") if l.strip()`.

tests/test_all_component_and_model_builder.py
```diff
@@ -0,0 +1,296 @@
...
+    assert len(repeat) == 4
```
aarchiba, line 270: Please test what repeat is more explicitly - if I run the test suite and find out it's not 4 I won't have any idea why it should be 4. I think here it should be the set `{JUMP, T2EQUAD, T2EFAC, ECORR}`?

tests/test_all_component_and_model_builder.py
```diff
@@ -0,0 +1,296 @@
...
+    comps, conflict, unknown_param = mb.choose_model(param_inpar)
+    tm = mb(io.StringIO(test_par1))
+    assert len(comps) == 14
```
aarchiba, line 273: Again I don't know why 14. If you're going to test against the explicit list of what's in the par file, please include that list and not just its length.

aarchiba:
I started trying to write code to rename maskParameters (`JUMP17` -> `JUMP_tel_ao`) but ran into some peculiar problems that arise from the existing model builder. I see you're not changing the timing model from parfile code much, so I hope there won't be too many surprises. 

Just to be clear: the role of the model builder is to construct a timing model with the correct components; this timing model is then passed the parfile. Is this correct? Specifically, is it the job of the timing model code to construct any needed parameters on those components? Or are the parameters constructed (but not assigned values) in the model builder?

luojing1211:
The current model_builder class only selects the components and adds the necessary parameters like JUMPs. The parameter values are read in by a method in the timing model.  So your change would be effected.

aarchiba:
> The current model_builder class only selects the components and adds the necessary parameters like JUMPs. The parameter values are read in by a method in the timing model. So your change would be affected.

Okay, just to check: the model builder assembles all the components that are needed and also adds instances of all the parameters, but does not set any values. The timing model function reads the par file again and imports the values (only), emitting warnings if anything unfamiliar is found. Is this correct?

Does it make sense to separate these two operations? Shouldn't all this happen with one pass through the par file?

luojing1211:
Your statement is correct. In principle, we can put it together. But I think keeping the parfile value parse code in the TImingModel also makes sense so that one can put a timing model together manually and then read the parameter value

aarchiba:
> Your statement is correct. In principle, we can put it together. But I think keeping the parfile value parse code in the TImingModel also makes sense so that one can put a timing model together manually and then read the parameter value. I know it creates difficulty for the new mask parameter naming. I would like to have it solved as well. I remember you proposed an idea to have maskParameter as a dictionary and I thought about it, it seems a neat idea, which can avoid this problem.

I really cannot see how one would ever find it useful to have a manually assembled model but then read the values from a par file. If I for some reason had to do that, I would just `get_model` the parfile. How would what you get ever differ from what `get_model` ever gave you?

aarchiba:
Also on the tests needed:

- Binary models that do not support parameters like `SINI` should raise exceptions if these are specified in par files.

tests/test_all_component_and_model_builder.py
luojing1211: That is a good point. This will have to modify the add_parameter code, which is quite critical. So I think it is better to do it in another PR.

src/pint/models/timing_model.py
luojing1211: This is to make sure the map in the prefix mapping is ordered by the index.

tests/test_model_manual.py
```diff
@@ -86,19 +86,10 @@ def test_valid_model(tmp_dir, func, name, expectation):
...
-        m_old.get_params_mapping().keys()
-    )
-    assert set(m_new.components.keys()) == set(m_old.components.keys())
-
-
 @pytest.mark.xfail(
     reason="This parfile includes both ecliptic and equatorial coordinates"
 )
-@pytest.mark.parametrize("gm", [get_model, get_model_new])
+@pytest.mark.parametrize("gm", [get_model])
```
aarchiba, line 92: No point parametrizing over one function.

tests/test_model_manual.py
```diff
@@ -61,13 +66,8 @@ def test_component_categories(model):
     (get_model, "BT", pytest.raises(MissingParameter)),
     (get_model, "ELL1", pytest.raises(MissingParameter)),
     (get_model, "ELL1H", pytest.raises(MissingParameter)),
```
aarchiba, line 68: `get_model` can be removed as a parameter here.

luojing1211:
> Also on the tests needed:
> 
> * Binary models that do not support parameters like `SINI` should raise exceptions if these are specified in par files.

Should this be in the test for the model builder? Checking parameters is in the Validate function. Should it be in the tests of components?

aarchiba:
> > Also on the tests needed:
> > 
> > * Binary models that do not support parameters like `SINI` should raise exceptions if these are specified in par files.
> 
> Should this be in the test for the model builder? Checking parameters is in the Validate function. Should it be in the tests of components?

Yes. If you attempt to load a parfile with a bogus parameter, `get_model` should emit an exception. Exactly where it should originate in the code doesn't matter much to me. But the whole point of this PR is that `get_model` does not behave the way it should, so we need tests to verify that now it does.

aarchiba:
> @aarchiba I would like to push this forward. Can you give some examples of "Parfiles that might be ambiguous about which components should be included get the correct components"? Does this require some more development of our code?

I don't know off the top of my head which COmponents share parameters, so most of these examples should probably emit exceptions, but we should be clear about what happens (and by "be clear" I mean, "obvious from the documentation and verified in test cases"):

- Par file with both ELAT/ELONG and RAJ/DECJ (should probably emit exception)
- Par file with ELAT/ELONG and PMRA/PMDEC (should probably emit exception)
- Par files that call for ECORR using PINT and/or T/T2 names for the parameters (including a mix; should just work)
- Par files that include duplicate parameter values (like `F0 1` and `F0 2`) by the same name and by aliases (should emit exception)
- Par files with bogus meaningless non-comment lines, formatted like parameters and not (should emit exception)
- Par files with malformed parameters (like too many/not enough entries for a maskParameter)
- Par files with legal and illegal mixtures of PB/PBDOT and FBn
- Par files missing necessary parameters for their models
- Par files with inappropriate parameters for their models (SINI for DDK for example)

One hope I had for this PR was that it would be easy for a user to quickly query, for each Component, which Parameters were unique to it and would force its inclusion. This might give a guide to where complexities might arise, and would make it easy to add to the list of Components which parameters would lead to inclusion of the relevant Component.

I don't know what code development is needed to support these things; if you have a convincing set of test cases that work, then none is needed; if the test cases don't work, then we presumably need to develop code to make them work.

src/pint/models/timing_model.py
```diff
@@ -2639,6 +2697,269 @@ def __init__(self):
...
+            PINT parameters, it will raise a `UnknownParameter` error.
+        first_init_par : str
+            The parameter name that is first initialized in a component. If the
+            paramere is non-indexable, it is the same as ``pint_par``, otherwrise
+            it returns the parameter with the first index. For example, the
+            ``first_init_par`` for 'T2EQUAD25' is 'EQUAD1'
+
+        Notes
+        -----
+        Providing a indexable parameter without the index attached, it returns
```
aarchiba, line 2899: What is the rationale for having it accept `F` and return `F0, F0` when `F7` would return `F7, F0`? Why accept un-numbered values at all?

src/pint/models/timing_model.py
```diff
@@ -2670,3 +2991,21 @@ def __str__(self):
         if self.msg is not None:
             result += "\n  " + self.msg
         return result
+
+
+class ConflictAlias(TimingModelError):
```
aarchiba, line 2996: AliasConflict would be clearer.

src/pint/models/binary_ell1.py
```diff
@@ -169,7 +169,7 @@ def change_binary_epoch(self, new_epoch):
         else:
             fbterms = [
                 getattr(self, k).quantity
-                for k in self.get_prefix_mapping("FB").values()
+                for k in self.get_prefix_mapping_component("FB").values()
```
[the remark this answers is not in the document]
luojing1211: The get_perfix_mapping_component() function enumerates all FBx parameters and put the value in the format of {x: FBx value}. So it should be a one-to-one mapping. If there is a missing FBx, the validation function should give an error.

src/pint/models/timing_model.py
luojing1211: For JUMP, T2EQUAD...

src/pint/models/binary_ell1.py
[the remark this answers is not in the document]
aarchiba: So to function correctly, this code relies on `.validate()` being run at some point? That seems fragile. Can you please rewrite it so it uses the keys to construct the derivatives appropriately? That would ensure that `fbterms` is always in the right order as well.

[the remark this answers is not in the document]
luojing1211: This function maps all the prefix type of parameters including, `DMX_ `. Maybe for the case of derivatives, we need another function. But let me add a check in this function. When we restructure the Parameters, we should have a better design for it.

aarchiba:
> Please include tests that verify that:
> 
>     * All valid par files included with PINT can be loaded successfully
> 
>     * Parameters specified by alias work, both when they are only part of an existing component and when the aliased name is the one that makes a new/different component necessary. (This is the bug that we are waiting for this PR to fix.)
> 
>     * Parfiles that might be ambiguous about which components should be included get the correct components

We still seem to lack these, most particularly tests that demonstrate that aliases can be used interchangeably with true names in par files. This was the motivation for this rewrite, so it is important that we have tests that clearly demonstrate it.

src/pint/models/binary_ell1.py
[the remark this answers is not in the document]
aarchiba: This code, right here, will fail badly if `get_prefix_mapping_component()` returns FBn parameters with a gap, or starting from a number other than 0, or if the dictionary `.values()` appear out of order. That makes this code depend for its correct functioning on subtle checks happening elsewhere in PINT. Please make this specific function actually use the keys to identify which parameter corresponds to which term in the Taylor series for orbital frequency.

src/pint/models/timing_model.py
```diff
@@ -2639,6 +2697,269 @@ def __init__(self):
...
+    """ A class for the components pool.
+
+    This object stores and manages the instances of component classes with class
+    attribute .register = True. This includes the PINT built-in components and
+    user defined components instances without any valid parameter values (
+    parameters are initialized when a component instance gets constructed,
+    however, the parameter values are unknown to the components at the moment).
+    Thus, runing `.validate()` function will fail. This class is designed for
+    helping model building and parameter seraching, not for direct data
+    analysis.
```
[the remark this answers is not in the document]
luojing1211: The model builder creates the instances of this class. If the user wants to have their own selection of the components using this class, they have to create their own instances. Most of the users shouldn't see this class, only for the experts.

src/pint/models/timing_model.py
```diff
@@ -2639,6 +2697,269 @@ def __init__(self):
...
+        This function only returns the pint defined parameter name, not
+        including the aliases.
+        """
+        component_special_params = defaultdict(list)
+        for param, cps in self.param_component_map.items():
+            if len(cps) == 1:
+                component_special_params[cps[0]].append(param)
+        return component_special_params
+
+    def _add_alias_to_map(self, alias, param_name, alias_map):
```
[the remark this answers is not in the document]
luojing1211: It is made for packing the same code. If you look at the code of the `_param_alias_map` this function is called again and again.

src/pint/models/timing_model.py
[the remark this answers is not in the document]
aarchiba: Please answer those questions in the documentation itself.

src/pint/models/timing_model.py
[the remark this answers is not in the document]
aarchiba: Shouldn't the map be a property (maybe lazy) that is computed on the fly instead of requiring users to manually add aliases one at a time?

luojing1211:
I am working on those tests.

luojing1211:
Just realized that this PR needs more work. It fails badly in the case that "EQUAD and T2EQUAD" are present in the same parfile. The read_parfile gets super confused.

luojing1211:
@aarchiba I have more updates on this PR. However, changing F0 from floatParameter to prefixParameter breaks your parameter table. Can you take a look at the updates and we can fix the parameter table.

src/pint/models/timing_model.py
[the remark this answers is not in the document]
luojing1211: This is a private function not meant for the users. The map is a lazy property. But the property call this function internally.

aarchiba:
> Your statement is correct. In principle, we can put it together. But I think keeping the parfile value parse code in the TImingModel also makes sense so that one can put a timing model together manually and then read the parameter value. I know it creates difficulty for the new mask parameter naming. I would like to have it solved as well. I remember you proposed an idea to have maskParameter as a dictionary and I thought about it, it seems a neat idea, which can avoid this problem.

What breaks? I'd expect there to be a single entry for the prefixParameter, as there is no, and no separate entry for F0.

luojing1211:
> > @aarchiba I have more updates on this PR. However, changing F0 from floatParameter to prefixParameter breaks your parameter table. Can you take a look at the updates and we can fix the parameter table.
> 
> What breaks? I'd expect there to be a single entry for the prefixParameter, as there is now, and no separate entry for F0.

The description and unit are different between the F0 and Fn. There is a check for the entries that fail in this case.

aarchiba:
> > > @aarchiba I have more updates on this PR. However, changing F0 from floatParameter to prefixParameter breaks your parameter table. Can you take a look at the updates and we can fix the parameter table.
> > 
> > 
> > What breaks? I'd expect there to be a single entry for the prefixParameter, as there is now, and no separate entry for F0.
> 
> The description and unit are different between the F0 and Fn. There is a check for the entries that fail in this case.

I'm a little unclear on this. F0 is the 0th derivative of spin frequency, and its units are Hs s^(-0), just as Fn is the nth derivative and its units are Hz s^(-n). 

Is the problem that you would like a different description on a separate line for F0 than for Fn, n>0? If so, why make it part of the prefix parameter?

Is the problem that something crashes when you try to make it a prefix parameter? If so, what?

luojing1211:
Here is the error message when I can F0 to a prefixParameter 
```
_____________________________ test_list_parameters _____________________________

    def test_list_parameters():
>       list_parameters()

/home/runner/work/PINT/PINT/tests/test_utils.py:729: 

class_ = 'pint.models.spindown.Spindown'

    def list_parameters(class_=None):
        """List parameters understood by PINT.
    
        Parameters
        ----------
        class_: type, optional
            If provided, produce a list of parameters understood by the Component type; if None,
            return a list of parameters understood by all Components known to PINT.
    
        Returns
        -------
        list of dict
            Each entry is a dictionary describing one parameter. Dictionary values are all strings
            or lists of strings, and will include at least "name", "classes", and "description".
        """
        if class_ is not None:
            from pint.models.parameter import (
                boolParameter,
                intParameter,
                maskParameter,
                prefixParameter,
                strParameter,
            )
    
            result = []
            inst = class_()
            for p in inst.params:
                pm = getattr(inst, p)
                d = dict(
                    name=pm.name,
                    class_=f"{class_.__module__}.{class_.__name__}",
                    description=pm.description,
                )
                if pm.aliases:
                    d["aliases"] = [a for a in pm.aliases if a != pm.name]
                if pm.units:
                    d["kind"] = pm.units.to_string()
                    if not d["kind"]:
                        d["kind"] = "number"
                elif isinstance(pm, boolParameter):
                    d["kind"] = "boolean"
                elif isinstance(pm, strParameter):
                    d["kind"] = "string"
                elif isinstance(pm, intParameter):
                    d["kind"] = "integer"
                if isinstance(pm, prefixParameter):
                    d["name"] = pm.prefix + "{number}"
                    d["aliases"] = [a + "{number}" for a in pm.prefix_aliases]
                if isinstance(pm, maskParameter):
                    d["name"] = pm.origin_name + " {flag} {value}"
                    d["aliases"] = [a + " {flag} {value}" for a in pm.prefix_aliases]
                if "aliases" in d and not d["aliases"]:
                    del d["aliases"]
                result.append(d)
            return result
        else:
            import pint.models.timing_model
    
            results = {}
            ct = pint.models.timing_model.Component.component_types.copy()
            ct["TimingModel"] = pint.models.timing_model.TimingModel
            for v in ct.values():
                for d in list_parameters(v):
                    n = d["name"]
                    class_ = d.pop("class_")
                    if n not in results:
                        d["classes"] = [class_]
                        results[n] = d
                    else:
                        r = results[n].copy()
                        r.pop("classes")
                        if r != d:
>                           raise ValueError(
                                f"Parameter {d} in class {class_} does not match {results[n]}"
                            )
E                           ValueError: Parameter {'name': 'F{number}', 'description': 'Spin-frequency 1 derivative', 'kind': 'Hz / s'} in class pint.models.spindown.Spindown does not match {'name': 'F{number}', 'description': 'Spin-frequency', 'kind': 'Hz', 'classes': ['pint.models.spindown.Spindown']}
```

src/pint/models/timing_model.py
luojing1211: This is maybe changed, but it works for the current setup. We will change it when the maskParameter use the keyvalue as name

src/pint/models/timing_model.py
```diff
@@ -2639,6 +2697,269 @@ def __init__(self):
         self.phase_derivs_wrt_delay = []
 
 
+class AllComponents:
```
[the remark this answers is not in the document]
luojing1211: This is fixed.

src/pint/models/timing_model.py
```diff
@@ -2639,6 +2697,269 @@ def __init__(self):
...
+                    repeatable.append(p)
+                    repeatable.append(par._parfile_name)
+                    # also add the aliases to the repeatable param
+                    for als in par.aliases:
+                        repeatable.append(als)
+        return list(set(repeatable))
+
+    @lazyproperty
+    def category_component_map(self):
+        """Return the mapping from category to component.
```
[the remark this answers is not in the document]
luojing1211: This information is in the Return part.

src/pint/models/timing_model.py
```diff
@@ -2639,6 +2697,269 @@ def __init__(self):
...
+        """
+        category = defaultdict(list)
+        for k, cp in self.components.items():
+            cat = cp.category
+            category[cat].append(k)
+        return category
+
+    @lazyproperty
+    def component_category_map(self):
+        """Return the mapping from component to category.
```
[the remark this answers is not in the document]
luojing1211: This information is in the Return section. Should I add it to the one-line description?

src/pint/models/timing_model.py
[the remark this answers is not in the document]
luojing1211: Does it answer your question? Or what is your suggestion?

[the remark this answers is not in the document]
aarchiba: I guess I don't understand why you're requiring other code to add aliases instead of, when you need to know what aliases exist, just going and querying the Parameters. Requiring the aliases to be added makes this a stateful process and leaves one to wonder what happens if someone forgets to add their alias, or how aliases are to be removed.

src/pint/models/timing_model.py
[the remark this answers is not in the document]
aarchiba: I'm just concerned that the one-line description is a little unclear. Maybe "A dict mapping category to component name." or something similar?

src/pint/models/timing_model.py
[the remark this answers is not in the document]
aarchiba: As the previous, the one-line description could easily contain this information, making it much more useful.

luojing1211:
> > Your statement is correct. In principle, we can put it together. But I think keeping the parfile value parse code in the TImingModel also makes sense so that one can put a timing model together manually and then read the parameter value. I know it creates difficulty for the new mask parameter naming. I would like to have it solved as well. I remember you proposed an idea to have maskParameter as a dictionary and I thought about it, it seems a neat idea, which can avoid this problem.
> 
> I really cannot see how one would ever find it useful to have a manually assembled model but then read the values from a par file. If I for some reason had to do that, I would just `get_model` the parfile. How would what you get ever differ from what `get_model` ever gave you?

I believe this is addressed by the new update.

luojing1211:
> > Here is the error message when I changed F0 to a prefixParameter
> > ```
> > _____________________________ test_list_parameters _____________________________
> > 
> >     def test_list_parameters():
> > >       list_parameters()
> > 
> > /home/runner/work/PINT/PINT/tests/test_utils.py:729: 
> > _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> > 
> > class_ = 'pint.models.spindown.Spindown'
> > 
> >     def list_parameters(class_=None):
> >         """List parameters understood by PINT.
> >     
> >         Parameters
> >         ----------
> >         class_: type, optional
> >             If provided, produce a list of parameters understood by the Component type; if None,
> >             return a list of parameters understood by all Components known to PINT.
> >     
> >         Returns
> >         -------
> >         list of dict
> >             Each entry is a dictionary describing one parameter. Dictionary values are all strings
> >             or lists of strings, and will include at least "name", "classes", and "description".
> >         """
> >         if class_ is not None:
> >             from pint.models.parameter import (
> >                 boolParameter,
> >                 intParameter,
> >                 maskParameter,
> >                 prefixParameter,
> >                 strParameter,
> >             )
> >     
> >             result = []
> >             inst = class_()
> >             for p in inst.params:
> >                 pm = getattr(inst, p)
> >                 d = dict(
> >                     name=pm.name,
> >                     class_=f"{class_.__module__}.{class_.__name__}",
> >                     description=pm.description,
> >                 )
> >                 if pm.aliases:
> >                     d["aliases"] = [a for a in pm.aliases if a != pm.name]
> >                 if pm.units:
> >                     d["kind"] = pm.units.to_string()
> >                     if not d["kind"]:
> >                         d["kind"] = "number"
> >                 elif isinstance(pm, boolParameter):
> >                     d["kind"] = "boolean"
> >                 elif isinstance(pm, strParameter):
> >                     d["kind"] = "string"
> >                 elif isinstance(pm, intParameter):
> >                     d["kind"] = "integer"
> >                 if isinstance(pm, prefixParameter):
> >                     d["name"] = pm.prefix + "{number}"
> >                     d["aliases"] = [a + "{number}" for a in pm.prefix_aliases]
> >                 if isinstance(pm, maskParameter):
> >                     d["name"] = pm.origin_name + " {flag} {value}"
> >                     d["aliases"] = [a + " {flag} {value}" for a in pm.prefix_aliases]
> >                 if "aliases" in d and not d["aliases"]:
> >                     del d["aliases"]
> >                 result.append(d)
> >             return result
> >         else:
> >             import pint.models.timing_model
> >     
> >             results = {}
> >             ct = pint.models.timing_model.Component.component_types.copy()
> >             ct["TimingModel"] = pint.models.timing_model.TimingModel
> >             for v in ct.values():
> >                 for d in list_parameters(v):
> >                     n = d["name"]
> >                     class_ = d.pop("class_")
> >                     if n not in results:
> >                         d["classes"] = [class_]
> >                         results[n] = d
> >                     else:
> >                         r = results[n].copy()
> >                         r.pop("classes")
> >                         if r != d:
> > >                           raise ValueError(
> >                                 f"Parameter {d} in class {class_} does not match {results[n]}"
> >                             )
> > E                           ValueError: Parameter {'name': 'F{number}', 'description': 'Spin-frequency 1 derivative', 'kind': 'Hz / s'} in class pint.models.spindown.Spindown does not match {'name': 'F{number}', 'description': 'Spin-frequency', 'kind': 'Hz', 'classes': ['pint.models.spindown.Spindown']}
> > ```
> 
> Ah. You now have two PrefixParameters with the same prefix pattern. I did not expect this ever to occur, and it never did before. It poses an ambiguity - if you want to make F17, which one should you use to make the new parameter from?

I removed F1 as a default built-in parameter. This problem should be fixed now

src/pint/models/model_builder.py
```diff
@@ -2,317 +2,377 @@
...
+
+    The ModelBuilder class helps building a TimingModel from a parameter file
+    (i.e., pulsar ephemerise or '.par' file).
+    It first maps the provided parameter names to the PINT defined parameter
+    names, if they are in the PINT parameter aliases list. Then, the
+    ModelBuilder selects model components based on the following rules:
+        * The components in the :py:attr:`~default_components` list will be selected.
+        * When a component get mapped uniquely by the given parameters.
+        * The pulsar binary component will be selected by the 'BINARY' parameter.
+    """
```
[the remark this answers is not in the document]
luojing1211: I added doc to __call__

src/pint/models/timing_model.py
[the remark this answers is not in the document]
luojing1211: I changed the name to '_check_alias_conflict'. Please take a look.

src/pint/models/timing_model.py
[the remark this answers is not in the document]
luojing1211: Doc added

src/pint/models/timing_model.py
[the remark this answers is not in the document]
luojing1211: Doc changed.

src/pint/models/timing_model.py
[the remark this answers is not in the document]
luojing1211: Doc changed.

src/pint/models/timing_model.py
```diff
@@ -2358,27 +2387,17 @@ def param_prefixs(self):
...
-           Name of prefix.
-
-        Returns
-        -------
-        dict
-           A dictionary with prefix pararameter real index as key and parameter
-           name as value.
+    @property_exists
+    def aliases_map(self):
+        """Return all the aliases and map to the PINT parameter name.
```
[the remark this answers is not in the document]
luojing1211: Doc and tests added.

src/pint/models/parameter.py
```diff
@@ -121,6 +121,7 @@ def __init__(
     ):
 
         self.name = name  # name of the parameter
+        self._parfile_name = name
```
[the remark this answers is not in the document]
luojing1211: comments added.

PR #1030, opened 2021-04-14
Allow TOAs to be written in index order

aarchiba:
Users can now write out par files in the order specified in the `index` column; this allows the tim file to reproduce the original input order; users can also set the order they like.


aarchiba:
One could argue that the default should be to preserve order; that will make most of our users see tim files written in a different order, though, so it's potentially disruptive.

PR #1032, opened 2021-04-20
Quick jump_flags_to_params fix

champagne-cmd:
There were issues with the .tim jumps and their key_value being changed from an int type to a string type, causing `select_toa_mask` to fail. I modified the key for these jumps to be -tim_jump, similar to the modification with the -gui_jump, to avoid this conflicting conversion. I'm sorry these keep having issues @scottransom , this should take care of everything!

### DIFF
--- a/src/pint/models/jump.py
+++ b/src/pint/models/jump.py
@@ -155,13 +155,17 @@ class PhaseJump(PhaseComponent):
             mask = jump_par.select_toa_mask(toas)
             # apply to dictionaries
             for dict in toas.table["flags"][mask]:
-                if "jump" in dict.keys():
+                if "jump" in dict.keys() and type(dict["jump"]) is list:
                     # check if jump flag already added - don't add flag twice
-                    if jump_par.index in dict["jump"]:
+                    if str(jump_par.index) in dict["jump"]:
                         continue
-                    dict["jump"].append(jump_par.index)  # otherwise, add jump flag
+                    dict["jump"].append(str(jump_par.index))  # otherwise, add jump flag
+                elif "jump" in dict.keys() and type(dict["jump"]) is str:
+                    if dict["jump"] == str(jump_par.index):
+                        continue
+                    dict["jump"] = [dict["jump"], str(jump_par.index)]
                 else:
-                    dict["jump"] = [jump_par.index]
+                    dict["jump"] = [str(jump_par.index)]
 
     def add_jump_and_flags(self, toa_table):
         """Add jump object to PhaseJump and appropriate flags to TOA tables.
@@ -221,8 +225,8 @@ class PhaseJump(PhaseComponent):
         # add appropriate flags to TOA table to link jump with appropriate TOA
         for dict1 in toa_table:
             if "jump" in dict1.keys():
-                dict1["jump"].append(ind)  # toa can have multiple jumps
+                dict1["jump"].append(str(ind))  # toa can have multiple jumps
             else:
-                dict1["jump"] = [ind]
-            dict1["gui_jump"] = ind  # toa can only have one gui_jump
+                dict1["jump"] = [str(ind)]
+            dict1["gui_jump"] = str(ind)  # toa can only have one gui_jump
         return name
--- a/src/pint/models/parameter.py
+++ b/src/pint/models/parameter.py
@@ -1377,12 +1377,9 @@ class maskParameter(floatParameter):
                         "are MJD, FREQ, NAME, TEL."
                     )
         self.key = key
-        if key == "-gui_jump" or key == "-tim_jump":
-            self.key_value = key_value
-        else:
-            self.key_value = [
-                key_value_parser(k) for k in key_value
-            ]  # retains string format from .par file to ensure correct data type for comparison
+        self.key_value = [
+            key_value_parser(k) for k in key_value
+        ]  # retains string format from .par file to ensure correct data type for comparison
         self.key_value.sort()
         self.index = index
         name_param = name + str(index)
--- a/src/pint/models/parameter.py
+++ b/src/pint/models/parameter.py
@@ -1377,7 +1377,7 @@ class maskParameter(floatParameter):
                         "are MJD, FREQ, NAME, TEL."
                     )
         self.key = key
-        if key == "-gui_jump":
+        if key == "-gui_jump" or key == "-tim_jump":
             self.key_value = key_value
         else:
             self.key_value = [
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -1257,9 +1257,7 @@ class TimingModel:
                     )
                     self.add_param_from_top(param, "PhaseJump")
                     getattr(self, param.name).frozen = False
-                    flag_dict[
-                        "tim_jump"
-                    ] = num  # this is the value select_toa_mask uses
+                flag_dict["tim_jump"] = num  # this is the value select_toa_mask uses
         self.components["PhaseJump"].setup()
 
     def delete_jump_and_flags(self, toa_table, jump_num):
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -1287,9 +1287,11 @@ class TimingModel:
                         type(dict["jump"]) is str
                     ):  # if only one jump will just hold a str of the index number, otherwise will hold a list of str indeces
                         del dict["jump"]
+                    elif len(dict["jump"]) == 1:
+                        del dict["jump"]
                     else:
                         dict["jump"].remove(str(jump_num))
-                if "gui_jump" in dict.keys() and dict["gui_jump"] == jump_num:
+                if "gui_jump" in dict.keys() and dict["gui_jump"] == str(jump_num):
                     del dict["gui_jump"]
                 # renumber jump flags at higher jump indeces in whole TOA table (held as str values, must convert to ints for subtraction and then back to str)
                 if "jump" in dict.keys():
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -1257,7 +1257,9 @@ class TimingModel:
                     )
                     self.add_param_from_top(param, "PhaseJump")
                     getattr(self, param.name).frozen = False
-                flag_dict["tim_jump"] = num  # this is the value select_toa_mask uses
+                flag_dict["tim_jump"] = str(
+                    num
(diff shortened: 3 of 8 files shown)


scottransom:
Getting close, @champagne-cmd !

So the select_toa_mask() is now not failing, but it is not giving all of the TOAs!  Only one TOA for each JUMP.  See this example with the M28N files I sent before.

```
Python 3.8.6 (default, Jan 27 2021, 15:42:20) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.18.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from pint.models import get_model_and_toas                                        

In [2]: m, t = get_model_and_toas("M28N.par", "M28N.tim")                                 
INFO: Parameter PBDOT's value will be scaled by 1e-12 [pint.models.parameter]
INFO: Parameter PBDOT's value will be scaled by 1e-12 [pint.models.parameter]
/home/sransom/git/PINT/src/pint/models/timing_model.py:303: UserWarning: PINT only supports 'T2CMETHOD IAU2000B'
  warn("PINT only supports 'T2CMETHOD IAU2000B'")
INFO: Using EPHEM = DE438 from the given model [pint.toa]
INFO: Using CLOCK = BIPM from the given model [pint.toa]
INFO: Applying clock corrections (include_gps = True, include_bipm = True) [pint.toa]
INFO: Observatory gbt, loading clock file 
	/home/sransom/git/PINT/src/pint/datafiles/time.dat [pint.observatory.topo_obs]
INFO: Applying observatory clock corrections. [pint.observatory.topo_obs]
INFO: Applying GPS to UTC clock correction (~few nanoseconds) [pint.observatory.topo_obs]
INFO: Observatory gbt, loading GPS clock file 
	/home/sransom/git/PINT/src/pint/datafiles/gps2utc.clk [pint.observatory.topo_obs]
INFO: Applying TT(TAI) to TT(BIPM) clock correction (~27 us) [pint.observatory.topo_obs]
INFO: Observatory gbt, loading BIPM clock file 
	/home/sransom/git/PINT/src/pint/datafiles/tai2tt_bipm.clk [pint.observatory.topo_obs]
INFO: Observatory meerkat, loading clock file 
	/home/sransom/src/tempo2/clock/mk2utc.clk [pint.observatory.topo_obs]
INFO: Applying observatory clock corrections. [pint.observatory.topo_obs]
INFO: Applying GPS to UTC clock correction (~few nanoseconds) [pint.observatory.topo_obs]
INFO: Observatory meerkat, loading GPS clock file 
	/home/sransom/git/PINT/src/pint/datafiles/gps2utc.clk [pint.observatory.topo_obs]
INFO: Applying TT(TAI) to TT(BIPM) clock correction (~27 us) [pint.observatory.topo_obs]
INFO: Observatory meerkat, loading BIPM clock file 
	/home/sransom/git/PINT/src/pint/datafiles/tai2tt_bipm.clk [pint.observatory.topo_obs]
INFO: Using EPHEM = DE438 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE438 [pint.toa]
INFO: Set solar system ephemeris to link:
	https://data.nanograv.org/static/data/ephem/de438.bsp [pint.solar_system_ephemerides]

In [3]: m.jump_flags_to_params(t)                                                         
INFO: PhaseJump component added [pint.models.timing_model]

In [4]: x = m['JUMP1']                                                                    

In [5]: x.select_toa_mask(t)                                                              
Out[5]: array([190])

In [6]: x = m['JUMP2']                                                                    

In [7]: x.select_toa_mask(t)                                                              
Out[7]: array([194])
```

scottransom:
I'm also a bit worried about the multiple special-casing of the flags used for different types of jumps.  I've forgotten the reason why there needs to be special casing at all?  Can either @champagne-cmd or @luojing1211 remind us of that?

I get that we need a list because TOAs can be in in multiple JUMPs.  But why can't all JUMPs use a string (for instance) as the flag?  (i.e. `key_value`)

scottransom:
@champagne-cmd To add some more information to the code that I ran above, check out what is in the lines of the TOAs 190 and 191 (note that both should be part of `JUMP1`)

```
In [10]: t.table[190]['flags']                                                            
Out[10]: 
{'format': 'Princeton',
 'ddm': 0.0,
 'jump': [1],
 'clkcorr': <Quantity 2.79333711e-05 s>,
 'tim_jump': 1}

In [11]: t.table[191]['flags']                                                            
Out[11]: 
{'format': 'Princeton',
 'ddm': 0.0,
 'jump': [1],
 'clkcorr': <Quantity 2.79332671e-05 s>}
```

So only the first TOA in the JUMP is getting the "tim_jump" flag set.

As to my previous question, why is that extra flag necessary at all since there is already `'jump' : [1]` in there?

scottransom:
Yes!  That seems to work now!  I still have those questions above, though....

champagne-cmd:
Hey @scottransom I'm really sorry this has been so buggy, I pushed a fix and tried out the code you posted above and things appear to be fixed. I am working on another testing PR right now, and I will expand on the tests for this function either in that PR or this one - perhaps I should just add them to this one to increase the confidence in this code for merging.

I have been struggling with a good, consistent implementation for jumps within the GUI for a while now, and I may have been creating these special case flags based on a misunderstanding I have about jumps used outside of pintk. 

The main problem is the data type of the key_value in the jump. In general, this is stored as a string because jumps from the .par files tend to be grouped by some sort of TOA property (i.e., the jump in the J1614-2230_NANOGrav_12yv3.wb.gls.par file has a key_value of Rcvr_800). Jumps imported from the .tim files or added in the GUI do not have this, so their key_value has just been the index of the jump. 

Before I added the conditional statement in the maskParameter initialization, these values were automatically converted to strings. Now they are kept as ints for the special case jump flags (-gui_jump and -tim_jump). This was because pintk manipulates the jump entries in the toa table and reads from them as ints. I can modify this so that the jump key_values are stored in the toa table as strings and convert to int within the pintk framework for operation/manipulation (for instance, If you added JUMP1 to TOA 1 and JUMP2 to TOA 2, then went back and deleted JUMP1, pintk will go into the TOA table and decrement the value of the jump stored in the TOA table to match the jump parameter). However, I didn't do that previously because I was afraid that pintk might read in a jump with a value like Rcvr_800 and try to convert that to an int. I now see that this wouldn't be an issue - since the jump to this particular .par file has a key '-fe' instead of '-jump', pintk won't make this mistake since it looks specifically for 'jump' within the TOA table.

So, before I proceed further, I need to clarify a few things about the jumps: 
1. Do all .par file jumps utilize a key other than just "-jump"? Is the "-jump" key unique to pintk/.tim file jump use? 
2. Should the key_value of all jumps be stored as strings in the TOA table and converted to ints as needed in pintk for reading and manipulation? Or should jump parameters with the "-jump" key store their key_values as ints? This depends on the answer to 1.

scottransom:
@luojing1211 See the questions above from @champagne-cmd 

Is there any reason why all jumps couldn't be controlled by a single "-jump" flag in the TOAs table?  If the key_values were always strings, they could be coded as needed based on the JUMP type (including as string versions of numbers if we just serially add JUMPs, as in reading in old-style jumps in .tim files).

If that were the case, I think it would make things cleaner as it would mean only one type of flag (with no special casing for GUI or timfile JUMPs) and all JUMPs would use strings as the IDs for the JUMP.

luojing1211:
The problem is some of our flag keys in TOA lines are listed as 430 for the backend. Allowing Int as key flag in the JUMP, is confusing.

scottransom:
@luojing1211 I'm confused for two reasons....  First off, the 430 is a *value* for the flag/key `-fe`, not a `-jump` flag.  Second, that 430 is not treated as an integer, nor do I want *any* JUMP flags to be integers.  I'd prefer them all to be strings (and yes, the string could be an integer represented as a string).  So I'm not seeing how the 430 affects this at all.  Am I missing something?

luojing1211:
key values should already be a string in maskParameter. But before that fix, 430 caused trouble for us. It is not directly JUMP-related. I was explaining why I choose to keep the key_value as a string.

scottransom:
@luojing1211 I think we are talking around each other.  :-)   The reason I asked about this is that currently GUI-created jumps and also timfile-based jumps (i.e. what this PR is about) are now both *not* using strings for their maskParameter key_values.  They are being special-cased as ints.  I'm bringing this up to see if we think there should be special casing or if all maskParam key_values should always be strings.  (Note that I'm also specifically asking whether all types of JUMPs should also just use the regular `-jump` flag with string keys rather than using special flags and keys).  Maybe we can talk about this on the telecon.

luojing1211:
Let us talk about this in the telecon.

scottransom:
OK, based on the conversation today, I want to make sure we are all on the same sheet of music for these `tim` and `gui` jumps for @champagne-cmd 's PR.

1.  In the short term, I'm fine if the flags that we use on the TOAs for `tim` and `gui` jumps are different from `-jump` since regular parfile-based JUMPs don't have a `-jump` flag.  However, the main thing is that the `key_value` used by those flags should always be a string.
2.  For timfile-based jumps, currently in toa.py (lines 774-776), a `-jump` flag is created and it is assigned a *list* and not a string.  I don't actually think that the list is necessary here as it is (I believe) impossible to assign TOAs to multiple JUMPs from timfiles (since paired JUMP flags define the JUMP).  So, if we want a special flag, I think we should change that flag to be `-tim_jump` and the key_value to be some unique string.
3.  For gui-based and parfile-based JUMPs, it *is* possible to have TOAs in multiple JUMPs.  So, once again, if we want a unique flag for GUI JUMPs, maybe we set that to "-gui_jump" and again, just assign a unique string key_value.
4. There is still the question as to what we should use for the key_values since an int-as-string can be deleted making the model appear to have "missing" JUMPs  (i.e.  JUMP1, JUMP3, JUMP4 but no JUMP2).  Maybe we use a unique 3-letter random string or something?  I think we might have to check for potential clashes before assigning a new one.

Anyway, hopefully this clears things up.

luojing1211:
If we don't associate the JUMP index with the gui-jump key value. we can break the confusion as you mentioned. How about we reserve some string for this special case, something like "GUI_1", "GUI_2". 

Or change maskParameter to multi-value parameter, which needs a bigger change, but the index problem is totally avoided.

champagne-cmd:
Modified `key_value` for all jumps to be a string and updated jump test accordingly. This iteration still has the JUMP index as the `key_value`, just in string form. Now looking into implementing unique string value separate from JUMP index.

PR #1033, opened 2021-04-26
Fix Noise design matrix index in the fitter

luojing1211:
The fitter gets the noise design matrix from the stack design matrix using the indices. The stack design matrix's index is labeled as 
+++++++++++++++++++++++++++++++++++++++++
           |    offset | fit parameters | Noise design matrix
+++++++++++++++++++++++++++++++++++++++++
TOAs |

In the red noise residual calculation, the offset column is not taken into consideration. The start index of the noise design matrix is 1 smaller than the real start position. This PR fixes it.

### DIFF
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -220,7 +220,7 @@ class Fitter:
         self.is_wideband = False
         self.converged = False
 
-    def fit_toas(self, maxiter=None):
+    def fit_toas(self, maxiter=None, debug=False):
         """Run fitting operation.
 
         This method needs to be implemented by subclasses. All implemenations
@@ -1018,6 +1018,7 @@ class DownhillFitter(Fitter):
         required_chi2_decrease=1e-2,
         max_chi2_increase=1e-2,
         min_lambda=1e-3,
+        debug=False
     ):
         """Carry out a cautious downhill fit.
 
@@ -1246,7 +1247,7 @@ class DownhillWLSFitter(DownhillFitter):
         )
         self.method = "downhill_wls"
 
-    def fit_toas(self, maxiter=10, threshold=None, **kwargs):
+    def fit_toas(self, maxiter=10, threshold=None, debug=False, **kwargs):
         """Fit TOAs.
 
         This is mostly implemented in
@@ -1264,7 +1265,7 @@ class DownhillWLSFitter(DownhillFitter):
             :func:`pint.fitter.DownhillFitter.fit_toas`
         """
         self.threshold = threshold
-        super().fit_toas(maxiter=maxiter, **kwargs)
+        super().fit_toas(maxiter=maxiter, debug=debug, **kwargs)
 
     def create_state(self):
         return WLSState(self, self.model)
@@ -1402,7 +1403,7 @@ class DownhillGLSFitter(DownhillFitter):
             self, self.model, full_cov=self.full_cov, threshold=self.threshold
         )
 
-    def fit_toas(self, maxiter=10, threshold=0, full_cov=False, **kwargs):
+    def fit_toas(self, maxiter=10, threshold=0, full_cov=False, debug=False, **kwargs):
         """Fit TOAs.
 
         This is mostly implemented in
@@ -1424,7 +1425,7 @@ class DownhillGLSFitter(DownhillFitter):
         """
         self.threshold = threshold
         self.full_cov = full_cov
-        r = super().fit_toas(maxiter=maxiter, **kwargs)
+        r = super().fit_toas(maxiter=maxiter, debug=debug, **kwargs)
         # FIXME: set up noise residuals et cetera
         # Compute the noise realizations if possible
         ntmpar = len(self.model.free_params)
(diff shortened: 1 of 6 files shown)


paulray:
This is one of the many reasons why the offset parameter should be *explicit* in the model and not implicit.

luojing1211:
I agree. But we have not decided on how to handle the absolute phase fitting.

tests/test_gls_fitter.py
```diff
@@ -60,6 +60,20 @@ def test_gls_fitter(self):
                     assert np.abs(v - val[0]) <= e, msg
                     assert np.abs(1 - val[1] / e) < 0.1, msg
 
+    def test_noise_design_matrix_index(self):
+        self.fit(full_cov=False)
```
aarchiba, line 64: Why is it necessary to do a fit here? That will be slow.

tests/test_gls_fitter.py
```diff
@@ -60,6 +60,20 @@ def test_gls_fitter(self):
...
+        ntmpar = len(self.f.model.free_params)
+        Mn = self.f.model.noise_model_designmatrix(self.f.toas)
+        M, params, units = self.f.get_designmatrix()
+        M_stack = np.hstack((M, Mn))
+        noise_dims = self.f.model.noise_model_dimensions(self.f.toas)
+        # This is how the gls fitter get the noise design matrix. I don't know
+        # if there is some other way to test the index.
```
aarchiba, line 71: For noise model basis functions that are easy to describe - like the constant offset one, or ECORR for a simple data set (merge two fake data sets so the ECORR basis functions are pairs) - you can confirm that the basis functions are what you think they should be. I wrote some examples of this in the tests for the downhill fitters, if I recall correctly.

tests/test_gls_fitter.py
```diff
@@ -60,6 +60,20 @@ def test_gls_fitter(self):
...
+        for comp in noise_dims.keys():
+            p0 = noise_dims[comp][0] + ntmpar + 1
+            p1 = p0 + noise_dims[comp][1]
+            assert np.all(M_stack[:, p0:p1] == Mn[:, p0 - ntmpar - 1 : p1 - ntmpar - 1])
```
aarchiba, line 75: I don't understand what relationship this is supposed to be testing.

tests/test_model_manual.py
```diff
@@ -89,7 +89,7 @@ def test_valid_model(tmp_dir, func, name, expectation):
 def test_compare_get_model_new_and_old():
     m_new = get_model_new(parfile)
     m_old = get_model(parfile)
-
+    print(parfile)
```
aarchiba, line 92: Do not print. You can always run `pytest --pdb` and inspect local variables in a failing test.

tests/test_gls_fitter.py
```diff
@@ -60,6 +60,20 @@ def test_gls_fitter(self):
                     assert np.abs(v - val[0]) <= e, msg
                     assert np.abs(1 - val[1] / e) < 0.1, msg
 
+    def test_noise_design_matrix_index(self):
```
aarchiba, line 63: Does this test fail when run against the code before the bug fix? I can't really tell what it is testing or whether it actually verifies correctness.

aarchiba, line 63: Did you also test the downhill GLS and the wideband (downhill and not) fitters for the same bug?

src/pint/fitter.py
```diff
@@ -1432,6 +1432,8 @@ def fit_toas(self, maxiter=10, threshold=0, full_cov=False, **kwargs):
             noise_dims = self.model.noise_model_dimensions(self.toas)
             noise_resids = {}
             for comp in noise_dims.keys():
+                # The first column of designmatrix is "offset", add 1 to match
+                # the indices of noise designmatrix
                 p0 = noise_dims[comp][0] + ntmpar
```
aarchiba, line 1437: Should this line see the same change as the others?

src/pint/fitter.py
```diff
@@ -1432,6 +1432,8 @@ def fit_toas(self, maxiter=10, threshold=0, full_cov=False, **kwargs):
             noise_dims = self.model.noise_model_dimensions(self.toas)
             noise_resids = {}
             for comp in noise_dims.keys():
```
aarchiba, line 1434: Just use `for comp in noise_dims:` if you're changing the code anyway.

luojing1211:
Our current test does not complain about the mismatch of indices. The only place of the usage is computing the red noise residual, shifting one index does not change the result dramatically. The test I put there is to use the same code from the fitter to produce the requested indices which are used to call the full design matrix. Then I compare if the sliced matrix the same as the noise design matrix.

tests/test_gls_fitter.py
luojing1211: The problem is to get the noise model design matrix values from the full design matrix (Timingmodel design matrix + noise model design matrix). The noise dim calculation is ok. The problem is at the fitter. I don't have a good way to get P0, and P1, listed below from the fitter.

aarchiba:
It is fairly easy to check red noise residuals if the basis functions are the top-hats you get from ECORR. This would explain some of the problems we've been having with whitened residuals. I wrote some relevant tests in #1000

tests/test_model_manual.py
scottransom: I don't understand the "Do not print."  Why is that?  It gets suppressed during normal usage, but you can see if if you want by using `pytest -s`.  It seems (and has been) useful to me.

PR #1039, opened 2021-04-28
Set minimum window size for pintk

### LINKED ISSUE #1036: pintk window initial size
On Macs, when you run `pintk` the initial GUI window size is tiny and needs to be expanded manually to even see the GUI buttons or anything.  It would be great to fix this.

paulray:
This sets a minimum size for a pintk window. This fixes #1036 at least on Mac.
Could people try on other platforms?

### DIFF
--- a/src/pint/scripts/pintk.py
+++ b/src/pint/scripts/pintk.py
@@ -160,10 +160,11 @@ def main(argv=None):
     args = parser.parse_args(argv)
 
     root = tk.Tk()
+    root.minsize(800, 600)
     if not args.test:
         app = PINTk(root, parfile=args.parfile, timfile=args.timfile, ephem=args.ephem)
         root.protocol("WM_DELETE_WINDOW", root.destroy)
-        root.mainloop()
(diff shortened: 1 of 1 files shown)


src/pint/scripts/pintk.py
```diff
@@ -160,10 +160,11 @@ def main(argv=None):
     args = parser.parse_args(argv)
 
     root = tk.Tk()
+    root.minsize(800, 600)
     if not args.test:
         app = PINTk(root, parfile=args.parfile, timfile=args.timfile, ephem=args.ephem)
         root.protocol("WM_DELETE_WINDOW", root.destroy)
-        root.mainloop()
+        tk.mainloop()
```
scottransom, line 167: Any idea why this needed to change from `root` to `tk`?  Seems strange since the other change just modified `root`....

paulray: I don't think it is needed, but the examples I saw all did it that way, so I changed it. Probably it is synonymous.

PR #1041, opened 2021-05-05
Allow a whole lot of X-ray missions

matteobachetti:
By the way, let us make PINT understand data from many more missions.

This should work with all currently supported missions, plus many more. And if a mission is not known, it just calls it "generic" and tries to read the data from the first FITS extension (of course, data have to be barycentered, orbit files would be trickier!).

- [x] Basic implementation
- [x] Make it work without HEADAS (minor fix)
- [x] Make missions in xselect.mdb non-generic (this has little use for now, but it can be important if we decide to filter by PI or perform some other data selection)

### DIFF
--- a/src/pint/event_toas.py
+++ b/src/pint/event_toas.py
@@ -1,4 +1,5 @@
 """Generic function to load TOAs from events files."""
+import os
 import astropy.io.fits as pyfits
 import numpy as np
 from astropy import log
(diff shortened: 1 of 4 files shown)


paulray:
Cool! I didn't know about xselect.mdb.  Will this now fail if someone doesn't have HEADAS set, however?

matteobachetti:
Yes, but this is a bug that I will fix shortly!

matteobachetti:
@paulray ready for review

PR #1042, opened 2021-05-08
Fix ddk 1/x

luojing1211:
This PR fixes the DDK 1/x error. Make stand-alone binary model update correct.

### DIFF
--- a/src/pint/models/binary_ddk.py
+++ b/src/pint/models/binary_ddk.py
@@ -102,3 +102,5 @@ class BinaryDDK(BinaryDD):
                 "DDK model does not accept `SINI` as input. Please"
                 " use `KIN` instead."
             )
+        if self.PX.value == 0.0 or self.PX.value is None:
+            raise ValueError("DDK model needs a valid `PX`. ")
--- a/src/pint/models/binary_ddk.py
+++ b/src/pint/models/binary_ddk.py
@@ -102,5 +102,10 @@ class BinaryDDK(BinaryDD):
                 "DDK model does not accept `SINI` as input. Please"
                 " use `KIN` instead."
             )
-        if self.PX.value == 0.0 or self.PX.value is None:
-            raise ValueError("DDK model needs a valid `PX`. ")
+
+        if hasattr(self._parent, 'PX'):
+            if self._parent.PX.value == 0.0 or self._parent.PX.value is None:
+                raise ValueError("DDK model needs a valid `PX`. ")
+        else:
+            raise MissingParameter('Binary_DDK', 'PX', 'DDK model needs PX from'
+                                   'Astrometry.')
--- a/src/pint/models/binary_ddk.py
+++ b/src/pint/models/binary_ddk.py
@@ -103,9 +103,10 @@ class BinaryDDK(BinaryDD):
                 " use `KIN` instead."
             )
 
-        if hasattr(self._parent, 'PX'):
+        if hasattr(self._parent, "PX"):
             if self._parent.PX.value == 0.0 or self._parent.PX.value is None:
                 raise ValueError("DDK model needs a valid `PX`. ")
         else:
-            raise MissingParameter('Binary_DDK', 'PX', 'DDK model needs PX from'
-                                   'Astrometry.')
+            raise MissingParameter(
+                "Binary_DDK", "PX", "DDK model needs PX from" "Astrometry."
+            )
--- a/src/pint/models/binary_ddk.py
+++ b/src/pint/models/binary_ddk.py
@@ -4,7 +4,7 @@ from astropy import log
 from pint.models.binary_dd import BinaryDD
 from pint.models.parameter import boolParameter, floatParameter
 from pint.models.stand_alone_psr_binaries.DDK_model import DDKmodel
-from pint.models.timing_model import MissingParameter
+from pint.models.timing_model import MissingParameter, TimingModelError
 
 
 class BinaryDDK(BinaryDD):
@@ -98,7 +98,7 @@ class BinaryDDK(BinaryDD):
                     "DDK", "DDK model needs proper motion parameters."
                 )
         if self.SINI.quantity is not None:
-            raise ValueError(
+            raise TimingModelError(
                 "DDK model does not accept `SINI` as input. Please"
                 " use `KIN` instead."
             )
--- a/src/pint/models/binary_ddk.py
+++ b/src/pint/models/binary_ddk.py
@@ -104,8 +104,8 @@ class BinaryDDK(BinaryDD):
             )
 
         if hasattr(self._parent, "PX"):
-            if self._parent.PX.value == 0.0 or self._parent.PX.value is None:
-                raise ValueError("DDK model needs a valid `PX`. ")
+            if self._parent.PX.value <= 0.0 or self._parent.PX.value is None:
+                raise ValueError("DDK model needs a valid `PX` value.")
         else:
             raise MissingParameter(
                 "Binary_DDK", "PX", "DDK model needs PX from" "Astrometry."
--- a/src/pint/models/binary_ell1.py
+++ b/src/pint/models/binary_ell1.py
@@ -262,16 +262,6 @@ class BinaryELL1H(BinaryELL1Base):
     def setup(self):
         """Parameter setup."""
         super(BinaryELL1H, self).setup()
-
-    def validate(self):
-        """Parameter validation."""
-        super(BinaryELL1H, self).validate()
-        # if self.H3.quantity is None:
-        #     raise MissingParameter("ELL1H", "H3", "'H3' is required for ELL1H model")
-        if self.SINI.quantity is not None:
-            warn("'SINI' will not be used in ELL1H model. ")
-        if self.M2.quantity is not None:
-            warn("'M2' will not be used in ELL1H model. ")
         if self.H4.quantity is not None:
             self.binary_instance.fit_params = ["H3", "H4"]
             # If have H4 or STIGMA, choose 7th order harmonics
@@ -283,3 +273,16 @@ class BinaryELL1H(BinaryELL1Base):
             self.binary_instance.ds_func = self.binary_instance.delayS_H3_STIGMA_exact
             if self.STIGMA.quantity <= 0:
                 raise ValueError("STIGMA must be greater than zero.")
+
+        self.update_binary_object()
+
+
+    def validate(self):
+        """Parameter validation."""
+        super(BinaryELL1H, self).validate()
+        # if self.H3.quantity is None:
+        #     raise MissingParameter("ELL1H", "H3", "'H3' is required for ELL1H model")
+        if self.SINI.quantity is not None:
+            warn("'SINI' will not be used in ELL1H model. ")
+        if self.M2.quantity is not None:
+            warn("'M2' will not be used in ELL1H model. ")
--- a/src/pint/models/binary_ell1.py
+++ b/src/pint/models/binary_ell1.py
@@ -9,7 +9,7 @@ from pint.models.pulsar_binary import PulsarBinary
 from pint.models.stand_alone_psr_binaries import binary_orbits as bo
 from pint.models.stand_alone_psr_binaries.ELL1_model import ELL1model
 from pint.models.stand_alone_psr_binaries.ELL1H_model import ELL1Hmodel
-from pint.models.timing_model import MissingParameter
+from pint.models.timing_model import MissingParameter, TimingModelError
 from pint.utils import taylor_horner_deriv
 
 
@@ -32,7 +32,7 @@ class BinaryELL1Base(PulsarBinary):
     """
 
     def __init__(self):
-        super(BinaryELL1Base, self).__init__()
+        super().__init__()
         self.add_param(
             MJDParameter(
                 name="TASC", description="Epoch of ascending node", time_scale="tdb"
@@ -78,11 +78,11 @@ class BinaryELL1Base(PulsarBinary):
         self.warn_default_params = []
 
     def setup(self):
-        super(BinaryELL1Base, self).setup()
+        super().setup()
 
     def validate(self):
         """Validate parameters"""
-        super(BinaryELL1Base, self).validate()
+        super().validate()
 
         for p in ["EPS1", "EPS2"]:
             if getattr(self, p).value is None:
@@ -190,7 +190,7 @@ class BinaryELL1(BinaryELL1Base):
     register = True
 
     def __init__(self):
-        super(BinaryELL1, self).__init__()
+        super().__init__()
         self.binary_model_name = "ELL1"
         self.binary_model_class = ELL1model
 
@@ -261,7 +261,7 @@ class BinaryELL1H(BinaryELL1Base):
 
     def setup(self):
         """Parameter setup."""
-        super(BinaryELL1H, self).setup()
+        super().setup()
         if self.H4.quantity is not None:
             self.binary_instance.fit_params = ["H3", "H4"]
             # If have H4 or STIGMA, choose 7th order harmonics
@@ -277,10 +277,10 @@ class BinaryELL1H(BinaryELL1Base):
 
     def validate(self):
         """Parameter validation."""
-        super(BinaryELL1H, self).validate()
+        super().validate()
         # if self.H3.quantity is None:
         #     raise MissingParameter("ELL1H", "H3", "'H3' is required for ELL1H model")
         if self.SINI.quantity is not None:
-            raise ValueError("'SINI' will not be used in ELL1H model. ")
+            raise TimingModelError("'SINI' will not be used in ELL1H model. ")
         if self.M2.quantity is not None:
-            raise ValueError("'M2' will not be used in ELL1H model. ")
+            raise TimingModelError("'M2' will not be used in ELL1H model. ")
--- a/src/pint/models/binary_ell1.py
+++ b/src/pint/models/binary_ell1.py
@@ -273,9 +273,7 @@ class BinaryELL1H(BinaryELL1Base):
             self.binary_instance.ds_func = self.binary_instance.delayS_H3_STIGMA_exact
             if self.STIGMA.quantity <= 0:
                 raise ValueError("STIGMA must be greater than zero.")
-
         self.update_binary_object()
-
 
     def validate(self):
         """Parameter validation."""
--- a/src/pint/models/binary_ell1.py
+++ b/src/pint/models/binary_ell1.py
@@ -273,7 +273,7 @@ class BinaryELL1H(BinaryELL1Base):
             self.binary_instance.ds_func = self.binary_instance.delayS_H3_STIGMA_exact
             if self.STIGMA.quantity <= 0:
                 raise ValueError("STIGMA must be greater than zero.")
-        self.update_binary_object()
+        self.update_binary_object(None)
 
     def validate(self):
         """Parameter validation."""
--- a/src/pint/models/binary_ell1.py
+++ b/src/pint/models/binary_ell1.py
@@ -281,6 +281,6 @@ class BinaryELL1H(BinaryELL1Base):
         # if self.H3.quantity is None:
         #     raise MissingParameter("ELL1H", "H3", "'H3' is required for ELL1H model")
         if self.SINI.quantity is not None:
-            warn("'SINI' will not be used in ELL1H model. ")
+            raise ValueError("'SINI' will not be used in ELL1H model. ")
         if self.M2.quantity is not None:
-            warn("'M2' will not be used in ELL1H model. ")
+            raise ValueError("'M2' will not be used in ELL1H model. ")
--- a/src/pint/models/pulsar_binary.py
+++ b/src/pint/models/pulsar_binary.py
@@ -213,9 +213,8 @@ class PulsarBinary(DelayComponent):
                 aliase = []
 
             # the _parent attribute should give access to all the components
-            if (
-                hasattr(self._parent, par)
-                or set(aliase).intersection(self._parent.params)
+            if hasattr(self._parent, par) or set(aliase).intersection(
+                self._parent.params
             ):
                 try:
                     pint_bin_name = self._parent.match_param_aliases(par)
--- a/src/pint/models/pulsar_binary.py
+++ b/src/pint/models/pulsar_binary.py
@@ -195,13 +195,13 @@ class PulsarBinary(DelayComponent):
 
         Parameters
         ----------
-        toas : `pint.toa.TOAs` object, optional
+        toas: `pint.toa.TOAs` object, optional
             The TOAs that need to pass to the stand alone model.Default value is
             None. If toas is None, this function only updates the parameter value.
             If 'acc_delay' is not provided, the stand alone binary receives the
             standard barycentered TOAs.
 
-        acc_delay: `Numpy.ndarray`
+        acc_delay: `numpy.ndarray`
             If provided, TOAs will be corrected by provided acc_delay instead of
             the standard barycentering. The stand alone binary receives the
             input TOAs - acc_delay.
--- a/src/pint/models/pulsar_binary.py
+++ b/src/pint/models/pulsar_binary.py
@@ -157,7 +157,7 @@ class PulsarBinary(DelayComponent):
             )
         # Update the parameters in the stand alone binary
         self.update_binary_object()
-        
+
     def validate(self):
         super(PulsarBinary, self).validate()
 
@@ -213,7 +213,10 @@ class PulsarBinary(DelayComponent):
                 aliase = []
 
             # the _parent attribute should give access to all the components
-            if hasattr(self._parent, par) or list(set(aliase).intersection(self._parent.params)) != []:
+            if (
+                hasattr(self._parent, par)
+                or list(set(aliase).intersection(self._parent.params)) != []
+            ):
                 try:
                     pint_bin_name = self._parent.match_param_aliases(par)
                 except ValueError:
--- a/src/pint/models/pulsar_binary.py
+++ b/src/pint/models/pulsar_binary.py
@@ -215,7 +215,7 @@ class PulsarBinary(DelayComponent):
             # the _parent attribute should give access to all the components
             if (
                 hasattr(self._parent, par)
-                or list(set(aliase).intersection(self._parent.params)) != []
+                or set(aliase).intersection(self._parent.params)
             ):
                 try:
                     pint_bin_name = self._parent.match_param_aliases(par)
--- a/src/pint/models/pulsar_binary.py
+++ b/src/pint/models/pulsar_binary.py
@@ -206,11 +206,13 @@ class PulsarBinary(DelayComponent):
             the standard barycentering. The stand alone binary receives the
             input TOAs - acc_delay.
 
-        Note
-        ----
+        Warns
+        -----
         If the 'toas' are not provided, the stand alone binary model will use
-        the TOAs were passed to it from last interation or no TOAs for stand
-        alone binary model at all.
+        the TOAs were passed to it from last interation (i.e. last barycnetered
+        TOAs) or no TOAs for stand alone binary model at all. This behavior will
+        cause incorrect answers. Allowing optional 'toas' input is for some
+        lower level tests. We do not recommend PINT user to use it.
         """
         # Don't need to fill P0 and P1. Translate all the others to the format
         # that is used in bmodel.py
--- a/src/pint/models/pulsar_binary.py
+++ b/src/pint/models/pulsar_binary.py
@@ -156,7 +156,7 @@ class PulsarBinary(DelayComponent):
                 self.binary_instance, list(FBXs.keys())
             )
         # Update the parameters in the stand alone binary
-        self.update_binary_object()
+        self.update_binary_object(None)
 
     def validate(self):
         super(PulsarBinary, self).validate()
@@ -186,7 +186,7 @@ class PulsarBinary(DelayComponent):
                 continue
             bparObj.value = bparObj.value * u.Unit(bparObj.units)
 
-    def update_binary_object(self, toas=None, acc_delay=None):
+    def update_binary_object(self, toas, acc_delay=None):
         """Update stand alone binary's parameters and toas from PINT-facing object.
 
         This function passes the PINT-facing object's parameter values and TOAs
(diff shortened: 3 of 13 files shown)


luojing1211:
@aarchiba, could you give a look about this PR?

tests/test_ddk.py
```diff
@@ -105,26 +123,40 @@ def test_sini_from_value(self):
...
+                    assert pint_par.value == standalone_par.value
+                else:
+                    assert pint_par.value == standalone_par
+
+    def test_zero_PX(self):
+        zero_px_str = self.temp_par_str.replace("PX  0.8211", "PX  0.0")
         with pytest.raises(ValueError):
-            mb.get_model(StringIO(temp_par_str))
+            m = mb.get_model(StringIO(zero_px_str))
+        test_par_str = self.temp_par_str + "\n KIN  71.969  1  0.562"
```
aarchiba, line 155: Should this be a separate test?

tests/test_ell1h.py
```diff
@@ -143,6 +144,17 @@ def test_J0613_STIG(toasJ0613, modelJ0613_STIG):
     f.fit_toas()
 
 
+def test_SINI_M2():
+    """Test SINI and M2 error.
+    """
+    SINI_par = simple_par.replace("H3 2.7507208E-7", "SINI 0.8")
+    with pytest.warns(None):
+        _ = model.get_model(StringIO(SINI_par))
```
aarchiba, line 152: Why assign to `_`? just don't assign anything.

tests/test_ell1h.py
```diff
@@ -143,6 +144,17 @@ def test_J0613_STIG(toasJ0613, modelJ0613_STIG):
     f.fit_toas()
 
 
+def test_SINI_M2():
+    """Test SINI and M2 error.
+    """
+    SINI_par = simple_par.replace("H3 2.7507208E-7", "SINI 0.8")
+    with pytest.warns(None):
```
aarchiba, line 151: What warning should be emitted? Please match on the message, not just to make the test better but also so I know what is being tested for.

tests/test_ell1h.py
```diff
@@ -143,6 +144,17 @@ def test_J0613_STIG(toasJ0613, modelJ0613_STIG):
     f.fit_toas()
 
 
+def test_SINI_M2():
```
aarchiba, line 147: Is this testing what happens when only one of SINI/M2 is present and not the other? Shoudl that be two separate tests?

tests/test_ell1h.py
```diff
@@ -216,3 +228,25 @@ def test_zero_H3_H4_fit_H3(toasJ0613):
     # This should work
     f.fit_toas()
     assert f.model.H3.value > 0.0
+
+
+@pytest.mark.parametrize("model", ["modelJ1853", "modelJ0613"])
+def test_stand_alone_model_params(model, request):
```
aarchiba, line 234: What property are you actually checking here? Looks like maybe the same `verify_` something function would cover most of this?

src/pint/models/binary_ddk.py
```diff
@@ -102,3 +102,11 @@ def validate(self):
                 "DDK model does not accept `SINI` as input. Please"
                 " use `KIN` instead."
             )
+
+        if hasattr(self._parent, "PX"):
```
aarchiba, line 106: I can't see where `PX` is actually used? Do you have a test that verifies whether it is - for example changing PX and confirming that the residuals from barycentered TOAs change?

luojing1211: Here is the usage of DDK

https://github.com/nanograv/PINT/blob/85febde32ded496950797e04ffcfdc81c1923390/src/pint/models/stand_alone_psr_binaries/DDK_model.py#L325-L343

aarchiba: Do I understand correctly that this change is supposed to pull the PX value from the parent of the PINT-facing object to the internal standalone binary model? Does that actually happen? Do you test that anywhere?

luojing1211: Yes. That is the "test_stand_alone_model_params" is doing

tests/test_ddk.py
```diff
@@ -105,26 +123,40 @@ def test_sini_from_value(self):
...
-               OM 176.19845  1  0.0013
-               M2  0.283395  1  0.0104
-               KOM   83.100  1  1.800
-               SINI  0.8     1  0.562
-               K96  1"""
+        test_par_str = self.temp_par_str + "\n SINI  0.8     1  0.562"
+        with pytest.raises(ValueError):
+            mb.get_model(StringIO(test_par_str))
+
+    def test_stand_alone_model_params(self):
```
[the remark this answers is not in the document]
luojing1211: It is more like verify_stand_alone_model_parameters_are_updated_from_pint_facing_object. I think this can be a generic function. Should I make a test_utils.py file to host this type of function?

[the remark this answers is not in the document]
aarchiba: If the tests that use such a function all live in the same file, support functions can just live in that file. Support functions that are needed more generally, I'm not sure what the best approach is - if the support functions are potentially useful for PINT users (rather than just PINT developers) they could just go in `pint.utils`; if they're really only useful in the context of PINT's test suite, there's some standard way to do that - maybe just a module in `tests/` that doesn't start with `test_` so it doesn't get picked up as containing test cases? Other test files could just import it then.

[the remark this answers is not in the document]
luojing1211: I made a tests/utils.py for hosting this type of function.

luojing1211:
@aarchiba are the current updates ok?

tests/test_dd.py
```diff
@@ -44,6 +45,15 @@ def test_B1855(self):
             np.abs(pint_resids_us.value - self.ltres) < 1e-7
         ), "DD B1855 TEST FAILED"
 
+    def test_stand_alone_model_params_updates(self):
```
aarchiba, line 48: Should this just iterate over `*.par`? That is, any reason not to test that all models correctly import all parameters? One would just trivially report success for non-binary models.

tests/test_ell1h.py
```diff
@@ -143,6 +145,20 @@ def test_J0613_STIG(toasJ0613, modelJ0613_STIG):
     f.fit_toas()
 
 
+def test_SINI_error():
```
aarchiba, line 148: No error is produced, and this test does not test for one. It's not very clear what is supposed to be signalled - please test on the content of the warning, both so the test is more stringent and so readers can see what is being tested.

tests/utils.py
```diff
@@ -0,0 +1,40 @@
...
+import warnings
+
+
+def verify_stand_alone_binary_parameter_updates(m):
+    """Verify if the stand alone binary parameters get updated.
+
+    m: `~PINT.TimingModel` object.
+        The timing model with binary
+    """
+    if not hasattr(m, "binary_instance"):
```
aarchiba, line 16: Should this return when this occurs?

tests/utils.py
```diff
@@ -0,0 +1,40 @@
...
+import logging
+import os
+from io import StringIO
+import unittest
+import pytest
+import warnings
+
+
+def verify_stand_alone_binary_parameter_updates(m):
+    """Verify if the stand alone binary parameters get updated.
```
aarchiba, line 11: What does "updated" mean? What exactly does this check? Does it check that all binary parameters that are in the par file make it into the internal model with the same values? Does it check whether any spurious ones appear? If this routine succeeds, what does one know?

luojing1211:
@aarchiba Is PR ok now?

tests/test_dd.py
aarchiba: Is this needed any more now that all par files are tested?

tests/test_ddk.py
```diff
@@ -105,26 +124,26 @@ def test_sini_from_value(self):
...
+
+    def test_stand_alone_model_params_updates(self):
+        test_par_str = self.temp_par_str + "\n KIN  71.969  1  0.562"
+        m = mb.get_model(StringIO(test_par_str))
+        verify_stand_alone_binary_parameter_updates(m)
+
+    def test_zero_PX(self):
+        zero_px_str = self.temp_par_str.replace("PX  0.8211", "PX  0.0")
+        with pytest.raises(ValueError):
+            m = mb.get_model(StringIO(zero_px_str))
```
aarchiba, line 139: `m` is never used

tests/test_ell1h.py
```diff
@@ -143,6 +145,28 @@ def test_J0613_STIG(toasJ0613, modelJ0613_STIG):
     f.fit_toas()
 
 
+def test_SINI_warning():
+    """Test SINI and M2 error.
+    """
+    SINI_par = simple_par.replace("H3 2.7507208E-7", "SINI 0.8")
+    with pytest.warns(None) as record:
```
aarchiba, line 152: You can include the message check in the `pytest.warns`.

tests/test_ddk.py
```diff
@@ -105,26 +124,26 @@ def test_sini_from_value(self):
...
-               A1 32.34  1  0.001
-               E  0.074  1  0.001
-               T0 55388.836  1  0.0002
-               PB 67.825129  1  0.0001
-               OM 176.19845  1  0.0013
-               M2  0.283395  1  0.0104
-               KOM   83.100  1  1.800
-               SINI  0.8     1  0.562
-               K96  1"""
+        test_par_str = self.temp_par_str + "\n SINI  0.8     1  0.562"
```
aarchiba, line 127: Why is this supposed to raise a ValueError? Can you enhance the exception check so that the test is clear?

tests/test_ell1h.py
```diff
@@ -216,3 +240,9 @@ def test_zero_H3_H4_fit_H3(toasJ0613):
     # This should work
     f.fit_toas()
     assert f.model.H3.value > 0.0
+
+
+@pytest.mark.parametrize("model", ["modelJ1853", "modelJ0613"])
```
aarchiba, line 245: Is this needed now that all par files are tested?

tests/utils.py
```diff
@@ -0,0 +1,51 @@
...
+import logging
+import os
+from io import StringIO
+import unittest
+import pytest
+import warnings
+
+
+def verify_stand_alone_binary_parameter_updates(m):
+    """Verify if the stand alone binary parameter values gets updated by the
```
aarchiba, line 11: This is not the format for docstrings.

tests/utils.py
```diff
@@ -0,0 +1,51 @@
...
+                "Timing model needs to have binary components"
+                " to test the stand alone binary parameters."
+                " update"
+            )
+        )
+        return
+    for binary_par in m.binary_instance.binary_params:
+        standalone_par = getattr(m.binary_instance, binary_par)
+        try:
+            pint_par_name = m.match_param_aliases(binary_par)
```
aarchiba, line 37: Are standalone binary models allowed to use aliases of PINT parameters? Why?

aarchiba: Does this do the right thing if `match_par_aliases` rerturns None?

tests/utils.py
```diff
@@ -0,0 +1,51 @@
...
+    if not hasattr(m, "binary_instance"):
+        warnings.warn(
+            UserWarning(
+                "Timing model needs to have binary components"
+                " to test the stand alone binary parameters."
+                " update"
+            )
+        )
+        return
+    for binary_par in m.binary_instance.binary_params:
```
aarchiba, line 34: If I understand correctly, every parameter on the standalone object should have a corresponding parameter (of the same name?) on the PINT-facing object (unless it's marked internal somehow)? Will this test fail if that condition is not met? Assuming the PINT-facing version exists, then this checks that if the PINT-facing version is set, the value should agree between the PINT-facing version and the standalone version. How are units handled? Do standalone parameters have units?

aarchiba:
Please add a test that changing PX (in the PINT-facing model) actually affects the residuals from a DDK model (on barycentric TOAs).

tests/utils.py
luojing1211: The standalone mode parameters are astropy quantities. The names are not required to be the same between the standalone model and pint-facing model. Your assumption is correct. If the PINT-facing model is set the standalone model should be updated to the PINT-facing model.

tests/utils.py
luojing1211: I allow the standalone binary to use the aliases for the reason that I don't know how the future user program the standalone model. Are they going to use the standalone model only? We don't have a clear API for it. Personally, I did not see a clear benefit to having the binary model stand alone.

luojing1211: If the return is None, the program will skip the test for that one parameter.

luojing1211:
@aarchiba I added a test for checking if the binary delay is changed When PX changed. PX only effect the ddk model when the TOAs are not barycentric TOAs. So simulated barycentric TOAs will not work. I checked baryceterED TOAs in the stand-alone model to be equal before and after.

luojing1211:
@aarchiba Is this PR in the mergeable status.

tests/test_ddk.py
```diff
@@ -49,6 +69,23 @@ def test_J1713(self):
...
+        self.modelJ1713.update_binary_object(toas=self.toasJ1713)
+        assert self.modelJ1713.binary_instance.PX.value == self.modelJ1713.PX.value
+        bdelay0 = self.modelJ1713.binary_instance.binary_delay()
+        b_time0 = self.modelJ1713.binary_instance.t
+        # Change PX value
+        self.modelJ1713.PX.value = 0.1
+        self.modelJ1713.update_binary_object()
+        b_time1 = self.modelJ1713.binary_instance.t
+        assert self.modelJ1713.binary_instance.PX.value == 0.1
+        # The stand alone binary model's input time should not change
```
aarchiba, line 82: Shouldn't the binary model's time variable be barycentered times - which depend on PX for non-barycentered TOAs? Why are these time variables the same?

tests/test_ddk.py
aarchiba: I can't tell what the error actually is - maybe something about having SINI? Above that seems to produce only warnings in `validate()`?

tests/utils.py
aarchiba: So if the internal model uses a name that the PINT-facing object doesn't have, the test passes? Shouldn't it fail? Shouldn't we also require standalone binary models to use the PINT names for things (and not aliases)?

tests/utils.py
aarchiba: Can you put an explanation in the docstring sufficient for readers to understand what is being tested? What does this function actually verify?

tests/test_ddk.py
luojing1211: I want to isolate the PX contribution in the barycentering and DDK model. I specifically designed the test this way.

tests/utils.py
luojing1211: We can require that. But I would rather use a new pull request not this bug-fix Pull request.

tests/test_ddk.py
aarchiba: Right, but I'm not clear why it works - why do the two `t` variables have the same value once `PX` has been changed?

luojing1211: The stand-alone model has a choice that not updating the input time only the parameter. Here the variable t does not get updated.

aarchiba: Is that documented anywhere? Could you please add a comment explaining that that's what's going on, and also document the update function so we can tell what's going on? That seems like a potentially surprising behaviour.

src/pint/models/binary_ell1.py
```diff
@@ -283,3 +273,14 @@ def validate(self):
             self.binary_instance.ds_func = self.binary_instance.delayS_H3_STIGMA_exact
             if self.STIGMA.quantity <= 0:
                 raise ValueError("STIGMA must be greater than zero.")
+        self.update_binary_object()
+
+    def validate(self):
```
aarchiba, line 278: Should users expect a warning or an exception if they try to use `SINI` and `M2` with `ELL1H`? I would prefer an exception.

luojing1211: It is in the Validate function. The changes are not showing everything. I can change it to an exception.

tests/test_ddk.py
luojing1211: Will add it.

luojing1211: The documentation is added.

src/pint/models/binary_ell1.py
luojing1211: It should be an exception now

src/pint/models/pulsar_binary.py
```diff
@@ -185,7 +187,31 @@ def apply_units(self):
...
+        acc_delay: `Numpy.ndarray`
+            If provided, TOAs will be corrected by provided acc_delay instead of
+            the standard barycentering. The stand alone binary receives the
+            input TOAs - acc_delay.
+
+        Note
+        ----
+        If the 'toas' are not provided, the stand alone binary model will use
+        the TOAs were passed to it from last interation or no TOAs for stand
+        alone binary model at all.
```
aarchiba, line 213: I'm confused by this - can this function accept arbitrary TOAs object, or is there some requirement that they already be barycentered? When no TOAs are passed, are the same TOAs as before used (so that changing the sky position would result in different residuals) or is some kind of intermediate result stored?

aarchiba:
https://nanograv-pint--1042.org.readthedocs.build/en/1042/api/pint.models.pulsar_binary.html The new docs show that the type indications are mis-formatted - if you give the full type name (e.g. `pint.toa.TOAs`) Sphinx will construct a link to the type definition.

src/pint/models/pulsar_binary.py
luojing1211: If the acc_delay is not given:
     This function does the barycentering itself. It accepts the observatory TOAs object.
If the acc_delay is given:
    This function pass the observatory - acc_delay to the stand alone binary model. 

The standalone model save the "barycentered TOA" in itself. If one changes the sky position, without passing TOAs. the binary delay would not change.

luojing1211: I know this could be risky. That is why I want to remove the stand-alone binary model.

tests/utils.py
luojing1211: I added more docstring to it.

tests/test_ddk.py
luojing1211: I believe this is changed to a TimingModelError.

aarchiba: Right, but what is actually wrong with the model? Why do we expect this to raise an exception? That should be clear from the test code.

tests/test_ell1h.py
```diff
@@ -143,6 +145,20 @@ def test_J0613_STIG(toasJ0613, modelJ0613_STIG):
     f.fit_toas()
 
 
+def test_SINI_warning():
```
aarchiba, line 148: Please change the name of this test to reflect the fact that it is checking for an exception.

tests/utils.py
```diff
@@ -0,0 +1,58 @@
...
+import warnings
+
+
+def verify_stand_alone_binary_parameter_updates(m):
+    """Generic test function for binary model tests.
+
+    Verify if the stand alone binary parameter values gets updated by the PINT
+    binary model objects. This test goes through the stand-alone binary model's
+    parameters and checks if the value is the same as corresponding PINT-face
+    object's parameter. If the PINT-face object is unset, it will skip the test.
```
aarchiba, line 16: What does it do if the PINT-facing object is missing the parameter? Does it check for PINT-facing parameters that don't exist in the stand alone binary parameters? Please clarify this in the docstring (and maybe those should be checked?)

src/pint/models/pulsar_binary.py
```diff
@@ -185,7 +187,31 @@ def apply_units(self):
...
+        toas: `~pint.toa.TOAs` object, optional
+            The TOAs that need to pass to the stand alone model.Default value is
+            None. If toas is None, this function only updates the parameter value.
+            If 'acc_delay' is not provided, the stand alone binary receives the
+            standard barycentered TOAs.
+
+        acc_delay: `~numpy.ndarray`
+            If provided, TOAs will be corrected by provided acc_delay instead of
+            the standard barycentering. The stand alone binary receives the
+            input TOAs - acc_delay.
```
aarchiba, line 207: Just to be clear, this means that, if provided, the binary object will use the TDB MJDs minus whatever numbers are in `acc_delay`? How would one obtain sensible numbers for `acc_delay`? Can you give an example of how one would correctly use this parameter?

src/pint/models/pulsar_binary.py
```diff
@@ -185,7 +187,31 @@ def apply_units(self):
             bparObj.value = bparObj.value * u.Unit(bparObj.units)
 
     def update_binary_object(self, toas=None, acc_delay=None):
-        """Update binary object instance for this set of parameters/toas."""
+        """Update stand alone binary's parameters and toas from PINT-facing object.
+
+        This function passes the PINT-facing object's parameter values and TOAs
+        to the stand-alone binary object. If the TOAs are not provided, it only
+        updates the parameters not the TOAs.
```
aarchiba, line 194: I think this is unclear. If I understand correctly, not providing TOAs means that the parameters are updated, but it is not the actual TOAs that are retained unchanged - this object retains barycentered TOAs internally, so that if you change the sky position (for example), the retained values are no longer the values you would obtain if providing the same TOAs. This is a very surprising behaviour that will affect anyone using this function with the default values, and it should be clearly stated in the documentation.

src/pint/models/binary_ddk.py
```diff
@@ -98,7 +98,15 @@ def validate(self):
...
                 )
         if self.SINI.quantity is not None:
-            raise ValueError(
+            raise TimingModelError(
                 "DDK model does not accept `SINI` as input. Please"
                 " use `KIN` instead."
             )
+
+        if hasattr(self._parent, "PX"):
+            if self._parent.PX.value == 0.0 or self._parent.PX.value is None:
```
aarchiba, line 107: Is a parallax value of zero really invalid? Is an infinite distance not a valid value for this model? If you're going to be forbidding zero, should you also forbid negative values of PX?

luojing1211: PX is zero is valid but not for DDK model. If you think the DDK model should accept zero, we should have another issue to discuss. This PR is designed to fix an internal bug and add some more checks. Having DDK accept zero PX requires a bigger discussion and more changes to the code.

src/pint/models/pulsar_binary.py
luojing1211: acc_delay follows the APIs in the TimingModel.delay() method which uses the acc_delay. This is designed for the case if some delay depends on the total delay of previous delay components. One can use this parameter to pass it into the function like:
```
delay = all_other_delay(toas, acc_delay=None)
dependent_delay = dependent_delay_fun(toas, acc_delay=delay)
```
This is a very low-level function, most users should not have to use it. I agree with you this is not the most optimized code which I developed years ago. If we want to optimize this part. Let us do it in another PR.

tests/test_ddk.py
luojing1211: We don't have a good translator from SINI--> KIN. With SINI, the code will get confused.

tests/utils.py
luojing1211: If PINT object does not exist, it will skip the test for now since there are parameters in the stand alone binay that PINT facing object does not have.

tests/test_ddk.py
aarchiba: Okay, thanks. Please make that clear in the test code, perhaps by verifying that the exception raised includes this information.

src/pint/models/pulsar_binary.py
aarchiba: `acc_delay` appears in various places in our API, including in TimingModel itself; currently the documentation there is just `???`. Until we're ready to remove it it needs documentation, since (among others) your new test is relying on its behaviour.

src/pint/models/binary_ddk.py
aarchiba: What about negative numbers?

tests/test_ell1h.py
```diff
@@ -143,6 +145,20 @@ def test_J0613_STIG(toasJ0613, modelJ0613_STIG):
...
 
+def test_SINI_error():
+    """Test SINI and M2 error.
+    """
+    SINI_par = simple_par.replace("H3 2.7507208E-7", "SINI 0.8")
+    with pytest.raises(ValueError, match="'SINI' will not be used in ELL1H model. "):
+        model.get_model(StringIO(SINI_par))
+
+
+def test_M2_warning():
```
aarchiba, line 156: Please change this function name as well, for exactly the same reason as the call above.

src/pint/models/binary_ddk.py
luojing1211: I changed the sign to <= 0.0. I am not sure why it does not show here.

src/pint/models/pulsar_binary.py
luojing1211: I added a warning.

src/pint/models/pulsar_binary.py
```diff
@@ -185,7 +187,33 @@ def apply_units(self):
...
+            the standard barycentering. The stand alone binary receives the
+            input TOAs - acc_delay.
+
+        Warns
+        -----
+        If the 'toas' are not provided, the stand alone binary model will use
+        the TOAs were passed to it from last interation (i.e. last barycnetered
+        TOAs) or no TOAs for stand alone binary model at all. This behavior will
+        cause incorrect answers. Allowing optional 'toas' input is for some
+        lower level tests. We do not recommend PINT user to use it.
```
aarchiba, line 215: Just to be clear: the recommendation is that PINT users always pass TOAs, right? Perhaps removing the default value of None would help ensure this - users who know they need to can still explicitly pass None.

src/pint/models/pulsar_binary.py
```diff
@@ -185,7 +187,33 @@ def apply_units(self):
...
-        """Update binary object instance for this set of parameters/toas."""
+        """Update stand alone binary's parameters and toas from PINT-facing object.
+
+        This function passes the PINT-facing object's parameter values and TOAs
+        to the stand-alone binary object. If the TOAs are not provided, it only
+        updates the parameters not the TOAs.
+
+        Parameters
+        ----------
+        toas: `~pint.toa.TOAs` object, optional
```
aarchiba, line 198: The backticks and tilde do not work in this context. Just giving the class name, undecorated, should be correctly interpreted by Sphinx.

src/pint/models/binary_ell1.py
```diff
@@ -283,3 +273,14 @@ def validate(self):
             self.binary_instance.ds_func = self.binary_instance.delayS_H3_STIGMA_exact
             if self.STIGMA.quantity <= 0:
                 raise ValueError("STIGMA must be greater than zero.")
+        self.update_binary_object(None)
+
+    def validate(self):
+        """Parameter validation."""
+        super(BinaryELL1H, self).validate()
```
aarchiba, line 280: Just use `super()`.

src/pint/models/binary_ell1.py
```diff
@@ -283,3 +273,14 @@ def validate(self):
...
+        # if self.H3.quantity is None:
+        #     raise MissingParameter("ELL1H", "H3", "'H3' is required for ELL1H model")
+        if self.SINI.quantity is not None:
+            raise ValueError("'SINI' will not be used in ELL1H model. ")
```
aarchiba, line 284: Should this be a `TimingModelError`?

tests/test_ddk.py
```diff
@@ -60,7 +97,7 @@ def test_J1713_deriv(self):
             if p in ["PX", "PMRA", "PMDEC"]:
                 continue
             par = getattr(self.modelJ1713, p)
-            if type(par).__name__ is "boolParameter":
+            if type(par).__name__ == "boolParameter":
```
aarchiba, line 100: Should this be `isinstance(par, boolParameter)`?

tests/test_ddk.py
```diff
@@ -105,26 +142,29 @@ def test_sini_from_value(self):
...
+        with pytest.raises(
+            TimingModelError,
+            match="DDK model does not accept `SINI` as input. Please use `KIN` instead.",
+        ):
+            mb.get_model(StringIO(test_par_str))
+
+    def test_stand_alone_model_params_updates(self):
+        test_par_str = self.temp_par_str + "\n KIN  71.969  1  0.562"
+        m = mb.get_model(StringIO(test_par_str))
+        verify_stand_alone_binary_parameter_updates(m)
```
aarchiba, line 155: Should this test also check that the internal model has a `KIN` parameter? If not the `verify...` will still pass.

tests/utils.py
```diff
@@ -0,0 +1,62 @@
+"""Utililty functions for the tests"""
+import logging
+import os
+from io import StringIO
+import unittest
+import pytest
+import warnings
+
+
+def verify_stand_alone_binary_parameter_updates(m):
```
aarchiba, line 10: Would this be more useful if it took specific pairs - `("KIN", "KIN")` and additionally checked whether those were copied over? As it stands this passes if a parameter doesn't exist on either the PINT-facing or standalone side, so it's not actually checking whether such a parameter is copied over. Above you call this to check whether `KIN` is copied over but it doesn't actually test that.

luojing1211: I agree this function can be improved.  Why don't we use another issue to discuss it? As it is, it tests the most copies and works for the KIN case, with two assertations in the test_ddk. 

ps: Can we remove the stand-alone binaries? I did not see any benefit, but a lot more risks and we need to test them.

aarchiba: Well, you will need to coordinate with Patrick, as he's written code that depends on them working as they do now.

More, there's a place for using objects in python to handle multiple computations that share subexpressions without recomputing those subexpressions. One would create such an object from a model and a set of TOAs, these would be fixed, and subexpressions would be cached (say using `lazyproperty`). One might also be able to use modern Python tools to compute derivatives without implementing them analytically or doing numerical differentiation. So some kind of internal object, carefully designed, might be a good idea.

luojing1211: Back to the topic of this PR. Do you think we have to change the function `verify_stand_alone_binary_parameter_updates(m)` to take specific pairs before merging this PR? I would say the current status of this test is sufficient for detecting the issue that is PR is aiming to fix.  I don't want to delay the 15 yr data analysis.

aarchiba: I think the function is not as useful as it looks without the ability to check for the existence of specific pairs, but it's not a blocker for this PR necessarily.

tests/test_ddk.py
luojing1211: I added two checks for the existence of `kin` before the  `verify` test

tests/utils.py
luojing1211: Let us fix this test function with the future development that is discussed in #1063. Maybe this function will not be needed any more.

Issue #1044, opened 2021-05-11
Allow direct computation of pulsar, companion masses


dlakaplan:
Currently in `utils.py` the pulsar and companion masses are computed with functions that use bisection to approximate a solution.  While not a major limit, there are analytic solutions to those problems that also enable the computation to retain units.  They are also vectorized, which can be handy.

Separately, the calculations do not check that the inputs are Quantities.

I have functions that do both calculations directly.  I _think_ these choose the correct roots, but I am not 100% sure.  If these seem useful I can submit a PR to replace the existing versions with these (and add in some tests).

```python
def companion_mass(x, PB, incl=60*u.deg, Mp=1.4*u.Msun):
    """
    Compute the companion mass from the orbital parameters

    Compute companion mass for a binary system from orbital mechanics,
    not Shapiro delay.

    Parameters
    ----------
    PB : Quantity
        Binary orbital period
    x : Quantity
        Projected semi-major axis (aka ASINI) in `pint.ls`
    incl : Angle, optional
        Inclination angle, in u.deg or u.rad. Default is 60 deg.
    Mp : Quantity, optional
        Pulsar mass in u.solMass. Default is 1.4 Msun

    Returns
    -------
    mass : Quantity in u.solMass
    """
    if not (isinstance(incl,angles.Angle) or isinstance(incl,u.quantity.Quantity)):
        raise ValueError(
                f"The inclination should be an Angle but is {incl}.")
    if not isinstance(x, u.quantity.Quantity):
        raise ValueError(
                f"The projected semi-major axis x should be a Quantity but is {x}.")
    if not isinstance(PB,u.quantity.Quantity):
        raise ValueError(
                f"The binary period PB should be a Quantity but is {PB}.")
    if not isinstance(Mp,u.quantity.Quantity):
        raise ValueError(
                f"The pulsar mass should be a Quantity but is {Mp}.")

    # this is the mass function
    Y=4*np.pi**2*x**3/((PB)**2*c.G)
    # This then forms a cubic equation of the form:
    # ca*Mc**3 + cb*Mc**2 + cc*Mc + cd = 0
    ca=np.sin(incl)**3
    cb=-Y
    cc=-2*Mp*Y
    cd=-Y*Mp**2
    # solution
    # https://en.wikipedia.org/wiki/Cubic_equation#General_cubic_formula
    delta0 = cb**2 - 3*ca*cc
    delta1 = 2*cb**3 - 9*ca*cb*cc + 27*ca**2*cd
[...]

PR #1046, opened 2021-05-12
Compute masses

dlakaplan:
This addresses https://github.com/nanograv/PINT/issues/1044. I added tests for the functions:
`utils/companion_mass()`
`utils/pulsar_mass()`

These are both vectorized and explicitly check for and use units.

I also looked over both calculations and verified that they pick the correct roots and should not suffer from numerical instabilities, but I would appreciate a second check of that.  Specifically, for the `companion_mass()` calculation that uses the analytic roots of a cubic, which has a few different ways to do it.  I algebraically simplified the calculations and verified that there are no differences among quantities that could come close to 0, and that the only issues with 0 in a denominator are when sin(i) -> 0.

### DIFF
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1260,6 +1260,8 @@ def pulsar_mass(pb, x, mc, inc):
     cb = 2 * massfunct * mc
     cc = massfunct * mc ** 2 - (mc * np.sin(inc)) ** 3
     # solve it directly
+    # this has to be the positive branch of the quadratic
+    # because the vertex is at -mc, so the negative branch will always be < 0
     return ((-cb + np.sqrt(cb ** 2 - 4 * ca * cc)) / (2 * ca)).to(u.Msun)
 
 
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -7,6 +7,7 @@ from io import StringIO
 import astropy.constants as const
 import astropy.coordinates as coords
 import astropy.units as u
+import astropy.coordinates.angles as angles
 import numpy as np
 import scipy.optimize.zeros as zeros
 from astropy import log
@@ -1208,10 +1209,9 @@ def mass_funct2(mp, mc, i):
 
 
 def pulsar_mass(pb, x, mc, inc):
-    """Compute pulsar mass from orbit and Shapiro delay parameters
+    """Compute pulsar mass from orbital parameters
 
     Return the pulsar mass (in solar mass units) for a binary.
-    Finds the value using a bisection technique.
 
     Parameters
     ----------
@@ -1228,13 +1228,26 @@ def pulsar_mass(pb, x, mc, inc):
     -------
     mass : Quantity in u.solMass
     """
+    if not (isinstance(inc, angles.Angle) or isinstance(inc, u.quantity.Quantity)):
+        raise ValueError(f"The inclination should be an Angle but is {inc}.")
+    if not isinstance(x, u.quantity.Quantity):
+        raise ValueError(
+            f"The projected semi-major axis x should be a Quantity but is {x}."
+        )
+    if not isinstance(pb, u.quantity.Quantity):
+        raise ValueError(f"The binary period pb should be a Quantity but is {pb}.")
+    if not isinstance(mc, u.quantity.Quantity):
+        raise ValueError(f"The companion mass should be a Quantity but is {mc}.")
+
     massfunct = mass_funct(pb, x)
 
-    # Do some unit manipulation here so that scipy bisect doesn't see the units
-    def localmf(mp, mc=mc, mf=massfunct, i=inc):
-        return (mass_funct2(mp * u.solMass, mc, i) - mf).value
-
-    return zeros.bisect(localmf, 0.0, 1000.0) * u.solMass
+    # This then forms a quadratic equation of the form:
+    # ca*Mp**2 + cb*Mp + cc = 0
+    ca = massfunct
+    cb = 2 * massfunct * mc
+    cc = massfunct * mc ** 2 - (mc * np.sin(inc)) ** 3
+    # solve it directly
+    return ((-cb + np.sqrt(cb ** 2 - 4 * ca * cc)) / (2 * ca)).to(u.Msun)
 
 
 def companion_mass(pb, x, inc=60.0 * u.deg, mpsr=1.4 * u.solMass):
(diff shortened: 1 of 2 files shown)


aarchiba:
Using `hypothesis` is a fairly effective way to uncover every numerical instability or weirdness anywhere in a function. There are a few examples elsewhere in the code base.

dlakaplan:
Thanks, @aarchiba.  I added some tests using `hypothesis` and they all pass for a wide (plausible) range of parameters.

src/pint/utils.py
```diff
@@ -1258,13 +1288,76 @@ def companion_mass(pb, x, inc=60.0 * u.deg, mpsr=1.4 * u.solMass):
...
+    if not isinstance(x, u.quantity.Quantity):
+        raise ValueError(
+            f"The projected semi-major axis x should be a Quantity but is {x}."
+        )
+    if not isinstance(pb, u.quantity.Quantity):
+        raise ValueError(f"The binary period pb should be a Quantity but is {pb}.")
+    if not isinstance(mpsr, u.quantity.Quantity):
+        raise ValueError(f"The pulsar mass should be a Quantity but is {mpsr}.")
+
+    """
```
aarchiba, line 1302: Is it worth putting this in a Note section of the docstring?

tests/test_utils.py
```diff
@@ -648,12 +649,22 @@ def test_psr_utils():
 
     # Mass function
```
aarchiba, line 650: Can y ou make each of these a separate test, so we can tell what broke and what didn't?

tests/test_utils.py
```diff
@@ -677,6 +688,73 @@ def test_psr_utils():
...
+)
+def test_masses(Mpsr, Mc, Pb, incl):
+    """
+    test pulsar and companion mass calculations for a range of values
+    """
+    assume(incl > 0)
+    assume(incl <= 90)
+    assume(Mpsr > 0)
+    assume(Mc > 0)
+    assume(Pb > 0)
```
aarchiba, line 717: These `assume`s aren't necessary - the `assume` statement exists to reject generated values that are for some reason not acceptable, while you have already ensured that these limits are met in generation.

tests/test_utils.py
```diff
@@ -677,6 +688,73 @@ def test_psr_utils():
...
+
+    from pint.utils import (
```
aarchiba, line 719: Please just put this in the normal imports section at the top of the file.

tests/test_utils.py
```diff
@@ -677,6 +688,73 @@ def test_psr_utils():
...
+        companion_mass,
+        pulsar_mass,
+    )
```
aarchiba, line 722: Would it be clearer to immediately attach units to Mc, Mpsr, et cetera? Then units could be more or less omitted from the rest of this.

tests/test_utils.py
```diff
@@ -677,6 +688,73 @@ def test_psr_utils():
...
+
+    Mtot = (Mc + Mpsr) * u.Msun
+    # full semi-major axis
+    a = (c.G * Mtot * (Pb * u.day / (2 * np.pi)) ** 2) ** (1.0 / 3)
+    # pulsar semi-major axis
+    apsr = (Mc * u.Msun / Mtot) * a
+    # projected
+    x = (apsr * np.sin(incl * u.deg)).to(pint.ls)
```
aarchiba, line 730: Does `np.sin` do the conversion to radians when presented with a quantity?

tests/test_utils.py
```diff
@@ -677,6 +688,73 @@ def test_psr_utils():
...
+
+    # computed pulsar mass
```
aarchiba, line 732: Could different functions, and different properties of functions, be tested in different tests?

tests/test_utils.py
```diff
@@ -677,6 +688,73 @@ def test_psr_utils():
...
+    assert np.isclose(
+        pulsar_mass(Pb * u.day, x, Mc * u.Msun, inc=incl * u.deg), Mpsr * u.Msun
+    )
+    # computed companion mass
+    assert np.isclose(
+        companion_mass(Pb * u.day, x, mpsr=Mpsr * u.Msun, inc=incl * u.deg), Mc * u.Msun
+    )
+    # and test exceptions
+    with pytest.raises(ValueError):
```
aarchiba, line 741: Could you put these in separate tests named by what they're testing? Like, why is a ValueError supposed to be emitted in each case? I don't think these need to be `hypothesis`-ized, either.

tests/test_utils.py
dlakaplan: Yes, I wrote these before I put in the full hypothesis limits, so I agree they are redundant.

tests/test_utils.py
dlakaplan: I was just following whoever had done this before, but I agree it seems odd.

tests/test_utils.py
dlakaplan: I was just adding to how it was before, but I agree it seems odd.  I'll break out at least some of these.

tests/test_utils.py
dlakaplan: yes
```
In [1]: print(np.sin(90*u.deg))
1.0
```

tests/test_utils.py
dlakaplan: Done

tests/test_utils.py
dlakaplan: Done (if I understand what you are suggesting)

tests/test_utils.py
aarchiba: I think maybe I didn't make my suggestion clear? My idea was that the first lines of the test would ensure that all the inputs were immediately converted to Quantities, rather than repeatedly converting them on the spot everywhere they are used.

dlakaplan: I thought you were talking about whether the function (not the test) accepted values that had no units and made assumptions about defaults.  

I am not entirely sure what you mean about the test.  Maybe better to discuss at the next meeting (next week?).

src/pint/utils.py
```diff
@@ -1204,14 +1211,20 @@ def mass_funct2(mp, mc, i):
...
+        raise ValueError(f"The inclination should be an Angle but is {i}.")
+    if not isinstance(mc, u.quantity.Quantity):
+        raise ValueError(f"The companion mass should be a Quantity but is {mc}.")
+    if not isinstance(mp, u.quantity.Quantity):
+        raise ValueError(f"The pulsar mass should be a Quantity but is {mp}.")
+
     return (mc * np.sin(i)) ** 3.0 / (mc + mp) ** 2.0
 
 
 def pulsar_mass(pb, x, mc, inc):
```
aarchiba, line 1224: I think that these functions now work on arrays of values (where they didn't before because of the `bisect`)? Might it be worth just doing a quick check in the test suite, just confirm they run on arrays?

tests/test_utils.py
```diff
@@ -630,54 +639,218 @@ def test_pmtot():
...
+def test_mass_function_error_without_quantity_pb():
+    pb = 1.0 * u.d
+    x = 2.0 * pint.ls
+
+    # Mass function
+    with pytest.raises(ValueError):
+        mass_funct(pb.value, x)
+
+
+def test_other_mass_function():
```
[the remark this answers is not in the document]
dlakaplan: good idea - done.

src/pint/utils.py
dlakaplan: good idea - done

tests/test_utils.py
aarchiba: My suggestion would be to add the lines, here at the top of the function, 
```python
Mpsr = Mpsr*u.solMass
Mc = Mc*u.solMass
Pb = Pb*u.day
incl = incl*u.deg
```
Then the rest of the function would use Quantities consistently without having to write units mixed into things like `a = (c.G * Mtot * (Pb * u.day / (2 * np.pi)) ** 2) ** (1.0 / 3)` (no units for `Mtot` because it got units when created, but units for `Pb` because it doesn't have them).

Alternatively a bit of functional-programming hocus-pocus could arrange for hypothesis to produce values that are already Quantities, but this is simpler I think.

dlakaplan: Sure - I don't see a big difference, and if this is clearer it's fine.  Just pushed up those changes.

src/pint/utils.py
dlakaplan: OK, just pushed up changes with everything in the docstrings

tests/test_utils.py
```diff
@@ -630,54 +639,218 @@ def test_pmtot():
...
     # Mass function, second form
     assert np.isclose(
         mass_funct2(1.4 * u.solMass, 0.2 * u.solMass, 60.0 * u.deg),
         0.0020297470401197783 * u.solMass,
     )
 
+
+def test_other_mass_function_error_without_quantity_mp():
+    pb = 1.0 * u.d
+    x = 2.0 * pint.ls
```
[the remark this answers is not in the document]
dlakaplan: I'm not sure I agree, but I also don't have a strong opinion.  So this is now updated (I hope) and pushed.

luojing1211:
Is this able to merge now?

dlakaplan:
I think so.  @aarchiba ?

PR #1047, opened 2021-05-19
Covariance matrix fixes

dlakaplan:
This is still a WIP, but I wanted to get in a PR for the next PINT meeting.  Major changes:
* Most of the fitters now use the `CovarianceMatrix` class to preserve parameter order along with the matrix
* Some UI changes for that class (or `PintMatrix` class), mostly to simplify, a couple to extend use
* Renamed `fitter.covariance_matrix` to `fitter.parameter_covariance_matrix` to make it clear which is which (and changed other calls, tests, examples)
* Implemented new random model (`utils/calculate_random_models`) to generate random models using parameter covariance matrix
* tests for the above, plus a worked example (`examples/check_phase_connection.py`)

Still TBD:
* check that this works OK with all fitters (tests pass, not sure if other checks are needed)
* check that matrix metadata is not needlessly regenerated (parameter labels are redone on every step for some fitters, but maybe this is needed)
* I made `fitter.covariance_matrix` redirect to `fitter.parameter_covariance_matrix` for compatibility along with a deprecation warning, but I don't get the warning.  So maybe I did it wrong, and maybe we don't want this at all.

### DIFF
--- a/src/pint/pint_matrix.py
+++ b/src/pint/pint_matrix.py
@@ -88,6 +88,10 @@ class PintMatrix:
         axis: int (optional)
             axis for the label lookup
 
+        Returns
+        -------
+        labels : list of label names for each dimension
+
         """
         labels = []
         if axis is None:
@@ -104,6 +108,10 @@ class PintMatrix:
 
     def get_all_label_names(self):
         """return all possible unique label names
+
+        Returns
+        -------
+        labels : set of unique label names across all dimensions
 
         TODO: preserve order?
         """
@@ -122,6 +130,11 @@ class PintMatrix:
             Name of the label.
         axis: int (optional)
             axis for the label lookup
+
+        Returns
+        -------
+        sizes : list of (start,stop) indices for the given label across each dimension
+
         """
         lb_sizes = []
         lbs = self.get_label(label, axis=axis)
@@ -142,6 +155,20 @@ class PintMatrix:
         return label_entry[1][0]
 
     def get_axis_labels(self, axis):
+        """Get the axis labels for the specified axis
+
+        Parameters
+        ----------
+        axis: int 
+            axis for the label lookup
+
+        Returns
+        -------
+        labels : list of (name, (start,stop,unit) for the given axis
+
+
+        TODO: does this just replicate self.axis_labels[axis]?  Is the sort meaningful?
+        """
         dim_label = list(self.axis_labels[axis].items())
         dim_label.sort(key=self._get_label_start)
         return dim_label
@@ -150,8 +177,19 @@ class PintMatrix:
         """Get the label entry and its dimension.
         If axis is specified, will only be along that axis
 
-        We assume the labels are unique in the matrix.
+        Parameters
+        ----------
+        label: str
+            label to lookup
+        axis: int, optional
+            axis for the label lookup
+
+        Returns
+        -------
+        labels : list of (name, axis, start,stop,unit) for the given label
         """
+
+        # We assume the labels are unique in the matrix.
         all_label = []
         for ii, dim in enumerate(self.axis_labels):
             if label in dim.keys():
@@ -181,7 +219,18 @@ class PintMatrix:
             )
 
     def get_label_slice(self, labels):
-        """Return the given label slices."""
+        """Return the given label slices.
+
+        Parameters
+        ----------
+        labels: list of str
+            labels to lookup
+
+        Returns
+        -------
+        slice, labels : slice into original matrix and labels of new (sliced) matrix
+
+        """
         dim_slices = dict([(d, slice(None)) for d in range(self.ndim)])
         new_labels = dict([(d, {}) for d in range(self.ndim)])
         for lb in labels:
@@ -200,7 +249,18 @@ class PintMatrix:
         return np.ix_(*list(dim_slices.values())), list(new_labels.values())
 
     def get_label_matrix(self, labels):
-        """Get a sub-matrix data according to the given labels."""
+        """Get a sub-matrix data according to the given labels.
+
+        Parameters
+        ----------
+        labels: list of str
+            labels to lookup
+
+        Returns
+        -------
+        PintMatrix : new matrix with only the specified labels
+
+        """
         slice, new_labels = self.get_label_slice(labels)
         return PintMatrix(self.matrix[slice], new_labels)
 
--- a/src/pint/pint_matrix.py
+++ b/src/pint/pint_matrix.py
@@ -5,6 +5,7 @@ import numpy as np
 from itertools import combinations
 import astropy.units as u
 from collections import OrderedDict
+from warnings import warn
 import copy
 
 
@@ -236,6 +237,10 @@ class PintMatrix:
 
         DEPRECATED - use get_label(..., axis=axis)
         """
+        warn(
+            "This method is deprecated. Use `parameter_covariance_matrix.get_label(..., axis=axis)[0]` instead of `parameter_covariance_matrix.get_label_along_axis(...)",
+            category=DeprecationWarning,
+        )
         label_in_one_axis = self.axis_labels[axis]
         if label_name in label_in_one_axis.keys():
             return (label_name, axis) + label_in_one_axis[label_name]
--- a/src/pint/pint_matrix.py
+++ b/src/pint/pint_matrix.py
@@ -183,8 +183,6 @@ class PintMatrix:
         -------
         labels : list of (name, (start,stop,unit) for the given axis
 
-
-        TODO: does this just replicate self.axis_labels[axis]?  Is the sort meaningful?
         """
         dim_label = list(self.axis_labels[axis].items())
         dim_label.sort(key=self._get_label_start)
--- a/src/pint/pint_matrix.py
+++ b/src/pint/pint_matrix.py
@@ -155,7 +155,7 @@ class PintMatrix:
     def match_labels_along_axis(self, pint_matrix, axis):
         """Match one axis' labels index between the current matrix and input pint matrix.
 
-        The labels will be matched along axises, not cross the axises.
+        The labels will be matched along axes, not cross the axes.
 
         Parameters
         ----------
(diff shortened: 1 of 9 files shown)


dlakaplan:
Also added:
```python
    def __getitem__(self, key):
        return self.matrix[key]

    def __setitem__(self, key, value):
        self.matrix[key] = value
``` 
to `PintMatrix` to allow seamless access.

dlakaplan:
Can probably also deprecate `utils/show_param_cov_matrix` since I just added those methods directly to the CovarianceMatrix class.

src/pint/pint_matrix.py
```diff
@@ -50,6 +51,12 @@ def __init__(self, matrix, axis_labels):
         # Check label index overlap TODO: should we allow overlap?
         self._check_index_overlap()
 
+    def __getitem__(self, key):
```
aarchiba, line 54: Could you add a docstring here to explain what `key` is supposed to be? Like, is this supposed to implement array indexing, or indexing by label, or both?

src/pint/pint_matrix.py
```diff
@@ -72,16 +79,65 @@ def label_units(self):
             units.append(self.get_axis_labels(dim))
         return units
 
-    def get_label_size(self, label):
-        """Get the size of the a label in each axis.
+    def get_label_names(self, axis=None):
+        """return only the names of the labels
```
aarchiba, line 83: Could you please put this in docstring format? First line is a short summary, in the form of a complete sentence with capital letter and period. This will show up correctly in the documentation.

src/pint/pint_matrix.py
```diff
@@ -72,16 +79,65 @@ def label_units(self):
...
+        along the specified axis if requested
+
+        Parameters
+        ----------
+        axis: int (optional)
+            axis for the label lookup
+
+        Returns
+        -------
+        labels : list of label names for each dimension
```
aarchiba, line 93: For each dimension or just  the chosen axis?

src/pint/utils.py
```diff
@@ -1552,3 +1558,98 @@ def remove_dummy_distance(c):
             "Do not know coordinate frame for %r: returning coordinates unchanged" % c
         )
         return c
+
+
+def calculate_random_models(fitter, toas, Nmodels=100, keep_models=True, params="all"):
+    """
+    calculates random models based on the covariance matrix of the `fitter` object
```
aarchiba, line 1565: Docstring format. (see PEP8)

src/pint/utils.py
```diff
@@ -1552,3 +1558,98 @@ def remove_dummy_distance(c):
...
+
+    Returns
+    -------
+    dphase : np.ndarray
+        phase difference with respect to input model, size is [Nmodels, len(toas)]
+    random_models : list (optional)
+        list of random models (each is a `pint.models.timing_model.TimingModel`)
+
+
+    Note: to calculate new TOAs, you can do:
```
aarchiba, line 1591: There's a format for a Note section (and for embedded source code).

src/pint/pint_matrix.py
dlakaplan: Done.  Also removed `__setitem__` since I realized it wouldn't make sense and could lead to weird behavior.

src/pint/pint_matrix.py
dlakaplan: updated

src/pint/pint_matrix.py
```diff
@@ -72,16 +79,65 @@ def label_units(self):
...
+        TODO: preserve order?
+        """
+        labels = self.get_label_names()
+        unique_labels = []
+        for l in labels:
+            unique_labels += l
+        return set(unique_labels)
+
+    def get_label_size(self, label, axis=None):
+        """Get the size of the a label for all axes, or just the specified
```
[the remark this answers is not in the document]
dlakaplan: This is nomenclature from whoever first wrote this.  It allows a label to apply to multiple indices in an array.  So size is (start, stop) indices.

src/pint/pint_matrix.py
```diff
@@ -99,26 +155,61 @@ def _get_label_start(self, label_entry):
         return label_entry[1][0]
 
     def get_axis_labels(self, axis):
+        """Get the axis labels for the specified axis
```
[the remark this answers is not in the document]
dlakaplan: I agree (perhaps I added that note).  But I don't know who wrote the original function so I wanted to discuss before removing.

src/pint/pint_matrix.py
```diff
@@ -99,26 +155,61 @@ def _get_label_start(self, label_entry):
...
+                )
         else:
             return all_label
 
     def get_label_along_axis(self, axis, label_name):
-        """Get the request label from on axis."""
+        """Get the request label from on axis.
+
+        DEPRECATED - use get_label(..., axis=axis)
+        """
```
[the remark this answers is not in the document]
dlakaplan: I wanted to discuss first, but then yes.

src/pint/pint_matrix.py
```diff
@@ -128,34 +219,55 @@ def get_label_along_axis(self, axis, label_name):
...
+        """Return the given label slices.
+
+        Parameters
+        ----------
+        labels: list of str
+            labels to lookup
+
+        Returns
+        -------
+        slice, labels : slice into original matrix and labels of new (sliced) matrix
```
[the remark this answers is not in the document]
dlakaplan: an actual slice object

src/pint/pint_matrix.py
```diff
@@ -548,15 +660,100 @@ def combine_design_matrices_by_param(matrix1, matrix2, padding=0.0):
...
             raise ValueError("The input matrix is not symmetric.")
         # Check if the labels are symmetric
         if len(labels[0]) != len(labels[1]):
-            raise ValueError("The input labels are not sysmmetric.")
+            raise ValueError("The input labels are not symmetric.")
         super(CovarianceMatrix, self).__init__(matrix, labels)
 
+    def prettyprint(self, prec=3, coordinatefirst=False):
+        """
+        prettyprint(self, prec=3, coordinatefirst=False)
```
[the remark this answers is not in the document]
dlakaplan: OK

src/pint/utils.py
dlakaplan: fixed.

tests/test_random_models.py
```diff
@@ -0,0 +1,77 @@
...
+import pint.models as tm
+from pint import fitter, toa
+from pinttestdata import datadir
+import pint.models.parameter as param
+from pint import ls
+from pint import utils
+
+
+@pytest.mark.skipif(
+    "DISPLAY" not in os.environ, reason="Needs an X server, xvfb counts"
```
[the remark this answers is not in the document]
dlakaplan: Agree.  I had taken that from `test_fitter.py/test_fitter()` as noted, but it is not necessary.

Also agree about testing the `CovarianceMatrix` class more generally, but I mostly just made small tweaks and it would take a while to come up with all possible tests.  I can start thinking about that...

src/pint/pint_matrix.py
```diff
@@ -72,16 +79,65 @@ def label_units(self):
...
+            if isinstance(axis, int):
+                return [x[0] for x in self.get_axis_labels(axis)]
+            else:
+                # assume it's an iterable (how to check?)
+                r = axis
+        for dim in r:
+            labels.append([x[0] for x in self.get_axis_labels(dim)])
+        return labels
+
+    def get_all_label_names(self):
```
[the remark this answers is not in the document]
aarchiba: I just worry that general deduplication will look like it's going to give just one copy for covariance matrices, until someone runs into a covariance matrix with repeated labels on the same axis, when it's going to do something different from what they assume.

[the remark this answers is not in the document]
dlakaplan: I think this facility is important because this works for other matrices as well which might have repeated labels.  There are differences between what I did (relating to parameter covariance matrices) and other sorts (data covariance matrices, ...).  This may be historical - I don't know how much this is used.

So I would be fine with changing this behavior but not without understanding what other uses are.

Alternately, could make a subclass that does not allow for repeated labels.

tests/test_random_models.py
```diff
@@ -0,0 +1,56 @@
...
+from pint import utils
+
+
+def test_random_models():
+    # taken from test_fitter.py/test_fitter()
+    # Get model
+
+    m = tm.get_model(os.path.join(datadir, "NGC6440E.par"))
+
+    # Get TOAs
```
scottransom, line 26: I think the TOAs portion could be simplified considerably by using `get_TOAs()`, which does all of those later steps.  Or, alternatively, get both the model and the TOAs all at once (and in one line -- which is now how we should, in general, be recommending that people load things) using `pm.get_model_and_toas()`.

src/pint/pint_matrix.py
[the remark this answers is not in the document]
scottransom: Seems like this should get zapped now rather than have it hanging around.

tests/test_random_models.py
dlakaplan: I think I copied this from another test routine, but now I've updated it.  will push it up soon.

scottransom:
I really like the new random models interface.  The dphase array is great!  And it seems like it is faster than the old one....    Things are looking good to me!

dlakaplan:
I know I put in the note that we might deprecate `get_axis_labels`, but now I'm less convinced.  Along with the ordering it repackages it from a dict (indexed by parameter name) to a list of tuples.  It's not a huge difference between them but it would involve a bunch of small changes.  So for now I suggest we leave this as is (but I'll update the doctoring to make this clear).  

[sorry - I couldn't figure out how to respond to that comment individually]

scottransom:
Did you ever push up that small change for the test so that it uses get_model_and_toas()?  Once you do, I'll merge the PR.

dlakaplan:
I forgot, but just did it now.

scottransom:
Thanks for this!

PR #1050, opened 2021-06-07
Add observatories from tempo2's observatories.dat

### LINKED ISSUE #1048: Add observatory metadata content
We put in a facility (#790) to allow observatories to have metadata describing where/when the info comes from.  But it looks like it isn't uniformly used.  We should have a guide for how to add an observatory and include this as part of it.

### LINKED ISSUE #1049: Support observatory AXIS
The IPTA mock data challenge and `libstempo` use a fictitious observatory AXIS; it is on the ITRF X-axis, but I don't know other details. They'll be in tempo2 config files. PINT does not currently support this observatory but it should.

aarchiba:
This PR also includes a function to check PINT's list of observatories against TEMPO and TEMPO2's and report ones that PINT is missing or thinks are different. With this patch the only differences appear to be errors in TEMPO2 or minor disagreements in position.

One possible warning is that TEMPO expects G1 to mean the GBT 140-foot while PINT previously defined G1 to be the GEO600 site. I removed the latter but left a comment. 

Now includes `origin` strings for most (all?) topographic observatories, describing the origin and/or comparison with TEMPO/TEMPO2 plus any other information that seemed relevant and easy to obtain.

Closes #1049
Closes #1048

### DIFF
--- a/src/pint/observatory/__init__.py
+++ b/src/pint/observatory/__init__.py
@@ -136,6 +136,7 @@ class Observatory:
         # We can't do this in the import section above because this class
         # needs to exist before that file is imported.
         import pint.observatory.observatories  # noqa
+        import pint.observatory.special_locations  # noqa
 
         if name == "":
             raise KeyError("No observatory name or code provided")
--- a/src/pint/observatory/__init__.py
+++ b/src/pint/observatory/__init__.py
@@ -1,9 +1,12 @@
 import os
 import textwrap
 from collections import defaultdict
+from io import StringIO
 
 import astropy.constants as const
+import astropy.coordinates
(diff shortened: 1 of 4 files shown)


aarchiba:
I have a notebook I use to check - just runs the two new functions really - but I'm not sure what to do with it. It's not really very friendly to go into our documentation alongside the (validation example)[https://nanograv-pint.readthedocs.io/en/latest/examples-rendered/paper_validation_example.html], and it also can't be run on the doc builder, but when it can be run it provides a timestamped comparison of PINT's observatories with those of TEMPO and TEMPO2. I'm not quite sure where best to put it.

PR #1055, opened 2021-06-20
Plotting Priors fitter method

devbhakt:
Added in a method object to the fitter to plot the prior probability distribution, and an initial gaussian probability distribution over the post MCMC samples histogrammed with maximum posterior values, the initial fit values, and 2 sigma standard deviation lines.


src/pint/scripts/event_optimize.py
```diff
@@ -411,6 +411,90 @@ def prof_vs_weights(self, nbins=50, use_weights=False):
             plt.savefig(ftr.model.PSR.value + "_htest_v_wgtcut_unweighted.png")
         plt.close()
 
+    def plot_priors(self, samples, weights=None, bins=100, scale=False, file=False):
+        """
+        Show binned samples, prior probability distribution and an initial gaussian
```
scottransom, line 416: We should be using the numpy docstring format for new functions and methods.  That is described here:
https://numpydoc.readthedocs.io/en/latest/format.html

src/pint/plot_utils.py
```diff
@@ -176,3 +177,127 @@ def phaseogram_binned(
...
+    ----------
+    model
+        Input timing model 
+    chains: dict    
+        Post MCMC integration chains that are input into a histogram and compared with the priors. Contain the fitkeys and the samples
+    fitvals
+        Initial Parameter values
+    fiterrs
+        Initial Parameter uncertainties
+    maxpost_fitvals
```
aarchiba, line 209: Can these be optional? If you don't have them, can the code still make sensible plots? That would make the function a little more convenient. Actually if all the fit values and errors were optional, drawn on the plot only if provided, that would make this function more convenient - just drop the chains in if they're all you have, or include more information for a more useful plot.

src/pint/plot_utils.py
```diff
@@ -176,3 +177,127 @@ def phaseogram_binned(
...
+        Maximum Posterior values
+    With scale is True
+        The priors and initial probability distributions are scaled to the peak of the histogram.
+    With scale is False
+        The priors and initial probability distributions are plotted without the scaling.
+    """
+    keys = []
+    values = []
+    for item in chains.items():
```
aarchiba, line 218: With `.items()` it's more conventional to write `for k, v in chains.items():`

src/pint/plot_utils.py
```diff
@@ -176,3 +177,127 @@ def phaseogram_binned(
...
+                x_range[i]
+            )
+        )
+        a, x = np.histogram(values[i], bins=bins)
+        counts.append(a)
+
+    fig, axs = plt.subplots(len(keys), figsize=(8, 11), constrained_layout=True)
+
+    for i, p in enumerate(keys):
+        if i != len(keys[:-1]):
```
aarchiba, line 240: Should this be `len(keys)-1`? This might behave a little different with a zero-length list of keys, but otherwise the subtraction is probably clearer. Or, actually, is this really supposed to run for all but the last key?

src/pint/plot_utils.py
```diff
@@ -176,3 +177,127 @@ def phaseogram_binned(
...
+                    x_range[i] - maxpost_fitvals[i],
+                    priors[i],
+                    label="Prior Probability",
+                    color="g",
+                )
+        else:
+            handles, labels = axs[0].get_legend_handles_labels()
+            axs[i].set_axis_off()
+            axs[i].legend(handles, labels)
+    if file:
```
aarchiba, line 298: I think these are not a good idea. Users might want to add more things to the figure, so calling `plt.close()` is going to cause them problems. And they may want to specify extra arguments (`dpi` for example) when saving to a file, so let them save the thing if they want to.

tests/J0030+0451_prof_pre.txt
```diff
@@ -0,0 +1,256 @@
+0.00000       1.90144
```
aarchiba, line 1: Is there actual test code that uses these files?

src/pint/plot_utils.py
```diff
@@ -176,3 +177,127 @@ def phaseogram_binned(
...
+    fiterrs,
+    maxpost_fitvals,
+    burnin=100,
+    bins=100,
+    scale=False,
+    file=False,
+):
+
+    """
+    Show binned samples, prior probability distribution and an initial gaussian
```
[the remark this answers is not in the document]
devbhakt: I will go in and update the docstring

src/pint/plot_utils.py
devbhakt: The fitvals and fiterrs are used to calculate an initial gaussian probability distribution. I can make it so that it plots just the chain values and the priors and add in optionals.

src/pint/plot_utils.py
devbhakt: I will update that

src/pint/plot_utils.py
devbhakt: It is supposed to run for all but the last key.

tests/J0030+0451_prof_pre.txt
devbhakt: I do not think the files are used in test code. Should I go ahead and delete them?

aarchiba: Please don't add files that are never used. (This PR seems to be adding that file.)

aarchiba:
https://nanograv-pint--1055.org.readthedocs.build/en/1055/_autosummary/pint.plot_utils.plot_priors.html#pint.plot_utils.plot_priors

Documentation is here.

- `model` should at least contain a type indication (TimingModel) and short description.
- This appears to support discarding the beginning of each chain (`burn`) but does it support thinning?
- `chains` is a dictionary, but what should it contain? How can I create a suitable dictionary to call this function with? Even just pointing to the MCMC fitter, if that's how, would be a help.
- `maxpost_fitvals` - what format should this be in? A list, I guess, but which parameter goes where in the list? Is there any reason this can't be obtained from the chains? What happens if they are not provided?
- `fitvals` - please tell users what will happen if these are not provided.
- The `scale` and `file` arguments appear to be accepted but not documented.

aarchiba:
Thank you for the documentation update. I have a question about the `scale` parameter. If it is not provided, how are the priors and histogram scaled? I know histograms can be produced in density units, which seems like it should agree with a prior that is a normalized PDF (though emcee does not need its priors to be normalized). Is that what is done? The documentation would be more helpful if it said what to expect if `scale=False` were provided.

devbhakt:
Currently the histograms are not scaled at all and are not produced in density units. The priors are not normalized. If `scale=True`, then the peak of the priors are scaled to the peak of the histogram. If `scale=False`, the priors will be plotted as is, completely independent of the histogram. 
The reason why this was put in is during my testing phase with real and simulated data, there were times where the priors were comparable to the histograms on the plots for certain parameters, like RAJ & DECJ, but other parameters, such as F0 or F1, had either the priors or histograms be significantly larger and thus you would have a plot of either the histogram, or a straight line with no curvature for the priors with the histograms not visible. 
I can go in and add in more detail for the documentation on the `scale=False` case

aarchiba:
> Currently the histograms are not scaled at all and are not produced in density units. The priors are not normalized. If `scale=True`, then the peak of the priors are scaled to the peak of the histogram. If `scale=False`, the priors will be plotted as is, completely independent of the histogram.
> The reason why this was put in is during my testing phase with real and simulated data, there were times where the priors were comparable to the histograms on the plots for certain parameters, like RAJ & DECJ, but other parameters, such as F0 or F1, had either the priors or histograms be significantly larger and thus you would have a plot of either the histogram, or a straight line with no curvature for the priors with the histograms not visible.
> I can go in and add in more detail for the documentation on the `scale=False` case

Using density mode for the histograms seems like a good idea, since it makes the values much less dependent on the number of bins, and gives them a chance of lining up with the priors. But using the "proper" normalization for the priors, even if available, is liable to be useless - if they are broad compared to the posterior distribution, the proper normalization will make their density tiny. So I'm not arguing against the `scale` parameter, just requesting documentation (and suggesting using density for histograms so that there's a fighting chance of getting the right normalization).

devbhakt:
I have set the histograms so that they use the density mode and I have updated the `scale` parameter docstring. The pull request is ready for review and merge.

src/pint/plot_utils.py
```diff
@@ -176,3 +177,125 @@ def phaseogram_binned(
...
+    bins : int
+        Number of bins used in the histogram
+    scale : bool
+        If True, the priors will be scaled to the peak of the histograms.
+        If False, the priors will be plotted independent of the histograms.
+        In certain cases, such as broad priors, the priors or histograms
+        might not be apparent on the same plot due to one being significantly
+        larger than the other. The scaling is for visual purposes to clearly 
+        plot the priors with the samples
+    plotfile : optional
```
aarchiba, line 228: I think you should drop the `plotfile` argument. You are replicating functionality users of matplotlib must already be familiar with.

src/pint/plot_utils.py
```diff
@@ -176,3 +177,125 @@ def phaseogram_binned(
...
+                )
+        else:
+            handles, labels = axs[0].get_legend_handles_labels()
+            axs[i].set_axis_off()
+            axs[i].legend(handles, labels)
+    if plotfile is not None:
+        plt.savefig(plotfile)
+        plt.close()
+    else:
+        plt.show()
```
aarchiba, line 301: Definitely not a good idea to include a `plt.show()`; what if users want to annotate the plots?

src/pint/plot_utils.py
devbhakt: The argument has been dropped in the latest commit

devbhakt:
I think this pull request is ready for final review and merge

Issue #751, opened 2020-07-16
improve model parameter derivative tests


scottransom:
The tests of our derivative functions is now using finite differences with hard-coded step sizes.  This is prone to issues and requires manual changes to the test file test_derivative_utils.py.

I recommend changing all of the derivative tests to use automatic differentiation code.

Some possibilities are:
JAX (possibly the best / most powerful choice?):   https://github.com/google/jax
autograd (an older version of what is in JAX):  https://github.com/HIPS/autograd
numdifftools:  https://github.com/pbrod/numdifftools

aarchiba:
I think JAX and autograd use automatic differentiation based on custom data types (numbers with infinitesimal components) while numdifftools does fairly exhaustive finite-difference calculations with step size extrapolation. The interface of numdifftools is quite annoying but it should be a pretty easy drop-in addition. I used it on the triple system model.

PR #1056, opened 2021-06-22
Clean up and test derivative calculations

### LINKED ISSUE #1053: designmatrix returns the wrong units for offset
All other units in the design matrix are appropriate for derivatives of TOA (or residual) with respect to the parameter, but the offset is unitless instead of having units of time. This is actually special-cased to be wrong in the code.

The documentation in the design matrix also fails to include mention of what a comment indicates is a surprising minus sign.

The code also ignores the units returned from derivative functions, so incompatible units do not lead to an exception, and unusual units (days instead of seconds) are not converted.

### LINKED ISSUE #1054: Tests are order-dependent
Some of PINT's tests depend on being executed in a specific order. This shows in that many tests fail if run in parallel with `pytest-xdist`, or if run in random order using the pytest plugin that does that, or if the specific tests are run in isolation with `pytest tests/test_file.py::test_specific_thing`. Since the PINT test suite is slow, `pytest-xdist` would otherwise allow speedup by about the number of cores available, and running individual tests is very helpful. More, tests that depend on other tests make interpreting test failures to localize bugs quite challenging.

Tests observed to fail under such conditions:

- `tests/test_toa_selection.py::TestTOAselection::test_hash`
- `tests/test_ddk.py::TestDDK::test_J1713_deriv`

aarchiba:
The goal is to document and fix problems in PINT's derivative calculations. This includes using numdifftools to check the analytical derivatives against finite differences computed from the model.

Closes #1053 
Closes #1054 (now you can use pytest-xdist to run the test suite faster!)

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -33,3 +33,4 @@ pygments
 ipython
 pathlib2
 twine
+numdifftools
--- a/src/pint/models/stand_alone_psr_binaries/binary_orbits.py
+++ b/src/pint/models/stand_alone_psr_binaries/binary_orbits.py
@@ -50,8 +50,10 @@ class Orbit:
         try:
             func = getattr(self, "d_orbits_d_" + par)
         except AttributeError:
+
             def func():
                 return np.zeros(len(self.tt0)) * u.Unit("") / par_obj.unit
+
         result = func()
         return result
 
@@ -61,8 +63,10 @@ class Orbit:
         try:
             func = getattr(self, "d_pbprime_d_" + par)
         except AttributeError:
+
             def func():
                 return np.zeros(len(self.tt0)) * u.day / par_obj.unit
+
         result = func()
         return result
 
@@ -90,6 +94,7 @@ class OrbitPB(Orbit):
     subtracted from PBDOT when computing the derivative of orbits with respect
     to T0. It is also not included when computing ``pbdot_orbit``.
     """
+
     def __init__(self, parent, orbit_params=["PB", "PBDOT", "XPBDOT", "T0"]):
         super().__init__("orbitPB", parent, orbit_params)
 
@@ -154,6 +159,7 @@ class OrbitPB(Orbit):
 
 class OrbitFBX(Orbit):
     """Orbits expressed in terms of orbital frequency and its derivatives FB0, FB1, FB2..."""
+
     def __init__(self, parent, orbit_params=["FB0"]):
         super().__init__("orbitFBX", parent, orbit_params)
         # add the rest of FBX parameters.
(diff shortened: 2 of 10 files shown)


scottransom:
Yay!  Nice job!  This is something that we've talked about for years!

aarchiba:
> Yay! Nice job! This is something that we've talked about for years!

Tons of failures :-P A few are numdifftools being ill-behaved, but others seem to be PINT problems - for example, if the A1DOT gets the magic TEMPO-compatibility rescaling, its derivative does not. (And recall that Fitters rely on these derivatives to update model values...) I'm not sure what the issue is with FBn, but the derivative seems to be wrong by several orders of magnitude.

scottransom:
Not surprised.  Which is exactly why we've been talking about this!  Thanks for the efforts!

aarchiba:
`A1DOT` is sort of okay, it was an unexpected interaction between `numdifftools` and PINT's automatic rescaling of `A1DOT` (see #1057). It looks like probably `H3` and `FBn` do have errors, though.

scottransom:
This PR should reference issue #751

scottransom:
I'm kind of surprised by the failure of the `FBX` derivs since those use `taylor_horner_deriv()`, which I thought was good!

scottransom:
Although there doesn't seem to be any tests for `taylor_horner_deriv()` as there is for `taylor_horner()` itself...  That is definitely an omission!

aarchiba:
It seems very possible that there is a unit or offset error in the calculation of the `FBn` derivatives - they seem to be off by orders of magnitude. Or my test points are too far from `T0` and round-off is producing junk, but they don't look *that* bad.

aarchiba:
https://nanograv-pint--1056.org.readthedocs.build/en/1056/api/pint.models.timing_model.html#pint.models.timing_model.TimingModel The cleanup is meant to improve the documentation, which you can check for the PR here.

aarchiba:
Status update:

- `FBn` - with appropriate step sizes for the derivative values seem to be... close. Unclear whether there are real discrepancies yet. The pure orbital code seems much more accurate.
- `H3` - it appears that `ELL1H` uses Fourier series unnecessarily, and uses different approximations in the derivative and in the delay calculation.

I am currently working on tools for using and testing `H3`/`H4`/`VARSIGMA` in all models, not just `ELL1H`, so I might just mark the `H3` tests as `xfail` until that code sorts it out.

aarchiba:
I sorted out the numerics for everything else, and set the H3 tests to `xfail`; they need to be fixed elsewhere.

scottransom:
Can you add a test to `taylor_horner_deriv()`?  Seems like that should be explicitly tested since it is used for our derivative calcs, in general.

aarchiba:
Good idea. I did some tests and it's fine - makes the FBn work - but testing it explicitly would be much better.

scottransom:
I;ll merge when the checks pass.  Thanks!

scottransom:
Is that `black` issue legit?

aarchiba:
Forgot to mention in the meeting - you can now use `pytest-xdist` to parallelize the test suite. So I do `pytest -n 80` on my monster server and the whole test suite is over in a couple of minutes. It's *great*.

aarchiba:
Should be good to merge once CI finishes.

Issue #1037, opened 2021-04-28
Editing models in pintk


paulray:
In `pintk` you can edit the model, but how it works can be confusing and some cases don't seem to work.

Say you want to add F3 to a model...

What works is if you start `pintk`
 fresh, open the model editor, add a line that says "F3 0 1" and then click "Apply Changes" then click "Fit".  This is fine.

What does not work:

1. Start `pintk`, run a fit. Then edit the *prefit* model to add "F3 0 1", "Apply changes" then "Fit".  No F3 is added
2. Start `pintk`, run a fit. Then edit the *postfit* model to add "F3 0 1", "Apply changes" then "Fit".  No F3 is added

The files I used for these tests are here.  Rename all.txt to all.tim and F2.txt to F2.par:

[all.txt](https://github.com/nanograv/PINT/files/6393538/all.txt)
[F2.txt](https://github.com/nanograv/PINT/files/6393539/F2.txt)

champagne-cmd:
I'm not sure if this will solve the above, but one problem that is contributing to the dissonance between the model editor and the PlkWidget is that after you click 'Reset' under the graph, the PlkWidget creates an entirely new pulsar object to store in 'self.psr'. When originally created in the pintk script, PlkWidget and the model editor held the same pulsar widget in each of their respective 'self.psr' attributes. Within the reset function of the PlkWidget, however, the line `self.psr = copy.deepcopy(self.base_state.psr)` (line 664 in plk.py) creates a new pulsar object based off the base state, disconnecting the pulsar controlled by the model editor from the one controlled by the rest of pintk. I'm not sure whether it would be better to return the pulsar object after this operation and route it to the model editor or update the PlkWidget pulsar without creating a new pulsar object.

champagne-cmd:
I believe this issue can be closed @paulray ?

PR #1058, opened 2021-07-02
Model Editor Pintk

champagne-cmd:
This is a fix to issue #1037 in which adding parameters through the model editor after a fit didn't work. This was fixed by updating the postfit model in the pulsar object if it had been fitted (before only the prefit model was added to). Since the prefit model is set equal to the postfit model before a fitting after the first fit, it was erasing the modifications made through the model editor; the above resolves this.

Additionally, I noticed that the pulsar pointed to by the PlkWidget changed after pressing the "Reset" button, whereas the model editor pulsar did not. This essentially disconnected the model editor from the rest of pintk (and the TOA editor). I included a fix for this in this PR, where the new pulsar created in the PlkWidget is returned to the model editor and TOA editor when updates are called.

Finally, there is one small bug fix in random_models.py - the file was attempting to access the covariance matrix directly, which apparently is no longer supported. It now uses the 'matrix' attribute instead.

### DIFF
--- a/src/pint/pintk/paredit.py
+++ b/src/pint/pintk/paredit.py
@@ -200,7 +200,10 @@ class ParWidget(tk.Frame):
         self.set_model()
         self.call_updates()
 
-    def set_model(self):
+    def set_model(self, newpsr=None):
+        # if the pulsar was updated in pintk, update here
+        if newpsr != None:
+            self.psr = newpsr
         choice = self.choiceWidget.choice.get()
         if choice == "postfit":
             if self.psr.fitted:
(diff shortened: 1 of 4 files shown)


luojing1211:
Hi @champagne-cmd, This is a good fix. At the same time, please keep an eye on the #1026 This will make modify the model much easier.

scottransom:
As for the random models, it would be good to move the current random models code into one of the pintk files since it is quite specific for its use, and to get rid of the last ~half of it, replacing it with the `calculate_random_models` code that is now in `utils.py`.  But we can do that in another PR.

PR #1062, opened 2021-07-08
Check for bogus values of binary parameters

aarchiba:
Fitters can waste a lot of time exploring models with `M2` less than zero. This PR makes `.validate()` reject such models; this improves the behaviour of fitters and forces them to search for an actual solution.

Includes substantial documentation of what the parameters for binary models actually are.

Test cases that demonstrate the Fitter behaviour are awkward to write. I could include some test cases with par files with bogus values and confirm that an appropriate exception is raised. This would be more useful if the exception were a more specific type than ValueError - do we have an exception type for bogus parameter values?

### DIFF
--- a/src/pint/models/binary_dd.py
+++ b/src/pint/models/binary_dd.py
@@ -1,3 +1,4 @@
+"""Damour and Deruelle binary model."""
 from pint.models.parameter import floatParameter
 from pint.models.pulsar_binary import PulsarBinary
 from pint.models.stand_alone_psr_binaries.DD_model import DDmodel
@@ -5,19 +6,18 @@ from pint.models.timing_model import MissingParameter
 
 
 class BinaryDD(PulsarBinary):
-    """This is a PINT pulsar binary dd model class a subclass of PulsarBinary.
-    It is a wrapper for independent DDmodel class defined in
-    ./stand_alone_psr_binary/DD_model.py
-    All the detailed calculations are in the independent DDmodel.
-    The aim for this class is to connect the independent binary model with PINT platform
-    DDmodel special parameters:
-    A0 Aberration
-    B0 Aberration
-    GAMMA Binary Einsten delay coeeficient
-    DR Relativistic deformation of the orbit
-    DTH Relativistic deformation of the orbit
+    """Damour and Deruelle binary model.
 
-   Parameters supported:
+    This binary model is described in Damour and Deruelle ????.
+    It is a parametrized post-Keplerian model that supports additional
+    effects and post-Keplerian parameters compared to :class:`pint.models.binary_bt.BinaryBT`.
+    It does not assume General Relativity in order to infer
+    values for any post-Keplerian parameters.
+
+    The actual calculations for this are done in
+    :class:`pint.models.stand_alone_psr_binaries.DD_model.DDmodel`.
+
+    Parameters supported:
 
     .. paramtable::
         :class: pint.models.binary_dd.BinaryDD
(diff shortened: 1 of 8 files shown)


aarchiba:
Some of the doc changes - the binary parameter lists in particular - will be superseded by #1065

aarchiba:
Includes doc improvements, an example of which is here: https://nanograv-pint--1062.org.readthedocs.build/en/1062/api/pint.models.binary_bt.html

And here: https://nanograv-pint--1062.org.readthedocs.build/en/1062/api/pint.models.pulsar_binary.html#pint.models.pulsar_binary.PulsarBinary

aarchiba:
This has now removed parameters like `SINI` from models like `ELL1H` that do not support them. Unfortunately the current `get_model` doesn't actually signal an error when they occur in par files, so some tests have to be marked xfail, but one can hope that #1026 will fix this.

aarchiba:
This is ready to merge.

src/pint/models/binary_dd.py
```diff
@@ -1,23 +1,23 @@
...
-    DDmodel special parameters:
-    A0 Aberration
-    B0 Aberration
-    GAMMA Binary Einsten delay coeeficient
-    DR Relativistic deformation of the orbit
-    DTH Relativistic deformation of the orbit
+    """Damour and Deruelle binary model.
 
-   Parameters supported:
+    This binary model is described in Damour and Deruelle ????.
```
paulray, line 11: I think this is the ref:
@article{AIHPA_1986__44_3_263_0,
     author = {Damour, Thibault and Deruelle, Nathalie},
     title = {General relativistic celestial mechanics of binary systems. II. The post-newtonian timing formula},
     journal = {Annales de l'I.H.P. Physique th\'eorique},
     pages = {263--292},
     publisher = {Gauthier-Villars},
     volume = {44},
     number = {3},
     year = {1986},
     zbl = {0617.70010},
     mrnumber = {846469},
     language = {en},
     url = {http://www.numdam.org/item/AIHPA_1986__44_3_263_0/}
}

src/pint/models/binary_ell1.py
```diff
@@ -14,30 +14,38 @@
...
+    except that it removes the polar orbital parameters:
 
     Parameters supported:
 
     .. paramtable::
         :class: pint.models.binary_ell1.BinaryELL1
+
+    Note
+    ----
+    The parameter T0 is permitted but only
```
paulray, line 39: This seems like a sentence fragment

src/pint/models/binary_ell1.py
```diff
@@ -265,6 +243,8 @@ def __init__(self):
                 description="Number of harmonics for ELL1H shapiro delay.",
             )
         )
+        self.remove_param("M2")
```
luojing1211, line 242: Should we have something that can covert H3/H4 to SINI, M2?

src/pint/models/binary_ell1.py
aarchiba: And also wrong. Good catch.

src/pint/models/binary_dd.py
aarchiba: I'll just link to ADS.

src/pint/models/pulsar_binary.py
```diff
@@ -163,7 +185,42 @@ def setup(self):
...
+                f"Projected semi-major axis A1 cannot be negative ({self.A1.quantity})"
+            )
+        if self.PB.value is not None:
+            if self.PB.value <= 0:
+                raise ValueError(
+                    f"Binary period PB must be non-negative ({self.PB.quantity})"
+                )
+            if self.FB0.value is not None:
+                raise ValueError("Model cannot have values for both FB0 and PB")
+        if self.FB0.value is not None and self.FB0.value <= 0:
```
luojing1211, line 220: Should we have a check that Both `FB` and `PB` are None?

src/pint/models/binary_ell1.py
luojing1211: Ok, Let us do it in another PR

src/pint/models/pulsar_binary.py
aarchiba: I think that's checked by the function immediately below this, `check_required_params`.

src/pint/models/binary_ell1.py
aarchiba: I don't think that's a useful part of this PR, but I do think it should go into PINT. I can make those functions their own PR if that would be useful.

luojing1211:
I think this can be merged. @paulray Can I merge it ?

paulray:
Yes, merge away

luojing1211:
I will wait for the tests finish

aarchiba:
Thanks!

PR #1065, opened 2021-07-09
Automatically generate table of parameters

### LINKED ISSUE #1064: Table of timing model parameters is not automatically generated
We have a table of all the parfile parameters that PINT supports: https://nanograv-pint.readthedocs.io/en/latest/timingmodels.html

Unfortunately this is not auto-generated. Autogeneration should be possible during doc generation by importing PINT objects and poking around inside them. This may depend on an updated ModelBuilder, but fundamentally each freshly created timing model component object lists all the parameters it understands, and those include descriptions and units.

I suggest the table should contain the following columns, for each parameter:
- PINT name
- Recognized aliases if any
- Description (including conflicts with other parameters, etc.)
- Default value, if any
- Units, if any, or type (e.g. MJD, boolean, etc.)
- Fittable Y/N
- Which Components support this (clickable links!)

A few notes/suggestions:
- If we encounter Parameter objects with the same name that are different - different description, say - in different Components, the table-making code should just list them twice, and the humans should reconcile the situation.
- A number of components have parameters they don't really accept (must be zero/unset); these can probably all be removed, making it easier for users to realize the parameter won't work.
- We could alphabetize the table but ideally it would go in the same order we emit parameters in par files. 
- mask and prefix parameters are going to require some special handling - prefix parameters sometimes have units that depend on the index, and mask parameters' names can be non-obvious.

aarchiba:
We now generate a table of all the parameters PINT supports every time the docs are built. Each model component now also lists all the parameters it understands.

Also greatly flesh out documentation of models and their components, and incidentally modernize many uses of `super()`.

Closes #1064 
Will have some conflicts with #1062 because that updates some of the same documentation.

### DIFF
--- a/docs/_ext/componentlist.py
+++ b/docs/_ext/componentlist.py
@@ -1,0 +1,46 @@
+from docutils import nodes
+from docutils.parsers.rst import Directive
+from docutils.parsers.rst.directives.tables import Table
+from docutils.parsers.rst.directives import unchanged_required
+from docutils.statemachine import ViewList
+import pint.utils
+
+
+class ComponentList(Directive):
+    has_content = False
+
+    def run(self):
+        content = ViewList()
+        source = "bogus.rst"
+        content.append("Components supported by PINT:", source)
+        content.append("", source)
+
+        import pint.models.timing_model
+        d = pint.models.timing_model.Component.component_types.copy()
+        for k in sorted(d.keys()):
+            class_ = d[k]
+            full_name = f"{class_.__module__}.{class_.__name__}"
+            if hasattr(class_, "__doc__") and class_.__doc__ is not None:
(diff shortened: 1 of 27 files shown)


aarchiba:
- Big table (and extra docs): https://nanograv-pint--1065.org.readthedocs.build/en/1065/timingmodels.html#
- Individual Component table: https://nanograv-pint--1065.org.readthedocs.build/en/1065/api/pint.models.glitch.html#pint.models.glitch.Glitch

aarchiba:
The tables have a tendency to be too wide to render nicely (even in unreasonably wide browser windows) so please check whether they look okay with your browser/screen setup.

aarchiba:
If #1026 is included and works, this could list, along with all the components, which parameters ensured their inclusion.

Issue #1038, opened 2021-04-28
Cluttered output in pintk


paulray:
The text output of `pintk` is very cluttered.  Here is a simple example output from running pintk, adding "F3 0 1" to the model and clicking Fit:

[out.txt](https://github.com/nanograv/PINT/files/6393554/out.txt)

There is lots of distracting and extraneous output. But the worst is all the INFO statements *after* the results of the model fit are printed. This often causes the useful output to be scrolled off the screen.  The last thing printed should definitely be the summary of the fitted model.  Perhaps `pintk` should suppress all INFO statements and just send them to a log file or something in case you want to inspect them (e.g. in a pane of the GUI).

champagne-cmd:
@paulray it seems astropy's setup doesn't make this as easy as you would think. It is easy enough to suppress the astropy log output below a given level and to output to a log file. However, if you use the 'setLevel' function to suppress INFO statements, it suppresses them both to stdout and the log file. There doesn't appear to be an easy way to separate what is suppressed in the log file and what is suppressed to stdout without going in and changing astropy itself. A couple solutions I've thought of:

1. Put an option in pintk for turning INFO statements on and off, defaulting to off. So if a user wants more output they can opt for it but it doesn't clutter the screen for users that don't need it. This would still clutter the screen if turned on, and you wouldn't be able to see the INFO statements anywhere for runs with it turned off.

2. Go in and change stdout for pintk runs to redirect all output to a log file instead of the terminal. For output we do want going to the terminal, such as the fit results, the output can be switched back to the terminal temporarily. We might have to pass the original parameter into certain functions depending on its scope and where the redirection is needed, which could be a little awkward. But this solution may be more in line with what you were thinking above.

What do you think?

paulray:
Since Option 1 seems easy and it might be sufficient, let's try that first.

PR #1066, opened 2021-07-09
Pintk output

champagne-cmd:
Set pintk's default operation to hide INFO log statements from the terminal and added a button to the pintk interface to turn these INFO statements on and off. Addresses issue #1038 .

### DIFF
--- a/src/pint/pintk/plk.py
+++ b/src/pint/pintk/plk.py
@@ -243,7 +243,10 @@ class PlkRandomModelSelect(tk.Frame):
             widget.grid_forget()
 
     def changedRMCheckBox(self):
-        log.info("Random Models set to %d" % (self.var.get()))
+        if self.var.get() == 1:
+            log.info("Random Models turned on.")
+        else:
+            log.info("Random Models turned off.")
 
     def getRandomModel(self):
         return self.var.get()
@@ -400,6 +403,9 @@ class PlkActionsWidget(tk.Frame):
         self.writeTim_callback = None
         self.saveFig_callback = None
         self.revert_callback = None
+        self.INFO_callback = None
+
+        self.infoShowing = False  # state of INFO logs
 
         self.initPlkActions()
 
@@ -419,6 +425,11 @@ class PlkActionsWidget(tk.Frame):
         button = tk.Button(self, text="Revert", command=self.revert)
         button.grid(row=0, column=4)
 
+        self.infobutton = tk.Button(
+            self, text="Show INFO", command=self.changeInfoStatus
+        )
+        self.infobutton.grid(row=0, column=5)
+
     def setCallbacks(self, fit, reset, writePar, writeTim, revert):
         """
         Callback functions
@@ -431,6 +442,9 @@ class PlkActionsWidget(tk.Frame):
 
     def setFitButtonText(self, text):
         self.fitbutton.config(text=text)
+
+    def setInfoButtonText(self, text):
+        self.infobutton.config(text=text)
 
     def fit(self):
         if self.fit_callback is not None:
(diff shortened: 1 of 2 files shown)


aarchiba:
Would it be more useful to have a dropdown box to select the log level? So users could enable DEBUG or disable WARNING if desired? I'm picturing replacing the INFO button with a "Log level INFO" dropdown that let you pick between debug/info/warning/error/nothing?

A flashier solution would be for pintk to keep its own logs and have a log viewer window that let you pick which to show after the fact, but that's a big job and probably not worth the trouble unless there's a pre-existing log viewer widget that does exactly this.

champagne-cmd:
How does this look @aarchiba @paulray ?

src/pint/pintk/plk.py
```diff
@@ -243,7 +243,10 @@ def clear_grid(self):
             widget.grid_forget()
 
     def changedRMCheckBox(self):
-        log.info("Random Models set to %d" % (self.var.get()))
+        if self.var.get() == 1:
+            log.info("Random Models turned on.")
```
aarchiba, line 247: Should this be info or debug? It seems like the UI should be making it pretty clear.

src/pint/scripts/pintk.py
```diff
@@ -77,6 +79,34 @@ def initUI(self):
...
+            command=lambda: self.changeLogLevel("WARNING"),
+            variable=self.log_levels["WARNING"],
+        )
+        self.logMenu.add_checkbutton(
+            label="ERROR",
+            command=lambda: self.changeLogLevel("ERROR"),
+            variable=self.log_levels["ERROR"],
+        )
+        self.logMenu.add_checkbutton(
+            label="NOTSET",
```
aarchiba, line 104: What does NOTSET do? If that's all the option says, the user may be puzzled. Does this provide a way for the user to log nothing at all?

src/pint/scripts/pintk.py
```diff
@@ -77,6 +79,34 @@ def initUI(self):
         self.helpMenu.add_command(label="Plk Help", command=lambda: print(helpstring))
         self.menuBar.add_cascade(label="Help", menu=self.helpMenu)
 
+        self.logMenu = tk.Menu(self.menuBar)
```
aarchiba, line 82: I'm a little confused - does this provide a list of individually selectable checkboxes, or a list of options for the user to choose between? If I understand the logging mechanism, setting its log level to "INFO" enables info messages and also all the more important categories, so these categories aren't really selectable independently. I think tkinter's Combobox widget allows selecting between a finite list of options in a familiar way.

src/pint/scripts/pintk.py
champagne-cmd: I will need to check back on this, I thought one thing when testing it out last night, but it may essentially be a no-op and better left off the list.

src/pint/scripts/pintk.py
champagne-cmd: This is a menu of checkboxes, and how I have it set up is it essentially acts like a Combobox widget. Clicking the Log Level menu drops down a list of options with the selected option having a checkmark by it. Selecting another option unselects the current option and places the checkmark by that option. This is the minimum log level outputted - maybe changing the menu name to "Minimum Log Level" would be more understandable?

aarchiba: Why not use a Combobox?

champagne-cmd: I can change it to a combobox, I was just maintaining uniform visual, but shouldn't be a problem to switch. I can do that later tonight.

champagne-cmd: I replaced the menu, I couldn't place it in the menu bar at the top of the window so I placed it at the bottom. It hangs off the bottom of the window when expanded, if that is bothersome I am open to suggestions for other placements.

aarchiba: The normal behaviour of combo boxes is to pop up a list to choose from, which indeed can hang outside the parent window. Would tkinter's OptionMenu suit your case better?

champagne-cmd: I can definitely use that, I don't really have a preference though. I suppose it comes down to if you want the drop-down to be at the top of the window or the bottom - do you or @paulray have a preference?

aarchiba: My only objection was to reimplementing Combobox/OptionMenu by hand when code already existed.

Issue #1067, opened 2021-07-13
Fiterrors in MCMCfitter and Eventoptimize


devbhakt:
The fiterrs acquired using get_fit_keyvals are not converted properly in certain cases, such as RAJ and DECJ. The fiterrs array has the model errors in houranglesecond and arcseconds instead of hourangle and degree.

scottransom:
Do those errors have units attached, though?  If they do, then they are fine, since you can convert them to whatever you need using Astropy units' `.to()` feature.  If they units are removed, though, then I agree this is a problem!

devbhakt:
Currently, the fiterrs are only taking the value of the uncertainty and the units are removed. My pull request adds in a fix by creating an array for the model fit value units and uses them to convert the uncertainty using Astropy units' `.to()`

aarchiba:
There's a mystery here, one we might should check on with @luojing1211 - why are the units of `model.RAJ.uncertainty.unit != model.RAJ.units`? Are there other kinds of parameter where that occurs? How should the code decide what units the uncertainty should be in?

scottransom:
Yeah, I agree that this is confusing.  @devbhakt and I were just talking about it.  I think the issue is that this is basically what happens in TEMPO/TEMPO2 parfiles.  The positions are specified in full string coords, but the errors are in seconds of time (for RAJ) and arcseconds (for DEC).  Seems like we should always keep the units of the uncertainty the same as the units of the parameters themselves, and change that (if needed, like for RAJ and DECJ) only on outputting to a parfile.

PR #1068, opened 2021-07-13
Fiterrs bug in MCMCfitter and Event Optimize

devbhakt:
A fix for issue #1067


src/pint/scripts/event_optimize.py
```diff
@@ -241,15 +241,18 @@ def get_fit_keyvals(model, phs=0.0, phserr=0.1):
     fitkeys = [p for p in model.params if not getattr(model, p).frozen]
     fitvals = []
     fiterrs = []
-    for p in fitkeys:
+    fitunits = []
+    for i, p in enumerate(fitkeys):
         fitvals.append(getattr(model, p).value)
-        fiterrs.append(getattr(model, p).uncertainty_value)
+        fitunits.append(getattr(model, p).units)
+        fiterrs.append(float(((getattr(model, p).uncertainty).to(fitunits[i])).value))
```
aarchiba, line 248: Do you actually need the list of fitunits? It seems like the units are available from `model.PARAM.uncertainty.unit`?

src/pint/scripts/event_optimize.py
```diff
@@ -241,15 +241,18 @@ def get_fit_keyvals(model, phs=0.0, phserr=0.1):
...
     # The last entry in each of the fit lists is our absolute PHASE term
     # Hopefully this will become a full PINT model param soon.
     fitkeys.append("PHASE")
     fitvals.append(phs)
     fiterrs.append(phserr)
-    return fitkeys, np.asarray(fitvals), np.asarray(fiterrs)
+    fitunits.append("pulse phase")
```
aarchiba, line 254: Does the string `"pulse phase"` work as an acceptable unit?

src/pint/scripts/event_optimize.py
devbhakt: I was originally using fitunits for axis labels within the plotting priors method. I found a work around to making a list of fitunits so I can remove that list.

I tried both `model.PARAM.uncertainty.unit` and `model.params.uncertainty` and received the errors: Attribute PARAM not found in TimingModel or any Component and 'list' object has no attribute 'uncertainty'. I am not sure of those are bugs or I made a mistake in loading in the model.

src/pint/scripts/event_optimize.py
devbhakt: I had originally used that as a place holder and am working on changing that as it is not an acceptable unit. Thank you for pointing that out!

src/pint/scripts/event_optimize.py
aarchiba: Oh, uh, I was taking a shortcut. What I meant was, if your parameter was RAJ, `model.RAJ.uncertainty.unit` or `model.RAJ.units`; these are the same thing you'd get with `getattr(model, "RAJ").uncertainty.unit` or whatever.

aarchiba:
I'm sorry we have to keep manually running your CI tasks - after we merge this PR it should be automatic, but until then each change has to be approved: https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/

This is annoying - surely after the first time I approve your CI runs it could keep running them? - but in the meantime `tox` should do a fairly thorough job of running CI jobs locally. (Just run `tox` and it should run a bunch of versions with different dependencies; there's a command-line option to parallelize.)

aarchiba:
Can you write a test case that would have failed before this change but passes now? As simple and fast as possible - maybe just one MCMC iteration, we don't need actual statistics.

devbhakt:
I think this pull request can be closed without merging. I was writing the test case when I found that the original way, that event optimize and the mcmc fitter were getting the uncertainties, was providing the uncertainty values with the right corresponding units. I think the original issue was corrected by the Parameter simplification pull request #1071

aarchiba:
Would you mind putting the test case in the PR? It was a bug once and might be again if we aren't careful. For that matter if someone accidentally fixed the code, someone (ahem) might accidentally break it again.

devbhakt:
Pull Request #1071 has the test_units_consistent testcase within tests_parameters.py which covers the check for the model parameter units and model error units.
I don't think we need to do a test case with a MCMC iteration because the MCMC fitter will run to completion before and after the fix.

PR #1069, opened 2021-07-14
Adding information/metadata strings to par and tim files

dlakaplan:
Adds information like:
```
C Created: 2021-07-09T13:52:43.873908
C PINT_version: 0.8.2+297.g5ddf3207.dirty
C User: dlk
C Host: margle-2.local
C OS: macOS-10.14.6-x86_64-i386-64bit
C Comment: trying combination
```
to par and tim files via their writers.  At the moment this is just discarded on reading although that could change.  You can supply the last (optional) comment via:
```
t.write_TOA_file("combined.tim",comment="trying combination")
```

### DIFF
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -2053,6 +2053,7 @@ class TimingModel:
         self,
         start_order=["astrometry", "spindown", "dispersion"],
         last_order=["jump_delay"],
+        include_info=True,
         comment=None,
     ):
         """Represent the entire model as a parfile string.
@@ -2063,12 +2064,17 @@ class TimingModel:
             Categories to include at the beginning
         last_order : list
             Categories to include at the end
+        include_info : bool, optional
+            Include information string if True
         comment : str, optional
             Additional comment string to include in parfile
         """
         self.validate()
-        info_string = pint.utils.info_string(prefix_string="# ", comment=comment)
-        result_begin = info_string + "\n"
+        if include_info:
+            info_string = pint.utils.info_string(prefix_string="# ", comment=comment)
+            result_begin = info_string + "\n"
+        else:
+            result_begin = ""
         result_end = ""
         result_middle = ""
         cates_comp = self.get_components_by_category()
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -2053,6 +2053,7 @@ class TimingModel:
         self,
         start_order=["astrometry", "spindown", "dispersion"],
         last_order=["jump_delay"],
+        comment=None,
     ):
         """Represent the entire model as a parfile string.
 
@@ -2062,9 +2063,12 @@ class TimingModel:
             Categories to include at the beginning
         last_order : list
             Categories to include at the end
+        comment : str, optional
+            Additional comment string to include in parfile
         """
         self.validate()
-        result_begin = ""
+        info_string = pint.utils.info_string(prefix_string="# ", comment=comment)
+        result_begin = info_string + "\n"
         result_end = ""
         result_middle = ""
         cates_comp = self.get_components_by_category()
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -2053,6 +2053,7 @@ class TimingModel:
         self,
         start_order=["astrometry", "spindown", "dispersion"],
         last_order=["jump_delay"],
+        *,
         include_info=True,
         comment=None,
     ):
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1700,15 +1700,11 @@ def info_string(prefix_string="# ", comment=None):
     Adds:
 
     - Creation date
-
     - PINT version
-
-    - Username (given by the `gitpython`_ global configuration ``user.name`` if available, in addition to :func:`getpass.getuser`).
-
+    - Username (given by the `gitpython`_ global configuration ``user.name``
+      if available, in addition to :func:`getpass.getuser`).
     - Host (given by :func:`platform.node`)
-
     - OS (given by :func:`platform.platform`)
-
     - plus a user-supplied comment (if present).
 
     Parameters
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1707,7 +1707,7 @@ def info_string(prefix_string="# ", comment=None):
         # user-level git config
         c = git.GitConfigParser()
         username = c.get_value("user", option="name") + f" ({getpass.getuser()})"
-    except configparser.NoOptionError, ImportError:
+    except (configparser.NoOptionError, ImportError):
         username = getpass.getuser()
 
     s = f"""
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1675,9 +1675,9 @@ def info_string(prefix_string="# ", comment=None):
     import platform
     import datetime
 
-    s = f"Created at: {datetime.datetime.now().isoformat()}\nPINT version: {pint.__version__}\nUser: {getpass.getuser()}\nHost: {platform.node()}\nOS: {platform.platform()}"
+    s = f"Created: {datetime.datetime.now().isoformat()}\nPINT_version: {pint.__version__}\nUser: {getpass.getuser()}\nHost: {platform.node()}\nOS: {platform.platform()}"
     if comment is not None:
-        s += f"\n{comment}"
+        s += f"\nComment: {comment}"
     if (prefix_string is not None) and (len(prefix_string) > 0):
         s = "\n".join([prefix_string + x for x in s.split("\n")])
     return s
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -3,6 +3,12 @@ import re
 from contextlib import contextmanager
 from copy import deepcopy
 from io import StringIO
+import getpass
+import platform
+import datetime
+import textwrap
+import git
+import configparser
 
 import astropy.constants as const
 import astropy.coordinates as coords
@@ -12,6 +18,7 @@ import numpy as np
 import scipy.optimize.zeros as zeros
 from astropy import log
 from scipy.special import fdtrc
+import pint
 
 import pint.pulsar_ecliptic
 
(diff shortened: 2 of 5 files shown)


aarchiba:
Can you add tests that confirm that the information is correctly written to output files? Can you also verify that par and tim files with these extra lines round-trip okay (might be tested by existing tests, but what about improbable/inappropriate comment strings?

src/pint/models/timing_model.py
```diff
@@ -2062,9 +2063,12 @@ def as_parfile(
             Categories to include at the beginning
         last_order : list
             Categories to include at the end
+        comment : str, optional
+            Additional comment string to include in parfile
         """
         self.validate()
-        result_begin = ""
+        info_string = pint.utils.info_string(prefix_string="# ", comment=comment)
+        result_begin = info_string + "\n"
```
aarchiba, line 2071: It might be nice to be able to turn off the comments - I often use `as_parfile` as a human-readable format for parfiles while debugging, where that might not be so useful.

src/pint/utils.py
```diff
@@ -1667,3 +1667,17 @@ def remove_dummy_distance(c):
             "Do not know coordinate frame for %r: returning coordinates unchanged" % c
         )
         return c
+
+
+def info_string(prefix_string="# ", comment=None):
+    import pint
+    import getpass
+    import platform
+    import datetime
```
aarchiba, line 1676: Is there a reason these should be local imports rather than at the top of the file as normal?

src/pint/utils.py
```diff
@@ -1667,3 +1667,17 @@ def remove_dummy_distance(c):
             "Do not know coordinate frame for %r: returning coordinates unchanged" % c
         )
         return c
+
+
+def info_string(prefix_string="# ", comment=None):
```
aarchiba, line 1672: How does this handle comments containing newlines?

src/pint/utils.py
```diff
@@ -1667,3 +1667,17 @@ def remove_dummy_distance(c):
...
+    import pint
+    import getpass
+    import platform
+    import datetime
+
+    s = f"Created: {datetime.datetime.now().isoformat()}\nPINT_version: {pint.__version__}\nUser: {getpass.getuser()}\nHost: {platform.node()}\nOS: {platform.platform()}"
```
aarchiba, line 1678: A triple-quoted string would be much easier here. `textwrap.dedent` exists to clean up spurious indentation in such strings.

aarchiba, line 1678: Can we do better than `getpass`? Unfortunately the containers used by NANOGrav and others often mean that `getpass.getuser()` will often return something unhelpful (`jovyan` for all users on the NANOGrav notebook server). Could one drop in the git user information if that was available?

src/pint/models/timing_model.py
dlakaplan: good idea.  done

src/pint/utils.py
dlakaplan: No, that's probably just a result of me testing and then not properly integrating. fixed.

src/pint/utils.py
dlakaplan: I think it's OK (it converts `\n` to `\\n`) but I don't know if broader sanitation is needed.

src/pint/utils.py
dlakaplan: I tried to find something that was cross-platform.  Maybe this isn't the answer.  I tried to get the git info using gitpython, but I'm not sure how it will work if the global git config isn't available

dlakaplan:
I added more explicit testing in `tests/test_infostrings.py`.  It compares the contents of par and tim files with and without info strings.  And it explicitly checks the contents of the info strings.

I agree that testing for a wider range of inappropriate strings would be useful, but I am less clear on what that might be.   I am familiar with sanitizing inputs for SQL or HTML but I don't know what the specific concerns are here.

dlakaplan:
Although my implementation of the `git` username relies on the `gitpython` module, which is not standard.  So we should discuss before it is required.

src/pint/utils.py
aarchiba: Well, I guess my question is, if I supply a multiline comment, do I get:
```
C Comment: line one
line two
line three
```
or
``` 
C Comment: line one
C Comment: line two
C Comment: line three
```
or something else?

A test case to check (and specify for the curious) would be good.

aarchiba:
> Although my implementation of the `git` username relies on the `gitpython` module, which is not standard. So we should discuss before it is required.

I think for such a small application, catching ImportError and falling back to `getpass` would be fine - this means container users can provide `gitpython` to resolve the problem while other users need do nothing. I mean unless they have a user account under their cat's name or something, Y'know, hypothetically.

src/pint/models/timing_model.py
```diff
@@ -2053,6 +2053,8 @@ def as_parfile(
         self,
         start_order=["astrometry", "spindown", "dispersion"],
         last_order=["jump_delay"],
+        include_info=True,
```
aarchiba, line 2056: maybe make these keyword-only arguments?

src/pint/utils.py
```diff
@@ -3,6 +3,11 @@
 from contextlib import contextmanager
 from copy import deepcopy
 from io import StringIO
+import getpass
```
aarchiba, line 6: It'd be good to run `isort` over this file to put the imports in standard order and format.

src/pint/utils.py
```diff
@@ -1667,3 +1673,83 @@ def remove_dummy_distance(c):
...
+    comment: str, optional
+        a free-form comment string to be included if present
+    
+    Returns
+    -------
+    s : str
+        informative string
+
+
+    Notes
```
aarchiba, line 1706: Better to put examples in a standard Examples section: https://numpydoc.readthedocs.io/en/latest/format.html#examples
Also to give some sample code to produce such an example (I know that all you can supply is the comment and the comment character) like:
```
>>> info_string(prefix_string = "C ", comment="this is a \n two-line comment"
C Created: ...
```

src/pint/utils.py
```diff
@@ -1667,3 +1673,90 @@ def remove_dummy_distance(c):
             "Do not know coordinate frame for %r: returning coordinates unchanged" % c
         )
         return c
+
+
+def info_string(prefix_string="# ", comment=None):
+    """
+    Returns an informative string about the current state of PINT.
```
aarchiba, line 1680: This should be on the first line of the docstring (same line as the triple-quote) according to docstring standards.

aarchiba:
You have a merge conflict, so the CI can't run.

src/pint/utils.py
```diff
@@ -1667,3 +1673,90 @@ def remove_dummy_distance(c):
...
+
+    Adds:
```
aarchiba, line 1682: Does this render okay in the docs? `make docs` may work if you have pandoc, or if you resolve the merge conflict and CI runs you can use the link in the readthedocs job to see what this docstring looks like. (Maybe it won't show up - is this function in `__all__`?)

src/pint/utils.py
```diff
@@ -1667,3 +1673,90 @@ def remove_dummy_distance(c):
...
+    try:
+        import git
+
+        # user-level git config
+        c = git.GitConfigParser()
+        username = c.get_value("user", option="name") + f" ({getpass.getuser()})"
+    except (configparser.NoOptionError, ImportError):
+        username = getpass.getuser()
+
+    s = f"""
```
aarchiba, line 1745: Maybe do without the initial blank line?

aarchiba:
`tox -e black` is the quickest way to check the blackness of the code matches what CI wants (and get a list of failing files). I had it set up to run on every git checkin on an old machine but keep forgetting now.

src/pint/utils.py
dlakaplan: Working on this.  Is there an easy way to run `make docs` just on 1 file (the whole build takes a while).

src/pint/utils.py
dlakaplan: fixed.

src/pint/utils.py
dlakaplan: fixed.

src/pint/utils.py
aarchiba: Yeah I know it's unpleasant. `(cd docs && make html)` is the best I've got - it doesn't clear out intermediate Sphinx files so it's a little faster but it still takes a while.

src/pint/utils.py
```diff
@@ -1686,6 +1694,103 @@ def remove_dummy_distance(c):
...
+    C PINT_version: 0.8.2+311.ge351099d
+    C User: David Kaplan (dlk)
+    C Host: margle-2.local
+    C OS: macOS-10.14.6-x86_64-i386-64bit
+    C Comment: Example multi-line comment
+    C Comment: Also using a different comment character    
+
+    Notes
+    -----
+    This can be called via ``t.write_TOA_file("combined.tim",comment="trying combination")``
```
aarchiba, line 1756: I like this! Maybe sneak in a parenthesized :func:`~pint.toa.TOAs.write_TOA_file` to get a clickable link?

src/pint/utils.py
```diff
@@ -1686,6 +1694,103 @@ def remove_dummy_distance(c):
...
+    ----------
+    prefix_string: str, default='# '
+        a string to be prefixed to the output (often to designate as a
+        comment or similar)
+    comment: str, optional
+        a free-form comment string to be included if present
+    
+    Returns
+    -------
+    s : str
```
aarchiba, line 1727: No need to give it a name if there isn't an obvious one (e.g. Mc for a companion mass) - format will be okay with just type and description.

src/pint/utils.py
```diff
@@ -1686,6 +1694,103 @@ def remove_dummy_distance(c):
         return c
 
 
+def info_string(prefix_string="# ", comment=None):
+    """Returns an informative string about the current state of PINT.
+
+    Adds:
+    
+    * Creation date
```
aarchiba, line 1702: There's a Sphinx/docutils list construct, I think it's lines starting with -, not single-spaced, with a blank line above and below.

src/pint/utils.py
dlakaplan: I sort of get what you are looking for, but the syntax is eluding me, and the testing cycle takes a while.  any concrete pointers to a working example?

src/pint/utils.py
dlakaplan: I thought I was doing that.  Lines starting with "-" seem to be the same as those starting with "*".  I'm basing this off https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#bullet-lists (as much as I can interpret it).

src/pint/utils.py
aarchiba: Oh, sorry, that's markdown confusing things:
```
this looks like a live link to the function name: :func:`~pint.toa.TOAs.write_TOA_file`; the tilde shortens the link so it just looks like write_TOA_file() but still links to the docs for that function. 
```
There should be examples various places under docs/, a grep for :func: would pull them up.

src/pint/utils.py
aarchiba: Ah okay it's not working, but there definitely shouldn't be blank lines between list items. (List items that extend onto more than one line would be indented to align with the first character after the *).

aarchiba:
https://nanograv-pint--1069.org.readthedocs.build/en/1069/api/pint.utils.html#pint.utils.info_string docs

src/pint/utils.py
aarchiba: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#lists-and-quote-like-blocks this is what I followed and it seems to work

src/pint/utils.py
```diff
@@ -1686,6 +1694,144 @@ def remove_dummy_distance(c):
         return c
 
 
+def info_string(prefix_string="# ", comment=None):
+    """Returns an informative string about the current state of PINT.
+
+    Adds:
+
+    - Creation date
```
aarchiba, line 1702: If you delete the spaces between lines this should render as an actual list in Sphinx.

dlakaplan: I think it's correct as-is.  The version at https://nanograv-pint--1069.org.readthedocs.build/en/1069/api/pint.utils.html#pint.utils.info_string seems outdated (maybe because CI is failing because of coverage).  But the version I build locally looks OK.

Screen Shot 2021-07-22 at 10 34 32 AM

aarchiba: Coverage doesn't prevent other jobs, and readthedocs seems to be mishandling this. Could there be a Sphinx version difference? Unfortunately the version that people actually read online is the readthedocs version.

dlakaplan: If there is a version issue, we should perhaps pin that in the development requirements.  I had found versions of the documentation that required blank lines between list elements, but I also found the opposite. So we'll try it the other way...

aarchiba: I don't know what's going on. I'm going to assume it's fine now and readthedocs has a caching problem.

PR #1070, opened 2021-07-14
Jump by group

dlakaplan:
* Removed automatic creation of `groups` (by time gap) when creating `TOAs` objects
* Also renamed these to `clusters` (to eliminate confusion with grouping by observatory)
* Now this is done manually by:
```python
cluster = t.get_clusters(2*u.hr, add_column=False)
```
if `add_column` is True, will also add a `clusters` column to the `TOAs.table` object, along with a metadata tag (`cluster_gap`) about the gap used to create that column.
* This can then be added to the `flags` column (for jumping etc) by:
```python
for i, v in zip(np.arange(len(t)), cluster):
    t.table[i]["flags"]["toacluster"]=str(v)
```
This adds a flag `toacluster` to the table.


aarchiba:
Would it be better to do this in two steps?

- Add a flag `-group N` to each TOA, recording what group they are in
- Add `JUMP -group N` for each N

This puts in the same set of JUMPs but if the tim file is written out and then read back in, the groups will be preserved no matter the settings upon read. Additional TOAs won't renumber all the groups, either, though when merging tim files the group numbers will collide if one isn't careful.

dlakaplan:
The `groups` that I was referencing were those automatically defined by PINT due to gaps in the TOA record.  I was going to do an approach like the one you suggest (defining my own groupings and then adding flags) but I saw that the TOA table already had the `groups` column which accomplished the same goal.  Since that seemed to be baked into PINT I decided to use that, which of course doesn't preclude somebody from JUMPing based on other flags.

I did make it more flexible so that the gap used to define the `groups` column was adjustable by the user.  

I don't know what other tasks use those groups, so for the moment I wouldn't remove them (just use them).

scottransom:
That grouping functionality was added by @clp3ef for her APT code, and it seemed like a potentially useful addition for other people (I can totally see using it for other reasons).  I like that you made it mode flexible.

aarchiba:
My worry is that because the group number is not recorded in the tim file, it is very easy to cause a re-numbering of the groups with apparently minor changes to the tim file - for example, excising a stray TOA or dropping an epoch would renumber all following groups, making the par file invalid. So if you want to do this, why not make a tim file with the group number recorded in the flags? That you can mangle pretty freely and the JUMPs will still apply to the same TOAs.

dlakaplan:
@scottransom : I can see why, since that was my rationale as well, although it was (I think) a little odd that the grouping was done by default and without notification.

aarchiba:
> That grouping functionality was added by @clp3ef for her APT code, and it seemed like a potentially useful addition for other people (I can totally see using it for other reasons). I like that you made it mode flexible.

`toas.get_groups()` allows arbitrary gaps already, though if the gap is exactly two hours it will return the appropriate column of the table (this code breaks this function), and that is what Patrick @poneill129  is using to define his groups - I think he's using five-day blocks. (He also gave each of them their own jump at some point with a simple loop.)

dlakaplan:
What do you mean by:
> (this code breaks this function)
`toas.get_groups()` is automatically called on creation of a `TOAs` object.  I don't think I broke it.  All I did was allow the gap to be specified upon creation as well as by hand.

I hadn't been aware that only the 2h group was supposed to be returned as part of the toa table.  I have changed it so that the default grouping (from ingest) is now saved as part of the table metadata.  Then `get_groups()` behaves appropriately, I believe.

scottransom:
I never knew that this code was always run when returning a TOAs object!   Oops.  This seems like something that a user should have to explicitly run themselves afterwards.

luojing1211:
Just want to get it clarified, the group we are talking about here is not the astropy table.group() right?  The TOA table has an internal grouping by the observatories, that is why there would be reindexing in the parfile, for computing the observatory location and clock correction.

dlakaplan:
@scottransom : interesting.  I have changed the behavior to make the grouping a separate step:
```python
```
will have no groups present, but after:
```python
groups = t.get_groups(2*u.hr, add_column=True)
```
they will be both returned and added to the TOA Table, while:
```python
groups = t.get_groups(2*u.hr, add_column=False)
```
means they will just be returned but not added to the Table.

scottransom:
I think that looks good, @dlakaplan.   I think this will break APT, so @clp3ef will need to update it and explicitly call the grouping.  And as for @luojing1211, this is a different grouping -- it is a way of semi-intelligently identifying TOAs that are grouped together in time (like from a single observation, or a consecutive but different observations).

dlakaplan:
If grouping by time gap was not broadly intended to be automatically run always, then I would be more inclined to agree with @aarchiba that jumping by the `groups` column may not make sense.  Perhaps something to discuss next week.  Better might be to just have a quick utility to do grouping and translate that into flags (which offer similar functionality for jumping).

aarchiba:
> If grouping by time gap was not broadly intended to be automatically run always, then I would be more inclined to agree with @aarchiba that jumping by the `groups` column may not make sense. Perhaps something to discuss next week. Better might be to just have a quick utility to do grouping and translate that into flags (which offer similar functionality for jumping).

The existing `get_groups` produces the data necessary to group TOAs appropriately, so all that is needed is a convenient way to add a flag to some or all TOAs. Unfortunately this is more annoying than one might wish:
```python
for group_num, flags in zip(toas.get_groups(), toas.table['flags']):
    flags['my_groups'] = str(group_num) 
```
It's a shame we can't turn all the flags into columns in `toas.table`, but Astropy tables do not do a good job of supporting columns with missing entries. (Or arbitrary-length strings.) Something might be possible using the indexing operator on TOAs objects, or we have a few awkward flag-manipulation functions already (https://nanograv-pint.readthedocs.io/en/latest/api/pint.toa.html#pint.toa.TOAs.get_flag_value can't be used for indexing directly or anything). Manipulations of the flags generally require looping and mutating objects, and don't make it especially clear that any value other than a string is problematic. Some of this could be more firmly enforced.

dlakaplan:
So what I am working on now is a quick function `TOAs.add_flags()` which can add flags (not hard to do otherwise, but useful to provide as an example).  e.g., 
```python
group = t.get_groups(2*u.hr, add_column=False)
group = [str(x) for x in group]
t.add_flags(np.arange(len(t)), "toagroup", group)
t.add_flags(np.arange(len(t)), "toagroup", group, overwrite=True)
t.add_flags(np.arange(len(t)), "comment", 'good')
t.add_flags(200, "comment", "bad", overwrite=True)
```
so you need to specify the index/indices for the TOAs to flag, the name of the flag, and the value(s).  Existing flags are left untouched by default.   Does this make sense?  A jump can then be added by something like:
```python
...
        name="JUMP", key="-toagroup", value=0.0, key_value=35, units=u.s, frozen=False
    )
```
If this sort of usage makes sense I can clean this up (remove the JUMP by `groups` I added yesterday, make the `get_groups()` function not add a column to the TOA table by default, update testing).

aarchiba:
It's worth keeping in mind the existence of `model.add_param_and_flags(toas)` - this currently takes a subset of TOAs and adds both a JUMP parameter and a `-gui_jump N` flag to those TOAs. It might as well allow specification of the flag and value.

I think I prefer the indexing approach I propose in #1073 to `add_groups` because it allows reading and writing flags in a non-surprising way that integrates with existing TOA-subset selection tools and mimics the behaviour of Astropy tables.

dlakaplan:
I agree that #1073 would do a lot of this.  It might be that they could exist in parallel (one need not preclude the other).

aarchiba:
> I agree that #1073 would do a lot of this. It might be that they could exist in parallel (one need not preclude the other).

To quote The Zen of Python, `There should be one-- and preferably only one --obvious way to do it.`

src/pint/toa.py
```diff
@@ -947,6 +948,25 @@ def get_freq_array(bfv, ntoas):
 
 
 def _group_by_gaps(t, gap):
+    """
+    A utility function to group times according to gap-less stretches.
```
aarchiba, line 952: First line of a docstring should go on the first line so it shows up as a short description in the docs. But this function is private so a docstring is less important.

src/pint/toa.py
```diff
@@ -1438,6 +1453,32 @@ def get_flags(self):
         """Return a numpy array of the TOA flags."""
         return self.table["flags"]
 
+    def add_flags(self, indices, flag, flag_values, overwrite=False):
```
aarchiba, line 1456: This replicates functionality in #1074 and adds to the complexity of TOAs objects.

tests/test_flagging_grouping.py
```diff
@@ -0,0 +1,106 @@
...
+        self.tim = tim
+        self.m = mb.get_model(self.par)
+        self.t = toa.get_TOAs(
+            self.tim, ephem="DE405", planets=False, include_bipm=False
+        )
+
+
+@pytest.fixture
+def setup_NGC6440E():
+    os.chdir(datadir)
```
aarchiba, line 31: Please do not change directories in a test. This is modifying global state and can break things if a subset of tests are run. Just use path names `os.path.join(datadir, "file.par")`.

tests/test_flagging_grouping.py
```diff
@@ -0,0 +1,106 @@
...
+    return SimpleSetup("NGC6440E.par", "NGC6440E.tim")
+
+
+def test_add_flags(setup_NGC6440E):
+    """
+    test adding a new flag
```
aarchiba, line 37: Docstrings in test functions are not displayed anywhere but their names are; please put the same information in the name (no one is calling these functions, so there's no problem if they're long to type).

tests/test_flagging_grouping.py
```diff
@@ -0,0 +1,106 @@
...
+    cp_copy = m_copy.components["PhaseJump"]
+    par_copy = p.maskParameter(
+        name="JUMP", key="-toagroup", value=0.2, key_value=41, units=u.s
+    )
+    # this should be identical to the group above
+    cp_copy.add_param(par_copy, setup=True)
+    assert (
+        np.array(cp.JUMP1.select_toa_mask(setup_NGC6440E.t))
+        == np.array(cp_copy.JUMP1.select_toa_mask(setup_NGC6440E.t))
+    ).all(), (
```
aarchiba, line 96: I think you get better error messages if you compare the `set()`s rather than making arrays. And `pytest` displays a lot of detail, more if you don't include a message with the assertion, so it's usually better not to use the message part of an assertion.

tests/test_flagging_grouping.py
```diff
@@ -0,0 +1,106 @@
...
+        "%s vs. %s"
+        % (
+            cp.JUMP1.select_toa_mask(setup_NGC6440E.t),
+            cp_copy.JUMP1.select_toa_mask(setup_NGC6440E.t),
+        )
+    )
+
+
+if __name__ == "__main__":
```
aarchiba, line 105: This has no effect, ever.

src/pint/toa.py
```diff
@@ -1491,31 +1532,41 @@ def get_dm_errors(self):
...
 
-    def get_groups(self, gap_limit=None):
+    def get_groups(self, gap_limit=2 * u.h, add_column=False):
         """Flag toas within gap limit (default 2h = 0.0833d) of each other as the same group.
 
         Groups can be larger than the gap limit - if toas are separated by a gap larger than
         the gap limit, a new group starts and continues until another such gap is found.
 
-        Groups with a two-hour spacing are pre-computed when the TOAs object is constructed,
+        Groups with a two-hour spacing (default) are pre-computed when the TOAs object is constructed,
```
aarchiba, line 1541: I think this is no longer true with this change? Perhaps an updated summary of the caching functionality would be more appropriate? (If a matching column exists it will be used; `add_column` allows construction of such a column.)

src/pint/toa.py
```diff
@@ -1491,31 +1532,41 @@ def get_dm_errors(self):
...
+            ("groups" not in self.table.colnames)
+            or ("group_gap" not in self.table.meta)
+            or (gap_limit != self.table.meta["group_gap"])
+        ):
+            groups = _group_by_gaps(
+                self.get_mjds().to_value(u.d), gap_limit.to_value(u.d)
+            )
+            if add_column:
+                self.table.add_column(groups, name="groups")
+                self.table.meta["group_gap"] = gap_limit
```
aarchiba, line 1567: Does this metadata survive table subset selection (for example due to selecting a time range or TOA excision)? Should it?

tests/test_flagging_clustering.py
```diff
@@ -0,0 +1,74 @@
...
+    cp_copy = m_copy.components["PhaseJump"]
+    par_copy = p.maskParameter(
+        name="JUMP", key="-toacluster", value=0.2, key_value=41, units=u.s
+    )
+    # this should be identical to the cluster above
+    cp_copy.add_param(par_copy, setup=True)
+    assert (
+        np.array(cp.JUMP1.select_toa_mask(setup_NGC6440E.t))
+        == np.array(cp_copy.JUMP1.select_toa_mask(setup_NGC6440E.t))
+    ).all(), (
```
aarchiba, line 64: Just comparing the sets should give better error messages from pytest, especially if you don't try to supply a message.

tests/test_flagging_clustering.py
```diff
@@ -0,0 +1,74 @@
...
+        self.tim = tim
+        self.m = mb.get_model(self.par)
+        self.t = toa.get_TOAs(
+            self.tim, ephem="DE405", planets=False, include_bipm=False
+        )
+
+
+@pytest.fixture
+def setup_NGC6440E():
+    os.chdir(datadir)
```
aarchiba, line 31: Do not change directory in tests, just use `os.path.join(datadir, filename)`; changing directories changes global state.

dlakaplan: sure, although I was just copying some other tests.

aarchiba: Yes, we have a lot of bad tests. I'm trying to avoid adding new bad ones.

dlakaplan: Sure, I'm not objecting.  I'm just pointing out that a more systematic change to tests could be worthwhile.

aarchiba: I thought about this when I came back to active work on PINT, but the large number of tests and the difficulty of understanding what they are testing makes it an extremely daunting task to even convert them to a modern style, let alone to understandable unit, integration, or comparison tests. So I have been following the Scouts' rule of leaving everywhere better than I found it - when I have to modify some code I clean it up along the way.

tests/test_flagging_clustering.py
aarchiba: I was suggesting that `assert set(cp.JUMP1.select_toa_mask(setup_NGC6440E.t)) == set(cp_copy.JUMP1.select_toa_mask(setup_NGC6440E.t))` would be more readable and also provide better reports if it failed. (`pytest`, when comparing sets, lists elements that are present in one and not the other.)

dlakaplan: OK, I didn't know that.

aarchiba:
Conflicts :-P on the plus side when you fix them there shouldn't be red coverage x's any more.

PR #1071, opened 2021-07-14
Parameter simplification

aarchiba:
This is meant to clean up Parameter objects, making better use of inheritance and improving their documentation. User-visible changes are minor, mostly involving more strict input type validation for parameters. Implementation is much simplified and the documentation is better. There are also tests of pickle-ability, though these are currently marked as expected failures for prefix and mask parameters.

We have many more suggestions in the discussion for improving Parameter objects but these are probably best deferred to later PRs so that this can go in and be used by (for example) #1072, which made substantial changes to maskParameters.

### DIFF
--- a/src/pint/models/parameter.py
+++ b/src/pint/models/parameter.py
@@ -22,14 +22,17 @@ parameters PINT understands.
 
 """
 import numbers
+from warnings import warn
 
 import astropy.time as time
 import astropy.units as u
 import numpy as np
 from astropy import log
 from astropy.coordinates.angles import Angle
+
 from pint import pint_units
 from pint.models import priors
+from pint.observatory import get_observatory
 from pint.pulsar_mjd import (
     Time,
     data2longdouble,
@@ -41,7 +44,6 @@ from pint.pulsar_mjd import (
 )
 from pint.toa_select import TOASelect
 from pint.utils import split_prefixed_name
-from pint.observatory import get_observatory
 
 
 class Parameter:
@@ -94,12 +96,6 @@ class Parameter:
     continuous : bool, optional
         A flag specifying whether derivatives with respect to this
         parameter exist.
-    print_quantity : method, optional
-        A function that converts the internal value to a string for output.
-    set_quantity : method, optional
-        A function that sets the quantity property
-    get_value:
-        A function that get purely value from quantity attribute
     use_alias : str or None
         Alias to use on write; normally whatever alias was in the par
         file it was read from
@@ -120,21 +116,12 @@ class Parameter:
         frozen=True,
         aliases=None,
         continuous=True,
-        set_quantity=lambda x: x,
-        get_value=lambda x: x,
         prior=priors.Prior(priors.UniformUnboundedRV()),
-        set_uncertainty=fortran_float,
         use_alias=None,
     ):
 
         self.name = name  # name of the parameter
         self.units = units  # Default unit
-        self.set_quantity = set_quantity
-        # Method to get value
-        self.get_value = get_value
-        # Method to get uncertainty from input
-        self.set_uncertainty = set_uncertainty
-        self.from_parfile_line = self.from_parfile_line_regular
         self.quantity = value  # The value of parameter, internal storage
         self.prior = prior
 
@@ -151,21 +138,54 @@ class Parameter:
         self.use_alias = use_alias
 
     @property
-    def prior(self):
-        """prior distribution for this parameter.
+    def quantity(self):
+        """Value including units (if appropriate)."""
+        return self._quantity
 
-        This should be a :class:`~pint.models.priors.Prior` object describing the prior
-        distribution of the quantity, for use in Bayesian fitting.
+    @quantity.setter
+    def quantity(self, val):
+        """General wrapper method to set .quantity.
+
+        For different type of
+        parameters, the setter method is stored at ._set_quantity attribute.
         """
-        return self._prior
+        if val is None:
+            if hasattr(self, "quantity") and self.quantity is not None:
+                raise ValueError("Setting an existing value to None is not allowed.")
+            else:
+                self._quantity = val
+                return
+        self._quantity = self._set_quantity(val)
 
-    @prior.setter
-    def prior(self, p):
-        if not isinstance(p, priors.Prior):
-            raise ValueError("prior must be an instance of Prior()")
-        self._prior = p
+    @property
+    def value(self):
+        """Return the value (without units) of a parameter.
 
-    # Setup units property
+        This value is assumed to be in units of ``self.units``. Upon setting, a
+        a :class:`~astropy.units.Quantity` can be provided, which will be converted
+        to ``self.units``.
+        """
+        if self._quantity is None:
+            return None
+        else:
+            return self._get_value(self._quantity)
+
+    @value.setter
+    def value(self, val):
+        if val is None:
+            if (
+                not isinstance(self.quantity, (str, bool))
+                and self._quantity is not None
+            ):
+                raise ValueError(
+                    "This parameter value is number convertible. "
+                    "Setting .value to None will lost the "
+                    "parameter value."
+                )
+            else:
+                self.value = val
+        self._quantity = self._set_quantity(val)
+
     @property
     def units(self):
         """Units associated with this parameter.
@@ -215,26 +235,90 @@ class Parameter:
             # Change uncertainty unit to new unit
             self.uncertainty = self.uncertainty.to(self._units)
 
-    # Setup quantity property
     @property
-    def quantity(self):
-        """Value including units (if appropriate)."""
-        return self._quantity
+    def uncertainty(self):
+        """Parameter uncertainty value with units."""
+        return self._uncertainty
 
-    @quantity.setter
-    def quantity(self, val):
-        """General wrapper method to set .quantity.
-
-        For different type of
-        parameters, the setter method is stored at .set_quantity attribute.
-        """
+    @uncertainty.setter
+    def uncertainty(self, val):
         if val is None:
-            if hasattr(self, "quantity") and self.quantity is not None:
-                raise ValueError("Setting an existing value to None is not allowed.")
+            if hasattr(self, "uncertainty") and self.uncertainty is not None:
+                raise ValueError("Setting an exist uncertainty to None is not allowed.")
             else:
-                self._quantity = val
+                self._uncertainty = self._uncertainty_value = None
                 return
-        self._quantity = self.set_quantity(val)
+        val = self._set_uncertainty(val)
+
+        if not val >= 0:
+            raise ValueError(f"Uncertainty cannot be set to negative {val}")
+            # self.uncertainty_value = np.abs(self.uncertainty_value)
+
+        self._uncertainty = val.to(self.units)
+
+    @property
+    def uncertainty_value(self):
+        """Return a pure value from .uncertainty.
+
+        This will be interpreted as having units ``self.units``.
+        """
+        # FIXME: is this worth having when p.uncertainty.value does the same thing?
+        if self._uncertainty is None:
+            return None
+        else:
+            return self._get_value(self._uncertainty)
+
+    @uncertainty_value.setter
+    def uncertainty_value(self, val):
+        if val is None:
+            if (
+                not isinstance(self.uncertainty, (str, bool))
+                and self._uncertainty_value is not None
+            ):
+                log.warning(
+                    "This parameter has uncertainty value. "
+                    "Change it to None will lost information."
+                )
+            else:
+                self.uncertainty_value = val
+        self._uncertainty = self._set_uncertainty(val)
+
+    def _get_value(self, quan):
+        """Extract a raw value from internal representation.
+
+        Generally just returns the internal representation, but some subclasses
+        may override this to, say, convert to the correct units and then discard
+        them.
+        """
+        return quan
+
+    def _set_quantity(self, val):
+        """Convert value to internal representation.
+
+        Subclasses may override this to, for example, parse Fortran-format strings into
+        long doubles.
+        """
+        return val
+
+    def _set_uncertainty(self, val):
+        """Convert value to internal representation for use in uncertainty."""
+        if val != 0:
+            raise NotImplementedError()
+
+    @property
+    def prior(self):
+        """prior distribution for this parameter.
+
+        This should be a :class:`~pint.models.priors.Prior` object describing the prior
+        distribution of the quantity, for use in Bayesian fitting.
+        """
+        return self._prior
+
+    @prior.setter
+    def prior(self, p):
+        if not isinstance(p, priors.Prior):
+            raise ValueError("prior must be an instance of Prior()")
+        self._prior = p
 
     def prior_pdf(self, value=None, logpdf=False):
         """Return the prior probability density.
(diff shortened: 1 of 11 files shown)


aarchiba:
Additional changes: 

- Make all floatParameters store long double for simplicity.
- Don't let `.value` accept units(?)
- Do the 1e-12 rescaling only upon loading from a par file line, not on any assignment (?)

luojing1211:
One idea I have is to add an attribute, called "indexed", to the Parameter class and move the index-related methods and attribute to the base parameter class as well.  This way we can remove the prefixParameter, which has a relatively complicated structure.  This will require some changes in our API, but rather a simplification I believe, since the prefixParameter need to give a type of internal parameter type. For instance, DMX _0001 is a prefixParameter but internally it is a floatParameter. However, DMXR1_0001 has the MJDParameter as the internal storage.

aarchiba:
I think if we're going to change the API we should think hard about how prefix parameters work. The fact that there has to be at least one bogus unset one upon creation to allow the model builder stuff to work is awkward for users (see various code that checks for and deletes this bogus parameter). The fact that it's a pain in the neck to create a new one is annoying too.

The user-facing API I'd like to see for prefix families of parameters is:

- An existing prefix parameter just looks like any other parameter (e.g. `model.F1` behaves exactly like any other floatParameter).
- Reference to a name that should be a prefix parameter that doesn't yet exist creates it: if I access `model.F4` it should bring `model.F4` into existence, with appropriate units but unset. 
- There is an easy way to find out what prefix parameter families are supported, and to ask questions about them ("what frequency derivative parameters are set?" "what should I call the 17th parameter in this family?" "what number does DMX_0018 represent?") Presumably each family should have a sort of family name, I guess the prefix, through which it might be easiest to access this. Perhaps something like `model.prefix_family("DMX_").make_name(18)`? Each Component would then have some (or no) prefix families as well as components.

To me, this suggests that there should be separate machinery for handling families of parameters, but that the parameters that machinery creates should be perfectly normal parameters of familiar types.

aarchiba:
A broader problem with the Parameter API is that Parameters should be associated with types, not with instances - every AstrometryEcliptic object is going to have a PX parameter, and I shouldn't have to make a bogus invalid instance to find that out. I realize this may take a lot of work to fix, and it maybe shouldn't be a priority. But if this information were attached to the type, it might be possible to reduce the headaches associated with knowing when to run `.setup()` - I have frequently made changes to timing models only for them to have no effect until `.setup()` is run. There's only so far we can take this because it's probably useful for users and Fitters to be able to modify parameter values, so we can't avoid a need to validate new sets of parameters.

aarchiba:
Also if we're changing the API, can we name these objects in a PEP8-compliant way? `floatParameter` -> `FloatParameter`, for example.

luojing1211:
> I think if we're going to change the API we should think hard about how prefix parameters work. The fact that there has to be at least one bogus unset one upon creation to allow the model builder stuff to work is awkward for users (see various code that checks for and deletes this bogus parameter). The fact that it's a pain in the neck to create a new one is annoying too.
> 
> The user-facing API I'd like to see for prefix families of parameters is:
> 
> * An existing prefix parameter just looks like any other parameter (e.g. `model.F1` behaves exactly like any other floatParameter).
      If all parameter classes can be `indexed`, the existing prefix parameter will just behave like the other parameter type. The prefixParameter only has the extra functionality to handle the indexes.  
   
> * Reference to a name that should be a prefix parameter that doesn't yet exist creates it: if I access `model.F4` it should bring `model.F4` into existence, with appropriate units but unset.
     The current Prefix parameter has half of the functionality in the `prefixParameter.new_param()` with the correct unit and the description. However, just calling `model.F4` to make the parameter is something we should do in the TimingModel class. 

> * There is an easy way to find out what prefix parameter families are supported, and to ask questions about them ("what frequency derivative parameters are set?" "what should I call the 17th parameter in this family?" "what number does DMX_0018 represent?") Presumably each family should have a sort of family name, I guess the prefix, through which it might be easiest to access this. Perhaps something like `model.prefix_family("DMX_").make_name(18)`? Each Component would then have some (or no) prefix families as well as components. Since we already have a customized `__getattr__`, searches for parameters that don't find what they're looking for can check with the `prefix_family` objects to see if they want to make one.
    These features also seem to belong to the TimingModel class API. 

> 
> To me, this suggests that there should be separate machinery for handling families of parameters, but that the parameters that machinery creates should be perfectly normal parameters of familiar types.
   Are you saying to make a class like `ParameterFamiliy` to handle this part of the code? Maybe have a parameter manager class to handle the prefixParameter and mask parameters?

luojing1211:
> A broader problem with the Parameter API is that Parameters should be associated with types, not with instances - every AstrometryEcliptic object is going to have a PX parameter, and I shouldn't have to make a bogus invalid instance to find that out. I realize this may take a lot of work to fix, and it maybe shouldn't be a priority. But if this information were attached to the type, it might be possible to reduce the headaches associated with knowing when to run `.setup()` - I have frequently made changes to timing models only for them to have no effect until `.setup()` is run. There's only so far we can take this because it's probably useful for users and Fitters to be able to modify parameter values, so we can't avoid a need to validate new sets of parameters.

Can you provide some code examples for this idea?

luojing1211:
> Also if we're changing the API, can we name these objects in a PEP8-compliant way? `floatParameter` -> `FloatParameter`, for example.

Of course, this is in the plan.

aarchiba:
> > * Reference to a name that should be a prefix parameter that doesn't yet exist creates it: if I access `model.F4` it should bring `model.F4` into existence, with appropriate units but unset.
> 
> The current Prefix parameter has half of the functionality in the `prefixParameter.new_param()` with the correct unit and the description. However, just calling `model.F4` to make the parameter is something we should do in the TimingModel class.
> 
> > * There is an easy way to find out what prefix parameter families are supported, and to ask questions about them ("what frequency derivative parameters are set?" "what should I call the 17th parameter in this family?" "what number does DMX_0018 represent?") Presumably each family should have a sort of family name, I guess the prefix, through which it might be easiest to access this. Perhaps something like `model.prefix_family("DMX_").make_name(18)`? Each Component would then have some (or no) prefix families as well as components. Since we already have a customized `__getattr__`, searches for parameters that don't find what they're looking for can check with the `prefix_family` objects to see if they want to make one.
> 
> These features also seem to belong to the TimingModel class API.
> 
> > To me, this suggests that there should be separate machinery for handling families of parameters, but that the parameters that machinery creates should be perfectly normal parameters of familiar types.
> 
> Are you saying to make a class like `ParameterFamiliy` to handle this part of the code? Maybe have a parameter manager class to handle the prefixParameter and mask parameters?

I would prefer to minimize additions to the TimingModel class API - it already supports a lot of different things, plus you can call all sorts of functions on them that are actually implemented and documented somewhere else.

I would suggest creating a PrefixFamily object and a MaskFamily class. Components would then have dicts `.prefix_families` and `.mask_families` to access the families that they support; a TimingModel object would then acquire these dictionaries as properties, implemented by forwarding to the components.

A PrefixFamily  object would support:

- List all parameters of the family that are present in the model. (Preferably as a dict so you could easily pull out both name and parameter object.)
- Add a new parameter by number to the model; return name and/or object.
- Easily map numbers to names and names to numbers for parameters whether they exist or not.

A MaskFamily object would support (we should check these against the needs of JUMPs in #1072):

- List all parameters of the family that are present in the model.
- Add a new parameter by flag and value.
- (maybe) Easily compare flag/value pairs with existing JUMPs (e.g. "do any jumps select on `-fish`?" "is there already a flag for `-fish carp`?)
- (maybe) Easily compare selected TOA subsets with existing JUMPs (e.g. "is there already a JUMP affecting exactly these TOAs?")
- (maybe) Given a flag and value and a set of TOAs, add the flag and value and make a new parameter affecting them

TimingModel objects would then provide the new `.prefix_family` and `.model_family` dicts, and would forward otherwise missed `__getattr__` to these families. I don't think we'd add anything else to the TimingModel or Component API, and we might be able to remove things - `new_param` would go away, to be handled by the appropriate family, for example, and `add_param` would be for use of Component implementors only, not general PINT users.

aarchiba:
> > A broader problem with the Parameter API is that Parameters should be associated with types, not with instances - every AstrometryEcliptic object is going to have a PX parameter, and I shouldn't have to make a bogus invalid instance to find that out. I realize this may take a lot of work to fix, and it maybe shouldn't be a priority. But if this information were attached to the type, it might be possible to reduce the headaches associated with knowing when to run `.setup()` - I have frequently made changes to timing models only for them to have no effect until `.setup()` is run. There's only so far we can take this because it's probably useful for users and Fitters to be able to modify parameter values, so we can't avoid a need to validate new sets of parameters.
> 
> Can you provide some code examples for this idea?

Right now, a Component declares what Parameters it has in its `__init__` method, so there's no way to know without instantiating them. But we could give the class itself a list of supported parameters and families, so that by inspecting the class we'd know what parameters every instance supported. You could then make instance `__init__` methods accept parameters with values. Something like:

```python
class AstrometryEcliptic(DelayComponent):
    parameters = [ AngleParameter("ELAT", ...), AngleParameter("ELONG", ...), FloatParameter("PX", ...)]
    def __init__(self, ELAT, ELONG, PX=None):
        super().__init__(self)
        ...
```
The `.setup()` method as it currently stands doesn't do much for most Components but for JUMPs it builds convenience tables of all the JUMPs (this is something we'd let the MaskFamily object keep track of) or for binary models it constructs the internal object. It's generally very unclear when `setup()` needs to be run and when it doesn't, so if we could move as much of its function into `__init__` or elsewhere that would be a great help.

aarchiba:
Worth adding: a test of whether Parameter objects can be pickled, and by extension full TimingModel objects. (#1076)

luojing1211:
To have a family is a good idea. The API between the parameter families and timing component is the key to this development.

aarchiba:
I think many of these changes are out of scope for this PR, both because they need discussion and because they depend on the new model builder, and I want the improvements already in this patch to go in soon, because they overlap with my other PRs. Perhaps a discussion of parameter families can be moved to an issue?

luojing1211:
> I think many of these changes are out of scope for this PR, both because they need discussion and because they depend on the new model builder, and I want the improvements already in this patch to go in soon, because they overlap with my other PRs. Perhaps a discussion of parameter families can be moved to an issue?

Agree. It is a bigger discussion.

aarchiba:
The pickleability issue for maskParameters is resolved in #1072. For prefixParameters it's more complicated as they are used to customize how parameter n is described and what its units are. This can still be replaced by inheritance but it would require creating a subclass for parameter families that do things differently. And we're talking about rewriting this code anyway. So I'm inclined to leave them as xfail even though the bug #1076 depends on fixing these.

aarchiba:
I'm inclined to say, let's review and merge this, and save further changes for another PR. These changes are starting to intersect with the other outstanding PRs.

src/pint/models/parameter.py
```diff
@@ -146,18 +138,60 @@ def __init__(
...
+    @value.setter
+    def value(self, val):
+        if val is None:
+            if (
+                not isinstance(self.quantity, (str, bool))
+                and self._quantity is not None
+            ):
+                raise ValueError(
+                    "This parameter value is number convertible. "
+                    "Setting .value to None will lost the "
```
scottransom, line 182: s/lost/lose/

src/pint/models/parameter.py
```diff
@@ -201,124 +235,41 @@ def units(self, unt):
...
-        """General wrapper setter for uncertainty.
-
-        The setting method is stored at .set_uncertainty attribute
-        """
         if val is None:
             if hasattr(self, "uncertainty") and self.uncertainty is not None:
-                raise ValueError(
-                    "Setting an exist uncertainty to None is not" " allowed."
-                )
+                raise ValueError("Setting an exist uncertainty to None is not allowed.")
```
scottransom, line 247: s/exist/existing/

src/pint/models/parameter.py
```diff
@@ -201,124 +235,41 @@ def units(self, unt):
...
+                self._uncertainty = self._uncertainty_value = None
                 return
-        self._uncertainty = self.set_uncertainty(val)
+        val = self._set_uncertainty(val)
 
-        # This is avoiding negtive unvertainty input.
-        if self._uncertainty is not None and self.uncertainty_value < 0:
-            self.uncertainty_value = np.abs(self.uncertainty_value)
+        if not val >= 0:
+            raise ValueError(f"Uncertainty cannot be set to negative {val}")
```
scottransom, line 254: Maybe"  "Uncertainties cannot be negative"?

src/pint/models/parameter.py
```diff
@@ -437,30 +441,42 @@ def from_parfile_line_regular(self, line):
...
         return True
 
+    def add_alias(self, alias):
+        """Add a name to the list of aliases for this parameter."""
+        self.aliases.append(alias)
+
     def name_matches(self, name):
         """Whether or not the parameter name matches the provided name"""
         return (name == self.name.upper()) or (
             name in map(lambda x: x.upper(), self.aliases)
```
scottransom, line 454: Might that be worth changing to a list comp?  Those seem way more readable to me than mapping with lambdas....

src/pint/models/parameter.py
```diff
@@ -480,11 +496,12 @@ class floatParameter(Parameter):
...
-    Example::
-        >>> from parameter import floatParameter
-        >>> test = floatParameter(name='test1', value=100.0, units='second')
-        >>> print test
-        test1 (s) 100.0
+    Example
+    -------
+    >>> from parameter import floatParameter
+    >>> test = floatParameter(name='test1', value=100.0, units='second')
+    >>> print test
```
scottransom, line 503: print(test)

scottransom:
After the normal review, I'm just testing some basic things, and `pintk` is now broken.  I did `pintk examples/NGC6440E.par examples/NGC644E.tim`, and got this during the fit:
```Number of TOAs:  62
Number of commands:  1
Number of observatories: 1 ['gbt']
MJD span:  {self.first_MJD.mjd:.3f} to {self.last_MJD.mjd:.3f}
Date span: {self.first_MJD.iso} to {self.last_MJD.iso}
{key['obs']} TOAs ({len(grp)}):
  Min freq:      {np.min(grp['freq'].to(u.MHz)):.3f}
  Max freq:      {np.max(grp['freq'].to(u.MHz)):.3f}
  Min error:     {np.min(grp['error'].to(u.us)):.3g}
  Max error:     {np.max(grp['error'].to(u.us)):.3g}
  Median error:  {np.median(grp['error'].to(u.us)):.3g}

RMS PINT residuals are 1090.580 us

Pre-Fit Chi2:		157920.62 us^2
Pre-Fit Weighted RMS:	1090.5802 us
Post-Fit Chi2:		59.574241 us^2
Post-Fit Weighted RMS:	21.182025 us
          Parameter                   Pre-Fit	                Post-Fit	     Uncertainty        Difference          Diff/Unc
------------------------------------------------------------------------------------------------------------------------------------
Exception in Tkinter callback
    return self.func(*args)
  File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 464, in fit
    self.fit_callback()
  File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 670, in fit
    self.psr.fit(self.selected)
  File "/home/sransom/git/PINT/src/pint/pintk/pulsar.py", line 442, in fit
    self.write_fit_summary()
  File "/home/sransom/git/PINT/src/pint/pintk/pulsar.py", line 227, in write_fit_summary
    line += "%24s\t" % pre.print_quantity(pre.quantity)
AttributeError: 'AngleParameter' object has no attribute 'print_quantity'
```

Note that the toas method `.print_summary()` is also now broken (see the output above).  Although it seems like you broke that in #1085!

scottransom:
When bit changes are made, please at least try and run our "main" examples (`fit_NGC6440E.py` and a `pintk` of the same files).  If those break, then basically all "regular" PINT uses will be brioken as well!

aarchiba:
> When big changes are made, please at least try and run our "main" examples (`fit_NGC6440E.py` and a `pintk` of the same files). If those break, then basically all "regular" PINT uses will be brioken as well!

Please add test cases that verify the things you want not to be broken. See #1079 for how most pintk code could be tested. 

The examples that are in notebook form are actually run as part of the CI process, and I believe (but am not sure) that uncaught exceptions in them will fail CI as well, so that's another way to ensure that code is tested, if writing test cases is too onerous.

(The `print_summary` that I modified is an addition to the TOAs API whose sole function is `print(toas.get_summary)` (also that's the implementation); I'm not convinced that the extra API complexity is worth the shortcut. But it stayed in because it made one of our notebooks fail.)

aarchiba:
Note that #1074 now has an example of how simple testing of pintk code can catch this sort of bug.

src/pint/models/parameter.py
aarchiba: Fixed.

src/pint/models/parameter.py
aarchiba: Fixed.

src/pint/models/parameter.py
aarchiba: Clarified.

src/pint/models/parameter.py
aarchiba: Indeed.

src/pint/models/parameter.py
aarchiba: Fixed.

scottransom:
Looks great!  Thanks for this!

PR #1072, opened 2021-07-14
OLD: Improve JUMP objects

### LINKED ISSUE #1078: pintk help is double-spaced
For some reason the "help" output (is there any way to find out from the GUI that help is available, and that it appears in the terminal?) appears double-spaced:
```
The following interactions are currently supported by the Plk pane in the PINTkinter GUI:

Left click:     Select a point

Right click:    Delete a point

r:              Reset the pane - undo all deletions, selections, etc.

k:              (K)orrect the pane - rescale the axes

f:              Perform a fit on the selected points

d:              Delete the highlighted points

u:              Undo the most recent selection

c:              Clear highlighter from map

j:              Jump the selected points, or unjump them if already jumped

v:              Jump all TOA groups except those selected

i:              Print the prefit model as of this moment

o:              Print the postfit model as of this moment (if it exists)

p:              Print info about highlighted points (or all, if none are selected)

m:              Print the range of MJDs with the highest density of TOAs

+:              Increase pulse number for selected points

-:              Decrease pulse number for selected points

>:              Increase pulse number for all points to the right of selection

<:              Decrease pulse number for all points to the right of selection

h:              Print help

```

aarchiba:
This PR is meant to clarify the behaviour of JUMPs in PINT, and to ensure that they behave as expected in cases like overlapping JUMPs. This clarification has forced some changes affecting pintk. There are now some tests of pintk's Pulsar wrapper.

This has some minor API changes that would require updates from NANOGrav's `timing_analysis` to match, which we don't want to do before about 2021 Sept 10 because v1.0 is in active development. Testing is still very welcome, but please don't merge this until v1.0 is ready.

Closes #1078 

2022 July 6: This may need very substantial updating to match the current state of the code. It is probably better to pull code and documentation snippets from this into a new PR. I do think a PR to do these same things would be a good idea.


aarchiba:
Docs: https://nanograv-pint--1072.org.readthedocs.build/en/1072/api/pint.models.jump.html#pint.models.jump.PhaseJump

aarchiba:
I'm really unclear on what `jump_params_to_flags` is supposed to accomplish? It seems to add a new set of flags to the TOAs object that don't actually do anything and in fact cannot be used to select TOAs for JUMPing?

scottransom:
Some of the `jump_params_to_flags` stuff is explained in PR #1032 and the comments therein.  The main thing is that we were trying to standardize on using flags for all types of JUMPs.  And there are multiple types (including those from .tim files, .par files (of various sorts), and interactive ones from `pintk`).  I'm still not sure if things are consistent or not.

aarchiba:
> Some of the `jump_params_to_flags` stuff is explained in PR #1032 and the comments therein. The main thing is that we were trying to standardize on using flags for all types of JUMPs. And there are multiple types (including those from .tim files, .par files (of various sorts), and interactive ones from `pintk`). I'm still not sure if things are consistent or not.

I don't really know what you mean - there are only two kinds of JUMPs:

1. legacy JUMPs, found in `.tim` files and converted to `-jump N` flags upon load, and
2. JUMPs, which select a subset of TOAs by flag/table values and add a constant phase to them.

As far as I can tell `jump_params_to_flags` takes the second kind, which already selects a well-defined subset, and adds additional flags that cannot be used to select a subset?

aarchiba:
You seem to wind up with tim files like:
```
FORMAT 1
unk 999999.000000 56999.9999999968946643 1.000 gbt  -pn -14395108811.0
unk 999999.000000 57333.2222222321138195 1.000 gbt  -pn 212132.0
unk 999999.000000 57666.4444444391813657 1.000 gbt  -pn 14395478645.0
unk 999999.000000 57999.6666666724900694 1.000 gbt  -pn 28790673268.0
unk 999999.000000 58332.8888888887791435 1.000 gbt  -jump ['1'] -pn 43185798351.0
unk 999999.000000 58666.1111111063743750 1.000 gbt  -jump ['1'] -pn 57580876784.0
unk 999999.000000 58999.3333333312934144 1.000 gbt  -jump ['1'] -pn 71975943102.0
unk 999999.000000 59332.5555555595535763 1.000 gbt  -jump ['1', '2'] -pn 86371035295.0
unk 999999.000000 59665.7777777672042362 1.000 gbt  -jump ['1', '2'] -pn 100766185158.0
unk 999999.000000 59999.0000000061153356 1.000 gbt  -jump ['1', '2'] -pn 115161408996.0
```
and there isn't a way to JUMP all the TOAs that have `'1'` in their list.

paulray:
I think @champagne-cmd might want to chime in on this discussion.

luojing1211:
The reason for the list containing quoted numbers not just numbers is that we use string in the maskParameter key-values to prevent confusion in the case of when the key-value is "430", which means 430MHz backend.

aarchiba:
> The reason for the list containing quoted numbers not just numbers is that we use string in the maskParameter key-values to prevent confusion in the case of when the key-value is "430", which means 430MHz backend.

Flag values can only contain strings when they come out of par files, so storing a list in there at all is going to be a problem. But in any case the "430" backend string never occurs in a list, and what confusion would arise if the list contained integers?

scottransom:
@aarchiba As mentioned in that PR, the `key_values` for all jump flags should be strings.  And having a list definitely works *in memory* to allow JUMPs.   Are you saying that it doesn't work if you write the TOAs out to a `.tim` file and read them back in?  That is certainly not what we intend with JUMP flags like that.  So I don't think that matters.

luojing1211:
The reason to have a list is to handle the case like "JUMP -mjd 54000 55000". But I can see that we can change "key_value=[54000, 55000] " to key_value1 = 54000, key_value2 = 55000

scottransom:
And in the past, interactively handled JUMPs set from `pintk` were different.  They were `gui_jumps` or somesuch.  And all three different JUMPs where handled in slightly different ways.  As you have noted, it is still confusing now.

scottransom:
The main reason there is a list is so that a single TOA can be associated with multiple different JUMPs.

aarchiba:
> And in the past, interactively handled JUMPs set from `pintk` were different. They were `gui_jumps` or somesuch. And all three different JUMPs where handled in slightly different ways. As you have noted, it is still confusing now.

The way that `add_jump_and_flags` works, GUI jumps are exactly like any other kind of jump; they use the flag `-gui_jump` and set the value to some large integer N, then create `JUMP -gui_jump N 0`. So they're not any different from any other kind of JUMP.

aarchiba:
> @aarchiba As mentioned in that PR, the `key_values` for all jump flags should be strings. And having a list definitely works _in memory_ to allow JUMPs. Are you saying that it doesn't work if you write the TOAs out to a `.tim` file and read them back in? That is certainly not what we intend with JUMP flags like that. So I don't think that matters.

I think it's a bad idea to add flags to TOAs objects that won't survive the round trip to tim files. And if these jump flags are useful (for what? I still don't know) why would you not save them in tim files? If they're meant to be ephemeral and have a different data type from what's in the tim file, what are they doing in the `flags` part of the data structure?

aarchiba:
Let me elaborate on that point a little: The `.tim` file I included above is not readable by PINT at all. PINT will discover things in the flags area that don't look like flags, because the lists include spaces, and fail to recover the information in the `.tim` file.

aarchiba:
> The reason to have a list is to handle the case like "JUMP -mjd 54000 55000". But I can see that we can change "key_value=[54000, 55000] " to key_value1 = 54000, key_value2 = 55000

`maskParameters` do operate in two modes (determined by whether the key is `mjd`/`freq` or anything else), one where they accept a pair of floating-point values and one where they accept a string to match exactly. This is a different issue from having non-string things stored in the `flags` column of `toas.table`.

aarchiba:
My basic question is: why is it desired to have a flag on each TOA that lists which JUMPs affect it?
```python
for j in model.jumps:
    toa_indices = getattr(model, j).select_toa_mask(toas)
    if my_toa_index in toa_indices:
        print(f"TOA {my_toa_index} is affected by jump {j}")
```

scottransom:
I'm not sure that's the right question, @aarchiba.  We need a way to select TOAs that are JUMPed based on the presence or value of a generic flag.  That's a requirement because it is something that TEMPO2 users use.  Given that we have to do that, and also that we need to support .tim-file JUMPs which need to be assigned to each TOA as they are read, it made sense to use flags for other kinds of JUMPs (like likely temporary ones from interactive use in `pintk`).  That is why we use flags for this -- even for ephemeral (and very common!)  gui JUMPs.

Note that we do *not* have what your question seems to be asking: we do not make a single flag and assign to it every single JUMP that affects that TOA.  Nor do we desire such a thing.

The big issue, and it was the one that came up for #1032, was how do we assign more than one of the same type of JUMPs to the same TOA interactively?  We decided to use a list, because we never even considered that someone would want to write those JUMPs out into a `.tim` file -- so that doesn't work right now.  But it could certainly be made to work with some small changes in the .tim reading and writing code.  So I don't think that is a huge problem (and I think people will write those TOAs out rarely if ever, truthfully).

Note:  I still think that we should require that that `key_value` for flags that affect JUMPs need to be (or be treated as) strings.  We've had problems with this multiple times in the past when that was not the case (jump "1" colloided with jump 1 etc).

aarchiba:
> I'm not sure that's the right question, @aarchiba. We need a way to select TOAs that are JUMPed based on the presence or value of a generic flag. That's a requirement because it is something that TEMPO2 users use. 

We have that. `JUMP -fish carp 0`.

>Given that we have to do that, and also that we need to support .tim-file JUMPs which need to be assigned to each TOA as they are read, it made sense to use flags for other kinds of JUMPs (like likely temporary ones from interactive use in `pintk`). That is why we use flags for this -- even for ephemeral (and very common!) gui JUMPs.

Currently when we load TOAs with JUMP blocks, those JUMP blocks are assigned numbers and the affected TOAs get a flag `-jump N` which we can then use to select them for JUMPing with `JUMP -jump N 0`. 

The easiest way to implement GUI jumps is to add a flag `-gui_jump yes` to all TOAs selected in the GUI (and no others), then add `JUMP -gui_jump yes 0` to the model. 

> Note that we do _not_ have what your question seems to be asking: we do not make a single flag and assign to it every single JUMP that affects that TOA. Nor do we desire such a thing.

`jump_params_to_flags` creates a flag `-jump` on each TOA that lists the indices of all the JUMPs that affect that flag. Various related functions attempt to maintain this invariant. I don't see the benefit of this invariant. (Note that this is a conflicting usage of the same `-jump` flag with the one created upon reading legacy `.tim` files.) I propose abolishing this invariant.
 
> The big issue, and it was the one that came up for #1032, was how do we assign more than one of the same type of JUMPs to the same TOA interactively? We decided to use a list, because we never even considered that someone would want to write those JUMPs out into a `.tim` file -- so that doesn't work right now. But it could certainly be made to work with some small changes in the .tim reading and writing code. So I don't think that is a huge problem (and I think people will write those TOAs out rarely if ever, truthfully).

How about, if you want two different subsets for two different GUI jumps, `-gui_jump_1 yes` and `-gui_jump_2 yes`? Then `JUMP -gui_jump_1 yes 0` and `JUMP -gui_jump_2 yes 0` affect them independently and they can overlap or not. If you want to structurally force them not to overlap you could use the same flag but different values (`-fish carp` and `-fish sturgeon`).

> Note: I still think that we should require that that `key_value` for flags that affect JUMPs need to be (or be treated as) strings. We've had problems with this multiple times in the past when that was not the case (jump "1" colloided with jump 1 etc).

I've got a PR under development that ensures we deal with strings in the flags column.

scottransom:
I think the `-fish carp` / `-fist sturgeon` idea is totally fine.  I don't like anything at all about the way we currently handle JUMPs as model params or as flags/key_values with indices, though.  Because JUMPs can come and go and the numbers get out of sync and are confusing.  This is been an issue with JUMPs since the beginning -- because they are `maskParams` with integer (and sometimes sequential) indices.

But please, once again, be gentle with these discussions?  The code that we have has been written by *numerous* students over the years -- and a lot of this discussion is on stuff that was done only a few months ago!  There are zillions of ways to implement these things and the use cases can be *really* confusing (and different people value certain things over others).

scottransom:
One important thing to remember about a use case is that in `pintk`, it needs to keep track of all the TOAs that are affected by the various JUMPs as well as the JUMPs as model params.  So it needs to be able to map them both ways and be able to change any parts of that mapping (i.e. removing TOAs, removing/adding JUMPs, turning on/off existing JUMPs).  I suspect that the functions that you are talking about (e.g. `jump_par`ams_to_flags` and related) were written for precisely that purpose.  And they work right now!  (TOA writing using that info is broken, however, as you noted.)

aarchiba:
> I think the `-fish carp` / `-fist sturgeon` idea is totally fine. I don't like anything at all about the way we currently handle JUMPs as model params or as flags/key_values with indices, though. Because JUMPs can come and go and the numbers get out of sync and are confusing. This is been an issue with JUMPs since the beginning -- because they are `maskParams` with integer (and sometimes sequential) indices.

JUMP indices can and should be ignored. They play no role at all in which TOAs the JUMPs match; they exist only to give all JUMPs distinct names in the model. They can and should be replaced so that JUMPs get named something like JUMP_flag_value.

Unfortunately, because they define the user-visible name of the parameter from python, the current naming scheme, and the fragile numeration, have gotten baked into places like the NANOGrav yaml files and the `model.compare` function (which helpfully compares whichever JUMP got loaded seventh in one par file to whichever JUMP got loaded seventh in the other par file).

I started writing code to abolish JUMP indices but the current model builder does something arcane with them and I don't want to go digging in there as it is about to be replaced.

aarchiba:
> One important thing to remember about a use case is that in `pintk`, it needs to keep track of all the TOAs that are affected by the various JUMPs as well as the JUMPs as model params. So it needs to be able to map them both ways and be able to change any parts of that mapping (i.e. removing TOAs, removing/adding JUMPs, turning on/off existing JUMPs). I suspect that the functions that you are talking about (e.g. `jump_par`ams_to_flags` and related) were written for precisely that purpose. And they work right now! (TOA writing using that info is broken, however, as you noted.)

To get all the JUMPs as model params, you can use `model.jumps` to get their names.

To get all the TOAs affected by a jump you can use `model.JUMP17.select_toa_mask(toas)`.

To get all the JUMPs affecting a TOA, you can write a loop like the one I wrote above. (The current alternative method to do this requires writing a loop to iterate over all sets of TOA flags and asking whether the index of the current JUMP is in that list.)

The current functions do not work. I loaded in some TOAs with two JUMP blocks, whose TOAs got corresponding `-jump 1` and `-jump 2` flags. Then I tried `model.jump_flags_to_params(toas)` and it did nothing because I had two unrelated JUMPs (that did not select those TOAs) already in the model. It confused the JUMP indices (arbitrary based on the order of things in the par file) with the values associated with the `-jump` flag; those numbers get assigned based on the JUMP blocks when reading a legacy tim file but are otherwise recorded stably in saved `.tim` files.

paulray:
I haven't followed all this but if the problem is just numbering jumps, why not change things like GUI jumps and old style TIM file jumps to use a hash key, which is (essentially) unique and doesn't require any kind of sequence? Those can be assigned without care about how many jumps there already are, and can be written/read to TIM files.

aarchiba:
> I haven't followed all this but if the problem is just numbering jumps, why not change things like GUI jumps and old style TIM file jumps to use a hash key, which is (essentially) unique and doesn't require any kind of sequence? Those can be assigned without care about how many jumps there already are, and can be written/read to TIM files.

JUMP numbers need to go away permanently. We should not have parameters called JUMP17.

There's a separate question, of if (say) pintk wants to use two different GUI jumps, how it should name the associated flags and values; I am agnostic on this but there's no reason pintk can't just keep its own internal list of numbers for its GUI jumps. `-gui_jump_1`...

champagne-cmd:
Hey everyone,

To answer your question @aarchiba , the jumps are all given the 'jump' flag because, as @scottransom mentioned earlier, pintk uses the flags to find where the jumps are in the TOAs. I believe this design was chosen so that, if you just have a portion of the TOAs chosen in pintk that you are fitting to or performing some other operation on, you could simply iterate over that portion's flags and easily see what jumps are in your selection/if they are all jumped/etc - using the TOA indices that we have on-hand to find the jumps as opposed to iterating through the jumps to see if we come across a selected TOA index. But using select_toa_mask may be better to avoid using confusing flags to track this. 

I can start this transition next and see if I come across any pitfalls. I was also planning to transition the gui/tim jump key_values to a randomized hex instead of the jump index, but if you've already started on that @aarchiba I can hold off on that.

aarchiba:
I can remove the `jump_params_to_flags` and replace the code that relies on the `-jump` lists, but I don't have any experience at all with pintk, so maybe most effective would be if someone who did would give this PR (once I do that) a solid test to verify that I didn't break anything?

aarchiba:
Just to check, though: one goal of pintk is to allow multiple GUI jumps simultaneously, right? Is there any interest in these being able to overlap? If no overlaps are needed, I suggest using the flag `-gui_jump` and values that are arbitrary but different numbers; if overlaps are needed, `-gui_jump_N` with the value `yes` probably makes more sense.

My approach to these would be that pintk would have a list of GUI jumps that it was currently managing, and this list would include the flag and value associated with each one. This list could even just be "all JUMPs whose flag starts with `gui_jump`" and one could use the list that the timing model keeps, but I imagine pintk wants to record extra information like a colour for each one. Wherever it's recorded, the list would contain the information of what flag and value were used for each one, and if a new GUI jump were created one could go through the list and select a value and/or flag that didn't overlap with any existing one. (Choose an N one more than the highest currently in use, say.)

aarchiba:
See #1074 where I've tried to make flag manipulation much more convenient.

champagne-cmd:
> 
> 
> Just to check, though: one goal of pintk is to allow multiple GUI jumps simultaneously, right? Is there any interest in these being able to overlap? If no overlaps are needed, I suggest using the flag `-gui_jump` and values that are arbitrary but different numbers; if overlaps are needed, `-gui_jump_N` with the value `yes` probably makes more sense.
>
My understanding is that non-GUI jumps can overlap, but GUI jumps should not. I am not sure of the reason behind this though.

aarchiba:
> > Just to check, though: one goal of pintk is to allow multiple GUI jumps simultaneously, right? Is there any interest in these being able to overlap? If no overlaps are needed, I suggest using the flag `-gui_jump` and values that are arbitrary but different numbers; if overlaps are needed, `-gui_jump_N` with the value `yes` probably makes more sense.
> 
> My understanding is that non-GUI jumps can overlap, but GUI jumps should not. I am not sure of the reason behind this though.

Since you're creating the GUI jump mechanism in pintk, it's kind of up to you whether you want them to be able to overlap. Which would be more convenient when doing interactive timing? Which makes a more comfortable user interface?

paulray:
To me, it seems entirely reasonable to restrict GUI jumps to be non-overlapping.

aarchiba:
New changes here are starting to overlap with #1071, so I'll try to get that one ready so it can be merged and I can take advantage of its improvments. (Specifically as I try to make JUMPs easier to keep track of I need to adjust how maskParameters behave, and that PR simplifies parameter behaviour.)

aarchiba:
I cleaned up how JUMPs work but I can't really tell whether I broke pintk.

scottransom:
I just checked with simple .par and .tim files that have a couple of .tim file JUMPs, and yes, it seems that `pintk` is broken now.  Let me know if you want me to send you the files.
```
/home/sransom/python_venvs/py39/lib/python3.9/site-packages/numpy/lib/arraysetops.py:583: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  mask |= (ar1 == a)
Exception in Tkinter callback
    return self.func(*args)
  File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 437, in fit
    self.fit_callback()
  File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 641, in fit
    self.psr.fit(self.selected)
  File "/home/sransom/git/PINT/src/pint/pintk/pulsar.py", line 343, in fit
    numjumps = self.prefit_model.components["PhaseJump"].get_number_of_jumps()
AttributeError: 'PhaseJump' object has no attribute 'get_number_of_jumps'
```
[](url)

aarchiba:
> I just checked with simple .par and .tim files that have a couple of .tim file JUMPs, and yes, it seems that `pintk` is broken now. Let me know if you want me to send you the files.
> 
> ```
> /home/sransom/python_venvs/py39/lib/python3.9/site-packages/numpy/lib/arraysetops.py:583: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
>   mask |= (ar1 == a)
> Exception in Tkinter callback
> Traceback (most recent call last):
>   File "/usr/lib/python3.9/tkinter/__init__.py", line 1892, in __call__
>     return self.func(*args)
>   File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 437, in fit
>     self.fit_callback()
>   File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 641, in fit
>     self.psr.fit(self.selected)
>   File "/home/sransom/git/PINT/src/pint/pintk/pulsar.py", line 343, in fit
>     numjumps = self.prefit_model.components["PhaseJump"].get_number_of_jumps()
> AttributeError: 'PhaseJump' object has no attribute 'get_number_of_jumps'
> ```

The problem isn't the files, the problem is figuring out how to operate pintk. I've asked some questions on the NANOGrav slack that have helped. Still unclear about how it's supposed to work.

scottransom:
The basic operation is pretty simple!  `pintk PARFILE TIMFILE` and then hit the "fit" button!

aarchiba:
> The basic operation is pretty simple! `pintk PARFILE TIMFILE` and then hit the "fit" button!

If that's all that's needed it's now working great. Now that I know to press `h` and go look in the terminal I suspect more operations are possible.

scottransom:
The example tim file that I used had multiple timfile-JUMPs in it.  And it barfed the above when I ran the above (i.e. started pintk and hit the fit button).  So it isn't handling "regular" JUMPs properly right now.

aarchiba:
In the code, there is a boolean array `self.jumped` that seems to track which points are selected for jumping (I think). Am I correct in understanding that this means that there can only ever be one set of TOAs affected by a single GUI jump? Or, what is this array supposed to represent?

aarchiba:
Addressing #1079 could help drastically reduce the breakages of pintk.

champagne-cmd:
> 
> 
> In the code, there is a boolean array `self.jumped` that seems to track which points are selected for jumping (I think). Am I correct in understanding that this means that there can only ever be one set of TOAs affected by a single GUI jump? Or, what is this array supposed to represent?

That array stores booleans that are true if the toa is jumped (i.e. has a jump flag) and false otherwise. The only two places its ever really used are in the 'd' keypress operation, which deletes TOAs (if a selected TOA is jumped the user is instructed to delete the jump first) and in colormodes.py, where the jumped TOAs are given the red color.

champagne-cmd:
> 
> 
> > I just checked with simple .par and .tim files that have a couple of .tim file JUMPs, and yes, it seems that `pintk` is broken now. Let me know if you want me to send you the files.
> > ```
> > /home/sransom/python_venvs/py39/lib/python3.9/site-packages/numpy/lib/arraysetops.py:583: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
> >   mask |= (ar1 == a)
> > Exception in Tkinter callback
> > Traceback (most recent call last):
> >   File "/usr/lib/python3.9/tkinter/__init__.py", line 1892, in __call__
> >     return self.func(*args)
> >   File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 437, in fit
> >     self.fit_callback()
> >   File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 641, in fit
> >     self.psr.fit(self.selected)
> >   File "/home/sransom/git/PINT/src/pint/pintk/pulsar.py", line 343, in fit
> >     numjumps = self.prefit_model.components["PhaseJump"].get_number_of_jumps()
> > AttributeError: 'PhaseJump' object has no attribute 'get_number_of_jumps'
> > ```
> 
> The problem isn't the files, the problem is figuring out how to operate pintk. I've asked some questions on the NANOGrav slack that have helped. Still unclear about how it's supposed to work.

This error is because the function 'get_number_of_jumps' was deleted in jumps.py - replacing that function with len(self.prefit_model.jumps) should do the same thing.

aarchiba:
> > In the code, there is a boolean array `self.jumped` that seems to track which points are selected for jumping (I think). Am I correct in understanding that this means that there can only ever be one set of TOAs affected by a single GUI jump? Or, what is this array supposed to represent?
> 
> That array stores booleans that are true if the toa is jumped (i.e. has a jump flag) and false otherwise. The only two places its ever really used are in the 'd' keypress operation, which deletes TOAs (if a selected TOA is jumped the user is instructed to delete the jump first) and in colormodes.py, where the jumped TOAs are given the red color.

So, true if any jump at all affects that TOA? Like, in a two-telescope data set it will be true for all the TOAs from one telescope? Does it depend whether the JUMP affecting those TOAs is frozen or free?

I'm not sure what you mean by a "jump flag" - most JUMPs operate based on flags like `-fe 430` or frequency ranges or MJD ranges. Are these included in the binary array? These are JUMPs where removing the flag is impossible or inappropriate. Is this array perhaps intended only for GUI jumps?

Why should deleting a TOA imply deleting a JUMP parameter? The fitting operation already disables any JUMP that does not affect any of the TOAs that are being fit, so that's not a problem.

If the array is used only in those places, perhaps it can be generated on the fly when needed? This saves the complexity of ensuring it is always updated whenever any JUMP changes.

aarchiba:
Right now I'm trying to track down the message `ERROR: None [pint.pintk.plk]`; I think it comes from some operation involving the `jumped` array.

The main thing is I'm trying to understand how pintk wants to handle JUMPs so that I can make that work. It seems like maybe pintk has been developed mostly or entirely with single-telescope datasets that don't need JUMPs except temporarily during phase connection?

champagne-cmd:
@aarchiba the jumped array looked for the '-jump' flag in the TOA table that we are now moving away from - it looks for all jumped TOAs, not just the gui jumps. I believe the select_toa_mask function could replace this functionality and used on the fly as you said.

champagne-cmd:
> 
> 
> Right now I'm trying to track down the message `ERROR: None [pint.pintk.plk]`; I think it comes from some operation involving the `jumped` array.
> 

I believe that is due to line 1098 in plk.py - I tried changing it once, but I can't remember why the changes weren't approved.

aarchiba:
Trying to clear things up, I created #1084 to clarify how JUMPing should work in pintk.

aarchiba:
Worth doing here: 

- get rid of the "-" in a maskParameter's `.key`
- check that a maskParameter's `.key_value` is a valid flag value (`FlagDict.validate...()`)
- rename `.key` -> `.flag` and `.key_value` -> `.flag_value`

aarchiba:
Worth discussing:

If the maskParameter's criterion is based on MJD should we:
- Require Time objects
- Require Quantities in units of days
- Require float()
- Allow some combination of the above (and return it as which?)

For comparison if the criterion is based on frequency we require and return Quantities in (convertible to) MHz. If it's an observatory we allow any name for the observatory and return the canonical name for the observatory. If it's anything else we require a string and we return a string.

aarchiba:
If someone could test that pintk works as expected, I think I've fixed what I broke. Also pintk, when showing JUMPs, no longer shows their indices but instead their selection criteria.

aarchiba:
> Worth discussing:
> 
> If the maskParameter's criterion is based on MJD should we:
> 
>     * Require Time objects
> 
>     * Require Quantities in units of days
> 
>     * Require float()
> 
>     * Allow some combination of the above (and return it as which?)
> 
> 
> For comparison if the criterion is based on frequency we require and return Quantities in (convertible to) MHz. If it's an observatory we allow any name for the observatory and return the canonical name for the observatory. If it's anything else we require a string and we return a string.

Discussion result: floats only, for now, and check what exactly the comparison uses - TEMPO and TEMPO2 probably use the raw, un-clock-corrected time on the TOA line, and this is probably what users expect (although roundoff error probably means they should use enough padding it won't matter). But definitely not barycentered MJD unless the input MJD was barycentered.

aarchiba:
I'll rebase and clean up and test once #1074 is in, that should help with some of the mask weirdnesses. At that point I'm inclined to call it good; further improvements are probably best discussed in #1092 and probably depend on #1026 being ready.

scottransom:
@aarchiba I saw your request in the PINT Slack channel to take a look at this PR.  I'm happy to do that and hopefully get it a thumbs-up.  But why did it get changed to "HOLD"?

paulray:
I just tried to play with this using `pintk`. I loaded NGC6440 and tried to add a jump via the "j" keyboard command.  It failed with this exception:
```
  File "/Users/paulr/env/pint/lib/python3.7/site-packages/matplotlib/cbook/__init__.py", line 224, in process
    func(*args, **kwargs)
  File "/Users/paulr/src/PINT/src/pint/pintk/plk.py", line 1256, in canvasKeyEvent
    self.fitboxesWidget.addFitCheckBoxes(self.psr.prefit_model)
  File "/Users/paulr/src/PINT/src/pint/pintk/plk.py", line 150, in addFitCheckBoxes
    par_text = f"{pm.origin_name} {pm.key} {pm.key_value}"
AttributeError: 'maskParameter' object has no attribute 'key'
```

paulray:
I also tried adding this line to NGC6440E.par:
```
JUMP MJD 53400 53500 0 1
```
and then just starting pintk immediately triggered that same exception.

aarchiba:
> @aarchiba I saw your request in the PINT Slack channel to take a look at this PR. I'm happy to do that and hopefully get it a thumbs-up. But why did it get changed to "HOLD"?

As I said in the edited initial description, this changes the API in a way that would force an update of timing_analysis and after discussing it with Joe Swiggum I don't want to do that while we're working on v1.0. Please still feel free to review it but I think we should hold off on merging it for now.

aarchiba:
> I just tried to play with this using `pintk`. I loaded NGC6440 and tried to add a jump via the "j" keyboard command. It failed with this exception:
> 
> ```
> Traceback (most recent call last):
>   File "/Users/paulr/env/pint/lib/python3.7/site-packages/matplotlib/cbook/__init__.py", line 224, in process
>     func(*args, **kwargs)
>   File "/Users/paulr/src/PINT/src/pint/pintk/plk.py", line 1256, in canvasKeyEvent
>     self.fitboxesWidget.addFitCheckBoxes(self.psr.prefit_model)
>   File "/Users/paulr/src/PINT/src/pint/pintk/plk.py", line 150, in addFitCheckBoxes
>     par_text = f"{pm.origin_name} {pm.key} {pm.key_value}"
> AttributeError: 'maskParameter' object has no attribute 'key'
> ```

Ah, this is obviously some kind of pintk summary code that isn't part of the test suite. This PR changes the name of the selection attributes from `.key` and `.key_value` to `.flag` and `.flag_value` - I found `.key` very confusing in light of python's use of keys in dictionaries, and in fact what goes in there should be the flag that the selection acts upon. I'll go in and make sure that part of pintk is in the test suite and then change the code.

scottransom:
I just used `ripgrep` to search for all uses of `.key_value` in all of pint and found there were two places.  One of them is above:
```
 >rg .key_value                                                                   10:54:41
src/pint/models/parameter.py
1427:    key_value :  list/single value optional

src/pint/pintk/plk.py
150:                    par_text = f"{pm.origin_name} {pm.key} {pm.key_value}"
```

aarchiba:
No need for this to be on hold any more; when I have time to work on it I'll fix the conflicts and make the changes suggested.

src/pint/models/jump.py
```diff
@@ -2,86 +2,84 @@
...
-import numpy
+import numpy as np
 
 from pint.models.parameter import maskParameter
 from pint.models.timing_model import DelayComponent, MissingParameter, PhaseComponent
 
 log = logging.getLogger(__name__)
 
-
-class DelayJump(DelayComponent):
```
luojing1211, line 13: Tempo actually has delay jumps for the observatory cable delays. I was thinking to switch the backend jump to delay jump. The jump value affects the binary delay calculation.

luojing1211:
This PR needs to rebase and fix the conflict.

Issue #1073, opened 2021-07-15
Manipulating flags is a pain in the neck


aarchiba:
When users need to work with the flags on TOAs objects, they generally need to write for loops for even basic operations (set all or some to TOAs to have a particular value, for example). #1070 has some discussion that demonstrates this. Some specific headaches:

- Flags can't be manipulated (read from, written to) in an array-like manner
- Flags and values are all strings when read from tim files, but users can and do store other things
- Flags appear in tim files with dashes (`-fish`) but in the flags dictionary without them (`d['fish']`) and users are not warned if they make errors in this; maskParameters actually do need the dash
- Users assigning to flag keys or values can easily produce values that don't survive writing to tim files

aarchiba:
One suggestion for easier access: augment the indexing operator on TOAs objects so that `toas['mjd']` or `toas['fish']` produces an array that looks like an appropriate column. This would enable:
```python
toas['fish', toas['mjd']>58000] = 'carp'
```
or
```python
some_toas = toas[toas['fish']=='carp']
```

aarchiba:
One suggestion for data validation: create a FlagsDict class that only accepts valid keys and values, and store that in the astropy table instead of the current plain dicts.

scottransom:
I really like the idea of moving the flag setting/getting loops into the indexing operation.  It is still looping (and I don't really see a good way around that given the way tables work), but at least it is nice syntactic sugar.

aarchiba:
Now in #1074; open for comments. But you can do `toas[10:20, "gui_jump"] = "yes"`.

We *could* make all flags into columns in the table. This would allow very uniform handling with some caveats: 

- We'd need to specify types for each flag (e.g. `-pp_dm` is a Quantity in pc/cm3) and corresponding ways to read and write them as strings,
- we'd need a well-defined "no value" placeholder, since Astropy tables do not provide this (the empty string is natural for string types), 
- we'd have to make adding a flag easier than constructing a new column for addition to the TOAs object, and
- this would really break how people access flags now, and it would take some serious witchcraft to provide a workalike API.

aarchiba:
Just to confuse the issue, if you ever did a `select_toas_mask` based on a flag, that function added a column to `toas.table` named after the flag and with its values. No attempt was made to keep this column up to date.

scottransom:
[...]

PR #1074, opened 2021-07-16
flag improvements

aarchiba:
This is intended to address #1073 and make flags simpler and more reliable to work with. Currently it enforces restrictions on what keys and values can be stored in a TOA's `flags` so that they can safely be written to `.tim` files. It also allows users to manipulate TOA flags and columns through fancy indexing, so you can do (for example) `toas['gui_jump', 17]  = "yes"`.

Notable changes:

- `flags` can no longer hold anything but strings; you may need to use `str` to convert numbers into strings and `float` or `int` to convert them back.
    - If you wanted to store Quantities (that is, with units), those couldn't be written out to `.tim` files. I've converted several examples into strings in conventional units (for example `-energy` is in MeV), but if you don't want it to  wind up in a `.tim` file, perhaps an extra column in `toas.table` would be more appropriate? Certainly it would be more efficient.
    - `get_flag_value` now accepts a data type to convert the flag values to.
- Indexing into TOAs objects can now select columns or pseudo-columns of flags for reading or writing, but there are a few gotchas where things don't work the way one might wish or error checking isn't as good as it could be.
- Since empty strings cannot occur as flag values in `.tim` files, assigning an empty string deletes a flag. This may be confusing.


aarchiba:
Exactly what should and should not be allowed with the indexing maybe needs a little discussion.

Specifically, Astropy tables allow `t["column_name"]` or `t[::2]` but not `t["column_name", ::2]`, whereas numpy structured arrays do allow mixing indexing like this. That said, because Astropy tables can provide columns as views, `t["fish"][0] = "carp"` can work, while for us there's no way to do this - `t["fish"]` has to provide a newly constructed object, and without some really heroic witchcraft there's no way for modifications to that object to affect the flags on the original TOAs object. So it helps us a lot to allow `t["column_name", ::2]`. But users are going to try `t["fish"][0] = "carp"` and it's not going to affect the TOAs object.

aarchiba:
A further step we could take, which would make TOAs objects cleaner, would be to put all the flags into `toas.table` as columns. This already establishes that flag values are always strings, and that an empty string is equivalent to the flag not being present. Astropy tables can efficiently store strings of a fixed maximum length or less efficiently store strings as objects; in any case as long as we are comfortable with the empty string as "no flag" we could just make the convention that extra columns in `self.table` were flags in the tim file. Then any flag and its value could be accessed as a column of `toas.table`. If we were willing to enforce types and units on flag values (for example: `-pp_dm` is always floating-point, with units of pc/cm3), we could make these flags-as-columns very easy to access.

aarchiba:
New docs: https://nanograv-pint--1074.org.readthedocs.build/en/1074/api/pint.toa.html#pint.toa.TOAs

Honestly I'm not sure what the best way is to document the indexing modes supported by TOAs objects; the `__getattr__` function doesn't get an entry in the docs and I'm not sure how best to make it findable in the TOAs docstring.

aarchiba:
Worth checking - do we duplicate columns (`tel`) in the flags? If so how do we keep those in sync?

aarchiba:
I think this is ready for review and its inclusion would benefit #1072 and #1071.

src/pint/toa.py
```diff
@@ -1336,34 +1404,187 @@ def __len__(self):
...
+
+        Returns
+        -------
+        pint.toa.TOAs or np.ndarray or astropy.table.Column
+            The selected part of the object.
+
+        Note
+        ----
+        This function does not currently support extracting a single :class:`~pint.toa.TOA` object,
+        to use integer indexing a column must be selected.
```
scottransom, line 1441: Can you give an example of the wrong way and then the right way to do that?

scottransom:
So `pintk` is failing when you fit with timfile JUMPs:
```
/home/sransom/python_venvs/py39/lib/python3.9/site-packages/numpy/lib/arraysetops.py:583: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  mask |= (ar1 == a)
Exception in Tkinter callback
    return self.func(*args)
  File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 464, in fit
    self.fit_callback()
  File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 670, in fit
    self.psr.fit(self.selected)
  File "/home/sransom/git/PINT/src/pint/pintk/pulsar.py", line 421, in fit
    dict["jump"] = [str(num)]
  File "/home/sransom/git/PINT/src/pint/toa.py", line 1018, in __setitem__
    self.__class__.check_allowed_value(key, val)
  File "/home/sransom/git/PINT/src/pint/toa.py", line 1012, in check_allowed_value
    raise ValueError(f"value {v} for key {k} must be a string")
ValueError: value ['1'] for key jump must be a string
```

scottransom:
Fitting with timfile jumps seems to work fine with normal fitting scripts as long as you call `m.jump_flags_to_params(t)` (and interestingly, `pintempo` doesn't call that, so it doesn't work for timfiles with JUMPs.  I'll make an issue for that (or maybe just make a direct bugfix -- it is a one liner).

aarchiba:
I tried to temporarily bodge all the usages of `-jump`-lists-all-jumps-affecting-this-toa to use strings containing comma-separated integers, but this is fundamentally incompatible with both timfile jumps and the way jumps select TOAs, so the real answer here is #1072. I'll do what I can here.

aarchiba:
> Fitting with timfile jumps seems to work fine with normal fitting scripts as long as you call `m.jump_flags_to_params(t)` (and interestingly, `pintempo` doesn't call that, so it doesn't work for timfiles with JUMPs. I'll make an issue for that (or maybe just make a direct bugfix -- it is a one liner).

Note that `jump_flags_to_params` as implemented here is broken if the par file already contains any JUMPs - it assumes that the first JUMP/JUMP par in the `.tim` file is controlled by the first JUMP to occur textually in the par file, even if that JUMP explicitly selects a completely different set of TOAs, and so it does not create a JUMP flag. Likewise, if the third JUMP/JUMP pair is selected by a `JUMP -jump 3` in the par file but there are only two JUMPs in the par file, `jump_flags_to_params` will create a duplicate JUMP. This is all fixed in #1072.

src/pint/toa.py
aarchiba: Done.

aarchiba:
> So `pintk` is failing when you fit with timfile JUMPs:
> 
> ```
> /home/sransom/python_venvs/py39/lib/python3.9/site-packages/numpy/lib/arraysetops.py:583: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
>   mask |= (ar1 == a)
> Exception in Tkinter callback
> Traceback (most recent call last):
>   File "/usr/lib/python3.9/tkinter/__init__.py", line 1892, in __call__
>     return self.func(*args)
>   File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 464, in fit
>     self.fit_callback()
>   File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 670, in fit
>     self.psr.fit(self.selected)
>   File "/home/sransom/git/PINT/src/pint/pintk/pulsar.py", line 421, in fit
>     dict["jump"] = [str(num)]
>   File "/home/sransom/git/PINT/src/pint/toa.py", line 1018, in __setitem__
>     self.__class__.check_allowed_value(key, val)
>   File "/home/sransom/git/PINT/src/pint/toa.py", line 1012, in check_allowed_value
>     raise ValueError(f"value {v} for key {k} must be a string")
> ValueError: value ['1'] for key jump must be a string
> ```

Test case created, catches this, now passes. I did rip out some automated JUMP-removal code in the `fit()` function because it's rewritten in #1072.

scottransom:
Looks good.  Again, thanks for this!  Big improvements!

scottransom:
Crap.  Although now there are conflicts that have to be resolved.  Sorry that this didn't get checked earlier!

aarchiba:
Sorry forgot to update the rebase; the thing where pintk cleans up the jumps has to go away awaiting #1072's reimplementation of fix-jumps-before-fitting.

scottransom:
Great!  Merging.  Thanks.

PR #1083, opened 2021-07-21
improving docstrings in `utils.py`

dlakaplan:
I went through and retroactively improved the docstrings for functions I wrote in `utils.py`.  I also improved docstrings for some functions I didn't write but which were very similar (for computing characteristic age, Edot, B, ...).

A couple of very minor changes to the functions themselves, mostly moving from `.to(...).value` to `.to_value(...)`

### DIFF
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1109,7 +1109,7 @@ def p_to_f(p, pd, pdd=None):
     p : astropy.units.Quantity
         pulsar period (or frequency)
     pd : astropy.units.Quantity
-        period derivative (or frequency derivative)        
+        period derivative (or frequency derivative)
     pdd : astropy.units.Quantity, optional
         period second derivative (or frequency second derivative)
 
@@ -1120,7 +1120,7 @@ def p_to_f(p, pd, pdd=None):
     fd : astropy.units.Quantity
         pulsar frequency derivative (or period derivative)
     fdd : astropy.units.Quantity
-        frequency second derivative (or period second derivative) if pdd is supplied        
+        frequency second derivative (or period second derivative) if pdd is supplied
     """
     f = 1.0 / p
     fd = -pd / (p * p)
@@ -1146,10 +1146,14 @@ def pferrs(porf, porferr, pdorfd=None, pdorfderr=None):
         pulsar period (or frequency)
     porferr : astropy.units.Quantity
         pulsar period uncertainty (or frequency uncertainty)
+        :math:`\sigma_P` or :math:`\sigma_f`
     pdorfd : astropy.units.Quantity, optional
         pulsar period derivative (or frequency derivative)
+        :math:`\dot P` or :math:`\dot f`
     pdorfderr : astropy.units.Quantity, optional
-        pulsar period derivative uncertainty (or frequency derivative uncertainty)
+        pulsar period derivative uncertainty
+        (or frequency derivative uncertainty)  :math:`\sigma_{\dot P}`
+        or :math:`\sigma_{\dot f}`
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


src/pint/utils.py
```diff
@@ -1120,6 +1139,28 @@ def pferrs(porf, porferr, pdorfd=None, pdorfderr=None):
 
     Calculate the period or frequency errors and
     the pdot or fdot errors from the opposite one.
+
+    Parameters
+    ----------
+    porf : :class:`astropy.units.Quantity`
+        pulsar period (or frequency)
```
aarchiba, line 1146: This is horrible. Can we please have an `fperrs` that maybe calls this one to do the calculation but both have sensibly named arguments and docstrings? `porf` -> `p` in one and `f` in the other for example?

src/pint/utils.py
```diff
@@ -1138,18 +1179,56 @@ def pulsar_age(f, fdot, n=3, fo=1e99 * u.Hz):
...
+        initial frequency
+
+    Returns
+    -------
+    age : :class:`astropy.units.Quantity`
+        pulsar age in ``u.yr``
+
+    Notes
+    -----
+    Calculates ``(-f/(n-1)/fdot) * (1 - (f/fo)**(n-1))``
```
aarchiba, line 1203: Our Sphinx supports LaTeX math, maybe that would be clearer?

src/pint/utils.py
```diff
@@ -1213,27 +1329,29 @@ def mass_funct2(mp, mc, i):
...
+    mc : :class:`astropy.units.Quantity`
+        Companion mass, typically in ``u.solMass``
+    i : :class:`astropy.coordinates.Angle`
+        Inclination angle, in ``u.deg`` or ``u.rad``
 
     Returns
     -------
-    f_m : Quantity
-        Mass function in `u.solMass`
+    f_m : :class:`astropy.units.Quantity`
```
aarchiba, line 1341: In Parameters and Returns sections, for types of parameters/return values, you shouldn't need `:class:` and backticks, it should automatically get parsed as a type.

src/pint/utils.py
```diff
@@ -1695,29 +1822,28 @@ def calculate_random_models(fitter, toas, Nmodels=100, keep_models=True, params=
 
     Parameters
     ----------
-    fitter: `pint.fitter` object
+    fitter: :class:`pint.fitter`
```
aarchiba, line 1825: `pint.fitter.Fitter`

src/pint/utils.py
dlakaplan: I didn't write this function, just documented it.  So I didn't want to change the API.

paulray: It is a direct port from psr_utils in presto. It would be fine to propose a new version, but that would be a different pull request.

src/pint/utils.py
dlakaplan: Sure. Do you have any examples you could point to, in order to get the backslashes etc correct?

aarchiba: Er. I believe, in python, if you make the docstring a "raw string" (start with r""" ... """) the backslashes should be no different from those in normal LaTeX. But I could be wrong.

src/pint/utils.py
scottransom: Haha!  This is copied from PRESTO.  It's been in there for *18 years!* and I've used it a lot.  Kills two birds with one stone!   Minimizes copied code!   Super clever!  :-)

aarchiba: Zero copied code:
```python
def fperrs(f, ferr, ...):
    return pferrs(f, ferr, ...)
```
but the names and docstring could be intelligible.

aarchiba:
Hooray dots! Greek letters! But it looks like a (probably minor) conflict.

dlakaplan:
Thanks!

PR #1085, opened 2021-07-22
Make codecov informational

### LINKED ISSUE #1082: Codecov is overly agressive
We should modify how Codecov is used, so that it doesn't cause CI tests to fail just for a tiny decrease in coverage.
Perhaps the Github Action can be made to always return True so the tests always pass, but the coverage report is still available for inspection?

aarchiba:
Closes #1082

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -29,9 +29,9 @@ jobs:
           - os: ubuntu-latest
             python: '3.8'
             tox_env: 'notebooks'
-          - os: ubuntu-latest
-            python: '3.8'
-            tox_env: 'docs'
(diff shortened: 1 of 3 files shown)


dlakaplan:
Why is `toa.py` changed?  Is that related?

aarchiba:
I was trying to get a code coverage report out of the codecov machine. It's a minor code tidy-up to use f-strings instead of the old % formatting. I'm not getting big red xs, so this achieves the main goal, but eventually we should get the codecov machine posting a comment to this PR describing the coverage changes.

dlakaplan:
This looks good to me.  Anything that helps CI seems worthwhile.  But I don't know enough about the details to know if subtleties remain.

PR #1087, opened 2021-07-23
Unbreak codecov

aarchiba:
There was a typo in the codecov settings file, and the error was not obviously reported. This should fix it, though it's worth checking the CI results in detail to make sure that the "upload results to codecov" block is being run for the CI jobs that end in `-cov`.


aarchiba:
Filed a bug report with the codecov Github action maintainer.

aarchiba:
I think that these coverage reports are giving trouble because master never uploaded a correct coverage report, so it should start to behave when this is merged.

aarchiba:
Ready to review and merge.

PR #1088, opened 2021-07-26
Post-Keplerian parameter implementations

dlakaplan:
Implementation of `OMDOT`, `PBDOT`, `GAMMA` based on pulsar mass, companion mass, orbital period, orbital eccentricity.

Right now they are in `utils.py`, but we could consider moving them.

### DIFF
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1424,7 +1424,7 @@ def mass_funct2(mp: u.Msun, mc: u.Msun, i: u.deg):
         Pulsar mass, typically in ``u.solMass``
     mc : astropy.units.Quantity
         Companion mass, typically in ``u.solMass``
-    i : astropy.coordinates.Angle
+    i : astropy.coordinates.Angle or astropy.units.Quantity
         Inclination angle, in ``u.deg`` or ``u.rad``
 
     Returns
@@ -1449,9 +1449,9 @@ def mass_funct2(mp: u.Msun, mc: u.Msun, i: u.deg):
     .. math::
         f(m_p, m_c) = \\frac{m_c^3\sin^3 i}{(m_c + m_p)^2}
         
-    See [1]_
+    See [2]_
 
-    .. [1] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.34 (LHS)
+    .. [2] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.34 (LHS)
 
     """
     return (mc * np.sin(i)) ** 3.0 / (mc + mp) ** 2.0
@@ -1472,7 +1472,7 @@ def pulsar_mass(pb: u.d, x: u.cm, mc: u.Msun, i: u.deg):
         Projected pulsar semi-major axis (aka ASINI) in ``pint.ls``
     mc : astropy.units.Quantity
         Companion mass in ``u.solMass``
-    i : astropy.coordinates.Angle
+    i : astropy.coordinates.Angle or astropy.units.Quantity
         Inclination angle, in ``u.deg`` or ``u.rad``
 
     Returns
@@ -1538,7 +1538,7 @@ def companion_mass(pb: u.d, x: u.cm, i=60.0 * u.deg, mp=1.4 * u.solMass):
         Binary orbital period
     x : astropy.units.Quantity
         Projected pulsar semi-major axis (aka ASINI) in ``pint.ls``
-    i : astropy.coordinates.Angle, optional
+    i : astropy.coordinates.Angle or astropy.units.Quantity, optional
         Inclination angle, in ``u.deg`` or ``u.rad.`` Default is 60 deg.
     mp : astropy.units.Quantity, optional
         Pulsar mass in ``u.solMass``. Default is 1.4 Msun
@@ -1573,13 +1573,12 @@ def companion_mass(pb: u.d, x: u.cm, i=60.0 * u.deg, mp=1.4 * u.solMass):
     - :math:`c = -2 M_p {\\rm massfunct}`
     - :math:`d = -{\\rm massfunct} M_p^2`
 
-    To solve it we can use a direct calculation of the cubic roots:
-    https://en.wikipedia.org/wiki/Cubic_equation#General_cubic_formula
+    To solve it we can use a direct calculation of the cubic roots [3]_.
+
 
     It's useful to look at the discriminant to understand the nature of the roots
-    and make sure we get the right one
+    and make sure we get the right one [4]_.
 
-    https://en.wikipedia.org/wiki/Discriminant#Degree_3
 
     :math:`\Delta = (b^2 c^2 - 4ac^3-4b^3d-27a^2d^2+18abcd)`
 
@@ -1588,6 +1587,9 @@ def companion_mass(pb: u.d, x: u.cm, i=60.0 * u.deg, mp=1.4 * u.solMass):
     and this should be < 0
     since this reduces to :math:`-27\sin^6 i {\\rm massfunct}^2 M_p^4 -4\sin^3 i {\\rm massfunct}^3 M_p^3`
     so there is just 1 real root and we compute it below
+
+    .. [3] https://en.wikipedia.org/wiki/Cubic_equation#General_cubic_formula
+    .. [4] https://en.wikipedia.org/wiki/Discriminant#Degree_3
 
     """
     massfunct = mass_funct(pb, x)
@@ -1672,9 +1674,9 @@ def pbdot(mp: u.Msun, mc: u.Msun, pb: u.d, e: u.dimensionless_unscaled):
 
     and :math:`T_\odot = GM_\odot c^{-3}`.
 
-    See [1]_
+    More details in :ref:`Timing Models`.  Also see [5]_.
 
-    .. [1] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.52
+    .. [5] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.52
 
     """
     f = (1 + (73.0 / 24) * e ** 2 + (37.0 / 96) * e ** 4) / (1 - e ** 2) ** (7.0 / 2)
@@ -1730,9 +1732,9 @@ def gamma(mp: u.Msun, mc: u.Msun, pb: u.d, e: u.dimensionless_unscaled):
 
     with :math:`T_\odot = GM_\odot c^{-3}`.
 
-    See [1]_
+    More details in :ref:`Timing Models`.  Also see [6]_
 
-    .. [1] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.49
+    .. [6] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.49
 
     """
     value = (
@@ -1786,9 +1788,9 @@ def omdot(mp: u.Msun, mc: u.Msun, pb: u.d, e: u.dimensionless_unscaled):
         
     with :math:`T_\odot = GM_\odot c^{-3}`.
 
-    See [1]_
+    More details in :ref:`Timing Models`.  Also see [7]_.
 
-    .. [1] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.48
+    .. [7] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.48
 
     """
     value = (
@@ -1817,7 +1819,7 @@ def a1sini(mp, mc, pb, i=90 * u.deg):
         companion mass
     pb : astropy.units.Quantity
         Binary orbital period
-    i : astropy.units.Quantity
+    i : astropy.coordinates.Angle or astropy.units.Quantity
         orbital inclination
 
     Returns
@@ -1841,9 +1843,9 @@ def a1sini(mp, mc, pb, i=90 * u.deg):
         \\frac{a_p \sin i}{c} = \\frac{m_c \sin i}{(m_p+m_c)^{2/3}}
         G^{1/3}\\left(\\frac{P_b}{2\pi}\\right)^{2/3}
 
-    See [1]_
+    More details in :ref:`Timing Models`.  Also see [8]_
 
-    .. [1] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.21, 8.22, 8.27
+    .. [8] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.21, 8.22, 8.27
 
     """
     return (
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1427,13 +1427,13 @@ def mass_funct2(mp: u.Msun, mc: u.Msun, i: u.deg):
     Inclination is such that edge on is ``i = 90*u.deg``
     An 'average' orbit has cos(i) = 0.5, or ``i = 60*u.deg``
 
-    Calculates :math:`m_c\sin^3 i / (m_c + m_p)^2`
+    Calculates :math:`m_c^3\sin^3 i / (m_c + m_p)^2`
     """
     return (mc * np.sin(i)) ** 3.0 / (mc + mp) ** 2.0
 
 
 @u.quantity_input
-def pulsar_mass(pb: u.d, x: u.cm, mc: u.Msun, inc: u.deg):
+def pulsar_mass(pb: u.d, x: u.cm, mc: u.Msun, i: u.deg):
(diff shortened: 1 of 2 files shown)


dlakaplan:
Also using `quantity_input` https://docs.astropy.org/en/stable/api/astropy.units.quantity_input.html so functions look like:
```python
@u.quantity_input
def OMDOT(mp: u.Msun, mc: u.Msun, pb: u.d, e: u.dimensionless_unscaled):
```
There does not need to be any extra input checking.  if (for example) `mp` has no units or has the wrong units (something that cannot be converted into mass) it will raise an error.

src/pint/utils.py
```diff
@@ -1580,21 +1599,206 @@ def companion_mass(pb, x, inc=60.0 * u.deg, mpsr=1.4 * u.solMass):
     return x1.to(u.Msun)
 
 
-def ELL1_check(A1, E, TRES, NTOA, outstring=True):
+@u.quantity_input
+def PBDOT(mp: u.Msun, mc: u.Msun, pb: u.d, e: u.dimensionless_unscaled):
+    """Post-Keplerian orbital decay PBDOT.
```
aarchiba, line 1604: Could you please add a short description - something like "compute the amount of orbital decay due to gravitational-wave losses assuming GR" - so it will be more obvious to non-expert pulsar folks what this is computing?

src/pint/utils.py
```diff
@@ -1580,21 +1599,206 @@ def companion_mass(pb, x, inc=60.0 * u.deg, mpsr=1.4 * u.solMass):
...
+        * f
+        * (mp * mc)
+        / (mp + mc) ** (1.0 / 3)
+    )
+    return value.to(u.s / u.s)
+
+
+@u.quantity_input
+def GAMMA(mp: u.Msun, mc: u.Msun, pb: u.d, e: u.dimensionless_unscaled):
+    """Post-Keplerian time dilation and gravitational redshift GAMMA
```
aarchiba, line 1653: I think this one could really use some explanation - in particular that it has no relationship to the PPN parameter gamma, and also how it enters into the PPK equations of motion?

tests/test_utils.py
```diff
@@ -899,10 +903,170 @@ def test_companion_mass_error_noquantity_Pb():
...
+def test_pbdot_error_noquantity_mc():
+    Mp = 1.3381 * u.Msun
+    Mc = 1.2489 * u.Msun
+    Pb = 0.10225156248 * u.d
+    e = 0.0877775
+    with pytest.raises(TypeError):
+        PBDOT(Mp, Mc.value, Pb, e)
+
+
+def test_pbdot_error_noquantity_pb():
```
aarchiba, line 1051: I think it's probably fair to assume that for anything using `quantity_input` the type checking works.

tests/test_utils.py
```diff
@@ -899,10 +903,170 @@ def test_companion_mass_error_noquantity_Pb():
...
+    Mp = 1.3381 * u.Msun
+    Mc = 1.2489 * u.Msun
+    Pb = 0.10225156248 * u.d
+    e = 0.0877775
+    with pytest.raises(TypeError):
+        PBDOT(Mp, Mc, Pb.value, e)
+
+
+def test_A1SINI():
```
aarchiba, line 1060: Is it worth this being a parametrized calculation? Like, it's a round-trip check, so it should work for any reasonable combination of input values. With `pytest.mark.parametrize` you could quickly check three or four such combinations; with `hypothesis` you could make it try lots of combinations (but it doesn't directly support units and is really aggressive about finding failures).

src/pint/utils.py
dlakaplan: sure. done.

src/pint/utils.py
dlakaplan: also done.  note that this (and the others) do include citations to the literature

tests/test_utils.py
dlakaplan: Agree.  I wrote some of the tests first before I switched to `quantity_input`, so I just kept them, but they could be removed for simplicity if that's better.

tests/test_utils.py
dlakaplan: Sure.  I added in `hypothesis`-based tests.

src/pint/utils.py
```diff
@@ -1580,21 +1620,246 @@ def companion_mass(pb, x, inc=60.0 * u.deg, mpsr=1.4 * u.solMass):
     return x1.to(u.Msun)
 
 
-def ELL1_check(A1, E, TRES, NTOA, outstring=True):
+@u.quantity_input
+def PBDOT(mp: u.Msun, mc: u.Msun, pb: u.d, e: u.dimensionless_unscaled):
```
aarchiba, line 1624: pep8 says functions should be named in lowercase, with words separated by underscores. We can break the rules if there's a good reason, and we don't always follow them in the existing code base regardless, but would a lowercase version be more inconvenient or confusing?

src/pint/utils.py
```diff
@@ -1580,21 +1620,246 @@ def companion_mass(pb, x, inc=60.0 * u.deg, mpsr=1.4 * u.solMass):
...
+    """
+    return (
+        (mc * np.sin(i))
+        * (const.G * (pb / (2 * np.pi)) ** 2 / (mp + mc) ** 2) ** (1.0 / 3)
+    ).to(pint.ls)
+
+
+@u.quantity_input
+def ELL1_check(
+    A1: u.cm, E: u.dimensionless_unscaled, TRES: u.us, NTOA: int, outstring=True
```
aarchiba, line 1847: Lowercase would be preferable for arguments, for pep8 reasons, unless there's a strong reason to use uppercase. I could see an argument for A1 or E, but NTOA and TRES needn't be uppercase.

src/pint/utils.py
```diff
@@ -1580,21 +1620,246 @@ def companion_mass(pb, x, inc=60.0 * u.deg, mpsr=1.4 * u.solMass):
...
+        * (-192 * np.pi / 5)
+        * f
+        * (mp * mc)
+        / (mp + mc) ** (1.0 / 3)
+    )
+    return value.to(u.s / u.s)
+
+
+@u.quantity_input
+def GAMMA(mp: u.Msun, mc: u.Msun, pb: u.d, e: u.dimensionless_unscaled):
```
aarchiba, line 1686: Do these behave okay with arrays? I'm guessing they just broadcast each thing against all the others and then it works, but it might be good to check it in the test cases and then say so.

src/pint/utils.py
dlakaplan: That's what I had originally, but I changed these to match the table you made:
https://nanograv-pint.readthedocs.io/en/stable/timingmodels.html
for the PK parameters.  So I don't feel strongly either way, and would be happy to stick with lowercase.

src/pint/utils.py
dlakaplan: I didn't write this function, just added decorators and improved the comments a bit (I think they need a reference or better explanation as well, but I didn't dig one up).  So I would be fine with lowercase but am wary about changing the API on other people's code.

src/pint/utils.py
aarchiba: I'd rather lowercase. I know the upper-case versions are more standard for the parameters themselves (though I'm not sure how much of that is a PINT convention - I don't think tempo/tempo2 care?) but that might even be a good thing, helping to keep the function names separate from the parameters?

aarchiba:
Is it maybe a good idea if these functions link to the big parameter table (or its description) and/or the PulsarBinary docstring, and vice versa?

aarchiba:
https://nanograv-pint--1088.org.readthedocs.build/en/1088/api/pint.utils.html#pint.utils.PBDOT Docs are here.

We should think about the the order of this file - functions appear in the order they are listed in `__all__`, so that should probably either be topical or flat-out alphabetic.

src/pint/utils.py
dlakaplan: What are you asking particularly?  The function or the `quantity_input`?  Using `quantity_input` works.  I just added some explicitly array-based tests for the `A1SINI` function that also uses `companion_mass` or `pulsar_mass`.  And the function itself is just math.  Is there some aspect you are suspicious of?

dlakaplan:
> Is it maybe a good idea if these functions link to the big parameter table (or its description) and/or the PulsarBinary docstring, and vice versa?

This would be good (certainly the big parameter table - not sure that I see how it would best work with `PulsarBinary`).  How is the big table generated (it's automated, right)?  Is there a way to put in anchors for the various parameters (to make the linking easy)?

src/pint/utils.py
aarchiba: No, I was pretty sure they ran fine with arrays - but when implementing a heap of Keplerian stuff, I found that converting between eccentric, mean, and true anomalies needed an iterative process that prevented functions from working array-wise. So it's helpful if functions have a short note to indicate whether they work array-wise to save the user guessing.

dlakaplan: OK, can do.  Since these are direct calculations there is no difficulty (but I have had similar experiences with the various anomalies).

dlakaplan:
> https://nanograv-pint--1088.org.readthedocs.build/en/1088/api/pint.utils.html#pint.utils.PBDOT Docs are here.
> 
> We should think about the the order of this file - functions appear in the order they are listed in `__all__`, so that should probably either be topical or flat-out alphabetic.

I rearranged slightly, but I think if we move somewhere else that might make sense.  Could use the `orbital` module (not sure about `Kepler.py` since some of these are PK).  Is there an objection to putting them in `orbital/__init__.py`?

aarchiba:
> > Is it maybe a good idea if these functions link to the big parameter table (or its description) and/or the PulsarBinary docstring, and vice versa?
> 
> This would be good (certainly the big parameter table - not sure that I see how it would best work with `PulsarBinary`). How is the big table generated (it's automated, right)? Is there a way to put in anchors for the various parameters (to make the linking easy)?

Adding links from the big table to here is a little awkward as I don't want to make the table too big to render okay. They could certainly be made into link targets, but it would take a little jiggery pokery in the `docs/_ext/paramtable.py`. So I thought it might be simpler to just link to/from the section that has the table in it? Sphinx links are confusingly documented but I think it should just be
```
More details in `Section Title`_
```

aarchiba:
> > https://nanograv-pint--1088.org.readthedocs.build/en/1088/api/pint.utils.html#pint.utils.PBDOT Docs are here.
> > We should think about the the order of this file - functions appear in the order they are listed in `__all__`, so that should probably either be topical or flat-out alphabetic.
> 
> I rearranged slightly, but I think if we move somewhere else that might make sense. Could use the `orbital` module (not sure about `Kepler.py` since some of these are PK). Is there an objection to putting them in `orbital/__init__.py`?

That's a good thought. And the stuff in `kepler.py` is not friendly but does do some related sorts of calculation. 

I'm just looking through the list, though, and while some of the stuff in `pint.utils` is really "plumbing" like `taylor_horner`, and some is clearly orbital like these, you also have `pint.utils.pmtot`, which calculates the total proper motion, or `pint.utils.p_to_f` or `pint.utils.pulsar_B` - functions to let users calculate handy quantities. "Pulsar pocket calculator"  functions, if you will. They're not necessarily orbital, but maybe it makes a kind of sense to put all those in a common location?

dlakaplan:
Yes, I was looking at some of those too.  Another possibility would be to break `utils.py` into its own module.  Maybe something like:
* `utils/binaries.py`: the various binary-related ones
* `utils/general.py`: things like age, B, Edot
* `utils/__init__.py`: keep the rest in here, to not break existing imports and if there isn't a critical mass for its own file?

dlakaplan:
OK, references to `Timing Models` work now.  Also fixed the footnotes (which were conflicting).

src/pint/utils.py
aarchiba: The right answer for the anomalies is probably either to explicitly write an array-friendly version or use `np.vectorize`. I've tried to build a third approach where functions only ever have to deal with a 1D array and the decorator handles all the broadcasting and reshaping but I never came up with anything really satisfactory.

aarchiba:
I think Where Stuff Should Go is more of a telecon sort of discussion, but I'm inclined to say let's merge this when it's ready and discuss moving things after.

tests/test_utils.py
```diff
@@ -899,10 +903,224 @@ def test_companion_mass_error_noquantity_Pb():
...
+    arrays(float, 10, elements=floats(0.1, 90)),
+)
+def test_a1sini_Mc_array(Mp, Mc, Pb, i):
+    """test a1sini by looking for consistency with companion mass calculation with array input."""
+    Mp = Mp * u.Msun
+    Mc = Mc * u.Msun
+    Pb = Pb * u.d
+    i = i * u.deg
+    x = a1sini(Mp, Mc, Pb, i=i)
+    assert (np.isclose(Mc, companion_mass(Pb, x, i=i, mp=Mp))).all()
```
aarchiba, line 1105: `numpy.testing.allclose(thing, other_thing)` provides better error reporting if the things aren't the same shape or aren't all close.

dlakaplan: good to know.

src/pint/utils.py
```diff
@@ -1433,56 +1499,48 @@ def pulsar_mass(pb, x, mc, inc):
...
     This forms a quadratic equation of the form:
-    ca*Mp**2 + cb*Mp + cc = 0
+    :math:`ca M_p^2 + cb M_p + cc = 0``
 
     with:
-    ca = massfunct
-    cb = 2 * massfunct * mc
-    cc = massfunct * mc ** 2 - (mc * sini) ** 3
+
+    - :math:`ca = {\\rm massfunct}``
```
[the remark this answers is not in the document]
dlakaplan: yes, although it seemed to render OK.  but fixed anyway.

dlakaplan:
I just added another convenience function `omdot_to_mtot` to invert the `omdot` calculation.  And a test to check the round-trip:
(Mp,Mc,Pb,e) -> omdot -> Mp+Mc

(again, useful for comparing and predicting and just making sure you get the algebra/units right without looking up stuff).

I think this is good to go.

aarchiba:
Nice addition! One could easily imagine routinely using it in a list of derived quantities from a fit.

dlakaplan:
Yes, exactly.

scottransom:
Note that it might be worthwhile to check to make sure that these useful calcs can accept uncertainties parameters (from the uncertainties module).  Given that many use parfile fit results, which have uncertainties, it is super nice if the returned values (like Mtot from OMDOT etc) would automatically give a good uncertainty.  It might be that astropy quantities don't play nice with uncertainties....

aarchiba:
> Note that it might be worthwhile to check to make sure that these useful calcs can accept uncertainties parameters (from the uncertainties module). Given that many use parfile fit results, which have uncertainties, it is super nice if the returned values (like Mtot from OMDOT etc) would automatically give a good uncertainty. It might be that astropy quantities don't play nice with uncertainties....

I'd like to see these able to produce derivatives as well - if we want to use them for DDGR we'll need those. But for now I think having them at all is a benefit. 

If we're hoping to feed these the output of a fit, can the uncertainties module handle correlations between parameters? We'd presumably want to use those as well. It might be easier, if we wanted uncertainties, to just generate a thousand parameter sets from the returned parameter covariance matrix, apply these functions array-wise, and take the standard deviations.

aarchiba:
In implementation terms, if we had to we could upgrade `quantity_input` with a PINT wrapper that supported value-with-uncertainty or automatic-differentiation objects in addition to Quantities.

scottransom:
Yeah, the uncertainties module takes into account correlations between params (via derivatives).  It is really sweet.  And @paulray is using already to produce uncertainties in summary outputs post-fit.  However, I just checked and the `@u.quantity_input` and astropy quantities, in general, don't play well with uncertainties...  :-(

aarchiba:
> Yeah, the uncertainties module takes into account correlations between params (via derivatives). It is really sweet. And @paulray is using already to produce uncertainties in summary outputs post-fit. However, I just checked and the `@u.quantity_input` and astropy quantities, in general, don't play well with uncertainties... :-(

One option would be for `pint.utils` to provide a version of `quantity_input` that behaved exactly like this one but also had a `.raw` attribute so that `pulsar_B(...)` required units but `pulsar_B.raw(...)` required floats, presumed to be in the right units. That wouldn't help with the return values, though. We might be better off implementing our own `_raw` versions, automatically tested against the non-`_raw` versions. (The non-`_raw` versions can benefit from internal units checking to catch our own coding errors; or for a complicated enough function they could just convert to the right input units, feed them through the `_raw` version, and apply the right unit to the result.) Then we could feed uncertainties or other python-magic quantities into the `_raw` version.

In any case I think this is out of scope for this PR; I think this is sufficiently useful - including the units - that we should merge it and solve the how-to-use-uncertainties later.

aarchiba:
filed a feature request on Astropy: https://github.com/astropy/astropy/issues/11987

We could rather easily produce a decorator `pint.quantity_input` that called `quantity_input` and added a `.raw` attribute to the resulting decorated function.

PR #1089, opened 2021-07-26
Make fit_NGC6440E into documentation

aarchiba:
In #1071 we had some problems where changes broke `fit_NCG6440E.py` but it was never tested as part of CI. This change makes it part of our documentation - as a tutorial, rendered as a notebook - and as a result failures to run will result in CI failures. This can still be run as a `.py` file (and is downloadable from the tutorial page) but will also be visible in our documentation.

I also updated certain code to more modern approaches.

### DIFF
--- a/docs/examples/fit_NGC6440E.py
+++ b/docs/examples/fit_NGC6440E.py
@@ -1,46 +1,95 @@
 #! /usr/bin/env python
-"""Demonstrate use of pint in a script."""
-from __future__ import print_function, division
-import pint.toa
-import pint.fitter
-import pint.residuals
-import pint.models as mm
+# ---
+# jupyter:
+#   jupytext:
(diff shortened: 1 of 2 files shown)


aarchiba:
The notebooks are the only test this could really break and those are fine - we're having ephemeris server problems :-P 

New tutorial at: https://nanograv-pint--1089.org.readthedocs.build/en/1089/examples/fit_NGC6440E.html

aarchiba:
The bogus `get_summary` result is fixed in #1071.

aarchiba:
Ready for review and merging.

dlakaplan:
This looks fine to me.  It's minor enough that I can merge if nobody objects.

aarchiba:
Go for it!

PR #1090, opened 2021-07-27
Update downloads to use astropy; also add CI check for NANOGrav ephem server

aarchiba:
Astropy as of 4.0 has a better system for cached downloads of ephemeris data than we do, so use that.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -28,6 +28,9 @@ jobs:
             tox_env: 'py38-test-cov'
           - os: ubuntu-latest
             python: '3.8'
+            tox_env: 'ephemeris_connection'
+          - os: ubuntu-latest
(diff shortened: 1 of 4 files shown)


aarchiba:
Ready for review and merging.

luojing1211 (approved):
Is the "Check_ephemeris_connection.py" run by tox or CI? If the connection is down, what type of massage it emits?

aarchiba:
> Is the "Check_ephemeris_connection.py" run by tox or CI? If the connection is down, what type of massage it emits?

You can run it yourself with tox, and it's part of a full tox run. But our CI runs individual tox runs as tasks, and it is also run as part of our CI, as you can see in "Show all checks". If it fails, the `urlopen` line emits an uncaught exception, so you get an exception and traceback, and the CI job fails.

PR #1093, opened 2021-07-28
Switch from astropy logging to python logging

### LINKED ISSUE #1075: Astropy's logging functions are not meant for others to use
Since we encountered some strange behaviour and checked with them, the Astropy development team have clarified that no one but astropy should use the astropy logging mechanisms:

https://github.com/astropy/astropy/pull/11562

We should switch to plain python logging. (There are better logging frameworks, but since we are a library we should be able to work with whatever log handling the top-level application's logging uses; the plain python system lets us emit log messages in a way that all such systems can cope with.)

aarchiba:
Astropy has clarified that their logging system is meant only for code that is part of astropy itself. This PR switches to plain python logging, using per-module loggers so that users can select the log level on a per-module level.

Closes #1075

### DIFF
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -29,6 +29,7 @@ See :ref:`Timing Models` for more details on how PINT's timing models work.
 import abc
 import copy
 import inspect
+import logging
 from collections import OrderedDict, defaultdict
 from functools import wraps
 from warnings import warn
@@ -36,7 +37,6 @@ from warnings import warn
 import astropy.time as time
 import astropy.units as u
 import numpy as np
-from astropy import log
 from scipy.optimize import brentq
 
 import pint
@@ -53,6 +53,8 @@ from pint.models.parameter import (
 from pint.phase import Phase
 from pint.toa import TOAs
 from pint.utils import PrefixError, interesting_lines, lines_of, split_prefixed_name
+
+log = logging.getLogger(__name__)
 
 __all__ = [
     "DEFAULT_ORDER",
@@ -1664,7 +1666,7 @@ class TimingModel:
                 "max"     - print all lines from both models whether they are fit or not (note that nodmx will override this); DEFAULT
                 "med"     - only print lines for parameters that are fit
                 "min"     - only print lines for fit parameters for which diff_sigma > threshold
-                "check"   - only print significant changes with astropy.log.warning, not as string (note that all other modes will still print this)
+                "check"   - only print significant changes with logging.warning, not as string (note that all other modes will still print this)
 
         Returns
         -------
@@ -1682,7 +1684,7 @@ class TimingModel:
 
         Note
         ----
-            Prints astropy.log warnings for parameters that have changed significantly
+            Prints logging warnings for parameters that have changed significantly
             and/or have increased in uncertainty.
         """
         import sys
@@ -1851,7 +1853,6 @@ class TimingModel:
                         except (ValueError, AttributeError):
                             newstr += " {:28f}".format(otherpar.value)
                         if otherpar.value != par.value:
-                            sys.stdout.flush()
                             if par.name in ["START", "FINISH", "CHI2", "NTOA"]:
                                 if verbosity == "max":
                                     log.info(
@@ -1873,7 +1874,6 @@ class TimingModel:
                                     % par.name
                                 )
                                 newstr += " !"
-                            log.handlers[0].flush()
                         if (
                             par.uncertainty is not None
                             and otherpar.uncertainty is not None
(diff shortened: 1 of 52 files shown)


aarchiba:
This would probably benefit from a short section in the docs on "controlling PINT output" that demonstrates per-module log levels and keeping amounts of output reasonable. I'm not sure where, though.

aarchiba:
One should be able to do:
```python
logger.getLogger("pint.fitter").setLevel("DEBUG")
```
to obtain a really stupendous amount of output from PINT's fitters without getting a similarly large amount of output from the rest of PINT.

aarchiba:
Docs: https://nanograv-pint--1093.org.readthedocs.build/en/1093/controlling-logging.html#id1

luojing1211:
I think this can be merged after the conflict gets resolved.

src/pint/models/timing_model.py
```diff
@@ -1851,7 +1853,6 @@ def compare(
                         except (ValueError, AttributeError):
                             newstr += " {:28f}".format(otherpar.value)
                         if otherpar.value != par.value:
-                            sys.stdout.flush()
```
paulray, line 1854: These flushes were needed because the warnings came out in the wrong order compared to the strings being written to stdout.  Does the new logging mechanism fix this somehow without all the flushes?

src/pint/pintk/plk.py
```diff
@@ -2,32 +2,28 @@
...
+import tkinter as tk
+import tkinter.filedialog as tkFileDialog
+import tkinter.messagebox as tkMessageBox
+from tkinter import ttk
+
+log = logging.getLogger(__name__)
 
-log.setLevel("INFO")
-log.debug("This should show up")
+# FIXME: do not set logging in an importable module. Why is this here? settings should go in the scripts/
```
paulray, line 24: The setLevel line was added by @aarchiba  in 2019, and the log.debug line was added by @champagne-cmd . I agree that they can probably both be removed.

src/pint/scripts/pintk.py
```diff
@@ -160,7 +158,7 @@ def main(argv=None):
     )
     args = parser.parse_args(argv)
 
-    log.setLevel("WARNING")  # only display warnings and higher in the terminal
+    logging.getLogger().setLevel("INFO")
```
paulray, line 161: Why did this change from WARNING to INFO? @champagne-cmd set the default level to WARNING to reduce the unnecessary messages when running pink that cluttered the output

src/pint/models/timing_model.py
aarchiba: Logging material may well not be emitted to stdout at all, for example being written to a log file. If you want something on stdout, write it there.

src/pint/scripts/pintk.py
aarchiba: It's kind of moot now that pintk has the level setting. But this was copying the level from the INFO in `plk.py`, which I think may be the one that took effect. But this could go back to WARNING if that's the right default.

src/pint/models/timing_model.py
paulray: That may be a better way to do it, but deleting the flushes without changing the logging to prints is definitely going to produce wrong results.  So I'd say either change the logging to print statements (with color to make them visible among the other output) or restore the flushes and file an issue that there is a better way to do it.

aarchiba: Restoring the flushes is a problem because `log.handlers[0].flush` is liable to break if anyone uses non-trivial logging arrangements - the `.handlers` attribute is no part of the documented interface, and goodness knows which one will be element zero in the list. Intermixing logging and writing to stdout and expecting them to appear in the same place in any particular order is a mistake. What was the goal with this intermixing?

aarchiba: As a concrete example, if this code were tested, it is unlikely the test would be able to detect the messages in the intended order, because `pytest` redirects both stdout and log messages. (This modified version of the code does not fail any test.)

paulray: Hmm. It looks like things have changed so maybe this is moot. I'm pretty sure that compare() used to print the output, but now it returns a list of strings, and the user has to print them manually.  So, the flushes don't actually do anything. The way it is now, the warnings all come out first. They never come out close to the line that they refer to, which is bad.  The solution, as you say, is to not use any logging in this code but just add the warning messages to the string. I don't know how to make them stand out with colors in that case, though.

src/pint/scripts/pintk.py
champagne-cmd: Actually, I don't think the INFO in plk.py did anything - the setLevel called within pintk took precedence, maybe because the setLevel wasn't included in the actual function called by pintk.py. I meant to delete that setLevel in plk.py.

src/pint/models/timing_model.py
aarchiba: Why does the order matter? Do the warnings not contain enough information for us to know what they refer to? That's a problem if they are going into the logs (which the rest of the output isn't).

I don't think we can rely on colours, because (for example) the `TimingModel.compare()` output goes into the summary PDF in a `verbatim` environment.

aarchiba: Perhaps this should go into an issue about how `TimingModel.compare` works? It doesn't seem like this PR affects the problem but it sounds like the comparison is not operating ans designed.

src/pint/scripts/pintk.py
```diff
@@ -158,7 +158,7 @@ def main(argv=None):
     )
     args = parser.parse_args(argv)
 
-    logging.getLogger().setLevel("INFO")
+    logging.getLogger().setLevel("DEBUG")
```
paulray, line 161: The default log level is supposed to be WARNING for pintk

src/pint/models/timing_model.py
paulray: Yeah, it is fine to delete the flushes since they aren't working.  We can consider separately how compare should work.

src/pint/scripts/pintk.py
aarchiba: Sorry I knew that but put this in instead.

paulray:
I think this is fine to merge

PR #1096, opened 2021-07-29
New tutorial for simulating data and making a mass-mass plot

dlakaplan:
Nothing too fancy.  Uses `pint.toa.make_fake_toas` to simulate data for PSR B1534+12 using the par file from Fonseca et al. (2014).  And then various utilities to get the post-Keplerian constraints.

So it's not quite the same (no intensive Shapiro delay campaigns, etc).  But still results in a nice plot.

### DIFF
--- a/docs/examples/Simulate_and_make_MassMass.py
+++ b/docs/examples/Simulate_and_make_MassMass.py
@@ -1,0 +1,340 @@
+# ---
+# jupyter:
+#   jupytext:
+#     formats: ipynb,py:percent
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.11.2
+#   kernelspec:
+#     display_name: Python 3 (ipykernel)
+#     language: python
+#     name: python3
+# ---
+
+# %% [markdown]
+# # Generate fake data on a relativistic double neutron star (PSR B1534+12)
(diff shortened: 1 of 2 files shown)


aarchiba:
This is great! I wonder if we should make the section "Tutorials" into "Tutorials and examples"? Then Tutorials would be actual pedagogical tutorials, and Examples would be miscellaneous cool things not necessarily focused on teaching readers?

aarchiba:
I notice you had to change the par file so filed #1097 and #1098.

aarchiba:
I think some of the formatting is mangled - notebook text areas are Markdown and support $\dot f$-style math (with dollar signs), and I *think* those are correctly translated by Sphinx.

dlakaplan:
Yes, in the readthedocs build the formatting is weird.  Not sure if this is the same inconsistency we've been seeing elsewhere.  But the notebook itself renders fine.  I could remove the math, or wait and see how this appears on the full build.

aarchiba:
If it's okay in the notebook, then it *should* be okay in local doc builds, and it *should* be okay in readthedocs (which is the one that matters most, realistically). I'm doing a local build to try to see what happens locally; if that's different from readthedocs I'll see if I can figure out why.

aarchiba:
Yeah it's fine locally but broken on readthedocs.

aarchiba:
Hooray coverage works! But is completely unhelpful here as we don't collect coverage information from the notebook runs.

PR #1102, opened 2021-07-29
Move utils.py to derived_quantities.py

dlakaplan:
This follows our discussion from 2021-07-28.  Most of the "calculator" functions from `utils.py` are now in `derived_quantities.py`. I also moved the relevant tests from `tests/test_utils.py` to `tests/test_derived_quantities.py`.  Made a few other associated changes as well.


aarchiba:
Looks good to me. Do you think we need some kind of deprecation something? A message in `pint.utils` docs, decorated copies in `pint.utils` that emit a DeprecationWarning then work, something?

dlakaplan:
A deprecation warning would be useful in the docs.  Not sure about decorated copies, but could do.

aarchiba:
I think `tox -e notebooks` lets you rest the notebook execution at home, but might produce spurious changes in the `.py` files for the notebooks.

src/pint/derived_quantities.py
```diff
@@ -0,0 +1,864 @@
+"""Functions to compute various derived quantities"""
```
aarchiba, line 1: It might be good to put a little further explanation here what we mean by "derived quantities"; this docstring goes at the top of the module doc page.

dlakaplan: Look at the latest (when the docs build). I put in a warning for `utils.py` and added better docstrings for this module.  I did it manually (maybe there is a way to automate that) but it's a useful way of keeping track of the different tools (as we were discussing)

aarchiba:
Just a quick check, is this based on a version of `master` that includes my (supposed) codecov fix? I see the codecov post isn't working but it can only work if both the base and tip generate usable coverage reports.

src/pint/derived_quantities.py
aarchiba: We can generate a table, like this, just fine. I'll look into it, and maybe we should file a bug report suggesting that so I don't forget. What I was suggesting was just a short couple of sentences to the effect that this module is for computing various physical pulsar-related quantities from other physical pulsar-related quantities.

dlakaplan:
> Just a quick check, is this based on a version of `master` that includes my (supposed) codecov fix? I see the codecov post isn't working but it can only work if both the base and tip generate usable coverage reports.

Yes, I could have done that, but I figured an explicit list wouldn't hurt.

aarchiba:
I created #1104 to as a reminder and to collect notes on how to fix the documentation; it includes links to Astropy docs that use tables and describes how they do it and how we might do it.

src/pint/derived_quantities.py
aarchiba: I think now we have #1105 maybe leave the short summaries out? https://nanograv-pint--1105.org.readthedocs.build/en/1105/_autosummary/pint.utils.html#module-pint.utils

dlakaplan: Yes, that looks good. Will do.

Issue #942, opened 2021-01-10
No way to access data files for example code


aarchiba:
There is no easy way for users to locate example `.par` and `.tim` files, for example if they want to run our example notebooks or other sample code. The tests internally have a function meant to locate suitable `.par` and `.tim` files but these files are not installed as part of PINT and they can't easily be located from user code. 

It would be nice if the following worked, for any user with a working PINT installation:
```python
m = get_model(example_directory() / "NGC6440E.par")
```
One could then update the tests to behave this way also.

dlakaplan:
Yes, this is actually something I was thinking of more broadly.  Can't we have the sample data installed in a similar way to how the ephemeris etc. are (make a `data/` stub in the installation area)?

aarchiba:
The `setup.py`/`setup.cfg` machinery does have support for installing files that aren't python source code; we use this for observatory clock files already. We could arrange to have the example notebooks (?), python files (?), and input data installed somewhere when users `pip install`, and make a function `example_directory()` that knew how to locate them.

aarchiba:
> Yes, this is actually something I was thinking of more broadly. Can't we have the sample data installed in a similar way to how the ephemeris etc. are (make a `data/` stub in the installation area)?

We certainly can; I have forgotten how we did it for the ephemeris stuff, maybe our code already just picks up every non-python file lying around? If it lives inside our source tree I believe we can find it programmatically using `__file__` and the like.

dlakaplan:
In another project I'm part of it's installed through `pyproject.toml` which has a line:
```
include = ["vasttools/data", "example_notebooks"]
```
And then we reference it with:
```
BASE_MOC_PATH = pkg_resources.resource_filename(
    __name__, "./data/mocs/"
)
```
I think we can do something similar.

dlakaplan:
OK, I have a WIP to address this in #1103 .  I haven't moved all of the files there but it works and we could choose appropriate name(s) to make `examples/`, `tests/`, etc all work.

abhisrkckl:
This feature is now available via `pint.config.examplefile`.

dlakaplan:
Yes, this is done

PR #1103, opened 2021-07-29
Putting in an installable data location

dlakaplan:
This helps address #942.

I made a directory:
`src/pint/data`.  All  files in there will get installed into `pint/data/`.  I added to the top-level `config.py` a definition:
```
pint.config.datadir()
pint.config.examplefile(filename)
pint.config.runtimefile(filename)
```
All of data files from `examples` are in `data/examples`.  All of the clock etc. files are in `data/runtime`.  They are accessed by the functions above.


aarchiba:
We might find that there are some files in our test data that are intentionally broken or weird or otherwise too specialized to be very useful for users. So it might be worth keeping a few (non-duplicate) files around in the test data directory (though lots of tests rely on `os.chdir()` there and might break). We might prefer that the files in our examples directory were "nice", not emitting EPHVER warnings or anything like that, while some tests might make use of more "challenging" files.

dlakaplan:
Definitely.  This is more a proof of concept.

aarchiba:
API-wise, do we want to make the thing that reports the data location a function instead of a variable? In case, down the road, we realize we need to include more plumbing to handle, say, being inside a conda package or a wheel or a zip file or some other unusual situation? Unfortunately unlike classes, modules can't have variables that are secretly computed by calling a function. We might also need to do some magic with `os.path.abspath()` or something to make sure the user can access the data files from wherever they are.

dlakaplan:
My impression was that `pkg_resources` handled most of the various installation methods for us. That's what I got from:
https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html

> Accessing Data Files at Runtime
> 
> Typically, existing programs manipulate a package’s __file__ attribute in order to find the location of data files. However, this manipulation isn’t compatible with PEP 302-based import hooks, including importing from zip files and Python Eggs. It is strongly recommended that, if you are using data files, you should use the ResourceManager API of pkg_resources to access them. The pkg_resources module is distributed as part of setuptools, so if you’re using setuptools to distribute your package, there is no reason not to use its resource management API. See also Importlib Resources for a quick example of converting code that uses __file__ to use pkg_resources instead.

It's possible that I'm still not doing it right.  Adding an `abspath` would obviously be trivial.  Not sure about the rest.

aarchiba:
Maybe a couple of tests, that just check that the files are accessible through `datadir` regardless of the current directory? Something like
```python
def test_accessible(tmpdir):
    d = os.getcwd()
    try:
        os.chdir(tmpdir)
        assert os.path.exists(os.path.join(pint.datadir, "some.tim"))
    finally:
        os.chdir(d)
```
(I keep meaning to figure out how to use pathlib https://docs.python.org/3/library/pathlib.html )

aarchiba:
> My impression was that `pkg_resources` handled most of the various installation methods for us. That's what I got from:
> https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html
> 
> > Accessing Data Files at Runtime
> > Typically, existing programs manipulate a package’s **file** attribute in order to find the location of data files. However, this manipulation isn’t compatible with PEP 302-based import hooks, including importing from zip files and Python Eggs. It is strongly recommended that, if you are using data files, you should use the ResourceManager API of pkg_resources to access them. The pkg_resources module is distributed as part of setuptools, so if you’re using setuptools to distribute your package, there is no reason not to use its resource management API. See also Importlib Resources for a quick example of converting code that uses **file** to use pkg_resources instead.
> 
> It's possible that I'm still not doing it right. Adding an `abspath` would obviously be trivial. Not sure about the rest.

I guess I'm just a little nervous about whether that `resource_filename` works for directories, and whether it's safe to run it once and for all at import time (or if a later change in directory would result in a different return value, say). Accessing pint data through a function, `pint.data_file("thing.tim")` or `pint.data_dir()` kind of future-proofs the API I think?

dlakaplan:
I guess an advantage of a class or a function is that it can have a doc string.

dlakaplan:
Not sure if this is an issue (I've seen other packages do it with a variable) but it also doesn't hurt.  So I defined two functions as you suggest:
`pint.data_dir()`
`pint.data_file(filename)`

setup.cfg
```diff
@@ -41,6 +41,9 @@ install_requires =
 [options.packages.find]
 where = src
 
+[options.package_data]
+* = *.tim, *.par
```
aarchiba, line 45: Do any of our examples use any other kind of file? It'd be preferable if we could specify a directory and everything in it for inclusion, rather than by extension. Not really a problem, though, as long as there's a note somewhere pointing to this as the rule.

dlakaplan: Pretty sure we can.  I think everything in `src/pint/data` could easily be included.  So maybe a better solution is to make that whole directory installed, and at the same time try to remove all of the separate copies of the par/tim files and update the code (as much as possible) to use this one location.  

A down side is that if we have intentionally bad files it would make them more visible (as you said).  But a fix for that would be a README file (or similar) describing the origin of each file.  In fact that would generally be very useful.

aarchiba: I agree that a README would be useful, but I think it's okay for the intentionally-bad files to live in the testing directory somewhere. I suspect that most of the intentionally broken files are created for a single test, or at most a few tests in a single file, and for these I think including them, triple-quoted, and using StringIO might be sufficient (and is there any reason to use a giant, broken, file? most brokenness can probably be tested with a short stub par or tim file). If there are any that are intentionally broken and widely used we probably need to revisit this discussion. 

Another category we need to figure out what to do with is the files for which TEMPO or TEMPO2 output is stored for comparison - do we want to make them and the comparison output part of our distribution, or should that remain internal to PINT testing? (These tests are a headache when we see a discrepancy - who knows why it happens or what causes it... but I agree we have a need for them.)

So I think in the short term I'm inclined to use `pint/data/*` for the things we need for our examples, and keep the testing datadir for weird stuff that doesn't make sense for anyone else to use. Longer-term, I think we should try to shrink the weird testing-only stuff, maybe to zero, but that's part of a longer-term "make our tests less horrible" project.

dlakaplan:
I moved all of the tim and par files from `examples/` to the new location. I verified that the example notebooks build (at least locally) and added a test script (not sophisticated).

There are still data files in:
* `profiling/`
* `validation/`
* `tests/datafile/`
Based on the discussion above, maybe we won't touch the `tests/` now.  But we could move the others if desired. 

Also added a `README.md` file in the `data/` directory, which can be filled in

dlakaplan:
Are the files in `validation/` used anywhere?

dlakaplan:
OK, looks like they are used by the code in `validation/`, so will leave them there.

src/pint/__init__.py
```diff
@@ -105,3 +107,31 @@
...
+    ----------
+    filename : str
+
+    Returns
+    -------
+    str
+        Full path to the requested file
+
+    """
+    return os.path.join(datadir(), filename)
```
aarchiba, line 137: Would this be better as `...resource_filename("data/" + filename)`? I know the `pkg_resources` docs make a big deal about how resource names aren't file names...

dlakaplan:
`docs/examples-rendered/paper_validation_example.py` seems to want `J1600-3053_NANOGrav_11yv1.gls.par`/`J1600-3053_NANOGrav_11yv1.tim` which I can't find.

tests/test_datafiles.py
```diff
@@ -0,0 +1,30 @@
+"""Test installation of PINT data files"""
+import os
+import unittest
+import pint
+
+
+class TestDatafiles(unittest.TestCase):
```
aarchiba, line 7: This would be nicer in a `pytest` style - just write a function that asserts whatever it is. Or if you want to share the file list, make that a fixture:
```python
@pytest.fixture
def files():
     ... stuff ...
     return list_of_filenames
```
Or maybe:
```python
files = """stuff"""
list_of_files = [f.strip() for f in files.split("\n")]

@pytest.mark.parametrize("filename", list_of_files)
def test_can_read(filename):
    open(filename).read()
```
to get separate tests if you think that's useful.

tests/test_datafiles.py
```diff
@@ -0,0 +1,30 @@
...
+        J1614-2230_NANOGrav_12yv3.wb.tim
+        NGC6440E.par
+        NGC6440E.tim
+        PSRJ0030+0451_psrcat.par
+        waves.par
+        waves_withpn.tim"""
+
+        cls.files = [f.strip() for f in files.split("\n")]
+
+    def test_datafiles(self):
```
aarchiba, line 28: Maybe have a test that uses `try: ... finally:` to switch to a temporary directory and see if it can still access the files? Like, is our retrieval mechanism going to work okay from unexpected directories? If one were especially distrustful one could try retrieving the filename, changing the directory, and then opening the file. But I do think it's important we check this mechanism doesn't rely on any particular location being the current directory.

src/pint/__init__.py
dlakaplan: Not sure that I fully grasp the difference, but sure.

tests/test_datafiles.py
dlakaplan: OK, sure.  Easier to read.

tests/test_datafiles.py
dlakaplan: I don't understand why this needs a `try...`. But I have added tests to change to temp directories both before and after determining the file names.

aarchiba: The try/finally is to ensure that the current working directory is restored whether the test succeeds or fails, or if some other event (control-C for example) occurs in the middle of it. Ideally there would be a context manager to do this but I haven't found one in the Python standard library.

src/pint/__init__.py
aarchiba: I don't think there should be a difference; it should all be moot. I don't think PINT can be installed from a ZIP file (but maybe?) so all of them probably just return the file name. But I'm trying to figure out how to use `pkg_resources` as close to the way it is intended as possible.

dlakaplan:
What about: https://nanograv-pint.readthedocs.io/en/latest/api/pint.config.html
that has very similar language (data files) but looks for the files needed at runtime rather than for examples.  Is there a way to clarify?

dlakaplan:
Tests are improved as suggested.  I also added a `docs/data-files.rst` file to explain how to use this.  However, a few lingering issues:
* I would like the `pint.datadir` and `pint.datafile` functions to appear in the readthedocs.  But I don't see them.  Still working on how to trigger that (I've put references to them but the links don't work)
* I would like to reference `examples/time_a_pulsar.ipynb` from `docs/data-files.rst`, but that isn't working either.  I get a warning about a missing link.  Clearly linking there *is* possible, but my syntax isn't right.  Do I make the target the `ipynb` file?  `html`?  `py`?

dlakaplan:
New doc file: https://nanograv-pint--1103.org.readthedocs.build/en/1103/data-files.html

aarchiba:
Maybe instead of going under Tutorials the new information could go under "How to Work with PINT example files"? Though I guess that's more about how to work with them during PINT development. 

It does look like they do appear in the docs for the module `pint`, so they're in the right place in the reference docs.

dlakaplan:
Do you mean:
https://nanograv-pint.readthedocs.io/en/latest/working-with-notebooks.html
My read on that was that it was more about writing notebooks, not using them.  I put the info under Tutorials since it can be useful there too, and that's the first bit that shows up.  But if you/others have additional suggestions I can change.

I did get them included in the docs.

dlakaplan:
To try to keep things clear, docs for `datadir`/`datafile` include disambiguation refs to `datapath` and vice-versa, but it's still confusing.  
* https://nanograv-pint--1103.org.readthedocs.build/en/1103/api/pint.config.html#pint.config.datapath
* https://nanograv-pint--1103.org.readthedocs.build/en/1103/api/pint.html#pint.datadir
Could we change one of the names?  Maybe make the `pint.config.datapath` into `pint.config.runtimedatapath` or similar?

aarchiba:
> Do you mean:
> https://nanograv-pint.readthedocs.io/en/latest/working-with-notebooks.html
> My read on that was that it was more about writing notebooks, not using them. I put the info under Tutorials since it can be useful there too, and that's the first bit that shows up. But if you/others have additional suggestions I can change.
> 
> I did get them included in the docs.

The idea was that someone would go looking at Tutorials when they didn't really know what PINT could do or how to go about it, which this isn't really - but maybe putting it in as part of the actual Tutorials page? "Tutorials - here are some walkthroughs of things you can do with PINT; the data files to use them should be installed as part of your copy of PINT and accessed through ... so you should be able to download and run these from anywhere convenient." perhaps?

I agree Howtos right now is all about contributing to PINT, but I was hoping we'd get things like "I need to use this par file to predict absolute phase for VLBI, how do I make PINT do that?"

dlakaplan:
Is the latest version of:
https://nanograv-pint--1103.org.readthedocs.build/en/1103/tutorials.html
along the lines of what you were suggesting?

aarchiba:
> Is the latest version of:
> https://nanograv-pint--1103.org.readthedocs.build/en/1103/tutorials.html
> along the lines of what you were suggesting?

I think that's good. I actually meant something less specific - something like:

These tutorials are in the form of Jupyter notebooks, downloadable from a link at the top of each page. (Also available in the same place is a plain-python script version, in case this is more convenient.) You should be able to download these files and run them from anywhere convenient (provided PINT is installed). The data files they use are installed as part of PINT and accessed through...

dlakaplan:
I added some of that to `tutorials.rst`, and also added some relevant info to https://nanograv-pint--1103.org.readthedocs.build/en/1103/working-with-notebooks.html.

aarchiba:
> What about: https://nanograv-pint.readthedocs.io/en/latest/api/pint.config.html
> that has very similar language (data files) but looks for the files needed at runtime rather than for examples. Is there a way to clarify?

Is it possible to combine the mechanisms, so that they both work the same way and draw from the same places? I think having two different kinds of installed data files accessed in two different ways is confusing, and documenting it is a weak second-best to making the mechanisms simpler. Is there any reason the examples can't go in with the runtime data files? If the runtime files have a search path they can be found along, well, that's probably okay for the examples too, isn't it?

I should say that our handling of, say, clock correction files is far from optimal - we can't necessarily use whichever of $TEMPO, $TEMPO2, or PINT-installed versions is newest. So if you needed to change the file-location system for that, that would be okay.

dlakaplan:
I have now included the clock files etc.:
* `data/examples/...` accessed via `pint.config.examplefile()`
* `data/runtimefile/...` accessed via `pint.config.runtimefile()`

I had to move these functions to the `pint.config` module, since otherwise there was a circular import (`pint` module imports `pint.pulsar_ecliptic` which needs files from `runtimefile` directory).  So this is slightly clumsier for the user, but I did import them at the top level so you can access them via `pint.examplefile()` like before - just the docs don't show up where I want.

Does that seem reasonable?

aarchiba:
> I have now included the clock files etc.:
> 
>     * `data/examples/...` accessed via `pint.config.examplefile()`
> 
>     * `data/runtimefile/...` accessed via `pint.config.runtimefile()`
> 
> 
> I had to move these functions to the `pint.config` module, since otherwise there was a circular import (`pint` module imports `pint.pulsar_ecliptic` which needs files from `runtimefile` directory). So this is slightly clumsier for the user, but I did import them at the top level so you can access them via `pint.examplefile()` like before - just the docs don't show up where I want.
> 
> Does that seem reasonable?

I have very mixed feelings about these convenience imports - as you say, the functions aren't where the docs say they are. We have some elsewhere, most notably `get_model`, and it's confusing to refer to them in the docs. Maybe we can live with `from pint.config import examplefile`? Then the examples have access to the files with just `examplefile()`?

dlakaplan:
I largely agree, and those imports are _actually_ why I needed to move the functions to `pint.config` in the first place (since imports of other functions led to circular import errors otherwise).

So for simplicity I can just keep them in `pint.config` and update the docs, if that's the preference.

aarchiba:
I think it's fine for these to live in `pint.config`.

dlakaplan:
Probably best to remove the convenience imports (can we remove them elsewhere too?) and just go for a simple version.  Can always add those in later

Issue #1104, opened 2021-07-29
PINT documentation can be unwieldy


aarchiba:
PINT has a number of cases where modules and/or classes start to have a lot of functions, and the current doc formatting documents each one after the other: https://nanograv-pint.readthedocs.io/en/latest/api/pint.utils.html

Astropy, for comparison, produces tables/lists: https://docs.astropy.org/en/stable/utils/index.html

I think the tables/lists would provide a more manageable overview of what PINT can do.

aarchiba:
PINT uses `sphinx.ext.autodoc` (and `napoleon`) to turn docstrings into RST; so does Astropy and everything else. 

PINT uses `sphinx-apidoc` to construct (every time) a tree of framework `.rst` files, one for each module, that invoke `autodoc` to fill in the docstrings. (I think this has the side effect of forcing `make` to regenerate everything every time you run it.)

Astropy has a manually constructed set of `.rst` files that, for each module, invoke `sphinx.ext.automodapi`; this turns into a table of functions/class methods that have the one-line descriptions and a link to a separate page for each member with the documentation details. (Personally I find the manually written docs a little hard to navigate.)

I *think* that what I'd like to do is replace PINT's use of `sphinx-apidoc` with something that generated a tree of rst files, each of which contains an `automodapi` command. But some experimentation will be needed.

aarchiba:
I think one can probably make `sphinx-apidoc` do the right thing - use `automodapi` - by supplying modified templates like: https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/templates/apidoc/module.rst_t

One could also, maybe, make our `run_apidoc` generate the API in a scratch directory then use a copy-over-only-if-changed approach to avoid confusing `make`. (Maybe this wouldn't matter, since the docstrings need to be interpolated and that might not be sensible about changes.)

aarchiba:
Incidentally, the main slow part of `make docs` is that it tries to execute all the notebooks. I'm not sure how feasible it would be to skip this step, say based on modification times, or allowing a manual skip.

aarchiba:
This gives examples of doing many of our documentation tasks: https://github.com/JamesALeedham/Sphinx-Autosummary-Recursion/tree/master/docs

PR #1105, opened 2021-07-30
WIP: Docs now use tables

aarchiba:
This is an attempt to address #1104 by arranging for the autogenerated PINT docs to construct tables of functions.

Currently it (accidentally) breaks inclusion of notebooks, which has the side effect of dramatically accelerating the process of running Sphinx.


aarchiba:
Doc example: https://nanograv-pint--1105.org.readthedocs.build/en/1105/_autosummary/pint.utils.html#module-pint.utils

dlakaplan:
That's great - will make things easier to read.  The notebooks breaking is bad/good, but hopefully fixable

(on that note: can we identify which notebooks take the longest and make them optional somehow?)

aarchiba:
I think the notebooks are fixed now, this branch should be okay to merge.

#1107 has some ideas on improving the notebook build experience.

dlakaplan:
This looks good.  It also looks like each function/class is now it's own page.  This makes some things easier but does make the menu bar on the left much longer.  Was that intentional?

aarchiba:
"Intentional" is a bit strong. I might have preferred something more like the class pages, where the function table links to the combined function docstrings at the bottom of the same page, but it's not clear to me that `autosummary` can be made to do that. (It generates the module and class pages from templates that are part of the PR, and it has templates for the single-function pages, but I'm not sure it can include the function docs on the module page.) 

For me, the way the left panel works is that it opens out the list of functions in a module only if you're actually looking at the module page or a function in it, and it's sort of handy the way it lets you jump around different functions in the same module.

dlakaplan:
Same here for the left panel - it is handy, but if you open a section with a lot of classes it can be *long*.  But overall I think this is very good.

dlakaplan:
I haven't dug in to the detailed changes (I don't speak Sphinx) but I can merge this if nobody else (@luojing1211 ?) is going to

PR #1112, opened 2021-08-04
Fix allow_local for nicer and others

paulray:
This fixes what I believe is a bug in event_toas.py.
It is supposed to set allow_local to True for the missions that have their own TOA loading method, but it was not working and was generating and ERROR message saying that NICER didn't support spacecraft local TOAs, which is not true.

### DIFF
--- a/src/pint/event_toas.py
+++ b/src/pint/event_toas.py
@@ -83,7 +83,14 @@ def create_mission_config():
 
     # Allow local TOAs for those missions that have a load_<MISSION>_TOAs method
     for mission in ["nustar", "nicer", "xte"]:
-        mission_config[mission]["allow_local"] = True
+        try:
+            # If mission was read from HEASARC, just update allow_local
+            mission_config[mission]["allow_local"] = True
+        except KeyError:
+            # If mission was not read from HEASARC, then create full new entry
+            mission_config[mission] = {}
+            mission_config[mission].update(mission_config["generic"])
+            mission_config[mission]["allow_local"] = True
 
     # Fix xte
     mission_config["xte"]["fits_columns"] = {"ecol": "PHA"}
--- a/src/pint/event_toas.py
+++ b/src/pint/event_toas.py
@@ -68,12 +68,6 @@ def create_mission_config():
         }
     }
 
-    # Allow local TOAs for those missions that have a load_<MISSION>_TOAs method
-    for mission in ["nustar", "nicer", "xte"]:
-        mission_config[mission] = {}
-        mission_config[mission].update(mission_config["generic"])
-        mission_config["allow_local"] = True
-
     # Read the relevant information from $HEADAS/bin/xselect.mdb, if present
     for mission, data in read_mission_info_from_heasoft().items():
         mission_config[mission] = {"allow_local": False}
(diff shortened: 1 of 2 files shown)


paulray:
@matteobachetti  could you look at this and see if you agree?

paulray:
I just pushed one other fix for @kerrm to look at.  I don't think prob=0 in sig2sigma should be a value error. Do you agree?

kerrm:
What happens when you give it 0?  Formally 0 <--> infinity sigma.

paulray:
> What happens when you give it 0? Formally 0 <--> infinity sigma.

```
In [7]: pint.eventstats.sig2sigma(0)                                                                                                                                                                                                                              
/Users/paulr/src/PINT/src/pint/eventstats.py:106: RuntimeWarning: divide by zero encountered in log
  x0 = (-2 * np.log(sig[mask] * (np.pi) ** 0.5)) ** 0.5
/Users/paulr/src/PINT/src/pint/eventstats.py:107: RuntimeWarning: invalid value encountered in true_divide
  results[mask] = x0 - np.log(x0) / (1 + 2 * x0)
Out[7]: nan
```
Maybe it should be made to return `np.inf`?

paulray:
The reason I'm trying to fix sig2sigma is that I got that exception and I think it was when the sig value underflowed to 0.

kerrm:
I'm fine with it returning np.inf.  It would just need to be vector valued, e.g. results[sig==0] = np.inf.

If you're dealing with small probs, you might want to try the logprob argument, if you can get the significance that way.  If something upstream is returning 0 for the p-value, it might be better to deal with it there, though.

aarchiba:
Annoyingly, scipy has `scipy.stats.norm.logsf` and `scipy.stats.norm.isf` but doesn't have `scipy.stats.norm.ilogsf`, which would make this much easier. Are there test cases for `sig2sigma` and `sigma2sig`? Having one for zero probability would be helpful, to specify this behaviour and make sure it works.

src/pint/event_toas.py
```diff
@@ -87,6 +81,17 @@ def create_mission_config():
         mission_config[mission]["fits_extension"] = ext
         mission_config[mission]["fits_columns"] = cols
 
+    # Allow local TOAs for those missions that have a load_<MISSION>_TOAs method
```
aarchiba, line 84: Can you use getattr to check for the existence of a `f"load_{mission}_TOAs` method directly, and save maintaining this explicit list?

paulray: That isn't trivial since the mission names are different than the load_XXX_TOAs names. Also, I'm not sure this whole thing is needed. The only purpose is to set `allow_local` and the only place that is used it to emit an error message that does not stop the code from running:
```
    if not mission_config[mission]["allow_local"] and timesys != "TDB":
        log.error("Raw spacecraft TOAs not yet supported for " + mission)
```
Maybe @matteobachetti has some future plans for `allow_local`?

paulray:
OK, I reverted the sig2sigma change. I'll see about fixing it upstream to make sure no 0 probabilities get passed.

aarchiba:
> OK, I reverted the sig2sigma change. I'll see about fixing it upstream to make sure no 0 probabilities get passed.

I think they should both be upgraded to work well with log probabilities (that might be too small for the probability to represent in floating-point - that's not that many sigma...). `scipy.stats.logsf` gets you one direction but the other is more cumbersome. And it's fine to consider those functions out of scope for this PR.

matteobachetti:
@aarchiba @paulray sorry for showing up late, I'm on vacation.
In Stingray we have a bunch of functions to treat sigmas and other significance measurements, taking care of most of the computing and rounding errors for small probabilities. Maybe it helps? 
https://github.com/StingraySoftware/stingray/blob/v0.2/stingray/stats.py

paulray:
Thanks @matteobachetti, on this PR the main questions for you were about the allow_local and if my fixes for that look OK. Sorry to have pestered you while you are on vacation!

matteobachetti:
@paulray I have no future plans for allow_local, it was just probably a bad design on my part :)

Issue #1113, opened 2021-08-05
companion_mass definition doesn't match call in fitter.py


nategarver-daniels:
There appears to be a mismatch in the way that derived_quantities.companion_mass is defined vs. how it is called in fitter.py:

```def companion_mass(pb: u.d, x: u.cm, i=60.0 * u.deg, mp=1.4 * u.solMass):```

vs the two calls in fitter.py (only one shown):

```
mcmed = pint.derived_quantities.companion_mass(
                        self.model.PB.quantity,
                        self.model.A1.quantity,
                        inc=60.0 * u.deg,
                        mpsr=1.4 * u.solMass,
                    )
```

PR #1114, opened 2021-08-05
Fix for improper arguments in fitter.get_summary()

dlakaplan:
The signature for `companion_mass` changed and it didn't change in `fitter.get_summary()`.  This wasn't caught by testing since the summary was only printed for a non-binary pulsar, and the companion/pulsar mass lines are only called for binaries (need to be `ELL1` and have certain parameters free).

This has been fixed, and a new test (`test_fitsummary_binary()`) has been added to `tests/test_fitter.py` to check for it.


dlakaplan:
Fixed #1113

nategarver-daniels:
Looks good, I've confirmed that this fixes the issue I was having.

PR #1115, opened 2021-08-05
Fix polycos

paulray:
The calls to model.phase in the polyco computation were missing abs_phase=True, so this add that.

### DIFF
--- a/src/pint/event_toas.py
+++ b/src/pint/event_toas.py
@@ -83,7 +83,14 @@ def create_mission_config():
 
     # Allow local TOAs for those missions that have a load_<MISSION>_TOAs method
     for mission in ["nustar", "nicer", "xte"]:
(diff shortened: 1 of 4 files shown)


paulray:
Oops, this accidentally includes the edit from #1112

aarchiba:
Can you include a test that shows what was going wrong when `abs_phase` wasn't being passed? Like, what happened? Can you write a test that shows it happening and that it was wrong?

paulray:
I think this is fine. The `test_polycos.py` is updated in this PR as well. It now ensures that phases computed with polycos match phases computed with `model.phase(ts,abs_phase=True)`, rather than `model.phase(ts,abs_phase=False)` which was giving meaningless absolute phases.

aarchiba:
I see, thanks!

Issue #848, opened 2020-12-03
random models is broken in pintk


scottransom:
I think that Note that #814 broke the random models code in pintk, since it changed the behavior of `make_fake_toas`.  If @clp3ef has a bit of time to check it, I suspect that it is probably a quick fix. If not, I'll try to take a look over the weekend.

clp3ef:
I looked and the issue is the default observatory used to make the fake TOAs was changed from "GBT" to "@". Changing the default obs in the function header back to "GBT" fixes the issue. @aarchiba, was there a reason the default observatory was switched to "@", or a reason it shouldn't be "GBT"?

paulray:
In my version I just changed `obs="@"` to `obs="GBT"` in `make_fake_toas()` and that didn't fix the problem for me.

scottransom:
@clp3ef hmmm...  that's strange.  I don't see why the default observatory should matter.  Because we want fake TOAs to compare with the timing model we are looking at.  And those TOAs should be able to come from any observatory.

clp3ef:
@paulray Huh, that's really strange. I just reset my changes and then changed the observatory to GBT and it fixes it for me, and there's no other edits made to any files. @scottransom If its something that isn't hard to check, does changing the observatory in make_fake_toas() make the random models in pintk work for you?

paulray:
Here is my diff, and what the pintk plot looks like when I run with random_models on the NGC6440E example.
```
% git diff
diff --git a/src/pint/toa.py b/src/pint/toa.py
index 2fd01c5f..734f663e 100644
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -514,7 +514,7 @@ def format_toa_line(
 
 
 def make_fake_toas(
-    startMJD, endMJD, ntoas, model, freq=999999, obs="@", error=1 * u.us
+    startMJD, endMJD, ntoas, model, freq=999999, obs="GBT", error=1 * u.us
 ):
     """Make evenly spaced toas with residuals = 0 and  without errors
 
```
![image](https://user-images.githubusercontent.com/1771428/101412392-8b748180-38b0-11eb-93bb-b6d1a64d7728.png)

scottransom:
I was just going to report the exact same thing.

clp3ef:
Huh, okay, I think that's a different issue with the pintk random models? For the NGC64440E example, turning off fitting for DM seems to fix the random models, so it goes from 
[...]

PR #1118, opened 2021-08-08
Fix random_models.py

champagne-cmd:
Updated random_models.py to use new covariance matrix machinery. Fixes issue #848 where random models didn't work properly with the DM parameter.
One thing to discuss: Should random_models.py be collapsed into the new calculate_random_models in utils.py? The random_models.py returns time residuals, while the calculate_random_models returns phase residuals.

### DIFF
--- a/src/pint/random_models.py
+++ b/src/pint/random_models.py
@@ -42,19 +42,24 @@ def random_models(
         TOAs object containing the evenly spaced fake toas to plot the random lines with
         list of residual objects for the random models (one residual object each)
     """
-    params = fitter.model.get_params_dict("free", "num")
-    mean_vector = params.values()
+    param_values = fitter.model.get_params_dict("free", "num")
+    cov_matrix = fitter.parameter_covariance_matrix
+    # this is a list of the parameter names in the order they appear in the coviarance matrix
+    param_names = cov_matrix.get_label_names(axis=0)
+    mean_vector = np.array([param_values[x] for x in param_names if not x == "Offset"])
     # remove the first column and row (absolute phase)
-    cov_matrix = (((fitter.covariance_matrix.matrix[1:]).T)[1:]).T
-    fac = fitter.fac[1:]
+    if param_names[0] == "Offset":
+        cov_matrix = cov_matrix.get_label_matrix(param_names[1:])
+        fac = fitter.fac[1:]
+        param_names = param_names[1:]
+    else:
+        fac = fitter.fac
     f_rand = deepcopy(fitter)
     mrand = f_rand.model
 
     # scale by fac
-    log.debug("errors", np.sqrt(np.diag(cov_matrix)))
-    log.debug("mean vector", mean_vector)
-    mean_vector = np.array(list(mean_vector)) * fac
-    cov_matrix = ((cov_matrix * fac).T * fac).T
+    mean_vector = mean_vector * fac
+    scaled_cov_matrix = ((cov_matrix.matrix * fac).T * fac).T
 
     toa_mjds = fitter.toas.get_mjds()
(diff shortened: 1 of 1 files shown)


aarchiba:
> Updated random_models.py to use new covariance matrix machinery. Fixes issue #848 where random models didn't work properly with the DM parameter.
> One thing to discuss: Should random_models.py be collapsed into the new calculate_random_models in utils.py? The random_models.py returns time residuals, while the calculate_random_models returns phase residuals.

Can you describe the differences between them? If that's the only difference then conversion is so easy I don't think both functions should exist. If there are more differences then perhaps it makes sense to integrate them into one function that can do both.

aarchiba:
Do you have tests that verify that this works now? Even just running it at all would be a help.

scottransom:
I talked to @clp3ef about this a bit, but I don't think she has had a chance to look into it.  But the codes are really similar, and do almost identical things.  So I certainly think that both places could use the new code.

aarchiba:
A simple test would be structural - hand it a timing model and a set of TOAs and ensure that the result was the right shape and had the right units. 

A more sophisticated could take a simple model with F0 and F1 and no correlation and check the random residuals against pen-and-paper uncertainties at various times. One could check that the results were within five sigma, or check that the mean and standard deviation of a modest sample were roughly right. (The latter would require some modest statistical deviousness, and any random operation should take a random seed or random state object for test repeatability.)

aarchiba:
Discussion says: absolutely combine the two. Is it possible for them to generate (lists of) Residuals objects? That would ensure that the conversion between time and phase was handled as well as we know how.

champagne-cmd:
It seems importing the Residuals class in utils.py creates a circular import as residuals.py imports `weighted_mean` from utils.py. So not sure if we can create a list of residual objects within calculate_random_models. I will go ahead and try to implement the input for choosing phase or time resid, and if another solution presents itself we can discuss and modify.

champagne-cmd:
There is something I'm a little puzzled by. In random_models.py, two fake toa objects are created:
```
x = toa.make_fake_toas(
        minMJD - spanMJDs * ledge_multiplier,
        maxMJD + spanMJDs * redge_multiplier,
        npoints,
        mrand,
    )
x2 = toa.make_fake_toas(minMJD, maxMJD, npoints, mrand)
```
and the second fake TOAs object is only used like so:
```
rs = mrand.phase(x, abs_phase=True) - fitter.model.phase(x, abs_phase=True)
rs2 = mrand.phase(x2, abs_phase=True) - fitter.model.phase(x2, abs_phase=True)
# from calc_phase_resids in residuals
rs -= Phase(0.0, rs2.frac.mean() - rs_mean)
```
The comment says "from calc_phase_resids in residuals", but I'm a little lost why two TOAs objects are used or the best way to incorporate this into calculate_random_models, as that method only uses one fake TOAs object.

aarchiba:
I have to say I would have just implemented this by jittering the model and computing Residuals objects for each jittered version of the model. That would only need one set of TOAs, whether real or fake, at which to show the effect of model changes. But this method seems to work, though I don't understand what it would mean to subtract phases evaluated at different times. The issue may be getting the constant offset right?

paulray:
> It seems importing the Residuals class in utils.py creates a circular import as residuals.py imports `weighted_mean` from utils.py. So not sure if we can create a list of residual objects within calculate_random_models. I will go ahead and try to implement the input for choosing phase or time resid, and if another solution presents itself we can discuss and modify.

Does this method help? https://stackoverflow.com/questions/22187279/python-circular-importing

aarchiba:
This does highlight a problem with PINT: PINT's TimingModels can capture parameter uncertainties but not parameter correlations. So we need a Fitter here just for the covariance matrix.

Issue #1122, opened 2021-08-12
Faking TOAs: zima vs. PINT


dlakaplan:
I'm investigating the two ways to fake TOAs: `pint.toa.make_fake_toas` and `zima`.  So far I don't see them as consistent. 

Procedure: calculate fake TOAs for a simple setup with (hopefully) identical input parameters.  Compare results.

A few notes: inside `zima`, it repeatedly calculates residuals manually via:
```
F_local = m.d_phase_d_toa(t_zima)
r_zima_manual = m.phase(t_zima).frac / F_local
```
It then uses those to adjust the TOAs.  So, just to make sure that's not the issue I am using that method together with the packaged `pint.residuals.Residuals` method.  Note that I'm doing this for the standard `'NGC6440E.par`, so a very simple model.

Commands (20 TOAs, 1 us error):
```
t_pint = make_fake_toas(55000, 58000, 20, model=m, freq=1400 * u.MHz, dm=10 * pint.dmu, obs='GBT')
t_pint.adjust_TOAs(TimeDelta(error * np.random.randn(len(t_pint))))
```
And zima:
`zima --freq 1400 --error 1.0 --ntoa 20 --startMJD 55000.00000000675 --duration 2999.9999999875 --ephem DE421 --obs GBT .../NGC6440E.par zima.tim`

Results.  Note that the 
```
PINT rms (manual calculation): 1.9251282477811997 us
ZIMA rms (manual calculation): 8240.313330113022 us
PINT rms (PINT calculation): 0.9677796770673149 us
ZIMA rms (PINT calculation): 5148.585677973798 us
```
Using the "manual" style residual calculation is worse than the standard `pint` version.  By about a factor of 2. The PINT/PINT version gives the expected 1us rms.

I also followed the `zima` script in plotting the residuals.  The PINT-generated values seem OK: with the "manual" style residuals there is an offset, but they are flat.  The zima-generated values are all over the place.  If I run zima and use `--plot` its output looks similar.

![random_toas](https://user-images.githubusercontent.com/5092062/129224296-3e317006-4631-4f0f-8e44-b0f90590e843.png)

paulray:
Yes, I was just using zima to make some fake TOAs for testing pintk. It seems buggy. Something is broken in it. This is a very good reason to have it use `make_fake_toas()`, which can be tested in CI.

dlakaplan:
[...]

Issue #1095, opened 2021-07-28
Add ability to simulate noise in Pint TOAs


Hazboun6:
A few of us developed code that is basically the Pint rewrite of the noise simulation pieces of [toasim.py](https://github.com/vallis/libstempo/blob/master/libstempo/toasim.py) in `libstempo`. This was developed by myself, @andrealommen and some of her students for adding noise to NICER X-ray TOAs, but should be generally applicable to any TOAs. 

New methods include:
```Python
make_ideal
add_efac
add_equad
add_ecorr
add_dm_rednoise
add_rednoise
```
I think the names are fairly understandable to those who are familiar with these common types of pulsar timing data noise. Both red noise methods add in time-correlated noise constructed using a Fourier basis. The `add_dm_rednoise` method adds in that noise with a dependence on the observation frequency of 1/RF^2, but that could easily be generalized to any 1/RF^p dependence. 

This does not include any gravitational wave simulation code, but that could definitely be added in the future. 

I'll post the code in a separate comment below. There is a chance the code is a bit dated at this point, but I'm happy to update and add this into Pint. 

Two questions:

1.  Given that I'll write some tests and update the code, is this something folks would like to see a part of the package? 
2. Where is the best place for this? I could follow the `libstempo` naming and call it `toa_sim`, or start a `noise.py`. There is the `toa.make_fake_toas` method, but it seems like there is plenty of code in there already.

Hazboun6:
The current state of the code is [here](https://github.com/Hazboun6/pta_sim/blob/master/pta_sim/pint_sim.py).

Hazboun6:
@aarchiba is this still something that would be useful to you?

aarchiba:
This does seem useful! I had envisioned PINT's noise model objects being able to generate noise of the kind they describe - so if you had a noise model describing red noise of a particular amplitude and spectral index, or an ECORR affecting a certain category of TOAs, you could just pass those models TOAs and they would generate adjustments to reflect an instance of the noise they describe.

It seems to me that the machinery here is more or less capable of doing that - that is, that given an appropriate description of the noise model they can generate the noise. So it might be just a matter of translating from the noise model parameters in the PINT objects to the appropriate values for these functions?

Hazboun6:
This makes sense to me. One could add a method to the Base class, something like `noise_model. NoiseComponent.add_fake_noise()`, and populate that with the various methods from above. 

[...]

PR #1123, opened 2021-08-12
Fix fake toas

dlakaplan:
Making sure that `zima` works correctly, using the same `make_fake_toas` functions as other parts of the code.

This led to some changes:
* Moved the functions (and `calculate_random_models`) to a new module `pint.simulation`.  The reason for this was that including them in `toa` led to circular import worries, and the same for including them in `utils`.  Those imports should really be sorted elsewhere but I didn't want to get into that.
* Added a base function `make_fake_toas` that accepts a `TOAs` object as input.  This then has specific uses:
* `make_fake_toas_uniform` (the old function - uniformly spaced)
* `make_fake_toas_fromtim` (`zima` did this too)
* added the ability to "fuzz" the uniform MJDs (replicating `zima`) 
* added the ability to actually include the error as noise on the TOAs, not just populate a column (again, `zima` did this, but it's now optional for both)
* checked that this works with `EFAC` and `EQUAD`
* added unit tests

I updated all of the existing tests and docs to reflect these changes.  

Usages:
```
toas = pint.simulation.make_fake_toas_uniform(57001, 58000, 200, model=model, error=1 * u.us,add_noise=True )
```
or
```
m, t = get_model_and_toas(
        pint.config.examplefile("B1855+09_NANOGrav_9yv1.gls.par"),
        pint.config.examplefile("B1855+09_NANOGrav_9yv1.tim"),)
# refit the data to get rid of a trend
f = GLSFitter(t, m)
f.fit_toas()
t_sim = pint.simulation.make_fake_toas(t, f.model, add_noise=True)
```

### DIFF
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -244,7 +244,7 @@ class Residuals:
 
         Parameters
         ----------
-        calctype : str, optional
+        calctype : {'modelF0', 'numerical', 'taylor'}
             Type of calculation.  If `calctype` == "modelF0", then simply the ``F0``
             parameter from the model.
             If `calctype` == "numerical", then try a numerical derivative
@@ -354,7 +354,7 @@ class Residuals:
 
         Parameters
         ----------
-        calctype : str, optional
+        calctype : {'modelF0', 'numerical', 'taylor'}
             Type of calculation.  If `calctype` == "modelF0", then simply the ``F0``
             parameter from the model.
             If `calctype` == "numerical", then try a numerical derivative
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -346,7 +346,7 @@ class Residuals:
             mean, err = weighted_mean(full, w)
         return full - mean
 
-    def calc_time_resids(self, calctype="modelF0"):
+    def calc_time_resids(self, calctype="taylor"):
         """Compute timing model residuals in time (seconds).
 
         Converts from phase residuals to time residuals using several possible ways
@@ -354,7 +354,7 @@ class Residuals:
 
         Parameters
         ----------
-        calctype : {'modelF0', 'numerical', 'taylor'}
+        calctype : {'taylor', 'modelF0', 'numerical'}
             Type of calculation.  If `calctype` == "modelF0", then simply the ``F0``
             parameter from the model.
             If `calctype` == "numerical", then try a numerical derivative
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -17,7 +17,7 @@ from scipy.linalg import LinAlgError
 
 from pint.models.dispersion_model import Dispersion
 from pint.phase import Phase
-from pint.utils import weighted_mean
+from pint.utils import weighted_mean, taylor_horner_deriv
 
 __all__ = [
     "Residuals",
@@ -239,21 +239,24 @@ class Residuals:
         wmean, werr, wsdev = weighted_mean(self.time_resids, w, sdev=True)
         return wsdev.to(u.us)
 
-    def get_PSR_freq(self, modelF0=True):
+    def get_PSR_freq(self, calctype="modelF0"):
         """Return pulsar rotational frequency in Hz.
 
         Parameters
         ----------
-        modelF0 : bool, optional
-            If True, simply read the ``F0`` parameter from the model. If False,
-            query the model for the spin period at the time of each TOA.
+        calctype : str, optional
+            Type of calculation.  If `calctype` == "modelF0", then simply the ``F0``
+            parameter from the model.
+            If `calctype` == "numerical", then try a numerical derivative
+            If `calctype` == "taylor", evaluate the frequency with a Taylor series
 
         Returns
         -------
-        freq : Quantity
+        freq : astropy.units.Quantity
             Either the single ``F0`` in the model or the spin frequency at the moment of each TOA.
         """
-        if modelF0:
+        assert calctype in ["modelF0", "taylor", "numerical"]
+        if calctype == "modelF0":
             # TODO this function will be re-write and move to timing model soon.
             # The following is a temproary patch.
             if "Spindown" in self.model.components:
@@ -265,7 +268,12 @@ class Residuals:
                     "No pulsar spin parameter(e.g., 'F0'," " 'P0') found."
                 )
             return F0.to(u.Hz)
-        else:
+        elif calctype == "taylor":
+            # see Spindown.spindown_phase
+            dt = self.model.get_dt(self.toas, 0)
+            fterms = [0.0 * u.dimensionless_unscaled] + self.model.get_spin_terms()
+            return taylor_horner_deriv(dt, fterms, deriv_order=1).to(u.Hz)
+        elif calctype == "numerical":
             return self.model.d_phase_d_toa(self.toas)
 
     def calc_phase_resids(self):
@@ -338,11 +346,32 @@ class Residuals:
             mean, err = weighted_mean(full, w)
         return full - mean
 
-    def calc_time_resids(self):
-        """Compute timing model residuals in time (seconds)."""
+    def calc_time_resids(self, calctype="modelF0"):
+        """Compute timing model residuals in time (seconds).
+
+        Converts from phase residuals to time residuals using several possible ways
+        to calculate the frequency.
+
+        Parameters
+        ----------
+        calctype : str, optional
+            Type of calculation.  If `calctype` == "modelF0", then simply the ``F0``
+            parameter from the model.
+            If `calctype` == "numerical", then try a numerical derivative
+            If `calctype` == "taylor", evaluate the frequency with a Taylor series
+
+        Returns
+        -------
+        residuals : astropy.units.Quantity
+
+        See Also
+        --------
+        :meth:`pint.residuals.get_PSR_freq`
+        """
+        assert calctype in ["modelF0", "taylor", "numerical"]
         if self.phase_resids is None:
             self.phase_resids = self.calc_phase_resids()
-        return (self.phase_resids / self.get_PSR_freq()).to(u.s)
+        return (self.phase_resids / self.get_PSR_freq(calctype=calctype)).to(u.s)
 
     def calc_chi2(self, full_cov=False):
         """Return the weighted chi-squared for the model and toas.
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -255,7 +255,7 @@ class Residuals:
         freq : astropy.units.Quantity
             Either the single ``F0`` in the model or the spin frequency at the moment of each TOA.
         """
-        assert calctype.lower() in ["modelF0", "taylor", "numerical"]
+        assert calctype.lower() in ["modelf0", "taylor", "numerical"]
         if calctype.lower() == "modelf0":
             # TODO this function will be re-write and move to timing model soon.
             # The following is a temproary patch.
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -255,8 +255,8 @@ class Residuals:
         freq : astropy.units.Quantity
             Either the single ``F0`` in the model or the spin frequency at the moment of each TOA.
         """
-        assert calctype in ["modelF0", "taylor", "numerical"]
-        if calctype == "modelF0":
+        assert calctype.lower() in ["modelF0", "taylor", "numerical"]
+        if calctype.lower() == "modelf0":
             # TODO this function will be re-write and move to timing model soon.
             # The following is a temproary patch.
             if "Spindown" in self.model.components:
@@ -268,12 +268,12 @@ class Residuals:
                     "No pulsar spin parameter(e.g., 'F0'," " 'P0') found."
                 )
             return F0.to(u.Hz)
-        elif calctype == "taylor":
+        elif calctype.lower() == "taylor":
             # see Spindown.spindown_phase
             dt = self.model.get_dt(self.toas, 0)
             fterms = [0.0 * u.dimensionless_unscaled] + self.model.get_spin_terms()
             return taylor_horner_deriv(dt, fterms, deriv_order=1).to(u.Hz)
-        elif calctype == "numerical":
+        elif calctype.lower() == "numerical":
             return self.model.d_phase_d_toa(self.toas)
 
     def calc_phase_resids(self):
@@ -368,7 +368,7 @@ class Residuals:
         --------
         :meth:`pint.residuals.get_PSR_freq`
         """
-        assert calctype in ["modelF0", "taylor", "numerical"]
+        assert calctype.lower() in ["modelf0", "taylor", "numerical"]
         if self.phase_resids is None:
             self.phase_resids = self.calc_phase_resids()
         return (self.phase_resids / self.get_PSR_freq(calctype=calctype)).to(u.s)
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -271,7 +271,18 @@ class Residuals:
         elif calctype.lower() == "taylor":
             # see Spindown.spindown_phase
             dt = self.model.get_dt(self.toas, 0)
-            fterms = [0.0 * u.dimensionless_unscaled] + self.model.get_spin_terms()
+            # if the model is defined through F0, F1, ...
+            if "F0" in self.model.params:
+                fterms = [0.0 * u.dimensionless_unscaled] + self.model.get_spin_terms()
+
+            # otherwise assume P0, PDOT
+            else:
+                F0 = 1.0 / self.model.P0.quantity
+                if "PDOT" in self.model.params:
+                    F1 = -self.model.PDOT.quantity / self.model.P0.quantity ** 2
+                else:
+                    F1 = 0 * u.Hz / u.s
+                fterms = [0.0 * u.dimensionless_unscaled, F0, F1]
             return taylor_horner_deriv(dt, fterms, deriv_order=1).to(u.Hz)
         elif calctype.lower() == "numerical":
             return self.model.d_phase_d_toa(self.toas)
--- a/docs/examples/Simulate_and_make_MassMass.py
+++ b/docs/examples/Simulate_and_make_MassMass.py
@@ -33,6 +33,7 @@ import io
 import pint.fitter
 from pint.models import get_model
 import pint.derived_quantities
+import pint.simulation
 
 
 # %% [markdown]
@@ -208,7 +209,7 @@ error = 5 * u.us
 Ntoa = 1000
 # make the new TOAs.  Note that even though `error` is passed, the TOAs
 # start out perfect
-tnew = pint.toa.make_fake_toas(
+tnew = pint.simulation.make_fake_toas_uniform(
     tstart.mjd * u.d, tstop.mjd * u.s, Ntoa, model=m, obs="ARECIBO", error=error
 )
 # So we have to still add in some noise
--- a/docs/examples/Simulate_and_make_MassMass.py
+++ b/docs/examples/Simulate_and_make_MassMass.py
@@ -210,7 +210,7 @@ Ntoa = 1000
 # make the new TOAs.  Note that even though `error` is passed, the TOAs
 # start out perfect
 tnew = pint.simulation.make_fake_toas_uniform(
-    tstart.mjd * u.d, tstop.mjd * u.s, Ntoa, model=m, obs="ARECIBO", error=error
+    tstart.mjd * u.d, tstop.mjd * u.d, Ntoa, model=m, obs="ARECIBO", error=error
 )
 # So we have to still add in some noise
 tnew.adjust_TOAs(astropy.time.TimeDelta(np.random.normal(size=len(tnew)) * error))
--- a/docs/examples/check_phase_connection.py
+++ b/docs/examples/check_phase_connection.py
@@ -25,7 +25,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 import astropy.units as u
 
-import pint.fitter, pint.toa
+import pint.fitter, pint.toa, pint.simulation
 from pint.models import get_model_and_toas
 from pint import utils
 import pint.config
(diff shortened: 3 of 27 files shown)


aarchiba:
How about the basic function accepts a TOAs object as input? And twiddles them minimally to produce zero residuals? And optionally adds noise? I'm thinking of cases that are awkward to generate now like:

- Generate fake TOAs with two different telescopes, with different frequencies, uncertainties, and maybe observing cadences, that fit to the model
- Generate TOAs according to some observing campaign but save a lot of time by not worrying about the residuals (this comes up in testing!)
- Generate narrowband-style TOAs in clumps

Some of this can be helped by `merge_toas` but absolute phase can be a headache. So if `pint.simulate` contained a bunch of functions to produce TOAs simulating various observing campaigns, one could use `merge_toas` to get even more complex campaigns, and then feed the giant mess to the `make_residuals_zero` function. The `add_noise` function here would presumably add the claimed uncertainties; as we were talking about in #1095 we'd eventually add `model.generate_noise(toas)` to handle EQUAD/ECORR/red noise.

dlakaplan:
Having the base function operate on a TOA object (rather than an array of MJDs) could work.  Agree that would make a wider range of tests possible.

aarchiba:
I think convenience functions are a good idea, but I like the idea of having three separate parts so people can mix and match: setting up TOAs to represent a set of observations, adjusting them so that they give zero residuals, and adding noise.

dlakaplan:
Yes, that's how I have it now:
* `make_fake_toas()`: accepts TOA object
* `make_fake_toas_uniform()`: old behavior (uniformly spaced)
* `make_fake_toas_timfile()`: probably could be replaced with just the first one.

Turns out handling `EFAC`/`EQUAD` is easy with `model.scaled_toa_uncertainty()`.

dlakaplan:
Addresses #1122

dlakaplan:
@paulray : would appreciate your opinion on this (esp. the interface changes).  There is now a CI test for `zima` as well.

aarchiba:
> Yes, that's how I have it now:
> 
>     * `make_fake_toas()`: accepts TOA object
> 
>     * `make_fake_toas_uniform()`: old behavior (uniformly spaced)
> 
>     * `make_fake_toas_timfile()`: probably could be replaced with just the first one.
> 
> 
> Turns out handling `EFAC`/`EQUAD` is easy with `model.scaled_toa_uncertainty()`.

That's not quite what I meant - the three parts I had in mind were:

1. Tools to generate TOAs objects describing various observation schemes (uniform single-frequency, narrowband-like, I'm not sure what)
2. A tool to tweak TOAs to force residuals to zero
3. Tools to add noise to a TOAs object (simple claimed uncertainty, or taking into account the noise model)

The convenience functions would just string a combination of these together to save typing.

dlakaplan:
I'm not sure that the functions you describe are needed.  Or maybe not quite in that breakdown.  For instance, number 3 already exists:
```
        tsim.adjust_TOAs(time.TimeDelta(err))
```
number 2 could be its own routine, but it's also easy enough to just keep that in the first function that I described and have everything call that.  As for the first, I agree that this could be useful, but it also seems like it could easily get very specific and might be better to leave most of the complexity to users (with a few examples).  

But I'm happy to discuss further.

aarchiba:
> I'm not sure that the functions you describe are needed. Or maybe not quite in that breakdown. For instance, number 3 already exists:
> 
> ```
>         tsim.adjust_TOAs(time.TimeDelta(err))
> ```
> 
> number 2 could be its own routine, but it's also easy enough to just keep that in the first function that I described and have everything call that. As for the first, I agree that this could be useful, but it also seems like it could easily get very specific and might be better to leave most of the complexity to users (with a few examples).
> 
> But I'm happy to discuss further.

Well, for example, I've often wanted to generate uniformly-spaced TOAs but don't care what their residuals are. (And that's kind of annoying to do by hand.) More-complicated patterns likewise. I'm not suggesting writing tools to make those as part of this PR, but making room for them in the logical structure seems to make sense, as I can see good reasons for someone to want to use PINT to simulate an observing campaign. 

It sounds like your `make_fake_toas` does nothing but set the residuals to zero, which suggests that maybe it should have a name that says that?

I agree that if you have an instantiation of noise, in seconds, it's pretty easy to add it to TOAs. The third category was meant to be tools for instantiating noise - you could want plain scatter, you could want the EFAC/EQUAD from the model, or you could want full-on ECORR/red noise too.

paulray:
I'm getting an exception and I've tried a few different times but can't make it go away.  I'm running in the tests/datafile directory:
```
% zima --obs @ --freq 0.0 --startMJD 58300.0 --duration 100 --error 100.0 --addnoise --plot --ntoa 100 ngc300nicer.par ngc300fake.tim
INFO: Reading model from ngc300nicer.par [pint.scripts.zima]
INFO: TZRFRQ was 0.0 or None. Setting to infinite frequency. [pint.models.absolute_phase]
INFO: Generating uniformly spaced TOAs [pint.scripts.zima]
INFO: Using CLOCK = TT(TAI), so setting include_bipm = False [pint.simulation]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: Applying clock corrections (include_gps = True, include_bipm = False) [pint.toa]
INFO: Special observatory location. No clock corrections applied. [pint.observatory.special_locations]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: tdb column already exists. Deleting... [pint.toa]
INFO: tdbld column already exists. Deleting... [pint.toa]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: Column ssb_obs_pos already exists. Removing... [pint.toa]
INFO: Column ssb_obs_vel already exists. Removing... [pint.toa]
INFO: Column obs_sun_pos already exists. Removing... [pint.toa]
INFO: tdb column already exists. Deleting... [pint.toa]
INFO: tdbld column already exists. Deleting... [pint.toa]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: Column ssb_obs_pos already exists. Removing... [pint.toa]
INFO: Column ssb_obs_vel already exists. Removing... [pint.toa]
INFO: Column obs_sun_pos already exists. Removing... [pint.toa]
INFO: tdb column already exists. Deleting... [pint.toa]
INFO: tdbld column already exists. Deleting... [pint.toa]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: Column ssb_obs_pos already exists. Removing... [pint.toa]
INFO: Column ssb_obs_vel already exists. Removing... [pint.toa]
INFO: Column obs_sun_pos already exists. Removing... [pint.toa]
INFO: tdb column already exists. Deleting... [pint.toa]
INFO: tdbld column already exists. Deleting... [pint.toa]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: Column ssb_obs_pos already exists. Removing... [pint.toa]
INFO: Column ssb_obs_vel already exists. Removing... [pint.toa]
INFO: Column obs_sun_pos already exists. Removing... [pint.toa]
INFO: tdb column already exists. Deleting... [pint.toa]
INFO: tdbld column already exists. Deleting... [pint.toa]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: Column ssb_obs_pos already exists. Removing... [pint.toa]
INFO: Column ssb_obs_vel already exists. Removing... [pint.toa]
INFO: Column obs_sun_pos already exists. Removing... [pint.toa]
INFO: tdb column already exists. Deleting... [pint.toa]
INFO: tdbld column already exists. Deleting... [pint.toa]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: Column ssb_obs_pos already exists. Removing... [pint.toa]
INFO: Column ssb_obs_vel already exists. Removing... [pint.toa]
INFO: Column obs_sun_pos already exists. Removing... [pint.toa]
INFO: tdb column already exists. Deleting... [pint.toa]
INFO: tdbld column already exists. Deleting... [pint.toa]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: Column ssb_obs_pos already exists. Removing... [pint.toa]
INFO: Column ssb_obs_vel already exists. Removing... [pint.toa]
INFO: Column obs_sun_pos already exists. Removing... [pint.toa]
INFO: tdb column already exists. Deleting... [pint.toa]
INFO: tdbld column already exists. Deleting... [pint.toa]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: Column ssb_obs_pos already exists. Removing... [pint.toa]
INFO: Column ssb_obs_vel already exists. Removing... [pint.toa]
INFO: Column obs_sun_pos already exists. Removing... [pint.toa]
INFO: tdb column already exists. Deleting... [pint.toa]
INFO: tdbld column already exists. Deleting... [pint.toa]
INFO: Using EPHEM = DE405 for TDB calculation. [pint.toa]
INFO: Computing PosVels of observatories and Earth, using DE405 [pint.toa]
INFO: Column ssb_obs_pos already exists. Removing... [pint.toa]
INFO: Column ssb_obs_vel already exists. Removing... [pint.toa]
INFO: Column obs_sun_pos already exists. Removing... [pint.toa]
  File "/Users/paulr/env/pint/bin/zima", line 33, in <module>
    sys.exit(load_entry_point('pint-pulsar', 'console_scripts', 'zima')())
  File "/Users/paulr/src/PINT/src/pint/scripts/zima.py", line 85, in main
    ts = pint.simulation.make_fake_toas_uniform(
  File "/Users/paulr/src/PINT/src/pint/simulation.py", line 263, in make_fake_toas_uniform
    return make_fake_toas(ts, model=model, add_noise=add_noise, name=name)
  File "/Users/paulr/src/PINT/src/pint/simulation.py", line 157, in make_fake_toas
    zero_residuals(tsim, model)
  File "/Users/paulr/src/PINT/src/pint/simulation.py", line 67, in zero_residuals
    ts.adjust_TOAs(time.TimeDelta(-r.time_resids))
  File "/Users/paulr/src/PINT/src/pint/toa.py", line 1868, in adjust_TOAs
    self.compute_posvels(self.ephem, self.planets)
  File "/Users/paulr/src/PINT/src/pint/toa.py", line 2279, in compute_posvels
    sun_obs = objPosVel_wrt_SSB("sun", tdb, ephem) - ssb_obs
  File "/Users/paulr/src/PINT/src/pint/solar_system_ephemerides.py", line 167, in objPosVel_wrt_SSB
    pos, vel = astropy.coordinates.get_body_barycentric_posvel(objname, t)
  File "/Users/paulr/env/pint/lib/python3.8/site-packages/astropy/coordinates/solar_system.py", line 337, in get_body_barycentric_posvel
    return _get_body_barycentric_posvel(body, time, ephemeris)
  File "/Users/paulr/env/pint/lib/python3.8/site-packages/astropy/coordinates/solar_system.py", line 291, in _get_body_barycentric_posvel
    for body_p_or_v, p_or_v in zip(body_posvel_bary,
  File "/Users/paulr/env/pint/lib/python3.8/site-packages/jplephem/spk.py", line 232, in generate
    raise OutOfRangeError(
jplephem.exceptions.OutOfRangeError: segment only covers dates 2305424.5 through 2525008.5
```

dlakaplan:
That command works for me, although I don't have that par file.    But:
`zima --obs @ --freq 0.0 --startMJD 58300.0 --duration 100 --error 100.0 --addnoise --plot --ntoa 100 src/pint/data/examples/NGC6440E.par zima.tim`
works.  Can you share your par file?

scottransom:
Is that because of DE405 not covering the dates in question somehow?  Maybe try a different ephem?

dlakaplan:
But the date limits that it's complaining about are like 1599 and 2201.  So something is going wrong somewhere.

dlakaplan:
I can change to DE405 in the NGC6440E par file and it works fine.  Maybe some unset parameter in @paulray 's file is ending up with MJD of 0?

paulray:
The file I used is in tests/datadir in PINT

dlakaplan:
OK, I see what's happening (but don't understand why yet).  The iteration that is supposed to end up at 0 residuals is growing larger rather than smaller, so it's pulling the times to before the SSB calculations are valid.

dlakaplan:
It's to do with the really large F1.

paulray:
Ah, well its a real pulsar so we should be able to handle that case.

dlakaplan:
So as I keep the par file the same and just change the F1 over a range, the max residual behavior changes. Typically it gets smaller with each iteration and converges quickly. For larger |F1|  is takes more iterations. But then larger values still blow up.  
fake_runaway

I'll get to this more on Monday.

dlakaplan:
OK, I figured out part of it.  `residuals.time_resids` uses `get_PSR_freq()`, which doesn't correctly account for spin down.  Replacing that by a calculating involving `model.d_phase_d_toa(toas)` works (mostly) for any F1.  However:
* the apparent frequency calculated this way is a factor of 2 lower than I'd think (but I haven't thought about it hard)
* the iteration now converges, but it doesn't look particularly pretty. The residuals are still too high.

paulray:
Note that `get_PSR_freq()` has a mode to do the calculation correctly:
```
    def get_PSR_freq(self, modelF0=True):
        """Return pulsar rotational frequency in Hz.

        Parameters
        ----------
        modelF0 : bool, optional
            If True, simply read the ``F0`` parameter from the model. If False,
            query the model for the spin period at the time of each TOA.

        Returns
        -------
        freq : Quantity
            Either the single ``F0`` in the model or the spin frequency at the moment of each TOA.
```

paulray:
Also, note that we discussed adding another option to get_PSR_freq() that is in between just grabbing F0 and doing the full d_phase_d_TOA calculation, where it would compute the frequency from the F0 through Fn Taylor series. That would be perfect for this case.

dlakaplan:
Yes, I see.  So maybe it's the fact that this seems to be off by a factor of 2?  Again, will investigate more on Monday

dlakaplan:
old way with a single frequency (from the F0 of the model):
```
In [3]: r.get_PSR_freq()
Out[3]: <Quantity 0.050375 Hz>
```

This is what's needed for a high-F1 pulsar.  But note it's off by a factor of 2:
```
In [4]: r.get_PSR_freq(modelF0=False)
...
Out[4]: 
<Quantity [0.0251875, 0.0251875, 0.0251875, 0.0251875, 0.0251875,
           0.0251875, 0.0251875, 0.0251875, 0.0251875, 0.0251875] Hz>
```
This may also be why `zima` wasn't working on its own (since it used similar routines)?

dlakaplan:
OK, I think I implemented the Taylor-series spin-down as suggested, making use of other pieces.  I don't think the numerical version (which was the alternate to `modelF0`) is right: it's off by a factor of 2 and very noisy.

Now there is:
```
get_PSR_freq(self, calctype="modelF0"):
        """Return pulsar rotational frequency in Hz.

        Parameters
        ----------
        calctype : str, optional
            Type of calculation.  If `calctype` == "modelF0", then simply the ``F0``
            parameter from the model.
            If `calctype` == "numerical", then try a numerical derivative
            If `calctype` == "taylor", evaluate the frequency with a Taylor series
```

and similarly 
```
    def calc_time_resids(self, calctype="modelF0"):
        """Compute timing model residuals in time (seconds).

        Converts from phase residuals to time residuals using several possible ways
        to calculate the frequency.

        Parameters
        ----------
        calctype : str, optional
            Type of calculation.  If `calctype` == "modelF0", then simply the ``F0``
            parameter from the model.
            If `calctype` == "numerical", then try a numerical derivative
            If `calctype` == "taylor", evaluate the frequency with a Taylor series
```
(arguably we might make the Taylor version the default).  

But this works with the high-F1 pulsar Paul had above.  

Screen Shot 2021-08-13 at 8 03 16 PM

dlakaplan:
A few related comments: 
* currently `residuals.get_PSR_freq()` is used by `residuals.get_time_resids()`.  It seems to me that the `get_PSR_freq` command might work better in the `models` package (although it does depend on both the TOAs and the model). 
* In either case we might want to have a persistent `@property` associated with it in case the calculations get lengthy.  But then we'd probably want to store the calculation type.
* We may want to store the calculation type anyway, since you can access the time residuals through the `resids` property and the `time_resids` attribute, but then you don't know what type is returned.
* Is 3 ways 1 (or 2) too many to access the same result?  Function, attribute, and parameter?
* Should we update the `residuals.update()` method to allow it to select the type of residual calculation?  
* Should we change the default frequency calculation to `taylor`?  It seems that this is still pretty fast and there are minimal downsides, except that it can return multiple values (1 per TOA) which may be unexpected.

dlakaplan:
A further note: it looks like the `WidebandDMResiduals` have the `resids` and `resids_value` properties, but they mean different things.

dlakaplan:
Maybe we should use cached properties?

dlakaplan:
Another thought: rather than define functions to convert between phase and frequency, should we use astropy equivalencies?  The following (inserted into `timing_model.py`) allows conversion between s, Hz, and cycles (dimensionless) like `(0.5*u.s).to(u.dimensionless_unscaled, equivalencies=m.equivalencies())`:

(we could also define separate equivalencies for spin and orbital conversions)
```
    def frequency(self, toas=None):
        """Return pulsar rotational frequency in Hz.

        Parameters
        ----------
        toas : pint.toa.TOAs, optional
            TOAs at which frequency should be computed.  If not supplied, uses ``F0`` or ``P0``

        Returns
        -------
        freq : astropy.units.Quantity
            Either the single ``F0`` in the model or the spin frequency at the moment of each TOA.
        """
        if toas is None:
            if "Spindown" in self.components:
                F0 = self.F0.quantity
            elif "P0" in self.params:
                F0 = 1.0 / self.P0.quantity
            else:
                raise AttributeError(
                    "No pulsar spin parameter(e.g., 'F0'," " 'P0') found."
                )
            return F0.to(u.Hz)
        # see Spindown.spindown_phase
        dt = self.get_dt(toas, 0)
        fterms = [0.0 * u.dimensionless_unscaled] + self.get_spin_terms()
        return taylor_horner_deriv(dt, fterms, deriv_order=1).to(u.Hz)

    def equivalencies(self, toas=None):
        freq = self.frequency(toas=toas).to_value(u.Hz)
        return u.equivalencies.Equivalency(
            [
                (u.s, u.Hz, lambda x: 1 / x, lambda x: 1 / x),
                (u.s, u.dimensionless_unscaled, lambda x: x * freq, lambda x: x / freq),
                (
                    u.Hz,
                    u.dimensionless_unscaled,
                    lambda x: x / freq,
                    lambda x: x * freq,
                ),
            ],
            "pulsar",
            {"toas": toas},
        )
```

scottransom:
OMG...  the cycles / freqs dimensionless equivalencies are terribly confusing.  We talked about that when they came out a few years ago and rejected the idea.  There is a bunch of very strange behavior and requirements.

dlakaplan:
@paulray : the test in `tests/test_fake_toas.py::test_zima()` does in fact do what you were asking about: it writes a tim file, reads it back in, and checks the rms.  

It is working fine for me.  I can read in the tim file and compute rms/plot outside of `zima` and it is consistent. Could you be more specific with the test that you did?

paulray:
Here is what I did:
```
cd tests/datafile
zima ngc300nicer.par ngc300fake.tim --addnoise --ntoa 100 --duration 100 --freq 0.0 --startMJD 58000.0 --obs @ --plot
```
That plot looks great.  But, when I do either of the following:
```
tempo2 -gr plk -f ngc300nicer.par ngc300fake.tim
pintk ngc300nicer.par ngc300fake.tim
```
I get a plot like the attached (first plot is from zima, second is from pintk).  Looking closer, I think the error bars are actually fine, but the residuals are coming out about 3x what they were in the plot zima made.
zima
pintk

dlakaplan:
That might be the same error in `pintk` when computing the residuals

paulray:
I just tried the same exercise with the NGC6440E.par and it works fine, so indeed it must be an issue with that particular par file. And, Tempo2 gets the same issue since it also uses the par file F0 for scaling, presumably.

paulray:
Indeed, changing the default calctype to "taylor" in `get_PSR_freq()` fixes the issue.
I think that should be the default, since it will be correct for many more cases than "modelF0", but it does break from the TEMPO/Tempo2 convention.  In general, I think our philosophy has been to make PINT correct where possible, but maintain backwards compatibility for those who need it.

dlakaplan:
Thanks, Paul.  That was my intent as well (default=`taylor`), but I didn't know if that would break things that expected only a single value.

my hope is that I will do a separate PR to fix some of the frequency calculation API

dlakaplan:
I tested the DM fitting we discussed.  It works fine (I believe).  What I did was:
* look at the NGC6440E dataset.  The last 8 TOAs are different freqs on the same day, so I selected only those
* I fit the model with only DM free to those data
* I then generated N (=30) fake TOA sets matching the 8 multi-free TOAs.  For each I again fit for only DM
* I verified that the rms of the fitted DM values is comparable to the uncertainty on the DM value from the original (real) fit.

I added a test for this (`test_fake_toas::test_fake_DMfit()`)

paulray:
Any reason not to merge this and #1129 now?

dlakaplan:
The latest push broke a notebook, so I'm fixing that now

paulray:
Cool, can you also add a CHANGELOG entry?

dlakaplan:
when we merge I will

dlakaplan:
I think this is ready to merge (@paulray or @scottransom )

src/pint/residuals.py
```diff
@@ -267,7 +270,23 @@ def get_PSR_freq(self, modelF0=True):
...
+            dt = self.model.get_dt(self.toas, 0)
+            # if the model is defined through F0, F1, ...
+            if "F0" in self.model.params:
+                fterms = [0.0 * u.dimensionless_unscaled] + self.model.get_spin_terms()
+
+            # otherwise assume P0, PDOT
+            else:
+                F0 = 1.0 / self.model.P0.quantity
+                if "PDOT" in self.model.params:
+                    F1 = -self.model.PDOT.quantity / self.model.P0.quantity ** 2
```
paulray, line 284: I know this computation is sort of obvious, but it might be good form to use `derived_quantities.p_to_f()`

paulray (approved):

PR #1129, opened 2021-08-18
Fix bug in d_phase_d_toa()

### LINKED ISSUE #1128: d_phase_d_toa wrong by a factor of 2
The test for this function is also broken, since it doesn't have an absolute value

```
import pint.residuals
import io
import numpy as np
from pint.toa import get_TOAs
from pint.polycos import Polycos
from pint.toa import  make_fake_toas

model = get_model(
    io.StringIO(
    """
PSRJ           J0055-3741
RAJ            00:55:04.86
DECJ           -37:41:43.7
CLK            TT(TAI)
MODE 1
UNITS          TDB
CORRECT_TROPOSPHERE  N
PLANET_SHAPIRO N
EPHEM          DE405
TZRMJD 58349.0
TZRSITE @
TZRFRQ 0.0
DM 100.0
PEPOCH  58155.6650707
F0  5.037499999999994   7.28302859375e-07  
F1  -2.745888834184557e-8   4.694641870347623e-10  
"""
    )
    )

toas = make_fake_toas(58300, 58400, 10, model=model, error=100*u.us
                          )
    
r = pint.residuals.Residuals(toas, model)
print(f"F (modelF0=True): {r.get_PSR_freq(modelF0=True)}")
print(f"F (modelF0=False): {r.get_PSR_freq(modelF0=False)}")
```

Gives:
```
F (modelF0=True): 5.037499999999994 Hz
F (modelF0=False): [2.34770733 2.33451599 2.32131764 2.3081133  2.2949043  2.28169312
 2.26848172 2.25527039 2.24205952 2.22884858] Hz
```

paulray:
Fixes #1128

This was a bug in both the computation and the test. Both are now fixed.

The other change I made was to change the finite difference dt from being 1000*P0 to 2*P0. Since PINT can compute accurate phases from one pulse to the next, 1000*P0 seems excessive and dangerous for long period pulsars. For a 1000s pulsar, that is 10 days away!!

The (broken) test used to try to compare to TEMPO polyco computation of local freq and insisted it was good to a relative tolerance of 1E-20.  I had to relax this to 1E-7 to make this work. I'm not sure what accuracy we should insist on for the two different computations of local frequency.  Thoughts?

### DIFF
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -1391,8 +1391,9 @@ class TimingModel:
         copy_toas = copy.deepcopy(toas)
         if sample_step is None:
             pulse_period = 1.0 / (self.F0.quantity)
-            sample_step = pulse_period * 1000
-        sample_dt = [-sample_step, sample_step]
+            sample_step = pulse_period * 2
+        # Note that sample_dt is applied cumulatively, so this evaulates phase at TOA-dt and TOA+dt
+        sample_dt = [-sample_step, 2 * sample_step]
 
         sample_phase = []
         for dt in sample_dt:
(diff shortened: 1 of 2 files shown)


dlakaplan:
This looks good to me. I can approve if you are ready.

PR #1131, opened 2021-08-19
Update Pintk random models operation

champagne-cmd:
This PR replaces #1118 and uses @dlakaplan's new `calculate_random_models` function, as well as new fake toas machinery. Time residuals are created in `calculate_random_models` based on the user inputting `return_time=True` to the function. The mean residual of each set of random residuals within the span of real data is then calculated within `pulsar.py` and subtracted out of each respective residual set before plotting.

### DIFF
--- a/src/pint/pintk/plk.py
+++ b/src/pint/pintk/plk.py
@@ -29,19 +29,18 @@ except ImportError:
     )
 
 
-# Where is this meant for? Maybe it belongs in application creation?
-# log.debug(
-#    "This should also show up. test click revert, turn params on and off, and prefit model"
-# )
+log.debug(
+    "This should also show up. test click revert, turn params on and off, and prefit model"
+)
 
 plotlabels = {
     "pre-fit": [
-        "Pre-fit residual",
+        r"Pre-fit residual ($\mu$s)",
         "Pre-fit residual (phase)",
         "Pre-fit residual (us)",
     ],
     "post-fit": [
-        "Post-fit residual",
+        r"Post-fit residual ($\mu$s)",
         "Post-fit residual (phase)",
         "Post-fit residual (us)",
     ],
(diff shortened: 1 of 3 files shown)


src/pint/pintk/plk.py
```diff
@@ -29,19 +29,18 @@
...
-#    "This should also show up. test click revert, turn params on and off, and prefit model"
-# )
+log.debug(
+    "This should also show up. test click revert, turn params on and off, and prefit model"
+)
 
 plotlabels = {
     "pre-fit": [
-        "Pre-fit residual",
+        r"Pre-fit residual ($\mu$s)",
```
scottransom, line 38: Does this mean that residuals are always in us now?

scottransom (commented):
All looks pretty good.  Two things I noticed while playing around (which if super easy, can just be fixed with this, otherwise I'll make a new issue) is that trying to make the x-axis "years" caused an error.  And also, using the menu to change the DEBUG/INFO/WARNING level didn't do anything.

src/pint/pintk/plk.py
champagne-cmd: Sorry - accidentally copied over old code! This should be fixed now - the y axis scales to us, ms, and s.

champagne-cmd:
I'm not getting any error when I switch over to years on the x axis, either with random models or without - maybe it was sue to the outdated code I had accidentally copied over?

As for the log level not working, I do not know why that is now failing. Everything looks right to me, but I had originally implemented that with the Astropy logger, perhaps the built-in logger has some extra steps needed?

paulray:
I just pulled this PR and I confirm that the log level selection does not seem to be working, but the years X-axis did not seem to cause any problem.

paulray:
I fiddled for a while and read some about the python logging facility, but I couldn't make it work.  I suggest we handle that as a separate PR.

scottransom:
Yeah, I can't reproduce the y-axis "year" thing now.  And I agree that the logging should go to another PR.  So I think this looks all good.

PR #1133, opened 2021-08-25
Fix remove fn

### LINKED ISSUE #1132: AttributeError after removing parameters from a model
After adding a higher order frequency derivative to a model and removing it, using the model with the toas will attempt to find the removed parameter. This causes an AttributeError. I have attached a screenshot of the error and some code to recreate the issue (though i had to change the extensions - remove_param_bug.txt should be remove_param_bug.py). 

Screen Shot 2021-08-25 at 14 48 54

[remove_param_bug.txt](https://github.com/nanograv/PINT/files/7049514/remove_param_bug.txt) 
[J1640+2224par.txt](https://github.com/nanograv/PINT/files/7049502/J1640%2B2224par.txt)
[junktim.txt](https://github.com/nanograv/PINT/files/7049503/junktim.txt)

aarchiba:
Incorrect caching made Spindown break when parameters were removed. This fixes that, and tests that it is fixed.

Closes #1132

### DIFF
--- a/src/pint/models/binary_ell1.py
+++ b/src/pint/models/binary_ell1.py
@@ -162,14 +162,10 @@ class BinaryELL1(PulsarBinary):
             dPB = PBDOT * dt_integer_orbits
             self.PB.quantity = self.PB.quantity + dPB
         else:
-            fbterms = [
-                getattr(self, k).quantity
-                for k in self.get_prefix_mapping("FB").values()
-            ]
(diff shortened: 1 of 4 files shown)


luojing1211:
I think this will have a conflict with the model builder PR. But let us merge this one. I will fix the model builder PR.

aarchiba:
Merge it when you're happy with it.

Issue #1076, opened 2021-07-19
Objects containing Parameters cannot be pickled


aarchiba:
Because Parameters internally store lambda functions, they cannot be pickled, and by extension `TimingModel` objects cannot be pickled. This complicates use of multiprocessing, and there is no reason it should not be possible to pickle them.

Tests in #1071 but the fix is a bit more complicated.

aarchiba:
The problem is that `prefixParameter` objects take function arguments to determine their descriptions and units as a function of the index. These are, in many cases, lambda functions, which cannot be pickled.

The easiest way to fix this is to just take all the lambda functions and put the code in named functions. A longer-term fix would be to improve the way prefix and mask parameters are created and used; see #1092 for thoughts on how this could be done.

dlakaplan:
Should be closed by #1140

Issue #1099, opened 2021-07-29
readthedocs does not support all formatting that local doc builds do


aarchiba:
In cleaning up the documentation for #1096 and #1083 it seems that the readthedocs builds don't render lists or math-from-notebooks correctly, but local doc builds do. 

A concrete example: https://nanograv-pint.readthedocs.io/en/latest/api/pint.utils.html#pint.utils.info_string is supposed to contain a list, (Creation date etc.) and indeed it does in my local build, but it does not on readthedocs. 

Readthedocs configuration is set in `.readthedocs.yml`, so it may be possible to make this reflect what we are using at home, or there may be some more troubling problem.

PR #1136, opened 2021-08-30
Fix gridding, allow gridding based on functions.

### LINKED ISSUE #1135: gridutils has indexing error
In `gridutils`/`grid_chisq`:
```
chi2 = np.zeros((len(par1_grid), len(par2_grid)))
    # Want par1 on X-axis and par2 on y-axis
    for ii, par1 in enumerate(par1_grid):
        getattr(ftr.model, par1_name).quantity = par1
        for jj, par2 in enumerate(par2_grid):
            getattr(ftr.model, par2_name).quantity = par2
            # Array index here is rownum, colnum so translates to y, x
            chi2[jj, ii] = ftr.fit_toas()
            ...
```
so it looks like `chi2` is initiated as being size (par1, par2) but then filled according to (par2, par1).  This is an easy fix, but I don't know if this was ever used (there are no tests for it).  If par1,par2 were the same size it wouldn't have come up.

dlakaplan:
Still very much WIP (no tests, need to work on MP versions).  Fixes #1135 but also does more.

For instance, previously `gridutils` could only define a grid corresponding to a 1d range of 2 parameters.  But what if you wanted a uniform grid over another parameter (cos(i) instead of sin(i))?  Or something involving more than one quantity?

I'm trying to add the capability to define the grid and the parameters separately, where the parameters are functions of the grid.  So:

```
pint.gridutils.grid_chisq_derived(fsave,
                                                'M2',
                                                lambda x,y: x,
                                                M2,
                                                'SINI',
                                                lambda x,y: np.sqrt(1-y**2),
                                                cosi)
```
defines a grid in cos(i) but then uses that to compute SINI.

This can be extended to more complicated parameterizations (e.g., define a grid in M1, but then compute M2 using some PK parameters).  There isn't a huge advantage to doing this for the simple version of the function (you can easily roll your own), but with the MP versions it may be worth it.


paulray:
Great, sounds like a good extension of the gridding utilities!

dlakaplan:
Actually, I can't even run the MP version on the data I'm using for testing because the `maskParameter` needed for a jump isn't picklable (it includes lambda's).  Has anybody used that function (`grid_chisq_mp`)?  I know there are ways around this (use `dill` instead of `pickle`) but I don't want to reinvent something that had already been solved.

dlakaplan:
I'm not sure this is the best way, but I have a version that works for MP now (using pathos).  

I also changed the grid behavior so that it can accept an arbitrary number of parameters to grid over

dlakaplan:
Now the interface looks like:
```
pint.gridutils.grid_chisq(fsave, ('M2','SINI'),
                                       (M2,sini))
```
so you can have 1 or 2 or 3 or more parameters to grid over.

aarchiba:
The answer is we need to fix #1076 so that PINT models can be pickled. It's not a particularly hard fix, I was just deferring it until we could fix it as part of #1071, but it seems to be more urgent than that.

aarchiba:
For multiprocessing, one tool I have found helpful is to use python's `concurrent.futures` model and have parallel functions accept an Executor object - this way someone can pass in a multiprocessing executor with an appropriate number of workers and your code will use it, or a multithreading executor if that's useful, or some distributed-computing executor; Executors provide a standardized interface to parallel processing that's powerful enough for what the gridutils want.

https://docs.python.org/3/library/concurrent.futures.html

dlakaplan:
Some updated timing:
```
old algorithm grid [5, 5] 1 cpus in 1.5876991748809814 s
old algorithm grid [5, 5] 4 cpus in 0.9443788528442383 s
new algorithm grid [5, 5] 1 cpus in 1.6766939163208008 s
new algorithm grid [5, 5] 4 cpus in 1.529547929763794 s
old algorithm grid [10, 10] 1 cpus in 6.481395959854126 s
old algorithm grid [10, 10] 4 cpus in 4.896639108657837 s
new algorithm grid [10, 10] 1 cpus in 7.438166856765747 s
new algorithm grid [10, 10] 4 cpus in 3.938685178756714 s
old algorithm grid [20, 20] 1 cpus in 31.33030915260315 s
old algorithm grid [20, 20] 4 cpus in 18.0501708984375 s
new algorithm grid [20, 20] 1 cpus in 27.26108193397522 s
new algorithm grid [20, 20] 4 cpus in 22.748576879501343 s
old algorithm grid [50, 50] 1 cpus in 172.6981749534607 s
old algorithm grid [50, 50] 4 cpus in 114.43707799911499 s
new algorithm grid [50, 50] 1 cpus in 170.2247428894043 s
new algorithm grid [50, 50] 4 cpus in 118.5853750705719 s
old algorithm grid [150, 150] 1 cpus in 1669.5488622188568 s
old algorithm grid [150, 150] 4 cpus in 948.948086977005 s
new algorithm grid [150, 150] 1 cpus in 1533.0161559581757 s
new algorithm grid [150, 150] 4 cpus in 943.6808099746704 s
new algorithm grid [5, 5, 5] 1 cpus in 7.619338035583496 s
new algorithm grid [5, 5, 5] 4 cpus in 5.0489091873168945 s
new algorithm grid [10, 10, 10] 1 cpus in 64.79371809959412 s
new algorithm grid [10, 10, 10] 4 cpus in 33.99533987045288 s
new algorithm grid [20, 20, 20] 1 cpus in 693.8935396671295 s
new algorithm grid [20, 20, 20] 4 cpus in 418.54967498779297 s
new algorithm grid [50, 50, 50] 1 cpus in 8233.895636081696 s
new algorithm grid [50, 50, 50] 4 cpus in 4726.437324762344 s
```
Harder to parse, but the new (cell-by-cell) vs. old (column) parallelization seems pretty comparable for the larger grids.

I have yet to change to the more generic Executor objects, though.

dlakaplan:
I re-wrote the MP stuff to use Executors, as Anne suggested.  And I cleaned up the interface a bit. There is still some old stuff and some WIP stuff in the end of the module (for continued testing) but I think this is getting better.

Note that I wrote an example, but I don't think it's getting build since the example notebooks are still not being listed properly (#1137?)

aarchiba:
Notebooks are not automatically listed, if you want it in the documentation you have to list it with the others (under tutorials.rst I think). (Which is not to say that our notebooks might not also be broken.)

src/pint/gridutils.py
```diff
@@ -1,16 +1,278 @@
...
+
+    # Freeze the  params we are going to grid over
+    for parname in parnames:
+        getattr(ftr.model, parname).frozen = True
+
+    # All other unfrozen parameters will be fitted for at each grid point
+    out = np.meshgrid(*parvalues)
+    chi2 = np.zeros(out[0].shape)
+    if ncpu > 1:
+        with executor(max_workers=ncpu) as e:
```
aarchiba, line 93: I would suggest accepting an already-created executor as an argument - executors might require various startup arguments that this function doesn't know about. So if you want to run this with a multiprocessing executor, you just call this function in a with statement and pass it the executor. This also helps keep executors around for multiple calls, if they're expensive to create (ssh to a bunch of machines...)

dlakaplan: I thought about that.  I think for more complicated cases that would definitely be better, but there might also be simpler cases where creating it is better.  I'll see if I can add a little flexibility to do both.

tests/test_grid.py
```diff
@@ -0,0 +1,169 @@
...
+
+    assert np.isclose(bestfit, chi2grid.min())
+
+
+def test_grid_oneparam():
+    parfile = pint.config.examplefile("NGC6440E.par")
+    timfile = pint.config.examplefile("NGC6440E.tim")
+    m, t = get_model_and_toas(parfile, timfile)
+
+    f = GLSFitter(t, m)
```
aarchiba, line 73: Why use the GLSFitter? Isn't it going to be slower and more numerically unstable? And not needed here?

src/pint/gridutils.py
aarchiba: Maybe I should be clearer: using an executor to run parallel tasks has a well-defined interface, but creating executors does not.

dlakaplan: OK, I didn't realize that.  I still think there should be a way to run this with no external work, so I'll see what looks reasonable.

dlakaplan:
This build is failing because some of the tests are failing.  This is related to the change to use Executors: it works sometimes, but when I try it with prefix parameters I _think_ it's hanging because of an unpicklability issue.  I don't know why it's hanging (which leads to tests being canceled) rather than crashing.  The functions work with other parameters but clearly we need to solve this.  So should probably delay this until after #1076 is solved.

aarchiba:
> This build is failing because some of the tests are failing. This is related to the change to use Executors: it works sometimes, but when I try it with prefix parameters I _think_ it's hanging because of an unpicklability issue. I don't know why it's hanging (which leads to tests being canceled) rather than crashing. The functions work with other parameters but clearly we need to solve this. So should probably delay this until after #1076 is solved.

I added comments on that issue on the easiest way to fix that. I won't be able to get to that for a week or two at the least.

dlakaplan:
Thanks.  It may be that we can push forward with this PR and just include a cautionary note that MP won't work with all models.  I also don't have a lot of time to start something new.

aarchiba:
Easiest in that case is to xfail the problem tests with an explanation involving pickling.

dlakaplan:
I have tried that, with a pytest timeout and a pytest fail.  That in itself works OK.  But the test still doesn't return - it doesn't give a pickle error like the other modules did, but just sits there.  So maybe better to do without that test.

dlakaplan:
Incorporates #1140 to enable use of pickled parameters.

Usage is now:
`grid_chisq(..., executor=, ncpu=,...)`
With the following behavior:
| `executor` | `ncpu` | result |
| ----------- | ------- | ------ |
| existing object | any | uses existing `executor` |
| None | 1 | uses single-processor |
| None | None | creates default `executor` with `cpu_count` workers |
| None | >1 | creates default `executor` with desired number of workers |

Does that make sense?  

Finally, I changed the test to use  2 CPUs (for the multi-CPU tests)

dlakaplan:
I put a table like the above in the doctoring.  On my local build it looks fine, but once again it isn't rendering properly at https://nanograv-pint--1136.org.readthedocs.build/en/1136/_autosummary/pint.gridutils.grid_chisq.html

(#1099)

dlakaplan:
I think this is ready for review, @paulray or @aarchiba

dlakaplan:
@aarchiba (or others) : do you know of appropriate Executor implementations for HPC environments?  Our local cluster is scheduled with slurm, but I can't find an obvious interface to use that.

scottransom:
@dlakaplan For most clusters I've seen, the scheduler is separate from the program execution.  So I would expect that you'd use a normal Python+MPI executor, which you would then call from your job script which slurm would schedule.

dlakaplan:
Current progress bar uses `tqdm` which is optional (#1144)

aarchiba:
> @dlakaplan For most clusters I've seen, the scheduler is separate from the program execution. So I would expect that you'd use a normal Python+MPI executor, which you would then call from your job script which slurm would schedule.

Yes, mpi4py provides an executor: https://mpi4py.readthedocs.io/en/stable/mpi4py.futures.html#mpipoolexecutor

Normally to use something like slurm and MPI you have to set up the slurm job in an appropriate way, and you have to write the top-level python script to decide whether each invocation is the main program thread or a worker. There will be instructions somewhere. But the message is, most high-powered parallelization requires some particular top-level structure. Fortunately there are mechanisms for writing parallel routines that can be agnostic about the top level.

There's also `dask`, which provides a rather different set of mechanisms for large-scale parallelism, and which might provide its own interface with slurm. I think with `dask` you can set up a main process and worker processes can come and go (as opposed to the lockstep required by MPI) but I don't have a lot of experience with it.

docs/examples/grid_parameters.py
```diff
@@ -0,0 +1,298 @@
...
+# %% tags=[]
+# 1, 2, and 3 sigma confidence limits
+nsigma = np.arange(1, 4)
+# these are the CDFs going from -infinity to nsigma.  So subtract away 0.5 and double for the 2-sided values
+CIs = (scipy.stats.norm().cdf(nsigma) - 0.5) * 2
+print(f"Confidence intervals for {nsigma} sigma: {CIs}")
+# chi^2 random variable for 2 parameters
+rv = scipy.stats.chi2(2)
+x = np.linspace(0, 20)
+contour_levels = np.interp(CIs, rv.cdf(x), x)
```
aarchiba, line 142: Why interpolate? Doesn't a chi-squared random variable support inverse CDF?

src/pint/gridutils.py
```diff
@@ -1,16 +1,400 @@
...
+        List of parameter values to grid over (each should be 1D array of astropy.units.Quantity)
+    """
+    # Make a full copy of the fitter to work with
+    myftr = copy.deepcopy(ftr)
+    for parname, parvalue in zip(parnames, parvalues):
+        # Freeze the  params we are going to grid over and set their values
+        # All other unfrozen parameters will be fitted for at each grid point
+        getattr(myftr.model, parname).frozen = True
+        getattr(myftr.model, parname).quantity = parvalue
+    myftr.fit_toas()
```
aarchiba, line 42: Do you want to include some kind of handling for fitter failures to converge? Return NaN? Passing errors through parallelization schemes is a messy business. Ideally I suppose this function would pass exceptions upwards and the parallel scheme would use a wrapper that does something more covnenient with them.

src/pint/gridutils.py
```diff
@@ -1,16 +1,400 @@
...
+    parvalues : list
+        List of parameter values to grid over (each should be 1D array of :class:`astropy.units.Quantity`)
+    executor : concurrent.futures.Executor or None, optional
+        Executor object to run multiple processes in parallel
+        If None, will use default :class:`concurrent.futures.ProcessPoolExecutor`, unless overridden by ``ncpu=1``
+    ncpu : int, optional
+        If an existing Executor is not supplied, one will be created with this number of workers.
+        If 1, will run single-processor version
+        If None, will use :func:`multiprocessing.cpu_count`
+    chunksize : int
```
aarchiba, line 66: What does this actually do? Since users are passing in existing executors does it do anything?

src/pint/gridutils.py
```diff
@@ -1,16 +1,400 @@
...
+        Size of the chunks for :class:`concurrent.futures.ProcessPoolExecutor` parallel execution.
+        Ignored for :class:`concurrent.futures.ThreadPoolExecutor`
+    printprogress : bool, optional
+        Print indications of progress (requires :mod:`tqdm`)
```
aarchiba, line 70: How does `tqdm` handle non-printing-friendly environments? Would it be sensible to drop this and make a progress bar if and only if tqdm is available?

src/pint/gridutils.py
```diff
@@ -1,16 +1,400 @@
...
+    >>> import numpy as np
+    >>> import pint.config
+    >>> import pint.gridutils
+    >>> from pint.fitter import WLSFitter
+    >>> from pint.models.model_builder import get_model, get_model_and_toas
+    # Load in a basic dataset
+    >>> parfile = pint.config.examplefile("NGC6440E.par")
+    >>> timfile = pint.config.examplefile("NGC6440E.tim")
+    >>> m, t = get_model_and_toas(parfile, timfile)
+    >>> f = WLSFitter(t, m)
```
aarchiba, line 88: Please use a downhill fitter in the example code. The WLS fitter does not guarantee convergence.

src/pint/gridutils.py
```diff
@@ -1,16 +1,400 @@
...
+    |                 |        | executor with          |
+    |                 |        | ``cpu_count`` workers  |
+    +-----------------+--------+------------------------+
+    | None	      | >1     | creates default        |
+    |                 |        | executor with desired  |
+    |                 |        | number of workers      |
+    +-----------------+--------+------------------------+
+    """
+
+    if isinstance(executor, concurrent.futures.Executor):
```
aarchiba, line 130: Python prefers "duck typing" - if it walks like a duck and quacks like a duck, it's a duck - to actual `isinstance`. Better to just ask if `exceutor is not None`.

src/pint/gridutils.py
```diff
@@ -1,16 +1,400 @@
...
+        # the executor has already been created
+        executor = executor
+    elif executor is None and (ncpu is None or ncpu > 1):
+        # make the default type of Executor
+        if ncpu is None:
+            ncpu = multiprocessing.cpu_count()
+        executor = concurrent.futures.ProcessPoolExecutor(max_workers=ncpu)
+
+    # Save the current model so we can tweak it for gridding, then restore it at the end
```
aarchiba, line 139: Why is this done? Doesn't the single-call procedure deepcopy the fitter? If you actually need to restore something at the end, use try/finally.

src/pint/gridutils.py
```diff
@@ -1,16 +1,400 @@
...
+    savemod = ftr.model
+    gridmod = copy.deepcopy(ftr.model)
+    ftr.model = gridmod
+
+    # Freeze the  params we are going to grid over
```
aarchiba, line 144: This is also done by the single-fit code.

src/pint/gridutils.py
```diff
@@ -1,16 +1,400 @@
...
+def grid_chisq_derived(
+    ftr,
+    parnames,
+    parfuncs,
+    gridvalues,
+    executor=None,
+    ncpu=None,
+    chunksize=1,
+    printprogress=True,
+):
```
aarchiba, line 203: This seems to involve a lot of code duplication - can the simpler interface be implemented on top of this one?

src/pint/gridutils.py
```diff
@@ -1,16 +1,400 @@
...
+    """Compute chisq over a grid of derived parameters
+
+    Parameters
+    ----------
+    ftr : pint.fitter.Fitter
+        The base fitter to use.
+    parnames : list
+        Names of the parameters to grid over
+    parfuncs : list
+        List of functions to convert `gridvalues` to quantities accessed through `parnames`
```
aarchiba, line 213: Can the docstring include an example maybe? I'm not really clear on what arguments these functions should accept (one list/array argument? as many arguments as there are dimensions in the grid?) or what they should return (an array of values in the same order as parnames?)

src/pint/gridutils.py
```diff
@@ -1,16 +1,400 @@
...
+    printprogress=True,
+):
+    """Compute chisq over a grid of derived parameters
+
+    Parameters
+    ----------
+    ftr : pint.fitter.Fitter
+        The base fitter to use.
+    parnames : list
+        Names of the parameters to grid over
```
aarchiba, line 211: Maybe it should be clear that these have to be the names recognized by the fitter, not the names of parameters you are actually gridding over?

tests/test_grid.py
```diff
@@ -0,0 +1,294 @@
...
+    m, t = get_model_and_toas(parfile, timfile)
+
+    f = WLSFitter(t, m)
+    f.fit_toas()
+    bestfit = f.resids.chi2
+
+    F0 = np.linspace(
+        f.model.F0.quantity - 3 * f.model.F0.uncertainty,
+        f.model.F0.quantity + 3 * f.model.F0.uncertainty,
+        15,
```
aarchiba, line 33: Would 3 and 5 grid points work just as well and much faster?

tests/test_grid.py
```diff
@@ -0,0 +1,294 @@
...
+        f.model.F1.quantity + 3 * f.model.F1.uncertainty,
+        17,
+    )
+
+    chi2grid = pint.gridutils.grid_chisq(f, ("F0", "F1"), (F0, F1), ncpu=1)
+
+    assert np.isclose(bestfit, chi2grid.min())
+
+
+def test_grid_multiprocessor():
```
aarchiba, line 46: Why not test the multiprocessor values to see if they're the same as the single-processor values?

tests/test_grid.py
```diff
@@ -0,0 +1,294 @@
...
+    f = WLSFitter(t, m)
+    f.fit_toas()
+    bestfit = f.resids.chi2
+
+    F0 = np.linspace(
+        f.model.F0.quantity - 5 * f.model.F0.uncertainty,
+        f.model.F0.quantity + 5 * f.model.F0.uncertainty,
+        21,
+    )
+    executor = concurrent.futures.ProcessPoolExecutor(max_workers=ncpu,)
```
aarchiba, line 105: You should almost always use a `with` statement to create and use an executor.

tests/test_grid.py
```diff
@@ -0,0 +1,294 @@
...
+        f.model.F1.quantity - 3 * f.model.F1.uncertainty,
+        f.model.F1.quantity + 3 * f.model.F1.uncertainty,
+        7,
+    )
+    DM = np.linspace(
+        f.model.DM.quantity - 3 * f.model.DM.uncertainty,
+        f.model.DM.quantity + 3 * f.model.DM.uncertainty,
+        5,
+    )
+    chi2grid = pint.gridutils.grid_chisq(f, ("F0", "F1", "DM"), (F0, F1, DM), ncpu=ncpu)
```
aarchiba, line 135: Better not to use parallelism unless it's actually part of what is being tested. Again a smaller grid would be much faster.

tests/test_grid.py
dlakaplan: Sure.  no real difference.

tests/test_grid.py
dlakaplan: I did see some very small differences (at the 1e-6 level).  I didn't track down the origin.

tests/test_grid.py
dlakaplan: yes, good point

tests/test_grid.py
dlakaplan: most of the grids are pretty small.  I've made them smaller where possible.

dlakaplan:
Trying to check chi^2 behavior for different fitters.  GLS and WLS fitter behave OK.  There are differences at the 1e-6 level or below in chi^2 between the best-fit and what I get for a grid that _should_ be centered near the best-fit, but they seem reasonable.  There are also differences between the multiprocessor and single-processor versions which I don't really understand (since the actual fits and initial conditions _should_ be the same).  But they are small.

```
Fitter: GLSFitter
MP grid - SP grid:  [ 0.00000000e+00  3.21074808e-07  2.84685896e-07 -7.93976838e-07
 -8.23938308e-07]
MP grid - bestfit:  8.5623758915021053184e-07
SP grid - bestfit:  5.715516929599706586e-07
Fitter: WLSFitter
MP grid - SP grid:  [ 0.00000000e+00  3.21074808e-07  2.84685896e-07 -7.93976838e-07
 -8.23938308e-07]
MP grid - bestfit:  8.5623758915021053184e-07
SP grid - bestfit:  5.715516929599706586e-07
```
[this is for a 5-point 1D grid in `F0`).

But I can't get good results for the Downhill Fitters.  DownhillWLS will run in SP mode, but when I run it in MP mode I get an error
```
  File "/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.8/concurrent/futures/process.py", line 239, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.8/concurrent/futures/process.py", line 198, in _process_chunk
    return [fn(*args) for args in chunk]
  File "/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.8/concurrent/futures/process.py", line 198, in <listcomp>
    return [fn(*args) for args in chunk]
  File "/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/gridutils.py", line 44, in doonefit
    myftr.fit_toas()
  File "/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/fitter.py", line 1346, in fit_toas
    super().fit_toas(maxiter=maxiter, debug=debug, **kwargs)
  File "/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/fitter.py", line 1131, in fit_toas
    new_state = current_state.take_step(step, lambda_)
  File "/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/fitter.py", line 1297, in take_step
    self.fitter, self.take_step_model(step, lambda_), threshold=self.threshold
  File "/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/fitter.py", line 1043, in take_step_model
    log.debug(f"Adjusting {getattr(self.model, p)} by {s}")
  File "/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/models/parameter.py", line 412, in __repr__
    out += "{:17s}".format(self.str_quantity(self.quantity))
  File "/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/models/parameter.py", line 1156, in str_quantity
    return quan.to_string(sep=":", precision=8)
  File "/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.8/site-packages/astropy/coordinates/angles.py", line 344, in to_string
    raise ValueError(
ValueError: 'hourangle' can not be represented in sexagesimal notation
```
I don't know why this is only coming up during the MP run.    I also can convert Angles with hour angle units to sexagesimal, but maybe that's not what it's trying to do (is it trying to convert the unit name itself)?

dlakaplan:
I get the same result with the DownhillGLSFitter.  I'll continue to dig, but if there's something obvious related to those fitters please let me know (got to go teach in a few).

dlakaplan:
P.S. A simple script to test the fitters is:
```
import astropy.units as u
import numpy as np
import pint.config
from pint.fitter import GLSFitter, WLSFitter, DownhillWLSFitter, DownhillGLSFitter
import pint.gridutils
import multiprocessing

if __name__ == "__main__":

    parfile = pint.config.examplefile("NGC6440E.par")
    timfile = pint.config.examplefile("NGC6440E.tim")
    f0 = WLSFitter(t, m)
    f0.fit_toas()
    F0 = np.linspace(
        f0.model.F0.quantity - 1 * f0.model.F0.uncertainty,
        f0.model.F0.quantity + 1 * f0.model.F0.uncertainty,
        5,
    )
    multiprocessing.freeze_support()
    bestfit = {}
    chi2grid_mp = {}
    chi2grid_sp = {}

    fitters = {
        "GLSFitter": GLSFitter,
        "WLSFitter": WLSFitter,
        "DownhillWLSFitter": DownhillWLSFitter,
        "DownhillGLSFitter": DownhillGLSFitter,
    }

    for fittername in fitters:
        print(f"Fitter: {fittername}")
        f = fitters[fittername](t, f0.model)
        f.fit_toas()
        bestfit[fittername] = f.resids.chi2

        chi2grid_sp[fittername] = pint.gridutils.grid_chisq(
            f, ("F0",), (F0,), printprogress=True, ncpu=1
        )
        chi2grid_mp[fittername] = pint.gridutils.grid_chisq(
            f,
            ("F0",),
            (F0,),
            printprogress=True,
        )

        print("MP grid - SP grid: ", chi2grid_mp[fittername] - chi2grid_sp[fittername])
        print(
            "MP grid - bestfit: ", chi2grid_mp[fittername].min() - bestfit[fittername]
        )
        print(
            "SP grid - bestfit: ", chi2grid_sp[fittername].min() - bestfit[fittername]
        )
```
(easier to run outside the `pytest` environment.)

docs/examples/grid_parameters.py
```diff
@@ -0,0 +1,298 @@
...
+quantity_support()
+import scipy.stats
+
+# %matplotlib inline
+
+# %%
+# Load in a basic dataset
+parfile = pint.config.examplefile("NGC6440E.par")
+timfile = pint.config.examplefile("NGC6440E.tim")
+m, t = get_model_and_toas(parfile, timfile)
```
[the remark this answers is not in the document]
dlakaplan: It still fits for RA, Dec, and DM.  I'll make that clear.

docs/examples/grid_parameters.py
dlakaplan: Yes, it does.  I couldn't find it initially (I think it used to be icdf) but I found it now.

dlakaplan:
I think the differences between SP and MP in terms of chi^2 with WLSFitter are related to the initial state of the fitter before the iterations start.  The MP fitters start from the same state, and the SP fitters change.  When I run them with a single value they return the same result.

dlakaplan:
So I definitely don't understand why the Downhill fitters don't work with MP.  If I comment out a couple of `debug` lines it all proceeds OK and gives sensible results.  But those lines shouldn't hurt anything.  They try to print RA as an hourangle.  The astropy code checks and sees if the units are hourangle, and if so it knows how to do it.  And the units _should_ be hourangle. In the SP version they are.  In the MP version they still look like hourangle but the comparison:
`quantity.unit is u.hourangle` 
fails.

Any idea why that is?

dlakaplan:
@scottransom (and others): our sysadmin pointed me to https://github.com/sampsyo/clusterfutures which is **great** as far as I can see.  I'm still debugging the exact usage but I think something like:
```
with cfut.SlurmExecutor(debug=True,keep_logs=True,additional_setup_lines=slurm_options) as executor:
	chi2, params = pint.gridutils.grid_chisq_derived(
	    f,
	    ("M2", "SINI", "PX", "PBDOT"),
            (lambda Mc, cosi, d: Mc, cg.sini, cg.parallax, cg.PBDOT),
            (Mc, cosi, distance),
	    executor=executor,
	    printprogress=False,
	)
```
works very well on our slurm system.

dlakaplan:
Using the `ClusterFutures` interface above, I ran a 50x50x50 grid on J0740.  The grid looped over distance, `cosi`, and `Mc`, and then computed `M2` (trivial) `SINI` (easy), `PX` (easy) and `PBDOT` (harder, needs to include other `PBDOT` contributions).  A single chi^2 took ~2min, and the grid of them took ~24h (running 100-200 concurrently).  The results look pretty believable.

J0740_pint

So I think this is close to mergeable.  The one issue that I haven't resolved is the failures around the `debug` statements in the Downhill Fitters (I put those in a separate `try/except`): I don't know why those happen, but they don't stop the operations.  Note that the example above used a `WidebandDownhillFitter`.

scottransom:
@dlakaplan That sounds encouraging!  Although the image that you uploaded looks completely white/blank to me.

dlakaplan:
oops... Compare to Figure 6 in Fonseca et al. (2021)
J0740_pint

docs/examples/grid_parameters.py
```diff
@@ -0,0 +1,298 @@
...
+    f.model.F1.quantity + 3 * f.model.F1.uncertainty,
+    27,
+)
+
+# %% [markdown]
+# # 1D Grid
+
+# %%
+# Do a 1D "grid"  Make sure that the parameters are supplied as tuples with length 1
+chi2_F0 = pint.gridutils.grid_chisq(f, ("F0",), (F0,))
```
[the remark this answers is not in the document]
dlakaplan: I think this is done.

src/pint/gridutils.py
dlakaplan: done

scottransom:
Looks good!  Thanks for all the effort!

PR #1140, opened 2021-09-03
Remove lambda

dlakaplan:
This helps address #1076.  It's not particularly elegant.  It replaces lambda functions with named functions where needed, and also removes some where the parameter construction doesn't require them.

There is a new test: `tests/tests_pickle.py` that tests whether or not timing models are serializable.  

Glitches and FD models still need to be tested for reproducing prior functionality (and then potentially for serializability too).


src/pint/models/frequency_dependent.py
```diff
@@ -8,6 +8,10 @@
 from pint.models.timing_model import DelayComponent, MissingParameter
 
 
+def _description_frequency_dependent_coefficients(x):
```
aarchiba, line 11: You can put this as a `classmethod` in the class `FD`, which would allow just naming it `_description_template` or whatever. I realize this doesn't fully cover all the naming complexities but it might help; also help make it clear the restricted scope/utility of these functions.

aarchiba:
There's at least one test marked `xfail` that checks pickleability, I added it in some of my parameter tidying. Probably #1071 . Wouldn't hurt to take the xfail off.

dlakaplan:
I have switch functions -> class methods in a few places.

I also un-xfailed `test_parameters::test_parameter_can_be_pickled`

aarchiba:
Looks good to me; merge as soon as it passes CI.

PR #1141, opened 2021-09-03
Small error check in model comparison

paulray:
Fixes an exception when doing `m1.compare(m2)` in the case that `m1` has equatorial coords and `m2` has ecliptic.

Also adds some CHANGELOG entries


almcewen0 (approved):
looks good to me!

PR #1143, opened 2021-09-10
New coordinates for AO, EFF, and JB from Ben Perera

aarchiba:
This PR updates the coordinates for several observatories to current VLBI coordinates. VLBI folks actually use position+velocity (and we know the velocities), but this just adds the positions as of March 2020. The old positions are preserved for comparison purposes.


aarchiba:
Some of the test failures are comparisons with TEMPO2; I think these tests are problematic, since they will fail if ever we do better than TEMPO2 does (which I think we do here) and since they provide no real information on what is "wrong".

I'm a little puzzled by the ecorr averaging test failure.

aarchiba:
An alternative approach to this would give the new coordinates new names, leaving the old observatories untouched. This could then be merged and allow easy testing and comparison; another PR when we're confident could swap in the new names.

aarchiba:
Failures:

- `tests/test_ecorr_average.py::TestEcorrAverage::test_ecorr_average` - compares to precomputed residuals averaged by TEMPO's `res_avg`
- `tests/test_times.py::test_times_against_tempo2` - compares observatory positions and velocities to tempo2
- `tests/test_fbx.py::TestFBX::test_J0023` - compares residuals to TEMPO2 at the 10-ns level

aarchiba:
Discussion says:

- Go ahead and merge as is with these names when feasible so that NANOGrav can use them in 15yr
- Mark the tests xfail
- Merge the GBT position at the same time (but don't expect much change)

GBT position: x y z = 882589.289 -4924872.368 3943729.418

aarchiba:
I also received and incorporated an updated position for CHIME from Bradley Meyers.

aarchiba:
The tests needed to be converted from `unittest` style so they could be `xfail`ed.

luojing1211:
> I also received and incorporated an updated position for CHIME from Bradley Meyers.

The new CHIME position is already merged by #1145

aarchiba:
> > I also received and incorporated an updated position for CHIME from Bradley Meyers.
> 
> The new CHIME position is already merged by #1145

Yes, I found that through merge conflicts. It's resolved now.

PR #1145, opened 2021-09-16
Updating CHIME observatory coordinates

bwmeyers:
As of January 2021 we have a better handle on the centre position of the CHIME telescope structure. I've also provided some details about where these new numbers come from, for posterity. I _think_ this is the only change needed, but happy for the devs to point me elsewhere and I can adjust as necessary. 

(And a minor correction to the expanded name.)

### DIFF
--- a/src/pint/observatory/observatories.py
+++ b/src/pint/observatory/observatories.py
@@ -228,11 +228,13 @@ TopoObs(
     "chime",
     tempo_code="y",
     itoa_code="CH",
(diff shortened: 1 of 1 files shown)


luojing1211:
Thank you @bwmeyers. We will let CI test run.

bwmeyers:
Just for reference, the new position has now been accepted into TEMPO and TEMPO2 (in case that was a consideration).

PR #1148, opened 2021-09-24
Transform models ECL <-> ICRS

dlakaplan:
This gives a way to convert TimingModels from ECL <-> ICRS.  The interface is the same regardless of the type of the initial model.  I played around with a few different implementations (ultimately complicated by circular import concerts with `utils.py`) but settled on this one.  

If you load in a model you can do:

`model.as_ICRS(epoch=epoch)`

which will give it to you as a model with `AstrometryEquatorial` components at the requested epoch.  Similarly, 

`model.as_ECL(epoch=epoch)` 

does the same for `AstrometryEcliptic` (with an optional specification of the obliquity).

The way that uncertainties on coordinates and proper motions are handled needs a bit more testing.  I currently make fake proper motions out of them and rely on `astropy` to transform.  This seems OK in principle (definitely not perfect, but hard to be perfect without refitting), but it would be good to verify `cos(Dec)` issues.  

Circular transformations end up pretty good, with small resulting changes in the uncertainties.


dlakaplan:
For a `TimingModel`, what are the units uncertainties on RA?  Is the `cos(Dec)` term included or not?  Same with uncertainties on `PMRA`?

kerrm:
> For a `TimingModel`, what are the units uncertainties on RA? Is the `cos(Dec)` term included or not? Same with uncertainties on `PMRA`?

I'm about 90% sure that in tempo2, the uncertainty on RA is in seconds and does not include cos(dec).  PMRA is really d(RA/dt)*cos(dec) in mas/yr so the uncertainty *does* include cos(dec).  I've just spent about 20 minutes looking at the code, and I'm still at about 90% confidence...  it's a bit of a mess.

dlakaplan:
Thanks.  And what about `AstrometryEcliptic`: is it the same for `ELONG`?

dlakaplan:
I think if this works, it deprecates `src/pint/modelutils.py` (which don't do the uncertainties)

aarchiba:
For testing value conversions, let me suggest taking a simple model of an isolated pulsar, generating fake TOAs, then computing residuals with respect to the converted model; they should be within a few nanoseconds at most. Easily parametrized over a few positions.

Checking the uncertainties is a bit messier, since it's an intrinsically lossy operation - you could generate a data set with uncertainties/noise and compare a the fit uncertainties between directly fitting in ECL and fitting in IERS and converting; the converted version should probably be pessimistic (since we have no way to represent correlations between parameters in models). Likewise round-trip conversions should yield wider error bars than the original.

luojing1211:
Is it a good idea to add a test that transforms from ICRS to ECL and a round trip back to see if the value and uncertainties has been changed?

dlakaplan:
I check it with residuals (as Anne suggested) rather than just model parameters.  I could add the pure model test as well if needed

dlakaplan:
Sure.  I actually had a round-trip test for the uncertainties (with a generous tolerance, since they can change).  I added a similar test for the parameter values.

PR #1158, opened 2021-12-05
add tempo/tempo2 parfile output

luojing1211:
This is a new pull request that is based on Nihan pol's parfile converter https://gitlab.nanograv.org/nano-time/timing_analysis/-/blob/15yr/src/timing_analysis/lite_utils.py. It converts the output parfile string to 'tempo/tempo2''s format. This is a simplified version for PR #1156 but loses the format flexibility.

### DIFF
--- a/src/pint/gridutils.py
+++ b/src/pint/gridutils.py
@@ -57,7 +57,7 @@ def doonefit(ftr, parnames, parvalues):
 
 
 def grid_chisq(
-    ftr, parnames, parvalues, executor=None, ncpu=None, chunksize=1, printprogress=True,
+    ftr, parnames, parvalues, executor=None, ncpu=None, chunksize=1, printprogress=True
 ):
     """Compute chisq over a grid of parameters
 
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -25,11 +25,7 @@ log = logging.getLogger(__name__)
 astropy_version = sys.modules["astropy"].__version__
 mas_yr = u.mas / u.yr
 
-__all__ = [
-    "AstrometryEquatorial",
-    "AstrometryEcliptic",
-    "Astrometry",
-]
+__all__ = ["AstrometryEquatorial", "AstrometryEcliptic", "Astrometry"]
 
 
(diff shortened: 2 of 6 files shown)


dlakaplan:
@luojing1211 and @scottransom : as we discussed I have tried to do a light-weight rewrite of this that doesn't add a new module but puts the formatting in the `parameter.as_parfile_line()` commands.  So now those accept a format argument, as does `as_parfile()`.  I think this is what we were talking about, but if this is not or if we'd prefer to use Jing's solution I can just stop what I'm doing.

I haven't made a PR yet, but you can see my version at:
https://github.com/dlakaplan/PINT/tree/new_T2_parfile

There is also a test (`tests/test_parfile_writing_format.py`) that tests these explicitly.

And there is a lightweight command-line utility:
`convert_parfile src/pint/data/examples/NGC6440E.par -f tempo2 -o test.par`

What do you think?

scottransom:
I like it!  Thanks for doing this!

luojing1211:
@dlakaplan Where do you do the ECL, HARMS value change?

dlakaplan:
https://github.com/dlakaplan/PINT/blob/new_T2_parfile/src/pint/models/parameter.py#L488. I've tested the ECL change (`tests/test_parfile_writing_format.py::test_ECL`).  `NHARMS` is harder but I think I got a parfile from Joe and it outputs OK.  As far as I can tell the main difference is that the values need to be integers?

For ECL, I'm a little confused.  As-is the code seems to change the value of ECL without actually changing the related parameters. Should we be converting between ecliptic models too?

luojing1211:
TEMPO2 only support ECL IERS2003, that is why we need to change the value and it is the reason that we always refit after we converted the parfile. But I believe converting between ecliptic models is a good idea.

scottransom:
I'm not sure that it is worth the hassle to convert between ecliptic models (unless it is really easy).  IMO we should probably output a warning when we write a parfile as a non-PINT format and say that the model should be re-fit using the non-PINT software before the model is considered valid and converged.  And I would want to do that even if we already did an ecliptic coord conversion.

PR #1160, opened 2021-12-22
Fixing readthedocs builds: removing lengthy tutorials

dlakaplan:
I dug more into the readthedocs timeouts, replicating it locally so that I could time every action.

I found that a small number of notebooks had long builds:
```
reading sources... [ 95%] examples/Grid_J0740
# 2021-12-21T13:57:26.401318 0:01:18.262379 0:02:27.128517

reading sources... [ 95%] examples/Wideband_TOA_walkthrough
# 2021-12-21T13:58:13.821306 0:00:25.441884 0:03:14.548505

reading sources... [ 96%] examples/example_dmx_ranges
# 2021-12-21T13:58:58.148910 0:00:25.678631 0:03:58.876109

reading sources... [ 97%] examples/grid_parameters
# 2021-12-21T13:59:47.920465 0:00:37.250885 0:04:48.647664

reading sources... [ 97%] examples/understanding_fitters
# 2021-12-21T14:00:29.584135 0:00:37.333979 0:05:30.311334
```
where the first number after the date is the time just for that step (e.g., 37s) and the second is the cumulative time (over 5min).  So I am trying to see if removing `examples/grid_parameters` and `examples/Grid_J0740` is sufficient to get our build time below the 15min limit (should save about 2min).

Along with this I moved those two notebooks to the wiki:
https://github.com/nanograv/PINT/wiki
This isn't perfect as they don't render, and the way I found to upload files requires cloning the wiki repository (and then one can get into versioning trouble etc).  But it may be simplest.  If this works I can add links to the documentation telling people to look there for notebooks (potentially with specific links for the relevant functions).

Note that the full build for the docs:
https://readthedocs.org/projects/nanograv-pint/builds/15507454/
has two steps, each of which took about 550s last time.  The first I can reproduce locally:
```
python -m sphinx -T -E -b html -d _build/doctrees -D language=en . _build/html
```

But the second I cannot, and it mostly seems to just make static versions of the various image results:
```
python -m sphinx -T -b readthedocssinglehtmllocalmedia -d _build/doctrees -D language=en . _build/localmedia
```
(when I run it, I get an error on the `readthedocssinglehtmllocalmedia` call).  This otherwise seems to duplicate a lot of the functionality of the previous call.  So it may be that we can save 9min somehow by removing/reducing this call, which would mean we could either add notebooks back in or just have more headroom.  Note that I don't this final call is done in the PR builds (e.g., https://readthedocs.org/projects/nanograv-pint/builds/15494361/) which may be why they don't time out.


dlakaplan:
So the RTD build for this PR took 507s, vs. 740s for 1158 (which included the longer notebooks).  This still isn't a full test of a real build since it is missing the final step, but it saves quite a bit of time.

Issue #1161, opened 2021-12-30
Observatory clock correction are not applied to the simulation.


luojing1211:
https://github.com/nanograv/PINT/blob/e0cc5b9fce4f108c243ba4b5285cae4f4da746c3/src/pint/simulation.py#L83
The current simulation code does not apply the sit correction to the TOAs. If we use the PINT simulated TOAs in TEMPO/TEMPO2, the result will be wrong.

scottransom:
Shouldn't they be wrong in PINT as well?

As for what should be happening, for the simulated observatory, we should apply clock corrections, do the calculations that create zero-residual TOAs for our input model, and then remove the clock corrections from the TOAs, right?  (Or at least that should be the default if we write out those TOAs).

luojing1211:
I don't think we have any tests that do the round-trip-- simulate TOAs, writes to parfile, and then reads back to PINT.

PR #1165, opened 2022-01-02
add site clock correction to the simulation

luojing1211:
Fixing issue #1161

### DIFF
--- a/src/pint/simulation.py
+++ b/src/pint/simulation.py
@@ -80,13 +80,11 @@ def zero_residuals(ts, model, maxiter=10, tolerance=1 * u.ns):
         )
 
 
-def update_fake_toa_clock(ts, model, include_bipm=False, include_gps=True):
-    """Update the clock settings (corrections, etc) for fake TOAs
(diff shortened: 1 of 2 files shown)


luojing1211:
Here is the demo that the simulate .tim does not have the clock correction in the master branch.  
[Test_simulation .pdf](https://github.com/nanograv/PINT/files/8074749/Test_simulation.pdf)
Steps: 
* Simulate TOAs from a very simply parfile 
* Writes TOAs out to a .tim file 
* Rerun the TOAs by TEMPO
* One can see the TEMPO residual is not at the expected level. 


However, this branch solves this problem using the function get_TOAs_list(). One can see the problem is solved by this PR. 
[Test_simulation  (1).pdf](https://github.com/nanograv/PINT/files/8074797/Test_simulation.1.pdf)

paulray:
@dlakaplan we should close this, right?

abhisrkckl:
Closing because an updated version of this has been merged.

PR #1166, opened 2022-01-05
Converting PINT parfiles to TEMPO/TEMPO2

dlakaplan:
This would replace #1158 .  Instead of a standalone function, the conversions are done in the `parameter.as_parfile_line()` commands. So now those accept a format argument, as does `as_parfile()`. 

There is also a test (`tests/test_parfile_writing_format.py`) that tests these explicitly.

And there is a lightweight command-line utility:
```convert_parfile src/pint/data/examples/NGC6440E.par -f tempo2 -o test.par```

This will issue a warning when converting the `ECL` parameter.  

I am not entirely sure what the change is for the `NHARMS` parameter.  It looks like it is just ensuring that the parameter is an integer: is there anything else?

### CODE REVIEW

**Comment by dlakaplan** [comment]
I added a small update so that the format will also be reflected in the commented info block at the beginning:
```
In [3]: print(m.as_parfile(format="tempo"))
# Created: 2022-01-26T10:39:08.656304
# PINT_version: 0.7+1883.g61d6e0d1
# User: David Kaplan (kaplan)
# Host: plock
# Format: tempo
```
etc.

**Comment by dlakaplan** [comment]
@luojing1211 : I think I added in the corrections for DDK:
```
>>> m = get_model("tests/datafile/J1713+0747_ddk_simple.par")
>>> print(m.as_parfile())
# Created: 2022-01-26T12:56:06.569931
# PINT_version: 0.7+1883.g61d6e0d1.dirty
# User: David Kaplan (kaplan)
# Host: plock
# Format: pint
...
KIN                                  71.4 0 0.7
KOM                                  83.0 0 3.0
```
but then:
```
>>> print(m.as_parfile(format="tempo"))
/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/models/parameter.py:502: UserWarning: Changing KIN from DT92 convention to IAU: this will not be readable by PINT
  warn(
/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/models/parameter.py:508: UserWarning: Changing KOM from DT92 convention to IAU: this will not be readable by PINT
  warn(
...
KIN                                 108.6 0 0.7
KOM                                   7.0 0 3.0
```
Does that look right?

**Comment by dlakaplan** [comment]
Question for @scottransom or @luojing1211 :

Based on the slide @scottransom showed yesterday on the MT meeting, the DT92 -> IAU change is needed when writing `tempo` for DDK.  But then it's also needed when writing T2 model for `tempo2`.  But can `pint` handle a T2 model internally?  If not, writing a `tempo2` file may need some other changes.

**Comment by luojing1211** [comment]
@dlakaplan  Just had a little test , here is what I found: 
1. If we want to output TEMPO2 format parfile with T2 binary model (maybe the detection group is interested in), here is what PINT has to do
    * Convert KOM KIN to IAU convention 
    * Change "BINARY DDK " to "BINARY T2" 
    * Add line "SINI   KIN"
    Note: this only applies PINT binary DDK --> TEMPO2 T2 model. Other PINT binaries to TEMPO2 T2 has not been tested yet.
2.  If we want to output TEMPO2 format parfile with DDK binary model:
   * Nothing needs to be changed in the binary parameters. 
3. If we want to output TEMPO format parfile with DDK binary model:
   *  Convert KOM KIN to IAU convention

File: src/pint/models/pulsar_binary.py [hunk 1]
```diff
@@ -348,13 +348,13 @@ def print_par(self):
                     f" does not match the binary"
                     f" component {self.binary_model_name}"
                 )
-            result = self._parent.BINARY.as_parfile_line()
+            result = self._parent.BINARY.as_parfile_line(format=format)
```
luojing1211, line 351: I guess we are not doing T2 BINARY model now. It is actually tricky to have it.

luojing1211 (commented):
I think the current status is ok. But I am not sure if we need T2 model for tempo2 parfile.

PR #1168, opened 2022-01-06
working on ECL -> ECL conversions

dlakaplan:
I added a matrix transform in `pulsar_ecliptic` to allow transformations between ecliptic models with different obliquities.

I also removed the error when trying to use `model.as_ECL()` when the requested obliquity was different than the original.  I put in code to make it work.

There was also a bug fix where the obliquity wasn't set correctly when going from ICRS -> ECL models.   I think it was computed OK but just not copied over.

I added tests for these, including iterating over all defined obliquities and checking that the coordinate conversions are OK.

### DIFF
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -141,7 +141,7 @@ class Astrometry(DelayComponent):
         return delay * u.second
 
     def get_d_delay_quantities(self, toas):
-        """Calculate values needed for many d_delay_d_param functions """
+        """Calculate values needed for many d_delay_d_param functions"""
         # TODO: Move all these calculations in a separate class for elegance
         rd = dict()
 
@@ -536,6 +536,7 @@ class AstrometryEquatorial(Astrometry):
         m_ecl.ELAT.quantity = c.lat
         m_ecl.PMELONG.quantity = c.pm_lon_coslat
         m_ecl.PMELAT.quantity = c.pm_lat
+        m_ecl.ECL.value = ecl
 
         # use fake proper motions to convert uncertainties on ELONG, ELAT
         # assume that ELONG uncertainty does not include cos(ELAT)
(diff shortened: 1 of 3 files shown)


dlakaplan:
I realize that this convert the positions/proper motions, but won't convert any angles defined relative to the various axes.  This is true in general (for other parfile conversions, say ECL <-> ICRS).  How many such angles are there?  That's something I could look into if needed.

src/pint/models/astrometry.py
```diff
@@ -903,25 +904,73 @@ def as_ECL(self, epoch=None, ecl="IERS2010"):
...
-                % (self.ECL.value, ecl)
-            )
-        m = copy.deepcopy(self)
-        if epoch is not None:
-            m.change_posepoch(epoch)
-        return m
+
+        # change epoch only
+        if ecl == self.ECL.value:
+            print("here")
```
luojing1211, line 917: Is this a left over debug line?

src/pint/models/astrometry.py
```diff
@@ -903,25 +904,73 @@ def as_ECL(self, epoch=None, ecl="IERS2010"):
...
+            m = copy.deepcopy(self)
+            if epoch is not None:
+                m.change_posepoch(epoch)
```
luojing1211, line 920: This should be covered in the change_posepoch tests. Don't we have tests for it?

src/pint/models/astrometry.py
dlakaplan: Yes, that was foolish.  I'll fix and push that back

src/pint/models/astrometry.py
dlakaplan: Not sure entirely what you mean.  This is just so that you don't get an error for a transformation (say) from IERS2010 to IERS2010.

luojing1211: In the changed files, there is a codecov warning saying "Added line #L920 was not covered by tests".

PR #1173, opened 2022-02-23
Update gridding

dlakaplan:
This does two things:
* it wraps the fitter inside an object after passing it to the gridding code.  This enables options to be supplied to the fitter (e.g., changing the number of iterations)
* it allows other parameters to be returned along the whole grid, not just chi^2.  So:
```
chi2, params, extraparams = pint.gridutils.grid_chisq_derived(
                f,
                ("M2", "SINI"),
                (lambda Mc, cosi: Mc, lambda Mc,cosi: np.sin(np.arccos(cosi))),
                (Mc, cosi),
                ('PB','A1','EPS1','EPS2'),
                executor=executor,
                printprogress=True,
                )
```
would return `PB`, `A1`, `EPS1`, and `EPS2` at every point.  This was requested by Thankful.

Right now the build is failing because of the notebooks, but I think this should be fixed by https://github.com/nanograv/PINT/pull/1172.  If that gets merged I think this should complete.


paulray (approved):
Looks like very useful additions!

PR #1174, opened 2022-03-01
Update GBT clock file

rossjjennings:
Updates the GBT clock file through 2022-02-28, using data from https://www.gb.nrao.edu/~fghigo/timer/time_gbt.dat. I had to do this locally to make use of TOAs from newer data, and I figured it would be good to contribute the results back to upstream PINT.


paulray:
I think most people get their clock corrections from either $TEMPO or $TEMPO2, so the one's that are distributed with PINT are rarely used.

rossjjennings:
@paulray PINT uses its own clock files by default, though, right? I thought that was the idea, so that it could be used even in the absence of the other timing packages.

paulray:
I'm not sure of the precedence order if both are present, but it should emit an INFO statement saying exactly what clock file was used: `"Observatory {0}, loading clock file \n\t{1}"`

PR #1176, opened 2022-03-02
Add warning when using A1DOT with DDK models

dlakaplan:
I have a working solution.  The challenge was to issue the warning when adding A1DOT without changing how A1DOT is stored.  Then also to only issue the warning once.

Not sure that I have a nice/optimal solution.


dlakaplan:
This could also be done using `warnings.warn` since those suppress multiple instances of the same warning by default.  not sure which is better.

dlakaplan:
Example:
```
In [1]: from pint.models import get_model

In [2]: m=get_model("test.par")
WARNING  (pint.models.stand_alone_psr_binaries.DDK_model): Using A1DOT with a DDK model is not advised.
```

aarchiba:
Is there any reason not to use `warnings.warn` instead of reimplementing the once-olny message thing? I have needed to turn off the once-only or reset it, and while that's a pain in the neck with `warnings.warn` at least you only need to figure it out once.

aarchiba:
Is there any reason not to use `warnings.warn` instead of reimplementing the once-olny message thing? I have needed to turn off the once-only or reset it, and while that's a pain in the neck with `warnings.warn` at least you only need to figure it out once.

dlakaplan:
I think now that we have a better logging/warning solution overall you are right that it would be better to do that.

src/pint/models/stand_alone_psr_binaries/DDK_model.py
```diff
@@ -46,6 +49,12 @@ def __init__(self, t=None, input_params=None):
         del self.param_default_value["SINI"]
         self.set_param_values()
 
+    def __setattr__(self, name, value):
```
aarchiba, line 52: Usually we do validation at the stage where we're evaluating the model. Other parameter errors and warnings are emitted there. Is it necessary to intercept `__setattr__`? This can make errors very difficult to track down.

aarchiba:
Should it have a CHANGELOG entry? In case users start asking where that new warning came from?

aarchiba:
Looks good to me. Merge as soon as CI passes. (Also curse the CHANGELOG-induced merge conflicts.)

dlakaplan:
Agreed about the conflict.  Hope I didn't mess it up.

PR #1177, opened 2022-03-03
Move logging to loguru, other improvements

dlakaplan:
I'm trying to update the logging throughout PINT.  I like the `loguru` package:
https://loguru.readthedocs.io/en/stable/index.html
It has nice options for customization & colors, and is easier to control than the standard logging (of course, it's another dependency).

This is somewhat experimental.  But I think I have changed all logging calls except those in `pintk` and some of the scripts to use `loguru`.  I also did a few extra things:
* In `logging.py`, I put in a small function to capture `warnings.warn` and send them through the logger.  So we can now use either and they will propagate in the same way
* In `logging.py` I put in a custom logging filter class.  This was to replicate some of what happens automatically for warnings, where (e.g.,) DeprecationWarnings are suppressed, and other warnings only show up once.  Since we are capturing the warnings that went away but now works.  I further added in the ability to suppress other messages, either never showing them or only showing them once (when working at the INFO level there are a _lot_ of messages that come through).  There is a default list of these, but it's further extensible by the user with something like:

```
import pint.logging

logfilter = pint.logging.LogFilter()

log.remove()
log.add(sys.stderr, level="INFO", filter=logfilter)
```
and this then works for everything later.  You can also change the logging format with environment variables:
```
export LOGURU_FORMAT="{time} | <lvl>{message}</lvl>"
export LOGURU_LEVEL=WARNING
```
and some nice things like that.

If people don't like this in general I'm happy to drop it, but if you have a minute please check it out with your common tasks.  In particular I haven't played at all with this in a notebook. 

[And a related note is that if you do run even with the standard logging at the INFO or DEBUG level, there are a lot of messages that come out.  We might want to re-think whether they are appropriate at the levels they are at]


dlakaplan:
So this works within a notebook, but the colorization of output does not.  there are some possibilities on how to do that but I don't want to go too far down unless there is broader support.

dlakaplan:
Here's what it looks like for a short fit with `level=DEBUG`:
Screen Shot 2022-03-03 at 2 44 26 PM

scottransom:
I like the way it looks and the additional customization abilities.  I'm not bothered at all about the extra dependency.  I think this gets a thumbs up from me.

dlakaplan:
The configuration has been moved from `__init__.py` to `logging.py`

dlakaplan:
@scottransom : I just updated `zima` to use this logging method.  I added some ways to adjust the verbosity as a command-line argument, which is slightly more complicated than the basic usage but is hopefully a good example.  I also updated some of the comments in `logging.py` to be more helpful: are there other questions that people have?

Examples (following the `zima` test options):

Default level (warning):
Screen Shot 2022-04-14 at 11 21 42 AM

`-v` (info):
Screen Shot 2022-04-14 at 11 21 51 AM

`-vv` (debug):
Screen Shot 2022-04-14 at 11 22 07 AM

[I also changed the default formatting]

paulray:
All those messages about columns existing and being deleted, should be changed to DEBUG instead of INFO, I think.

An as an aside, how do you read that blue text on black background! Those lines basically disappear to me.
Of course, I use white backgrounds, so it will look different for me...

dlakaplan:
I didn't do all of those `INFO` -> `DEBUG` changes, but I agree in principle

dlakaplan:
And my terminal colors are inherited from Joe Lazio from ~1996.

scottransom:
I was thinking the same thing about the DEBUG colors.  They are jarring...  Is a mid-tone green visible on both dark and light backgrounds?

scottransom:
As an FYI, both `ipython` and  `pytest` use --log-level for command-line setting.  So that's what I have already done for `pintempo` and `pintk`:

```
    parser.add_argument(
        "--log-level",
        help="Logging-level to use ('DEBUG', 'INFO', 'WARN', 'ERROR')",
        default=None,
        dest="loglevel",
    )
```

scottransom:
Oh, and as for default formatting.  Can you get the module on the output? Maybe after the level?  For `logging` the format variable is `%(name)s.  I've been playing with the following format:
```
format="%(levelname)s (%(name)s): %(message)s"
```

dlakaplan:
* Default formatting: sure
* Command-line arguments: OK (do you then make it a `choices` type?)
* Default colors: Can you suggest a specific color?  RGB, hex, ... (https://loguru.readthedocs.io/en/stable/api/logger.html#color)

scottransom:
I think it could be a choices type.  What I did above was just playing around.  If you put in an unknown level you get a ValueError:
```
❯ pintk --log-level scott NGC6440E.par NGC6440E.tim                                                                           14:49:10
  File "/home/sransom/python_venvs/py39/bin/pintk", line 33, in <module>
    sys.exit(load_entry_point('pint-pulsar', 'console_scripts', 'pintk')())
  File "/home/sransom/git/PINT/src/pint/scripts/pintk.py", line 168, in main
    logging.basicConfig(
  File "/usr/lib/python3.9/logging/__init__.py", line 2022, in basicConfig
    root.setLevel(level)
  File "/usr/lib/python3.9/logging/__init__.py", line 1421, in setLevel
    self.level = _checkLevel(level)
  File "/usr/lib/python3.9/logging/__init__.py", line 198, in _checkLevel
    raise ValueError("Unknown level: %r" % level)
ValueError: Unknown level: 'scott'
```

scottransom:
Just tried this with my `pintk` changes and it looks (and acts) great!  I'm totally thumbs up for it.  Very nice!

scottransom:
And BTW, both the blue and yellow look great in my normal terminal coloring...

dlakaplan:
What color does INFO appear in your terminals?  It looks white to me: is it black if you have a white background?

dlakaplan:
Actually, I guess INFO is just bold, so that will be whatever your default fg color is

dlakaplan:
Dawn suggests https://www.google.com/search?si=ANhW_NonC3uofTTjpwVeu5wkQU_Ekh9huyLrbt148qVE5sXFJW-Vgt3IKPJfFRQiI2sTg7YoQ0gzPBxUgGooPaGrSxrPyLAzntnB55CJdkmgFpIunyz-mVvFuhpqXFMlghqkAas0-pDy&hl=en-US&kgs=d15bac56792de783&shndl=21&source=sh/x/fbx/1&entrypoint=sh/x/fbx (hex #b790d4)  for DEBUG.

scottransom:
That's a great color.  I like it.  Although the colors that I'm seeing by default on my terminal are already great:
image

scottransom:
Another one that I'm seeing a bunch of instances of is: "Skipping Shapiro delay for Barycentric TOAs"

dlakaplan:
Just updated the format and colors a bit. Let me know what you think.  I also added that other message that was annoying Scott to the "only once" list.

Note that in the example below you can see a warning coming from `warnings.warn`:
Screen Shot 2022-04-15 at 9 22 30 AM
 This ends up coming out twice, once using it's own method and then it's captured by the loguru logger and reissued.  I can turn off the second issue if desired.  What I don't know how to do yet is make it look like it's coming from the actual module that issues it rather than `pint.logging`.

dlakaplan:
It looks like `pinttempo` still uses `astropy.log`, but maybe @scottransom  has already fixed.

scottransom:
@dlakaplan Yes, I'm handling the `pintempo` and all the `pintk` changes (and have them all done, just waiting for this to get merged).  There is one more common warning that happens with random models, and that is:
"Special observatory location. No clock corrections applied."
So maybe add that to the list as well?  I like that we can turn the filters off fairly easily if we want to see everything.
This is really looking good!

scottransom:
Another thing that I added which might make more sense to go in yours, was this extra debugging info that explains the solar system barycentering for the absolute phase TOA.
```
Δ src/pint/models/absolute_phase.py
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

──────────────────────────────────────┐
• 94: class AbsPhase(PhaseComponent): │
──────────────────────────────────────┘
        # Otherwise we have to build the TOA and apply clock corrections
        # NOTE: Using TZRMJD.quantity.jd[1,2] so that the time scale can be properly
        # set to the TZRSITE default timescale (e.g. UTC for TopoObs and TDB for SSB)
        log.debug("Creating and dealing with the single TZR_toa for absolute phase")
        TZR_toa = toa.TOA(
            (self.TZRMJD.quantity.jd1 - 2400000.5, self.TZRMJD.quantity.jd2),
            obs=self.TZRSITE.value,

───────────────────────────────────────┐
• 107: class AbsPhase(PhaseComponent): │
───────────────────────────────────────┘
            ephem=toas.ephem,
            planets=toas.planets,
        )
        log.debug("Done with TZR_toa")
        self.tz_cache = tz
        self.tz_hash = hash((self.TZRMJD.value, self.TZRSITE.value, self.TZRFRQ.value))
        self.tz_clkc_info = clkc_info
```

dlakaplan:
Added.  What about the other scripts?  I don't use most of them but as long as there are tests defined so that I can figure out reasonable inputs I can change them.  I see:
* `convert_parfile.py` (I actually wrote that, so I'll fix it)
* `event_optimize.py`
* `event_optimize_MCMCFitter.py`
* `event_optimize_multiple.py`
* `fermiphase.py`
* `photonphase.py`
so it looks like most of those are single-photon related.  Should I work on these or leave them for @paulray ?

scottransom:
Oh, and one other thing I found was at least one (in residuals.py) `log.warn(` rather than `log.warning(`

scottransom:
Good question about the other scripts.  I definitely think those should be fixed too.  If you can easily and quickly make the additions, I'd say go for it.  Or I can do it with my pintk bugfixes (which I wasn't going to put a PR in for).  Since you have a PR, maybe that makes sense for you to just add them?

dlakaplan:
I added those debug statements, the message to skip, and fixed `convert_parfile`

paulray:
Yeah, if you can do the basic changes to the scripts that would be great. Once this is merged, I'm happy to tweak the log levels.

scottransom:
OK.  So @dlakaplan as soon as you are happy, let me know and I'll merge this.  Since I've been using it and have looked it over I think it is ready.

dlakaplan:
OK, scripts in progress.  I'm making the default log level WARNING - easy to change.

dlakaplan:
@paulray : I'm also slightly changing the `argparse` for the scripts to use:
```
        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
```
so that it prints out the default automatically.  Hopefully this is OK.

[as a separate note: for `event_optimize` I notice you've implemented your own progress updating.  You might try `tqdm`]

dlakaplan:
OK, I updated:
* event_optimize.py
* event_optimize_MCMCFitter.py
* event_optimize_multiple.py
* fermiphase.py
* photonphase.py
where possible I tested using the arguments in the test suite, but I couldn't do that for `event_optimize_multiple` or `event_optimize_MCMCFitter`: the former crashed, and the latter didn't seem to generate an executable.

For the others the testing was minimal: I ran one of the test sequences and it ran with no error messages, but I didn't examine the output in detail. But the standard tests passed, so I don't think I broke anything.

paulray:
Yeah, I'm not sure about those other two event_optimize scripts. I don't think they are really supported.

dlakaplan:
Tests have now passed.  Probably OK to merge.

scottransom:
Excellent!  Thanks a bunch for this, David.  I've looked over the code and tested it.  Merging now.

Issue #1181, opened 2022-03-09
Add selection by cluster to maskParameter


dlakaplan:
Should be able to select TOAs for a Jump etc. by cluster number (used to be group number)

dlakaplan:
I think the clustering can set a column in the TOA table but doesn't have to.  But you are right - this will be an issue when writing out the data so perhaps that's a better solution.

PR #1182, opened 2022-03-09
WIP: maskParameter (e.g., for jump) select by clusters

dlakaplan:
You can define TOA clusters (using gaps etc) by:
```
clusters = t.get_clusters(add_column=True)
```
It seems that you should be able to use that column (when added) to select TOAs for jumping.  Previously this wasn't possible (I think) but this seems to work.

E.g., the old way:
```
    "JUMP",
    key="mjd",
    value=0.0,
    key_value=[t[clusters==3].get_mjds().min().value,
               t[clusters==3].get_mjds().max().value],
    units=u.s,
    frozen=False,
    )
```

And I get the same results via:
```
    "JUMP",
    key="cluster",
    value=0.0,
    key_value=3,
    units=u.s,
    frozen=False,
    )
```

Was I missing anything in another way to do this?

I think this works, just need to test more.

### DIFF
--- a/src/pint/models/absolute_phase.py
+++ b/src/pint/models/absolute_phase.py
@@ -2,6 +2,7 @@
 import logging
 
 import astropy.units as u
+from astropy.table import Column
 
 import pint.toa as toa
 from pint.models.parameter import MJDParameter, floatParameter, strParameter
@@ -109,6 +110,8 @@ class AbsPhase(PhaseComponent):
             ephem=toas.ephem,
             planets=toas.planets,
         )
+        if "clusters" in toas.table.colnames:
+            tz.table.add_column(Column([-1], name="clusters"))
         self.tz_cache = tz
         self.tz_hash = hash((self.TZRMJD.value, self.TZRSITE.value, self.TZRFRQ.value))
         self.tz_clkc_info = clkc_info
@@ -117,7 +120,7 @@ class AbsPhase(PhaseComponent):
         return tz
 
     def make_TZR_toa(self, toas):
-        """ Calculate the TZRMJD if one not given.
+        """Calculate the TZRMJD if one not given.
 
         TZRMJD = first toa after PEPOCH.
         """
--- a/src/pint/models/parameter.py
+++ b/src/pint/models/parameter.py
@@ -455,9 +455,10 @@ class Parameter:
 
         .. [1] https://github.com/nanograv/PINT/wiki/PINT-vs.-TEMPO%282%29-par-file-changes
         """
-        assert format.lower() in _parfile_formats, (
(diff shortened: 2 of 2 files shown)


scottransom:
Didn't we make it so that all of our `key_values` are supposed to be strings?  Or am I misremembering?  Or maybe that is still in an unmerged PR?  I believe that has been the intent.

dlakaplan:
MJDs can definitely be floats.  I think all "flags" are strings.

dlakaplan:
I guess the other solution to this was to add a flag and jump based on that:
```
cp.add_jump_and_flags(setup_NGC6440E.t.table["flags"][selected_toa_ind])
```
(from `test_jumps.py`).  Is that better?

Also, when should one use a `PhaseJump` vs. a `DelayJump`? I would have thought this was obvious via units, but I seem to be able to add a `PhaseJump` specified in `u.s`.

dlakaplan:
I added tests to explicitly compare selecting the TOAs to jump by:
* range of MJDs
* cluster value
* using `add_jump_and_flags` (which is most of what's tested in `test_jumps`)

It checks that both the TOA selection and the fit results are the same.

paulray:
I didn't even know about get_clusters!

dlakaplan:
Looking at the docs:
```
class DelayJump(DelayComponent):
    """Phase jumps

    Parameters supported:

    .. paramtable::
        :class: pint.models.jump.DelayJump

    Note
    ----
    This component is disabled for now, since we don't have any method
    to identify the phase jumps and delay jumps.
    """
```
although it does work.  but then:
```
class PhaseJump(PhaseComponent):
    """Arbitrary jumps in pulse phase.

    In spite of the name, the amounts here are specified in seconds and
    converted to phase using F0.
```
so they are effectively the same but the `PhaseJump` has additional methods:
```
get_number_of_jumps
get_jump_param_objects
jump_params_to_flags
add_jump_and_flags
```
Should we remove `DelayJump` to remove the confusion?  Make `PhaseJump` actually specified in cycles (and make it use a more accurate frequency calculation)?

dlakaplan:
Unless I missed another way to do this, I think this is ready for review

scottransom:
@dlakaplan Have you seen the "jump all groups" code in `pintk`?
https://github.com/nanograv/PINT/blob/28cb52d001d6a5eb3d7c6e06e0d550c4877c0dd9/src/pint/pintk/plk.py#L1442

Can your code make that code better?  Or are they linked in some way?  I'll try to take a look at this over the next day or two, but wanted you to see it.  (and things in `pintk` are changing fast)

aarchiba:
Some discussion on #1181

PR #1189, opened 2022-03-25
Update gridding

dlakaplan:
Add some improvements to how the gridding is done:
* Allows one to specify other parameters to return at each grid location with `extraparnames`.  
* Can also iterate over specific tuples instead of a full grid (to save time or allow non-uniform sampling) with `tuple_chisq`/`tuple_chisq_derived`.  
* Can pass additional arguments to the fitter, like `maxiter=...`
* Deals with fitters that don't converge, either returning the best current solution (if perhaps it's not converged) or NaN (if max iterations reached or another exception encountered)
* Plus bug fixes (handling of NaNs etc)


dlakaplan:
I also pulled in #1187 because I wanted to be able to run an example with both of those:
```alts = m.alternative_solutions()
    f = fitter.DownhillGLSFitter(t, m)
    
    parnames = ("KIN","KOM")
    chi2, _ = pint.gridutils.tuple_chisq(f, parnames, alts)
```

dlakaplan:
Maybe @paulray you can check this one?

paulray:
I reviewed the code and it all looks good to me. Very useful additions!

PR #1190, opened 2022-03-28
Fix a couple bugs

paulray:
This PR fixes an incorrect variable name and a poor usage of `np.where()` in some random models code that caused a crash in pintk sometimes.


scottransom:
Awesome.  Thanks.

PR #1193, opened 2022-03-31
Fixes profiling code

### LINKED ISSUE #1191: profiler not running
Looks like it's missing some data files.  Also the `gprof2dot` module should be included in the dev requirements.

paulray:
Fixed calls to new gridding API
Added README with instructions for some dependencies
Added par file that was missing

Did not add the large tim file, but put the curl command to get it into the README.

Fixes #1191

### DIFF
--- a/profiling/NGC6440E.par
+++ b/profiling/NGC6440E.par
@@ -1,0 +1,20 @@
+PSR              1748-2021E
+RAJ       17:48:52.75  1 0.05
+DECJ      -20:21:29.0  1 0.4
+F0       61.485476554  1 5e-10
+F1         -1.181D-15  1 1e-18
+PEPOCH        53750.000000
+POSEPOCH      53750.000000
+DM              223.9  1 0.3
+SOLARN0               0.00
+EPHEM               DE421
+CLK              UTC(NIST)
+UNITS               TDB
+TIMEEPH             FB90
+T2CMETHOD           TEMPO
+CORRECT_TROPOSPHERE N
+PLANET_SHAPIRO      N
+DILATEFREQ          N
+TZRMJD  53801.38605120074849
+TZRFRQ            1949.609
+TZRSITE                  1
--- a/profiling/README.txt
+++ b/profiling/README.txt
@@ -1,0 +1,28 @@
+This directory contains a suite of benchmark codes for PINT. It only lives here and is not installed as part of
+any PINT packages.
+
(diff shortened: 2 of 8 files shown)


paulray:
@dlakaplan if you are OK with this, feel free to merge.
Obviously the test failure is the broader problem with astropy IERS downloads.

dlakaplan:
is the `black` failure only the click thing?  Should you fix that here too or just do that with a separate PR?

paulray:
There were actually a couple of non-black files. Those are now fixed.  I thought the click fix was in your PR. Should I add it here?

dlakaplan:
Yes, it is in my PR.  I think it's fine to leave it there.  But more tests have failed now.

paulray:
Seems to be just that IERS file issue. I wonder if we can tweak astropy in the CI to not do that download?

dlakaplan:
I'm running the tests on your branch locally just to verify that most things pass.  maybe we should discuss this more broadly just so we don't set a bad precedent?  Or wait until we have a fix for the `astropy` issue since this isn't super urgent (I might be able to work on the astropy thing next week).

paulray:
I'm in favor of merging now and dealing with the astropy issue separately. 
`make test` runs all pass on my machine, and the profile code runs, so I think this is ready.

dlakaplan:
Do you also include `gprof2dot` in `requirements_dev.txt`?  I didn't see that in the list of changes.  It also uses the command `dot`, and I don't know where that is from.

paulray:
I just wrote this in the README:
```# Also, a couple of other packages are required:
# First, you need the "dot" program, which is part of the graphviz package (try your package manager to install this)
# Then use pip to install the following:
pip install gprof2dot py-cpuinfo```
The dot and gprof2dot codes are only required to run `run_profile.py` and not for the main `high_level_benchmark.py` so I wasn't sure if I should add them to the global PINT `requirements_dev.tex`

dlakaplan:
I guess if they are pip-installable, I would favor adding them.  If not the README is OK.  But if you think it's cleaner to do it all through the README I can be convinced.

paulray:
I hate to add too many dependencies, but since this is just `requirements_dev.tex` it won't affect most users.
I'll add the two pip packages.

dlakaplan:
OK, I'll merge despite the errors.

PR #1195, opened 2022-04-11
Feature photonphase polycos

sashalvna:
A polycos argument has been added to photonphase.py, which uses polycos to calculate pulse phases without using the full model of the pulsar. Useful for large files, where photonphase.py previously used to crash/time out. A test of the polycos argument has also been added to test_photonphase.py.


paulray:
What you might do is make a new branch and make sure it is fully up to date with the current nanograv/master, then just commit your changes to `photonphase.py` and `test_photonphase.py`?

paulray:
Also, the B1509 datafile already exists in `tests/datafile`, so should not be included in your PR.

scottransom:
The way that you would normally handle accounting for new changes to the repo is simply to rebase your branch on PINT master.

paulray:
> The way that you would normally handle accounting for new changes to the repo is simply to rebase your branch on PINT master.

But, in this case it looks like the commits in the PR revert desired changes in PINT, like using  `logging` instead of `astropy.log`.  So, it seems like git rebase won't have the desired effect.

PR #1201, opened 2022-04-14
Allow reading of archival tim files that use 'w' for ncyobs

### LINKED ISSUE #1200: PINT fails reading ncyobs TOAs that use the 'w' observatory code
The special NUPPI-specific Nancay observatory "ncyobs", which requires special clock corrections, has archival TOAs that use the (likely old and custom) single-character observatory code "w".  IPTA DR2 is an example of published (and public) data where this is the case.  A fix is simply to add "w" as an alias for "ncyobs" (alongside "nuppi").  I will make a PR for that ASAP.

scottransom:
This PR fixes #1200.  Tempo2 recognizes 'w' as "ncyobs" and we should too, especially since that is the observatory code used in IPTA DR2.

### DIFF
--- a/src/pint/observatory/observatories.py
+++ b/src/pint/observatory/observatories.py
@@ -127,7 +127,7 @@ TopoObs(
 )
 TopoObs(
     "ncyobs",
-    aliases=["nuppi"],
+    aliases=["nuppi", "w"],
     itrf_xyz=[4324165.81, 165927.11, 4670132.83],
     clock_fmt="tempo2",
     clock_file=["ncyobs2obspm.clk", "obspm2gps.clk"],


paulray:
I never knew about this alias!
What is the deal with WSRT?

scottransom:
I was wrong about the issue being WSRT.  I was thinking it was a WSRT given the "w" flag was what was causing the issues, and I thought that it was dying on the WSRT file, but it was dying on the previous one, a NRT file!  So I screwed that up and WSRT is OK.  But "ncyobs" is the one that needs the "w" alias.  (and note that when that alias is in place, the B1937+21 TOAs look *way* better than if you treat the TOAs as WSRT ones!).

PR #1202, opened 2022-04-14
WIP: keep Kopeikin corrections in ECL coordinates

dlakaplan:
Note that it is not smart enough to keep in ICRS if the model is ICRS (yet) but that is the eventual goal

So far I didn't recheck all of the equations but just did `alpha` -> `ELONG` and `delta` -> `ELAT`

### DIFF
--- a/src/pint/models/stand_alone_psr_binaries/DDK_model.py
+++ b/src/pint/models/stand_alone_psr_binaries/DDK_model.py
@@ -33,8 +33,8 @@ class DDKmodel(DDmodel):
         self.param_default_value.update(
             {
                 "KIN": 0 * u.deg,
-                "PMRA_DDK": 0 * u.mas / u.year,
-                "PMDEC_DDK": 0.5 * u.mas / u.year,
+                "PMELONG": 0 * u.mas / u.year,
+                "PMELAT": 0.5 * u.mas / u.year,
                 "PX": 0 * u.mas,
                 "KOM": 0 * u.deg,
                 "K96": True,
@@ -76,26 +76,29 @@ class DDKmodel(DDmodel):
         (Kopeikin 1995 L6 Eq10)
         """
         self._psr_pos = val
-        self._sin_delta = self._psr_pos[:, 2]
-        self._cos_delta = np.cos(np.arcsin(self._sin_delta))
-        self._sin_alpha = self._psr_pos[:, 1] / self._cos_delta
-        self._cos_alpha = self._psr_pos[:, 0] / self._cos_delta
+        # it appears that this is the vector K0
+        # K0 = (cos(alpha)*cos(delta), sin(alpha)*cos(delta), sin(delta))
+        # in equatorial coords
+        self._sin_elat = self._psr_pos[:, 2]
+        self._cos_elat = np.cos(np.arcsin(self._sin_elat))
+        self._sin_elong = self._psr_pos[:, 1] / self._cos_elat
+        self._cos_elong = self._psr_pos[:, 0] / self._cos_elat
 
     @property
-    def sin_delta(self):
-        return self._sin_delta
+    def sin_elat(self):
+        return self._sin_elat
 
     @property
-    def cos_delta(self):
-        return self._cos_delta
+    def cos_elat(self):
+        return self._cos_elat
 
     @property
-    def sin_alpha(self):
-        return self._sin_alpha
+    def sin_elong(self):
+        return self._sin_elong
 
     @property
-    def cos_alpha(self):
-        return self._cos_alpha
+    def cos_elong(self):
+        return self._cos_elong
 
     @property
     def SINI(self):
@@ -125,7 +128,7 @@ class DDKmodel(DDmodel):
         d_KIN = (-PMRA_DDK * sin(KOM) + PMDEC_DDK * cos(KOM)) * (t-T0)
         """
         d_KIN = (
-            -self.PMRA_DDK * self.sin_KOM + self.PMDEC_DDK * self.cos_KOM
+            -self.PMELONG * self.sin_KOM + self.PMELAT * self.cos_KOM
         ) * self.tt0
         return d_KIN.to(self.KIN.unit)
 
@@ -142,7 +145,7 @@ class DDKmodel(DDmodel):
     def d_SINI_d_KOM(self):
         if self.K96:
             d_si_d_kom = (
-                (-self.PMRA_DDK * self.cos_KOM - self.PMDEC_DDK * self.sin_KOM)
+                (-self.PMELONG * self.cos_KOM - self.PMELAT * self.sin_KOM)
                 * self.tt0
                 * np.cos(self.kin())
             )
@@ -154,7 +157,7 @@ class DDKmodel(DDmodel):
     def d_SINI_d_T0(self):
         if self.K96:
             d_si_d_kom = -(
-                -self.PMRA_DDK * self.sin_KOM + self.PMDEC_DDK * self.cos_KOM
+                -self.PMELONG * self.sin_KOM + self.PMELAT * self.cos_KOM
             )
             return d_si_d_kom.to(u.Unit("") / self.T0.unit)
         else:
@@ -170,7 +173,7 @@ class DDKmodel(DDmodel):
 
     def d_kin_proper_motion_d_KOM(self):
         d_DKIN_d_KOM = (
-            -self.PMRA_DDK * self.cos_KOM - self.PMDEC_DDK * self.sin_KOM
+            -self.PMELONG * self.cos_KOM - self.PMELAT * self.sin_KOM
         ) * self.tt0
         return d_DKIN_d_KOM.to(
             self.KIN.unit / self.KOM.unit, equivalencies=u.dimensionless_angles()
@@ -178,7 +181,7 @@ class DDKmodel(DDmodel):
 
     def d_kin_proper_motion_d_T0(self):
         d_DKIN_d_T0 = -1 * (
-            -self.PMRA_DDK * self.sin_KOM + self.PMDEC_DDK * self.cos_KOM
+            -self.PMELONG * self.sin_KOM + self.PMELAT * self.cos_KOM
         )
         return d_DKIN_d_T0.to(
             self.KIN.unit / self.T0.unit, equivalencies=u.dimensionless_angles()
@@ -201,8 +204,8 @@ class DDKmodel(DDmodel):
         """The correction on a1 (projected semi-major axis)
         due to the pulsar proper motion
         (KOPEIKIN. 1996 Eq 8.)
-        d_x = a1 * cot(kin) * (-PMRA_DDK * sin(KOM) + PMDEC_DDK * cos(KOM)) * (t-T0)
-        d_kin = (-PMRA_DDK * sin(KOM) + PMDEC_DDK * cos(KOM)) * (t-T0)
+        d_x = a1 * cot(kin) * (-PMELONG * sin(KOM) + PMDEC_DDK * cos(KOM)) * (t-T0)
+        d_kin = (-PMELONG * sin(KOM) + PMDEC_DDK * cos(KOM)) * (t-T0)
         d_x = a1 * d_kin * cot(kin)
         """
         a1 = self.a1_k(False, False)
@@ -253,14 +256,14 @@ class DDKmodel(DDmodel):
         """The correction on omega (Longitude of periastron)
         due to the pulsar proper motion
         (KOPEIKIN. 1996 Eq 9.)
-        d_omega = csc(i) * (PMRA_DDK * cos(KOM) + PMDEC_DDK * sin(KOM)) * (t-T0)
+        d_omega = csc(i) * (PMELONG * cos(KOM) + PMDEC_DDK * sin(KOM)) * (t-T0)
         """
         kin = self.kin()
         sin_kin = np.sin(kin)
         omega_dot = (
             1.0
             / sin_kin
-            * (self.PMRA_DDK * self.cos_KOM + self.PMDEC_DDK * self.sin_KOM)
+            * (self.PMELONG * self.cos_KOM + self.PMELAT * self.sin_KOM)
         )
         return (omega_dot * self.tt0).to(self.OM.unit)
 
@@ -271,7 +274,7 @@ class DDKmodel(DDmodel):
         d_omega_dot = (
             -cos_kin
             / sin_kin ** 2
-            * (self.PMRA_DDK * self.cos_KOM + self.PMDEC_DDK * self.sin_KOM)
+            * (self.PMELONG * self.cos_KOM + self.PMELAT * self.sin_KOM)
         )
         return (d_omega_dot * self.tt0).to(
             self.OM.unit / self.KIN.unit, equivalencies=u.dimensionless_angles()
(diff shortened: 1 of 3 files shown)


dlakaplan:
I think this now works with ICRS and ECL models.  But the testing isn't there yet.  A review would be helpful.

dlakaplan:
I'm still not fully convinced by the comparison, but I now have separate ECL and ICRS par files for the unit tests. I then computed the residuals binary delays using `libstempo` (which I guess uses `tempo2`) for each model.  The agreement is good but not perfect:
* residual max difference is 50ns (ECL) and 10ns (ICRS).  
* binary delay max difference is 5us (ECL) and 6us (ICRS).
Why should those be so high?  They were 0.5us in the previous tests, where I guess the argument was that since the PINT and `tempo` calculations used different reference frames it was OK if they were not the same.  Now they should be even closer, not worse.

[I am trying to do all of these comparisons pre-fit.  Hopefully that is working as desired.]

dlakaplan:
I'm also unsure of how the 5us differences in the binary delay are compatible with 10ns residual differences...

dlakaplan:
Here are the residuals comparing PINT (with the two different coordinate systems) and libstempo.  I did the same comparison with J1600 and could reproduce the results at https://nanograv-pint.readthedocs.io/en/latest/examples-rendered/paper_validation_example.html, so I think I'm doing this correctly. 

Neither one is large in an absolute sense, but I'm still a bit confused as to why the ECL version is so much worse and why they are both worse than the DD comparison for J1600.  Does it have to do with extra coordinate conversions/comparisons when determining the orientation angle of the orbit?

DDK_comparison
[Note: these are pre-fit residuals]

src/pint/models/stand_alone_psr_binaries/DDK_model.py
```diff
@@ -251,14 +257,14 @@ def delta_omega_proper_motion(self):
         """The correction on omega (Longitude of periastron)
         due to the pulsar proper motion
         (KOPEIKIN. 1996 Eq 9.)
-        d_omega = csc(i) * (PMRA_DDK * cos(KOM) + PMDEC_DDK * sin(KOM)) * (t-T0)
+        d_omega = csc(i) * (PMLONG_DDK * cos(KOM) + PMDEC_DDK * sin(KOM)) * (t-T0)
```
swiggumj, line 260: Also here: shouldn't `PMDEC_DDK` be `PMLAT_DDK`?

src/pint/models/stand_alone_psr_binaries/DDK_model.py
```diff
@@ -199,8 +205,8 @@ def delta_a1_proper_motion(self):
         """The correction on a1 (projected semi-major axis)
         due to the pulsar proper motion
         (KOPEIKIN. 1996 Eq 8.)
-        d_x = a1 * cot(kin) * (-PMRA_DDK * sin(KOM) + PMDEC_DDK * cos(KOM)) * (t-T0)
-        d_kin = (-PMRA_DDK * sin(KOM) + PMDEC_DDK * cos(KOM)) * (t-T0)
+        d_x = a1 * cot(kin) * (-PMLONG_DDK * sin(KOM) + PMDEC_DDK * cos(KOM)) * (t-T0)
```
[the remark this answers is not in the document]
dlakaplan: Those are just in docstrings, but fixed.

src/pint/models/stand_alone_psr_binaries/DDK_model.py
dlakaplan: Same

dlakaplan:
I've tried to go through and convert all of the math in the docstrings to actual sphinx math.  It could probably be done better, explaining things more at the top level and then just referring to those equations later.  But I think it's at least better now.

Note: a good way to make the docs easier to read would be to do this uniformly for all binary models.

dlakaplan:
Note that I can reproduce the binary delays used in the previous DDK tests using `libstempo`.  I can't quite get the residual calculation to match: it's off by +/-5ns.  I don't understand that since it should be a `libstempo` to `libstempo` comparison.

src/pint/models/stand_alone_psr_binaries/DDK_model.py
```diff
@@ -12,16 +12,34 @@
 
 class DDKmodel(DDmodel):
     """DDK model, a Kopeikin method corrected DD model.
-    The main difference is that DDK model considers the annual parallax of earth and
-    the proper motion of the pulsar.
-    effects on the pulsar binary parameters.
-    Speical parameters are:
+    The main difference is that DDK model considers the effects on the pulsar binary parametersfrom the  annual parallax of earth and    the proper motion of the pulsar.
```
aarchiba, line 15: Spacing problems

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - Multiple bug fixes in get_summary()/get_derived_params(), especially for binary calculations
 ### Changed
 - Changed logging to use `loguru`
+- DDK model will now use ICRS or ECL coordinates depending on what the input model is
```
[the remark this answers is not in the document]
dlakaplan: I think this only applies to J1713.

aarchiba:
> I could write that test, but I didn't think it would be sufficiently independent (since the KOM calculation would be PINT-based in both cases).

I think that is an essential sanity check - that PINT is computing DDK the same way whether you use ECL or ICRS. It's not enough, but if PINT was doing DDK right before in ICRS, this ensures that it's doing DDK correctly in ECL. The TEMPO2 weirdness suggests that perhaps PINT would fail this test, and one could then debug the problem without worry about what's going on in libstempo2.

dlakaplan:
So I have this:
```
import io
import numpy as np
import copy

from astropy import units as u, constants as constants
import pint.toa as toa
import pint.simulation
import pint.fitter
import pint.residuals

temp_par_str_ECL = """
    PSR  J1713+0747
    LAMBDA 256.66  1 0.001
    BETA 30.70036  1 0.001
    PMLAMBDA 5.2671  1  0.0021
    PMBETA  -3.4428  1  0.0043
    PX  0.8211  1  0.0258
    F0  218.81  1  0.01
    PEPOCH  55391.0
    BINARY  DDK
    A1 32.34  1  0.001
    E  0.074  1  0.001
    T0 55388.836  1  0.0002
    PB 67.825129  1  0.0001
    OM 176.19845  1  0.0013
    M2  0.283395  1  0.0104
    KOM   83.100  1  1.800
    KIN  71.969  1               0.562
    K96  1
"""

mECL = get_model(io.StringIO(temp_par_str_ECL))
tECL = pint.simulation.make_fake_toas_uniform(
    50000, 60000, 100, mECL, error=0.5 * u.us, add_noise=True
)
prefit = pint.residuals.Residuals(tECL, mECL)
print(f"Prefit chi^2 in ECL {prefit.calc_chi2()}")
# get proper motion vector and normalize
pm_ECL = np.array(
    [mECL.coords_as_ECL().pm_lon_coslat.value, mECL.coords_as_ECL().pm_lat.value]
)
pm_ECL /= np.sqrt(np.dot(pm_ECL, pm_ECL))

mICRS = mECL.as_ICRS()
pm_ICRS = np.array(
    [mICRS.coords_as_ICRS().pm_ra_cosdec.value, mICRS.coords_as_ICRS().pm_dec.value]
)
pm_ICRS /= np.sqrt(np.dot(pm_ICRS, pm_ICRS))
# get the angle between proper motion vectors, which should be the difference between KOMs
angle = np.arccos(np.dot(pm_ECL, pm_ICRS)) * u.rad

rICRS = pint.residuals.Residuals(tECL, mICRS)
print(
    f"Prefit chi^2 in ICRS with the same KOM ({mICRS.KOM.quantity}) {rICRS.calc_chi2()}"
)
mICRS_newKOM = copy.deepcopy(mICRS)
mICRS_newKOM.KOM.quantity -= angle
rICRS_newKOM = pint.residuals.Residuals(tECL, mICRS_newKOM)
print(
    f"Change KOM by {angle.to(u.deg)} to {mICRS_newKOM.KOM.quantity}, now chi^2 in ICRS is {rICRS_newKOM.calc_chi2()}"
)

f = pint.fitter.WLSFitter(tECL, mICRS_newKOM)

f.fit_toas()
print(
    f"After fitting, KOM is now {f.model.KOM.quantity} and chi^2 is {f.resids.calc_chi2()}"
)
```
which mostly does that.  And it seems to make sense all up to the last step where the fit drives both KIN and KOM to a strange place.  chi^2 does decrease, but the values don't make sense.

For reference when I just ran it I got:
```
Prefit chi^2 in ECL 106.76000833279625
Prefit chi^2 in ICRS with the same KOM (83.1 deg) 214.45704197341502
Change KOM by 5.315190059272063 deg to 77.78480994072792 deg, now chi^2 in ICRS is 107.99086669016373
After fitting, KOM is now 90.39162210740693 deg and chi^2 is 90.96982258395022
```

dlakaplan:
So the chi^2 after adjusting KOM by the expected amount agrees with the pre-fit value, but a fit takes it elsewhere

dlakaplan:
Also, if I refit in just the ECL model (which is what I used for the simulation) I still get new values of KIN,KOM with a lower chi^2.

```
Simulated TOAs in ECL with (KIN,KOM) = (71.969 deg,83.1 deg)
Prefit chi^2 in ECL 93.55836628060239
Prefit chi^2 in ICRS with the same KOM (83.1 deg) 230.05499699280733
Change KOM by 5.315190059272063 deg to 77.78480994072792 deg, now chi^2 in ICRS is 93.2744366120018
After fitting, (KIN,KOM) in ECL model is now (71.56777518947246 deg,81.11508636814123 deg) and chi^2 is 84.80118708590243
After fitting, (KIN,KOM) is now (71.50780414031163 deg,76.25187154200293 deg) and chi^2 is 84.45013111351614
```

aarchiba:
So converting ECL to ICRS and converting KOM as part of the process decreases the chi2 by 0.28 - seems like a lot for supposedly identical models.

Are both initial and post-fit models fully converged? WLSFitter need not converge in one step the way you're calling it.

What chi2 do you get when you convert the post-fit ICRS model back to ECL? Does if still differ by about 0.28?

It seems to me your ECL -> IERS conversion is done correctly but I'd need to sit down and do some careful 3D math to check it. Could you cross-check by making a unit vector based on KOM and doing an astropy position-and-proper-motion conversion?

dlakaplan:
I changed it so the KOM is now transformed along with the model coordinates/proper motions in the `as_ICRS()` and `as_ECL()` methods, and then did a round-trip check on the value.  That looks fine.

dlakaplan:
If I start in ECL, transform to ICRS, and back to ECL I get the identical chi^2 without fitting.

dlakaplan:
I also changed to downhill fitters with `maxiter=10`, and I think they are converged.   Results are still the same.

So, summary:
* The coordinate conversion will now do KOM, and the difference agrees with what we found empirically
* But the chi^2 isn't exactly the same between the ICRS and ECL versions (differs by <1 usually)
* And if we fit both models independently, then end up with an empirical difference close to the prediction but they don't end up exactly where we simulated them (and again, chi^2 is different by 1 or 2)

dlakaplan:
Looking at the residuals when the TOAs are perfect, it looks like there is an error in the proper motion:
sim_DDK_comparison
Note that I don't think this is an overall problem with the coordinate conversion, because if I do the same test with a DD model it looks fine:
sim_DD_comparison

Is there some small component of the binary motion that ends up in the proper motion which may not be transformed perfectly?

dlakaplan:
And here's a comparison as a function of orbital phase.  Clearly important, but not the whole answer.
sim_DDK_orbitalcomparison

aarchiba:
>     * The coordinate conversion will now do KOM, and the difference agrees with what we found empirically
> 
>     * But the chi^2 isn't exactly the same between the ICRS and ECL versions (differs by <1 usually)
>
>     * And if we fit both models independently, then end up with an empirical difference close to the prediction but they don't end up exactly where we simulated them (and again, chi^2 is different by 1 or 2)


If the chi-squared is different by more than about 0.01 the fitters are certainly going to settle somewhere different. This is just the result of the chi-squared difference, which seems to be the result of the difference in residuals between ECL  and IERS versions of what should be the same model.

So it seems that the puzzle is: why does changing the coordinate system result in residuals that differ by tens of nanoseconds? Is this an approximation built into the DDK mathematics or an error in our implementation of the DDK mathematics or a subtle error in our conversion?

aarchiba:
Would it be quick to copy and rename to get DDK_OLD_PINT that works the way PINT used to? I'm curious whether that gave identical residuals in ECL and ICRS (with no conversion of KOM, obviously). Might narrow down the possible source of the discrepancy?

dlakaplan:
Doing that in the master branch, the residuals agree to  <1ns.    Which I guess shows that the coordinate conversion part is OK.

aarchiba:
Just thinking of teasing apart the different terms - if you set proper motion to zero, do you see discrepancies due to orbital effects? What if proper motion  is nonzero but ELAT is 90 degrees (so no Earth parallax effects I think)? Or am I confused about what effects DDK includes?

dlakaplan:
I'm gradually doing those tests.  So far:
* parallax increases: residuals ~ parallax
* ELAT changes: residuals go down as ELAT goes up, but not ~cos(ELAT) as might expect just from the parallax term
* A1 changes: residuals ~ A1
* KIN changes: residuals ~ 1/sin(KIN) (or at least close to that)
So it's likely some interplay between the standard parallax and orbital parallax (not unexpectedly)

dlakaplan:
* pm -> 0: no change
* ELAT -> 89 (going to 90 creates issues): smaller, but still not 0

aarchiba:
That would seem to indicate that the problem is in one of the terms from Kopeikin 1996 - just a computed A1DOT or OMDOT, depending on the angle between the orientation of the orbit and the proper motion. Is it possible the difference is some second-order term not included in the model? I can't see how, the model should be the same in either coordinate system. It may be necessary to go digging into the guts of DDK and compare it to Kopeikin 1996.

dlakaplan:
I think it's the other way.  Changing proper motion doesn't change the difference, while parallax does.  Specifically:

Changing inclination, A1, and parallax change the amplitude of the difference.

Changing omega or KOM change the phase/shape of the difference (amplitude is fixed).

So it has to be some term that incorporates all of those.  I'm therefore focusing on orbital parallax terms (Kopeikin 1995).

dlakaplan:
This plot also looks *way* better than the ones I had above.  Still not perfect, but I could believe the difference here is a small coordinate difference.
J1713_DDK_comparison_MJD

dlakaplan:
The plot above is still vexing me.  The ICRS comparison looks very good.  The ECL comparison looks mostly as good, but some observations (individual day/receiver sets) are off.  Arecibo seems especially off after MJD 54000:
J1713_DDK_comparison_MJD_telescope
However, since this is a comparison of residuals between tempo2 and PINT, instrumental issues shouldn't really be a problem unless they are somehow different between clock files or something like that.  

Another question is why the GBT differences get worse from MJD 53800 to 54000.  53800 is where the input TOAs become out of order and need to be sorted (since libstempo does this implicitly).  This is done correctly as far as I can tell (comparing TOAs and frequencies).  And it looks fine for the ICRS comparison, so I'm not sure why the ECL comparison won't work.  I've double-checked for differences between the par files a few times and not found anything.

Any thoughts on what could be the issue(s)?  How far is this worth pursuing?  We know that internally to PINT the ICRS and ECL comparisons are very good (with simulated data), and that comparing real data in ICRS between PINT and tempo2 also looks good.  And that real data in ECL look mostly good.

dlakaplan:
I think this is now ready for merging.

aarchiba:
Changelog entry is in the wrong place (claims to be in the released version) but otherwise I'd say go ahead and merge.

dlakaplan:
Thanks - I fixed that.  It was a result of me putting in the change prior to the last release but not updating it.

aarchiba:
I think I may have fixed the network stuff in #1254 ; if it passes CI, we might should merge it and then everyone rebase/merge.

dlakaplan:
Here's the PINT vs PINT internal comparison showing good agreement.
Screen Shot 2022-05-20 at 10 03 46 AM

Issue #1188, opened 2022-03-23
DM derivative output format from TEMPO not recognized by PINT


scottransom:
While PINT can handle DM time derivatives for fitting, the format of the parameters as output by TEMPO (i.e. `DM001`, `DM002`, ...) is not recognized.  PINT expects and ouputs `DM1`, `DM2`, etc.  That fomat *is* accepted by TEMPO as input.  So we just need to improve the parameter parsing (for prefix parameters, I suppose) to recognize the format with the preceeding zeros.

Here is an error example:
```
/home/sransom/python_venvs/py39/lib/python3.9/site-packages/pint/models/model_builder.py:114: UserWarning: Unrecognized parfile line 'DM001 -1.694891583439D-04 1 2.283525222389D-04'
  warnings.warn(f"Unrecognized parfile line '{p_line}'", UserWarning)
/home/sransom/python_venvs/py39/lib/python3.9/site-packages/pint/models/model_builder.py:114: UserWarning: Unrecognized parfile line 'DM002 1.246760990803D-03 1 4.987934997885D-05'
  warnings.warn(f"Unrecognized parfile line '{p_line}'", UserWarning)
```

tcromartie:
Hey @scottransom, I'm trying to figure out the best way to do this. Should it be a change to the way prefixes work (i.e. always ignore the zeros)? I can't think of another parameter that would affect, but I'm not sure. Should it instead be a hardcoded change to make DM00n an alias for DMn? Should it affect both reading and writing? I was sort of thinking this would be as easy as adding an alias somewhere but that doesn't seem to be the case.

tcromartie:
Yeah I'm confused about that too, I thought the second splitting case would take care of it

scottransom:
I think the issue might be because `DM1` is explicitly defined in `dispersion_model.py` without an alias to `DM001`.  If you look in `parameter.py`, you can define aliases to a parameter.  So maybe just try putting that in?  And then maybe the other ones which are not already defined will just work?

scottransom:
Hmmm.  That didn't seem to help...  It is some subtle thing deep in there.  It will probably be a trivial change once we figure out where it goes!

scottransom:
Next place to look is the loop starting on line 337 in `model_builder.py`.  I suspect we aren't treating something correctly there.

scottransom:
Ah-ha...  OK.  This should help.
```
In [3]: ac = AllComponents()

[...]

PR #1204, opened 2022-04-15
Adding special case to handle tempo-style DM001 (vs. DM1)

tcromartie:
This addresses issue #1188 by adding a special case to `alias_to_pint_param` for `DM0*`.


scottransom:
Which line is the one that fixes the issue?  All the changes look like whitespace changes to me.  Maybe I'm missing something really subtle?  And can you please add to one of the test files and include a test that ensures that this parses things correctly?

tcromartie:
@scottransom Lines 3028-3036 and 3043-3046 are new and hopefully address the issue. I don't know why blackening the code resulted in so many changes; I certainly didn't touch many of those lines. I guess the formatting checks must have been added since the last time this was modified? And I marked it as WIP because I know I need to add a test still; let me know if I should have just waited on the PR instead.

scottransom:
Gotcha -- and yeah, I'm not sure how I missed those lines!  I'll take a look at the change.  But you definitely need a test or two!

src/pint/models/timing_model.py
```diff
@@ -3024,13 +3025,25 @@ def alias_to_pint_param(self, alias):
                 first_init_par = None
                 # Handle the case of start index from 0 and 1
                 for start_idx in [0, 1]:
-                    first_init_par_alias = prefix + "{1:0{0}}".format(fmt, start_idx)
+                    # Ensure compatibility with tempo-style "DM001"
+                    if "DM0" in alias:
+                        first_init_par_alias = prefix + "{1:0{0}}".format(
```
scottransom, line 3030: Might be good to change the string formatting here to use f-strings.  They are a lot easier to read.  Same with the string formatting in the "else" case below.

tcromartie: Just so you know, I pushed a change just now that only applies the old black version, it doesn't address anything else. I'll unmark as WIP when I've made all changes and added tests.

tcromartie:
@scottransom I've unmarked as WIP and it's passing checks. Let me know if you have suggestions for more thorough tests or if anything else is un-PINT-like.

scottransom:
Quick comments on your test(s):  given that I think you had an issue where your code clashed with DMX (I think), can you put a test in to show that that doesn't happen now?

And your change is really and truly a special case for DMn, right?  Was it too hard to make sure that this doesn't happen for other parameters in the future (or, for example, to make sure that F002 -> F2)?

tcromartie:
So @scottransom, I made a different change to `alias_to_pint_param` that doesn't require a special case for DMn; it seems to handle DMn, FBn, DMX, etc. just fine (in that what's returned, i.e. `pint_par` and `first_init_par`, is correct). However, in the cases where the listed parameter was in the DM001 format, it isn't setting the value even when it's given one (i.e. the model says "DM1 UNSET" even when a value is read in). It's interesting because in the case where the par file looks like this:

DM001    0.01   1   0.0
DM2        0.01   1   0.0
DM003   0.01   1   0.0

The resulting model shows:
DM1        UNSET
DM2       0.1    (units etc.)
DM3       UNSET

I assume the issue is related to `parameter.py` but given my relative unfamiliarity with the code, I'm curious if you know immediately what I might have done. If the solution isn't obvious I'll just keep debugging.

scottransom:
@tcromartie Did you push your change that doesn't special case DMn up?  I'm still seeing a special case in `alias_to_pint_param()`

scottransom:
So I was just looking at this and the total solution to the problem might be an addition of only 2 characters in one of the regular expressions `prefix_pattern` in utils.py.

The pattern in question is:
```
    re.compile(r"^([a-zA-Z]+)(\d+)$"),  # For the prefix like F12
```
and it gets modified to be
```
    re.compile(r"^([a-zA-Z]+)0*(\d+)$"),  # For the prefix like F12
```
That `0*` ignores any zero characters before the final decimal `(\d+)`, which will match.

The tests seem to be running without errors, and having `DM1` or `DM001` or `DM000001` in a parfile all give the same correct response.  It also doesn't seem to mess up `F0`.

If that works, we should probably add a test explicitly showing that explicitly checks that each of the above really do work!

Note also that we might want to change the next regular expression similarly:
```
    re.compile(r"^([a-zA-Z0-9]+_)(\d+)$"),  # For the prefix like DMXR1_3
```
to
```
    re.compile(r"^([a-zA-Z0-9]+_)0*(\d+)$"),  # For the prefix like DMXR1_3
```
and make sure that that doesn't break anything.  Could even do the first regular expression as well....

scottransom:
Assuming this works, we should probably also add tests for `F10`, `F010`, `DM100`, `DM00100` or similar to make sure that they are all doing what we think they should...

scottransom:
Note that the tests for this don't actually need to include parfiles, I don't think.  And just the `split_prefixed_name` function from utils.py.  So in `test_utils.py`, after adding `split_prefixed_name` to the import list from utils.py at the top, you could simply have a test like this:
```
def test_split_prefixed_name():
    assert split_prefixed_name("DM1") == ("DM", "1", 1)
    assert split_prefixed_name("DM10") == ("DM", "10", 10)
    assert split_prefixed_name("DM100") == ("DM", "100", 100)
    assert split_prefixed_name("DM01") == ("DM", "1", 1)
    assert split_prefixed_name("DM010") == ("DM", "10", 10)
    assert split_prefixed_name("DM0100") == ("DM", "100", 100)
    assert split_prefixed_name("DM001") == ("DM", "1", 1)
    assert split_prefixed_name("DM0010") == ("DM", "10", 10)
    assert split_prefixed_name("DM00100") == ("DM", "100", 100)
    assert split_prefixed_name("F0") == ("F", "0", 0)
    assert split_prefixed_name("F1") == ("F", "1", 1)
    assert split_prefixed_name("F10") == ("F", "10", 10)
    assert split_prefixed_name("F010") == ("F", "10", 10)
```
With additional ones related to `DMX_`, for instance, if we change that regex too.

Also note that the docstring for `split_prefixed_name` will need a slight tweak because of this:
```
        >>> split_prefixed_name("DMX_0123")
        ('DMX_', '0123', 123)
```
becomes
```
        >>> split_prefixed_name("DMX_0123")
        ('DMX_', '123', 123)
```

tcromartie:
Hey Scott, sorry for my late response. No, I hadn't pushed the changes because it wasn't working yet; I thought I should avoid doing that. But I'll try out your change, write some tests, etc. I can try to compare some edge cases and see which fix is more flexible. Thanks a lot for looking into it.

tcromartie:
@scottransom This solves the problem very similarly to how mine does, but I'm still seeing the same issues with UNSET parameters using your fix, too. I've reverted all my changes and implemented yours in utils.py (just changing the one regular expression). It works to strip the zeros just as before, but with this input:

DM001 0.01 1 0.0
DM2 0.01 1 0.0
DM003 0.01 1 0.0

The resulting model still shows:
DM1 UNSET
DM2 0.1 (units etc.)
DM3 UNSET

I'm going to ask about it on the call in a sec. After that I'll push one of the two changes so it's actually clear what I've tried; I was just trying to avoid pushing changes if I know they're broken.

scottransom:
@tcromartie Sorry for the delays due to my travels!  Can you send me your script that you are using for testing?  Because I was *not* (at least I don't think) seeing what you report above that the new small fix was still causing UNSET params.  But I'd like to test it the same way that you are.

src/pint/models/timing_model.py
```diff
@@ -3037,13 +3037,21 @@ def alias_to_pint_param(self, alias):
                 first_init_par = None
                 # Handle the case of start index from 0 and 1
                 for start_idx in [0, 1]:
-                    first_init_par_alias = prefix + "{1:0{0}}".format(fmt, start_idx)
+                    # Ensure compatibility with tempo-style "DM001"
+                    if "DM0" in alias:
```
aarchiba, line 3041: I don't think  it's a good idea to special-case DM parameters - it will confuse users that DM accepts three digits but DMX (say) doesn't.

scottransom:
@tcromartie I think that git error is just that the master branch has changed, so your commits are stale.  I think you just need to checkout master, do a git pull, then check out your branch with the fixes and do "git merge master" (or whatever you have called the upstream master.  Then you can git push your branch and it should automatically update the PR.  I'm going to test the changes (including my recommended ones) today.

scottransom:
@tcromartie OK.  I can confirm what you say about things still being UNSET.  Looking at the code, I think that we should remove (and fix) the vast majority of `alias_to_pint_param()` as much of it deals with leading zeros, which my proposed changes completely ignore (and they aren't used anywhere else in PINT, I don't think).  That should simplify the code quite a bit.

tcromartie:
`alias_to_pint_param()` deals with checking if the param is supposed to be indexed at 0 or 1 or whatever too, only the beginning of the function strips the zeros. I think though that removing the leading zero stuff from `alias_to_pint_param()` and implementing your change to the regular exp in utils.py should work. Is that what you're suggesting?

scottransom:
Yeah, the code is just way more complicated than it needs to be (which is true for a lot of the model building and handling stuff).  So I think we should simplify if we can.  I think I can make `alias_to_pint_param()` about 1/4 the size.  Although I'm still not sure if that will fix the UNSET issue...

src/pint/models/timing_model.py
```diff
@@ -3035,21 +3035,16 @@ def alias_to_pint_param(self, alias):
...
                 for start_idx in [0, 1]:
-                    first_init_par_alias = prefix + "{1:0{0}}".format(fmt, start_idx)
+                    first_init_par_alias = prefix + f"{start_idx:0{0}}"
                     first_init_par = self._param_alias_map.get(
                         first_init_par_alias, None
                     )
                     if first_init_par:
                         # Find the first init par move to the next step
-                        pint_par = split_prefixed_name(first_init_par)[0] + idx_str
+                        pint_par = split_prefixed_name(first_init_par)[0] + str(idx)
```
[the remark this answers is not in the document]
tcromartie: Oh yeah that's leftover from before Scott's change in utils.py. I think it's redundant now.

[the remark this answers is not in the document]
scottransom: Yeah.  And I have (I think) better code for the rest of that method... (assuming that we go with no-leading-zeros for our internal parameter names).

tcromartie:
I think/hope the UNSET issue is fixed... goodness knows it took me long enough to track it down!

src/pint/models/model_builder.py
```diff
@@ -436,18 +436,19 @@ def _setup_model(
...
             Whether to run the setup function in the timing model.
         validate : bool, optional
             Whether to run the validate funciotn in the timing model.
         """
-        if original_name is not None:
-            use_alias = True
-        else:
-            use_alias = False
-
+        ### This is a little weird. When would original_name be None if there are entries like F0:F0?
```
aarchiba, line 446: `original_name` is an optional parameter, what if it is never set?

src/pint/models/timing_model.py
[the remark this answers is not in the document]
aarchiba: I still think prefixParameters should have a setting, on construction, for the minimum number of digits - if 1 (the default), just do this; if larger, format with leading zeros as needed.

tests/test_process_parfile.py
```diff
@@ -95,6 +95,18 @@ def test_parse_parfile_index_param():
     assert pint_dict["DMX_0025"] == ["2 1 3"]
 
 
+def test_parse_dm_derivs():
```
aarchiba, line 98: Perhaps it would be good to include a test that verifies the PINT name of the parameters? Like just reference m.DM1, m.DM2, m.DM3 (and check that m.DM001 does not exist).

aarchiba:
> I think/hope the `UNSET` issue is fixed... goodness knows it took me long enough to track it down! ETA: It's failing the check of `DMX_0016`. It seems this was happening before the recent `model_builder` commit, it's a result of the leading zeros change.

I find it's really helpful to put things like this in the test suite before I've solved them. Then I can tinker with my code and keep running `pytest tests/test_my_new_thing.py` until it passes. (You can also call a specific test function with `pytest tests/test_my_new_thing.py::test_specific_thing`.) Iterating with `pytest` lets me know if I've succeeded and if I've broken something I thought I'd fixed, and the `--ff` and `--pdb` let me poke around in the broken test.

tcromartie:
Note: This now fails the `test_retain_aliases` check in `test_parfile_writing.py` for PSR/PSRJ and A1DOT/XDOT

aarchiba:
> Note: This now fails the `test_retain_aliases` check in `test_parfile_writing.py` for PSR/PSRJ and A1DOT/XDOT

So the idea of the aliases was: if someone read in a par file with `XDOT` for the orbital amplitude derivative, we'd translate it to `A1DOT` for use inside PINT, but when they wrote out their par file they probably wanted it to be `XDOT` like it was when they read it in. So I made parameters remember what alias they were called on read, and use that on write. If you wanted to forget what it was called on read you could do `m.A1DOT.original_alias = None` and it would use its default.

The output format stuff did not exist then, and I wasn't around when that stuff was written, so I don't know how they interact. Maybe the `original_alias` is no longer working, or even no longer relevant. But that test is testing whether it works.

Of course your case is one where we might well want to write out a different name than we read in, so you're going to have to do something with `original_alias`. What that should be is not so clear - maybe `original_alias` just covers the name part and not the number? Do any prefixParameters even have aliases?

tests/test_process_parfile.py
tcromartie: I hope my change just now addresses this

src/pint/models/parameter.py
```diff
@@ -583,8 +583,10 @@ def add_alias(self, alias):
 
     def name_matches(self, name):
         """Whether or not the parameter name matches the provided name"""
-        return (name == self.name.upper()) or (
-            name in [x.upper() for x in self.aliases]
+        return (
+            (name == self.name.upper())
+            or (name in [x.upper() for x in self.aliases])
+            or (name.replace("0", "") == self.name.upper())
```
aarchiba, line 589: Isn't this going to reduce `DM10` to `DM1`?

tcromartie: This was a very silly thing for me to do in retrospect, I didn't even implement the bad workaround as I meant to. I think I have something better though. Just working on the tests.

aarchiba:
It would be good to have a test to confirm what happens if you read in `DM001` then write it out again - I'm not even sure what's supposed to happen, so a test would be a good way to make that definite. Should the user expect `DM001` back or `DM1`?

tcromartie:
Currently all three formats for `m.as_parfile` read out the DMn params exactly as they were read in. My inclination is that the expected behavior would be for `format='pint'` to read out DM1, DM10, etc. and for tempo(2) to read out whatever the input was because tempo(2) accepts both. Does that seem right? Or should we stick with not changing the format?

aarchiba:
> Currently all three formats for `m.as_parfile` read out the DMn params exactly as they were read in. My inclination is that the expected behavior would be for `format='pint'` to read out DM1, DM10, etc. and for tempo(2) to read out whatever the input was because tempo(2) accepts both. Does that seem right? Or should we stick with not changing the format?

You can make parameters forget what they were called on input by setting `original_name` (?) to `None`, which gets them called what PINT calls them internally. If they specify `tempo` or `tempo2` format we should write the parameter out in that program's preferred format. What `pint` format should produce is less clear to me. But I think users would mostly prefer for things to stay the way they were in the input parfile - maybe the default format should be `None` to produce this behaviour, and PINT format should be to use PINT's internal names? 

I don't think you should solve this here, but I do think it would be good to verify that the output parameter name is something sensible. Just checking that they are written out the way they were read in would clarify things for now.

tests/test_parfile_writing_format.py
```diff
@@ -225,6 +263,8 @@ def test_NHARMS():
         NE_SW                                 0.0
         SWM                                   0.0
         DM                  149.60232712555309087
+        DM1                                   0.0
```
tcromartie, line 266: oops

aarchiba:
Looks good!

PR #1208, opened 2022-04-21
Small changes to photonphase and polycos

paulray:
* Added progress bar for polyco generation.  Maybe this should be made optional?
* Made sure that TOAs used in polyco generation use the EPHEM specified in the model. This is more correct and gets rid of a lot of warnings.
* Converted a small python loop into an array operation for efficiency

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -28,6 +28,7 @@ jupytext
 pdbpp
 tox
 pre-commit
+typed-ast>=1.5.0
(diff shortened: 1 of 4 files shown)


paulray:
@sashalvna can you take a look at this.  Look OK?

sashalvna:
> @sashalvna can you take a look at this. Look OK?

Yes all of this looks good, and I will look into computing polycos only for GTIs.

PR #1210, opened 2022-04-22
Clean-up covariance matrix pretty-printing.

scottransom:
Removed multiply-reproduced code. 
Add color option for correlation matrix (use it in pintk).
Minor pintk improvements (updating fit info in output parfiles).

An example of the colorized correlation matrix is below.  (There is one more level, for |val| > 0.999, which is red background with white text).

image

### DIFF
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -555,8 +555,6 @@ class Fitter:
         self.model.FINISH.value = self.toas.last_MJD
         self.model.NTOA.value = len(self.toas)
         self.model.EPHEM.value = self.toas.ephem
-        if chi2:
-            self.model.CHI2.value = self.resids.chi2
         self.model.DMDATA.value = hasattr(self.resids, "dm")
         if not self.toas.clock_corr_info["include_bipm"]:
             self.model.CLOCK.value = "TT(TAI)"
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -555,6 +555,8 @@ class Fitter:
         self.model.FINISH.value = self.toas.last_MJD
         self.model.NTOA.value = len(self.toas)
         self.model.EPHEM.value = self.toas.ephem
+        if chi2:
+            self.model.CHI2.value = self.resids.chi2
         self.model.DMDATA.value = hasattr(self.resids, "dm")
         if not self.toas.clock_corr_info["include_bipm"]:
             self.model.CLOCK.value = "TT(TAI)"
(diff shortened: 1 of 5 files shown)


dlakaplan:
The results look pretty nice here, but is there a way to put the color definitions somewhat more understandable/editable?  Maybe just a dictionary at the top of the file, or we could even have a config file (probably not).  And some help in understanding/changing the values since the syntax isn't obvious?

scottransom:
Moving the definitions is easy.  But unless you want another dependency (which might be OK), then the color settings are *incredibly* opaque.  Especially since different terminals (and their default color selections) can change some of the colors.  For example, the code that I chose for correlation values from 0.5-0.9 was *supposed* to be yellow.  But my terminal treats it as that blue.  The orange and red colors *did* come out as planned.  So I'm not sure how to make it easily changeable for people *and* easy to understand what color you are actually going to get.  I suppose we could include a "color test" script that shows all the colors that your terminal produces and the codes.  There are multiple examples of that online.  Or maybe we could just point to one of those in the code and/or docs?  (Note:  I *really* like the improved correlation matrix formatting, though.

Here is a code snippet that prints all the options:
```
def print_format_table():
    """
    prints table of formatted text format options
    """
    for style in range(8):
        for fg in range(30,38):
            s1 = ''
            for bg in range(40,48):
                format = ';'.join([str(style), str(fg), str(bg)])
                s1 += '\x1b[%sm %s \x1b[0m' % (format, format)
            print(s1)
        print('\n')

print_format_table()
```

dlakaplan:
Maybe just move that up top and include the snippet you suggest in a docstring so that it will be available in rtd?

Probably environment variables are also too complicated for this, but I worry about people with unusual terminal setups and/or colorblindness.  I guess that not being able to see distinctions between the different colors wouldn't be as big a problem as bad contrast between the text & background.  And they can always turn off colors entirely.

dlakaplan:
Or just put that in the how-to (which we may need to organize a bit).

scottransom:
@dlakaplan OK.  So I added 2 functions to pint.utils.  A `colorize()` one that can change foreground, background, and text type (i.e. bold, normal, underline...), as well as a `print_color_examples()` utility.  I think those should get picked up by Sphinx.  And I'm using those in the correlation matrix pretty-printing.  For colors, I changed to using "normal", green, yellow, red, and inverse-red, as they apparently work in all terminals (terminals apparently determine which of the 16 base colors work in their terminals and set those appropriately.  The `colorize()` routine only uses those colors, even though many terminals support 256 colors.

dlakaplan:
I think if you change:
```
`colorize()` 
```
to
```
:func:`~pint.utils.colorize`
```
it will link better.  I also don't know if you should put in a reverse link from `colorize` to `print_color_examples`?

Finally, will this work with both standard strings and unicode?

scottransom:
I just checked and yes, it works perfectly with unicode strings as well.  I'll update the doscstrings for that (and with better cross links).

dlakaplan:
Great - that `colorize` function could be nice elsewhere!

dlakaplan:
Looks good to me.  I can merge when you are happy.

scottransom:
I'd say merge away.  I'm happy.

dlakaplan:
ok, will just wait for the checks to finish

PR #1213, opened 2022-04-22
Update GPS clock file

rossjjennings:
Merges in updates to the GPS clock file from the Tempo2 version, extending the cutoff date from 2019 July 30 (MJD 58694) to 2020 Oct 27 (MJD 59149).


dlakaplan:
Until we get a permanent solution this looks fine to me.

PR #1216, opened 2022-04-27
Pintk: change the fitter

dlakaplan:
Allows the user to:
* change the fitter from the command line
* change the fitter from a drop-down menu (with only NB/WB as appropriate)
* plus the log level drop-down initial state now follows the command-line selection


scottransom:
Looks good!  Thanks for this!

PR #1217, opened 2022-04-27
PINTK can automatically select right fitter to use

dlakaplan:
Can do:
`pintk ... --fitter=auto`
and it will select the correct fitter based on whether or not the data are WB and whether or not they have correlated errors.

Similarly,
`pintk ... --fitter=downhill`
will do the same but with a Downhill fitter.


dlakaplan:
I think I've tested this for WB/NB and w/ and w/o ECORR, and it looks OK.

dlakaplan:
Now there is a function as well:
`pint.fitter.Fitter.auto()` that can do this, along with appropriate unit tests.  Not sure this is the best API

paulray:
This looks great to me!

PR #1221, opened 2022-04-28
Update black and reblacken all code

paulray:
Updated required version of `black>22.0`
Reformatted all files with new version of black.
~~ToDo: Add commit hash to `.git-blame-ignore-revs`~~


paulray:
Note that (as described in the PINT docs) in order for blame to actually ignore the blackening commits, you need to do this in your local repo:
```git config blame.ignoreRevsFile .git-blame-ignore-revs```
This does not propagate from one clone to another.

scottransom:
Just checking:  do we need to do anything specific to make github handle the `.git-blame-ignore-revs`?  Because I mostly use `git blame` directly on github.

paulray:
> Just checking: do we need to do anything specific to make github handle the `.git-blame-ignore-revs`? Because I mostly use `git blame` directly on github.

Seems to be in work: https://github.com/github/feedback/discussions/5033

scottransom:
When the tests pass, should we just merge this so that people with open PRs can re-blacken and this PR won't mess with them (or vise-versa) any more?  Seems reasonable to me.

paulray:
Dunno. How long does @dlakaplan think we should give developers to raise a concern? Should we wait until tomorrow to give @aarchiba time to see it in Europe?

dlakaplan:
Probably waiting a day would be good.

scottransom:
That's fine waiting a day, I think. But I don't think we should merge anything else until this PR gets merged.  Maybe we can aim for tomorrow morning.  Anne will have hopefully seen it by then.

dlakaplan:
I guess there has been no feedback?
@paulray : could you be a bit more explicit on what another user has to do in order to update their forks/branches?  Is it just:
* "Fetch Upstream" on my fork to sync my master branch with nanograv/PINT. 
* Run `git config blame.ignoreRevsFile .git-blame-ignore-revs` (before or after pulling to my local repo?)
* Repeat with any other branches I have (for PRs in progress)
* Upgrade my local black according to the new `requirements_dev.txt`.  
* blacken my PR contributions with the new black

Are those the right steps?  Is that the right order?

paulray:
Yes, that looks about right.
* The Fetch Upstream on github will just sync your master branch, not any other branches
* `git config blame.ignoreRevsFile .git-blame-ignore-revs` can be done anytime (or not at all). It only affects the output of `git blame` so you only need to do this if you use blame locally
* Update requirements with `pip install -r requirements_dev.txt --upgrade` to get black and anything else.
* Checkout a branch that you will later want to merge, then `black src tests examples`
* Hopefully then `git diff master` will only have your changes and not a bunch of whitespace changes!

scottransom:
I was just going to comment almost the same exact things that @paulray just said.  So I think that should be good.

Issue #1218, opened 2022-04-28
Reloading prefix parameters un-freezes them


dlakaplan:
When you have a prefix parameter in a par file that is initially frozen/0, such as:
```
F2 0
```
reloading it will unfreeze it as it inherits those properties from F0.  I think this is done at:
https://github.com/nanograv/PINT/blob/master/src/pint/models/parameter.py#L1530 

We should change the behavior so that `frozen` at least is not inherited.

PR #1222, opened 2022-04-28
Fix prefixParameter frozen inheritance

dlakaplan:
changed prefixparameter creation, added test

Fixes https://github.com/nanograv/PINT/issues/1218


dlakaplan:
@scottransom : I tested this with `pintk` and recentering `PEPOCH` in the pareditor and it no longer fits for F2 etc.

dlakaplan:
What this doesn't do is deal with any more subtle changes like those @paulray was hinting at.

src/pint/models/parameter.py
```diff
@@ -1253,9 +1253,9 @@ class prefixParameter:
     F22. Appropriate units will be inferred,
 
     To create a prefix parameter with the same prefix but different index, just
-    use the ``.new_param`` method. It will return a new prefix parameter with the
+    use the :meth:`pint.parameter.prefixParameter.new_param` method. It will return a new prefix parameter with the
     same setup but the index. Some parameters' unit and description will
-    be changed once the index has been changed. In order to get the right units
+    be changed once the index has been changed. The new parameter will not inherit the ``frozen`` status of its parent by default.  In order to get the right units
     and description, ``.unitTplt`` and ``.descriptionTplt`` should be provided. If
```
scottransom, line 1259: Just noticed this, but I think the docs are outdated here.  Should be `unit_template` and `description_template`

dlakaplan:
OK, that's fair.  I can see about updating the docs tomorrow.

paulray:
I had a problem today with pintk where some parameters I had added like `F3 0` came into pint as unfrozen, which is definitely a bug.  I'm not sure if this PR fixes that. I'll try again when this is merged.

scottransom:
Thanks, David K!

Issue #1224, opened 2022-05-02
Warnings that need to get suppressed


dlakaplan:
Please list here any warning messages you see that you don't want to see (info about what creates them might help, but isn't crucial).  I can then add them to the filter.

dlakaplan:
`ERFA function "dtf2d" yielded 1 of "dubious year (Note 6)"`

paulray:
```
The datapath rcparam was deprecated in Matplotlib 3.2.1 and will be removed two minor releases later.
WARNING  (pint.logging                  ): None 
The verbose.level rcparam was deprecated in Matplotlib 3.1 and will be removed in 3.3.
WARNING  (pint.logging                  ): None 
The verbose.fileo rcparam was deprecated in Matplotlib 3.1 and will be removed in 3.3.
WARNING  (pint.logging                  ): None 
The text.latex.unicode rcparam was deprecated in Matplotlib 3.0 and will be removed in 3.2.
WARNING  (pint.logging                  ): None 
The savefig.frameon rcparam was deprecated in Matplotlib 3.1 and will be removed in 3.3.
WARNING  (pint.logging                  ): None Setting the datapath via matplotlibrc is deprecated 3.2 and will be removed in in 3.3
. 
WARNING  (pint.logging                  ): None 
The datapath rcparam was deprecated in Matplotlib 3.2.1 and will be removed two minor releases later.
WARNING  (pint.logging                  ): None 
The datapath rcparam was deprecated in Matplotlib 3.2.1 and will be removed two minor releases later.
WARNING  (pint.logging                  ): None 
The savefig.frameon rcparam was deprecated in Matplotlib 3.1 and will be removed in 3.3.
WARNING  (pint.logging                  ): None 
The text.latex.unicode rcparam was deprecated in Matplotlib 3.0 and will be removed in 3.2.
WARNING  (pint.logging                  ): None 
The verbose.fileo rcparam was deprecated in Matplotlib 3.1 and will be removed in 3.3.
WARNING  (pint.logging                  ): None 
The verbose.level rcparam was deprecated in Matplotlib 3.1 and will be removed in 3.3.
WARNING  (pint.logging                  ): None 
The datapath rcparam was deprecated in Matplotlib 3.2.1 and will be removed two minor releases later.
WARNING  (pint.logging                  ): None 
The verbose.level rcparam was deprecated in Matplotlib 3.1 and will be removed in 3.3.
WARNING  (pint.logging                  ): None 
[...]

PR #1225, opened 2022-05-03
Tweak logging

dlakaplan:
Add more types of warnings to ignore, pay attention to the standard `warnings` filter.

Fixes some of https://github.com/nanograv/PINT/issues/1224


dlakaplan:
This isn't crucial, but I think it could be merged with no harm done.  I will look at tracking down some of the other Erfa warnings later (and adding the ability to raise exceptions from warnings).

PR #1230, opened 2022-05-04
Fix DMDATA output for non-pint format

### LINKED ISSUE #1228: Conversion to tempo(2) ignores DMDATA
PINT writes DMDATA to Y/N, while tempo(2) expect this field to have an integer value. This should be captured in the model conversion functionality since it causes errors when trying to fold with, e.g. `fold_psrfits`.

dlakaplan:
Fixed #1228

### DIFF
--- a/src/pint/models/parameter.py
+++ b/src/pint/models/parameter.py
@@ -507,6 +507,8 @@ class Parameter:
             log.warning(
                 f"Changing KOM from DT92 convention to IAU: this will not be readable by PINT"
             )
+        elif self.name == "DMDATA" and not format.lower() == "pint":
+            line = "%-15s %d" % (self.name, int(self.value))
 
         if self.uncertainty is not None:
             line += " %d %s" % (
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -366,6 +366,12 @@ class TimingModel:
             self.T2CMETHOD.value = "IAU2000B"
         if self.UNITS.value not in [None, "TDB"]:
             raise ValueError("PINT only supports 'UNITS TDB'")
+        if not self.START.frozen:
+            warn("START cannot be unfrozen...setting START.frozen to True")
+            self.START.frozen = True
+        if not self.FINISH.frozen:
+            warn("FINISH cannot be unfrozen...setting FINISH.frozen to True")
+            self.FINISH.frozen = True
 
         for cp in self.components.values():
             cp.validate()
--- a/tests/test_parfile_writing_format.py
+++ b/tests/test_parfile_writing_format.py
@@ -95,6 +95,14 @@ def test_ECL():
             assert l.split()[-1] == "IERS2003"
 
 
+def test_DMDATA():
+    """Should be an integer for TEMPO/TEMPO2"""
+    m = get_model(os.path.join(datadir, "J0030+0451_post.par"))
+    for l in m.as_parfile(format="tempo").split("\n"):
+        if "DMDATA" in l:
+            assert isinstance(int(l.split()[-1]), int)
+
+
 def test_formats():
     m = get_model(os.path.join(datadir, "B1855+09_NANOGrav_9yv1.gls.par"))
     with pytest.raises(AssertionError):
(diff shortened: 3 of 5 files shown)


tests/test_parfile_writing_format.py
```diff
@@ -95,6 +95,14 @@ def test_ECL():
             assert l.split()[-1] == "IERS2003"
 
 
+def test_DMDATA():
+    """Should be an integer for TEMPO/TEMPO2"""
+    m = get_model(os.path.join(datadir, "J0030+0451_post.par"))
+    for l in m.as_parfile(format="tempo").split("\n"):
+        if "DMDATA" in l:
+            assert isinstance(int(l.split()[-1]), int)
```
aarchiba, line 103: I don't think there's ever a situation where the function `int` returns a non-integer; should this test just be to call `int` and see if it raises an exception? How does this handle `0.0` and the like - floating-point numbers that are integers? I assume TEMPO/TEMPO2 don't like those?

src/pint/models/parameter.py
```diff
@@ -507,6 +507,8 @@ def as_parfile_line(self, format="pint"):
             log.warning(
                 f"Changing KOM from DT92 convention to IAU: this will not be readable by PINT"
             )
+        elif self.name == "DMDATA" and not format.lower() == "pint":
```
aarchiba, line 510: Don't we have an `intParameter` class? If so the right way to make this change would be to make `DMDATA` an `intParameter`. Or perhaps a `boolParameter`, since it is, and those accept `0`/`1` as input.

dlakaplan: I'm not sure.  `DMDATA` already is a `boolParameter`. The question isn't whether or not it can be read in correctly but how to change the output depending on desired format.  If all `boolParameters` want to be `T/F` for `PINT` and `1/0` for `TEMPO(2)` then we can do a more general change (and maybe that's the case - I haven't dug into it).  But this just fixes the specific output case that was causing a problem downstream.

dlakaplan: So parsing the input (0.0 vs. 0 vs. F) should still be the same.

dlakaplan:
The last pushes were to address a separate issue (#1237) but I put them in here.  I can separate the PRs if needed.

I'll add a test for the START/FINISH issue soon

src/pint/models/parameter.py
aarchiba: Is there any reason not to have DMDATA, at least, always output 0/1? PINT reads those happily. Perhaps when a boolParameter is created one could specify whether it is written as Y/N or 0/1? So where DMDATA is created we'd specify that it was just always written as 0/1?

src/pint/models/timing_model.py
```diff
@@ -366,6 +366,12 @@ def validate(self):
             self.T2CMETHOD.value = "IAU2000B"
         if self.UNITS.value not in [None, "TDB"]:
             raise ValueError("PINT only supports 'UNITS TDB'")
+        if not self.START.frozen:
```
aarchiba, line 369: I think this sort of thing would be better put in the mjdParameter when it was created? This would avoid special-casing anything here.

aarchiba:
We already have one place where the code has to specify all sorts of stuff about an individual parameter: when the parameter object is created and attached to the timing model. Since the parameter object handles all changes, the parameter object is also a good place to catch any illegal changes immediately when they happen. So I think as a general strategy it would be good to allow parameter objects to implement their own special handling based on the parameters they were passed when they were constructed. For example:

```python
        self.add_param_from_top(
            MJDParameter(name="START", description="Start MJD for fitting"), "", fittable=False,
        )
```
and
```python
        self.add_param_from_top(
            boolParameter(
                name="DMDATA",
                value=False,
                description="Was the fit done using per-TOA DM information?",
                fittable=False,
            ),
            "",
        )
```

dlakaplan:
These solutions may not be ideal.  Your suggestion maybe better.  But these will work for the specific bugs that are needed and they work now.  Personally I would suggest that move ahead with one or both of these fixes for the moment but leave the issues open, so that if somebody has time later they can do a better job.

aarchiba:
I'm trying to limit special-casing of parameters for the general health of PINT. Perhaps raise an issue pointing to the specific places these special cases occur so we can at least keep track of them?

PR #1231, opened 2022-05-05
Adding tempo(2) par/tim output to pintk

dlakaplan:
Also some other menu tweaks and an icon


src/pint/pintk/plk.py
```diff
@@ -1276,7 +1271,7 @@ def clickAndDrag(self, event):
         """
         Call this function when the mouse is clicked and dragged
         """
-        log.debug("You clicked and dragged in mode '%s'" % self.plkToolbar.mode)
+        log.debug(f"You clicked and dragged in mode '{self.pltToolbar.mode}'")
```
scottransom, line 1274: New bug!  Note that plkToolBar became pltToolbar

Issue #1226, opened 2022-05-04
WLSFitters try to take bogus step while GLSFitters claim convergence


dlakaplan:
If I start a `DownhillWLSFitter` at (or presumably near) a good solution, it will fail to converge and throw an exception.  The other fitters are fine.  I'm guessing this is because chi^2 is not decreasing but it should behave better.

```
import pint.fitter
import pint.models

m, t = pint.models.get_model_and_toas("DownhillWLSTest.par.txt", "DownhillWLSTest.tim.txt")
f = pint.fitter.DownhillWLSFitter(t, m)
f.fit_toas()
```
results in:
```
MaxiterReached                            Traceback (most recent call last)
~/askap/lmc/test_fitters.py in <module>
     28 m, t = pint.models.get_model_and_toas("DownhillWLSTest.par.txt", "DownhillWLSTest.tim.txt")
     29 f = pint.fitter.DownhillWLSFitter(t, m)
---> 30 f.fit_toas()

~/pythonpackages/PINT_dlakaplan/src/pint/fitter.py in fit_toas(self, maxiter, threshold, debug, **kwargs)
   1462         """
   1463         self.threshold = threshold
-> 1464         super().fit_toas(maxiter=maxiter, debug=debug, **kwargs)
   1465 
   1466     def create_state(self):

~/pythonpackages/PINT_dlakaplan/src/pint/fitter.py in fit_toas(self, m
axiter, required_chi2_decrease, max_chi2_increase, min_lambda, debug)
   1317             ) from exception
   1318         if not self.converged:
-> 1319             raise MaxiterReached(f"Convergence not detected after {maxiter} steps.")
   1320         return self.converged
   1321 

MaxiterReached: Convergence not detected after 10 steps.
```
when all other fitters converge very quickly (giving essentially the same chi^2 pre- and post-fit).

par/tim files attached:
[DownhillWLSTest.par.txt](https://github.com/nanograv/PINT/files/8621272/DownhillWLSTest.par.txt)
[DownhillWLSTest.tim.txt](https://github.com/nanograv/PINT/files/8621273/DownhillWLSTest.tim.txt)
(renamed to `.txt` so GitHub would accept them).

aarchiba:
Can this par and tim file go into the test suite? I'd like to see what a single DownhillWLS step looks like compared to the WLS, and I think there's already a test for this.

dlakaplan:
Sure.

aarchiba:
[...]

PR #1232, opened 2022-05-06
Remove ERFA "pmsafe" warnings

dlakaplan:
When you run for some pulsars you might get one (or many) warnings like:
```
WARNING  (pint.logging                  ): <class 'erfa.core.ErfaWarning'> ERFA function "pmsafe" yielded 1 of "distance overridden (Note 6)"
/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/logging.py:118: ErfaWarning: ERFA function "pmsafe" yielded 1 of "distance overridden (Note 6)"
  warn_(message, *args, **kwargs)
```
These should not be happening: we add a dummy distance in before applying proper motions.  But it seems that if the proper motion is large enough that still fails somehow on the `astropy` side.  For example

```
import astropy.coordinates as coords
import numpy as np

from astropy.time import Time
from pint.utils import add_dummy_distance, remove_dummy_distance

# normally using the dummy distances should lead to no ERFA pmsafe warning
sc = coords.SkyCoord(
    ra=180 * u.deg,
    dec=20 * u.deg,
    pm_ra_cosdec=100 * (u.mas / u.yr),
    pm_dec=100 * (u.mas / u.yr),
    obstime=Time(50000, format="pulsar_mjd", scale="tdb"),
)

sc2 = add_dummy_distance(sc)
# this should not lead to a warning
sc2.apply_space_motion(new_obstime=sc.obstime)

# but with very big proper motions, even if applied over 0 time
# we get the warning
sc = coords.SkyCoord(
    ra=180 * u.deg,
    dec=20 * u.deg,
    pm_ra_cosdec=1000 * (u.mas / u.yr),
    pm_dec=1000 * (u.mas / u.yr),
    obstime=Time(50000, format="pulsar_mjd", scale="tdb"),
)
sc2 = add_dummy_distance(sc)
# this shoudn't lead to a warning but it does
sc2.apply_space_motion(new_obstime=sc.obstime)
```
the second attempt emits a warning, and the only difference is that the proper motion is larger.

So we explicitly catch Erfa warnings in the `astrometry` module to suppress those.  Warnings issued elsewhere will still come through.

### DIFF
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -358,9 +358,12 @@ class AstrometryEquatorial(Astrometry):
             position_now = add_dummy_distance(self.get_psr_coords())
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore", ErfaWarning)
-                position_then = remove_dummy_distance(
-                    position_now.apply_space_motion(new_obstime=newepoch)
-                )
+                # for the most part the dummy distance should remove any potential erfa warnings
+                # but for some very large proper motions that does not quite work
+                # so we catch the warnings
+                position_then = position_now.apply_space_motion(new_obstime=newepoch)
+                position_then = remove_dummy_distance(position_then)
+
             return position_then
 
     def coords_as_ICRS(self, epoch=None):
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -1,6 +1,7 @@
 """Astrometric models for describing pulsar sky positions."""
 import copy
 import sys
+import warnings
 
 import astropy.constants as const
 import astropy.coordinates as coords
@@ -9,6 +10,11 @@ import numpy as np
 from astropy.time import Time
 
 from loguru import logger as log
+
+try:
+    from erfa import ErfaWarning
+except ImportError:
+    from astropy._erfa import ErfaWarning
 
 from pint import ls
 from pint.models.parameter import (
@@ -350,9 +356,11 @@ class AstrometryEquatorial(Astrometry):
             else:
                 newepoch = Time(epoch, scale="tdb", format="mjd")
             position_now = add_dummy_distance(self.get_psr_coords())
-            position_then = remove_dummy_distance(
-                position_now.apply_space_motion(new_obstime=newepoch)
-            )
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", ErfaWarning)
+                position_then = remove_dummy_distance(
+                    position_now.apply_space_motion(new_obstime=newepoch)
+                )
             return position_then
 
     def coords_as_ICRS(self, epoch=None):
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1201,9 +1201,9 @@ def add_dummy_distance(c, distance=1 * u.kpc):
             "No proper motions available for %r: returning coordinates unchanged" % c
         )
         return c
-    if c.obstime is None:
-        log.warning("No obstime available for %r: returning coordinates unchanged" % c)
-        return c
+    # if c.obstime is None:
+    #    log.warning("No obstime available for %r: returning coordinates unchanged" % c)
+    #    return c
 
     if isinstance(c.frame, coords.builtin_frames.icrs.ICRS):
         if hasattr(c, "pm_ra_cosdec"):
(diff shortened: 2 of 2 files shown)


aarchiba:
That's weird, any idea why ERFA is emitting the message? Maybe it's a spherical thing? But I think getting rid of the warning is the only reasonable option.

src/pint/models/astrometry.py
```diff
@@ -10,6 +11,11 @@
 
 from loguru import logger as log
 
+try:
+    from erfa import ErfaWarning
+except ImportError:
+    from astropy._erfa import ErfaWarning
```
aarchiba, line 17: Is this still relevant for our oldest acceptable version of astropy?

src/pint/utils.py
```diff
@@ -1201,9 +1201,6 @@ def add_dummy_distance(c, distance=1 * u.kpc):
             "No proper motions available for %r: returning coordinates unchanged" % c
         )
         return c
-    if c.obstime is None:
```
aarchiba, line 1204: Why are these going away?

src/pint/models/astrometry.py
dlakaplan: I'm not sure.  I copied it from elsewhere just to be safe.  I guess I could try deleting it and see if the relevant tests pass.

dlakaplan:
I don't know why ERFA is emitting this.  I found some discuss online in various astropy forums.  The dummy distance should handle it, except that for very large proper motions it is not (even for dt=0, I think).  I suspect it's some issue in the spherical<->cartesian conversion.  Most pulsars may not encounter this situation in reality but it could happen if a fit goes slightly off the rails (what happened to me).

src/pint/models/astrometry.py
aarchiba: I'm not sure we have relevant tests for this. In any case if it's still littered throughout the code the right answer is, if it's unnecessary, a separate PR to get rid of it.

aarchiba:
According to https://pyerfa.readthedocs.io/en/latest/api/erfa.pmsafe.html there are three possible situations that set the warning flag:

1. A small (?) or zero or negative parallax (reciprocal distance presumably); treated as close enough to keep velocity to 0.1c: warning status 1
2. Space velocity a large fraction of c; space velocity set to zero (!); warning status 2
3. Relativistic correction process fails to converge; warning status 4

It looks like we're seeing condition number 2 - but that proper motion may be being zeroed out in this case. Perhaps our fake distance needs to be really small to avoid this weirdness? Probably necessary to suppress the warning also, but it seems that we may be being warned of an actually problematic situation.

aarchiba:
Actually, the message says it's "note 6", that is, the claimed parallax is "small", zero, or negative. Are we feeding distances to astropy based on PX in our par files? If we're always using fake distances, maybe they need to be smaller?

dlakaplan:
The distance is 1 kpc by default.

aarchiba:
According the the `pmsafe` docs, it tries to deal with relativistic velocities somehow (handling superluminal motion?) so it's the *velocity*, not the amount of motion, that matters - so `dt=0` isn't better.

But the specific error I see (in the example files for bug #1226) specifies that it's "Note #6", that is, `pmsafe` is taking a "small or zero or negative" parallax and chopping it to a distance that keeps the velocity about 0.1c. So this has nothing direct to do with the velocity.

Are you seeing errors out of `pmsafe` with different messages (alluding to different Notes)? Those would be different anomalous conditions and require different fixes.

dlakaplan:
This isn't fully related to the initial issue but now all of the `loguru` config is only performed when you do:
```
import pint.logging
pint.logging.setup()
```
you can specify level and format at that time if desired.

dlakaplan:
This PR has mostly turned into logging fixes, with the other things pretty minor.  So I'm happy to discuss the logging fixes/improvements here and only merge if/once people are happy.  Separating the PRs is a bit of a pain (which I understand I caused for myself...)

aarchiba:
I think the "don't use colours unless asked" change should be discussed on the call before we go further. Having some examples handy where the current approach breaks things would be helpful.

aarchiba:
Trying to sort out the network problems in #1254

dlakaplan:
I agree that we need to discuss a lot of this.

dlakaplan:
I updated all of the example notebooks with basic logging usage

dlakaplan:
Here's what you get in an interactive session if you don't run `logging.setup()`: still plausible
Screen Shot 2022-06-02 at 12 22 34 PM

and if you do:
Screen Shot 2022-06-02 at 12 23 52 PM

and then reconfiguring:
Screen Shot 2022-06-02 at 12 26 45 PM

paulray:
Looks great!

PR #1234, opened 2022-05-09
Convert a cluster into a flag

dlakaplan:
To aid in selecting data by "clusters", convert clusters to flags that can easily be used for JUMP/EFAC/etc.

e.g., 
```
t.get_clusters(add_flag = True)
```
then the TOA table looks like:
```
t.table['flags']
                                          {'format': 'Tempo2', 'name': 'ALL.T16Fp', 'clkcorr': '2.800489029156601e-05', 'cluster': '4'}
                                          {'format': 'Tempo2', 'name': 'ALL.T16Fp', 'clkcorr': '2.800489016085896e-05', 'cluster': '4'}
                                         {'format': 'Tempo2', 'name': 'ALL.T16Fp', 'clkcorr': '2.8004890030153187e-05', 'cluster': '4'}
...
                                         {'format': 'Tempo2', 'name': 'ALL.T16Fp', 'clkcorr': '2.8004888308595357e-05', 'cluster': '4'}
                                         {'format': 'Tempo2', 'name': 'ALL.T16Fp', 'clkcorr': '2.8004888185360825e-05', 'cluster': '4'}
                           {'format': 'Tempo2', 'name': 'J0523-7125_PSMC.rf.f4t4', 'clkcorr': '2.6823950390193516e-05', 'cluster': '0'}
                                                                                                                                    ...
 {'format': 'Tempo2', 'name': 'uwl_220130_090227_2.calib.0000_0000.low.paz1.pFT8', 'clkcorr': '2.760637484357018e-05', 'cluster': '13'}
```

and then:
```
m.add_component(DelayJump(), validate=False)
    "JUMP",
    key="-cluster",
    value=0.0,
    key_value="1",
    units=u.s,
    frozen=False,
)
m.components["DelayJump"].add_param(par, setup=True)
```

### DIFF
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -1658,7 +1658,7 @@ class TOAs:
         add_column : bool, optional
             Whether or not to add a ``clusters`` column to the TOA table (default: False)
         add_flag : str, optional
-            If not ``None``, will add a flag with that name to the TOA table (default: None)
+            If not ``None``, will add a flag with that name to the TOA table whose value is the cluster number (as a string, starting at 0) (default: None)
 
         Returns
         -------
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -1637,7 +1637,7 @@ class TOAs:
             raise AttributeError("No DM error is provided.")
         return np.array(result)[valid] * pint.dmu
 
-    def get_clusters(self, gap_limit=2 * u.h, add_column=False, add_flag=False):
+    def get_clusters(self, gap_limit=2 * u.h, add_column=False, add_flag=None):
         """Identify toas within gap limit (default 2h = 0.0833d)
         of each other as the same cluster.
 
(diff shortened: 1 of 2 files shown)


dlakaplan:
Does some of the same things as #1182 but in a way that is preserved in `tim` file read/write (as @aarchiba suggested).

aarchiba:
How about allowing the user to specify what flag? `t.get_clusters(add_flag = "session")`? Avoids the user having to guess what flag will be used and gives them a choice, whether to support different granularities of clustering on the same data set or to avoid overwriting an existing flag on the input data?

aarchiba:
Fixes #1181

dlakaplan:
Good idea.   Implemented.

src/pint/toa.py
```diff
@@ -1675,6 +1677,13 @@ def get_clusters(self, gap_limit=2 * u.h, add_column=False):
             if add_column:
                 self.table.add_column(clusters, name="clusters")
                 self.table.meta["cluster_gap"] = gap_limit
+                log.debug(f"Added 'clusters' column to TOA table with gap={gap_limit}")
+            if add_flag is not None:
```
aarchiba, line 1681: What happens if they pass in True or 1? That seems like a likely error. Probably True will produce `-True 7` and `1` should produce an exception because `-1` is not a valid TOA flag but it's worth checking.

aarchiba: It should be the case that if you try to set an invalid flag you get an exception that tells you you did that, which would be a totally reasonable outcome.

aarchiba: Er, what I meant was, please try your code (maybe in the test suite!) with some bogus values and check that meaningful exceptions are emitted.

dlakaplan: OK, I see.  `check_allowed_key` did those checks, and the exceptions are properly raised.

PR #1236, opened 2022-05-10
Report the observatory when TEMPO/TEMPO2 needed.

aarchiba:
Currently if you try to load TOAs from an observatory PINT knows but that needs the TEMPO or TEMPO2 runtime files, and the corresponding environment variable is not set, you get an error message that just tells you the environment variable is missing. This code reports the name of the observatory that is calling for TEMPO/TEMPO2 runtime files.


aarchiba:
If the variable is set but the file isn't found, the missing filename is shown, which probably gives the user enough information to tell what's going on.

paulray:
Looks helpful to me!

PR #1239, opened 2022-05-11
Improve clock file handling

aarchiba:
This is meant to be a modest change, not yet implementing any global clock-file updating scheme as suggested in #1180 but addressing as far as possible the issue of finding any local clock files you do have. It also does something to address #1235 by pulling in all observatories that are in the TEMPO/TEMPO2 official repositories.

Notable changes:

- PINT now includes all clock correction files from TEMPO/TEMPO2 (including Jodrell corrections more recent than 1997)
- You can now specify what to do about missing clock corrections when loading TOAs: `get_TOAs("file.tim", limits="error")`
- You can now use telescopes without clock corrections; you just get warnings, the same as if you had data past the end of your clock corrections.
- You can now ask a clock correction file or an observatory when the last MJD it covers is
- `check_for_new_clock_files_in_tempo12_repos` is a first attempt at comparing PINT's clock corrections to the versions on the Web
- Demo notebook shows the status of PINT's clock corrections (visible in PR rendered docs: https://nanograv-pint--1239.org.readthedocs.build/en/1239/examples/check_clock_corrections.html)

Goals:

- [x] PINT reports something useful when clock corrections are missing
- [x] PINT includes all clock corrections from TEMPO/TEMPO2
~~- [ ] PINT pulls newer/extra clock corrections from `$TEMPO/` or `$TEMPO2/` if available~~
- [x] PINT handles missing clock corrections the same way as TOAs past the end of available clock corrections
- [x] Users have a way to request exception/warning behaviour for missing clog corrections

Closes #1241 #1235

PR #1244, opened 2022-05-12
Use EQUADs and EFACs in WLS fitters

### LINKED ISSUE #1226: WLSFitters try to take bogus step while GLSFitters claim convergence

aarchiba:
Ensure fitters correctly take into account model-based adjustments to TOA uncertainties.

Minor change, strict bug fix, resolves frustrating misbehaviour, has tests.

~~Still needs tests.~~ ~~Code currently depends on #1239 as my test case needed telescopes I don't have timing data for.~~

Closes #1226 #1245

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 ### Added
 - Warning when A1DOT parameter used with DDK model
 ### Fixed
+- WLS fitters no longer ignore EFAC/EQUAD (bug #1226)
 ### Changed
 
 ## [0.8.6 == 0.8.7] 2022-05-10
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -55,7 +55,7 @@ Fitters in use::
 To automatically select a fitter based on the properties of the data and model::
 
     >>> fitter = Fitter.auto(toas, model)
-    
+
 """
 import collections
 import copy
@@ -1336,7 +1336,9 @@ class WLSState(ModelState):
             toas=self.fitter.toas, incfrozen=False, incoffset=True
         )
         # Get residuals and TOA uncertainties in seconds
-        Nvec = self.fitter.toas.get_errors().to(u.s).value
+        Nvec = (
+            self.model.scaled_toa_uncertainty(self.fitter.toas).to(u.s).value
+        )
         scaled_resids = self.resids.time_resids.to(u.s).value / Nvec
 
         # "Whiten" design matrix and residuals by dividing by uncertainties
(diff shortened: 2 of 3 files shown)


aarchiba:
This bug will have affected all NANOGrav pulsars that use WLSFitter (those with no red noise), slowing or preventing convergence (and reporting this, to the frustration of users).

aarchiba:
I should note that without this fix the WLSFitter would run fine and would appear to settle on a set of parameters, but they would be the no-EFAC best-fit. DownhillWLSFitter might do this too, but it would sometimes (for example if starting from a non-PINT or GLS best fit) notice that a step was making the (including-EFAC) chi-squared worse and refuse to take the step, leading to convergence failures. This is what ultimately led to the discovery and fix of the bug, but it was annoying to users.

PR #1247, opened 2022-05-18
Global clock corrections

aarchiba:
This is a PR designed to run PINT with no observatory clock correction files at all. All are obtained by downloading them (via the astropy caching mechanism) from https://github.com/ipta/pulsar-clock-corrections .

Notable features:
- All clock files are downloaded to the astropy cache, and updated only if you use MJDs they don't cover, or if the repository publishes an invalidation notice (say an error was noticed in old corrections).
- You can export the clock files you are actually currently using with `pint.observatory.export_clock_files(dir)`.
- You can set the environment variable PINT_CLOCK_OVERRIDE to make PINT use clock files from a specific directory in preference to ones in the repository.
- The default BIPM TAI to TT corrections are the BIPM2021 ones.
- You can preload all clock files into the cache with `pint.observatory.update_clock_files()`.


aarchiba:
Currently it depends on #1239 ; depending on what happens to that one it will need updating.

aarchiba:
Current state:
- All TopoObs try to use the global repository; if their clock files aren't listed then they fall back to whatever's included in PINT. (Currently no observatory needs to do this, and anyway there's nothing in PINT.)
- PINT uses global clock corrections only for observatory clock corrections, not for BIPM or anything.
- PINT initially requests a clock correction file only if it has nothing for that observatory; otherwise it is content with an old one.
- PINT requests a newer clock correction file if a clock correction is needed for a time after the end of the clock corrections it has.
- PINT doesn't request a clock file from the Net unless the old one is out of date by an amount specified in the repository (currently mostly 7 days but Arecibo files never expire).
- Testing is currently limited but there is machinery in place to make testing possible (using a local "repository").

aarchiba:
Needs a URL change to reflect the repository move.

aarchiba:
A few needed things:

- How can users inspect the clock files they are using? With this they live in the astropy cache.
- How can users preload the cache with everything they need? This is an explicit request for Docker containers, or for use in HPC environments.
- How can users supply their own local clock corrections, overriding the global lookup?

scottransom:
All of those last few questions could be handled via two things, I think.
1. A utility function that exports the clock file from the cache.  (I suspect that would be a trivial addition)
2. A config file for clock corrections in a simple format (i.e. yaml?) that is optional, but if it exists, specifies exactly where to get clock corrections for each observatory.  We could make a default version that is kept in PINT source which is used instead of hardcoded `clock*` info in `observatories.py` and a user could copy and edit that an place it where they wanted (possibly with an environment variable letting PINT know it exists and where).

Also, just a general comment:  we don't need for everything to be complete before we accept/merge this.  All we need is something that is better than what we have now.  We can improve it incrementally.

aarchiba:
To do soon:

- [ ] PINT_CLOCK_OVERRIDE environment variable
- [ ] Export clock file, individually or en masse
- [ ] Preload cache
- [ ] Support date-based invalidation based on `index.txt` `invalid_if_older_than`
- [ ] Deal with `tests/datafile/time_wsrrt.dat`
- [ ] More complete testing

aarchiba:
Currently `gps2utc.clk` is depressingly out of date and limits our clock correction span. Clearly an issue for the repository.

aarchiba:
There are actually two different versions of that correction, depending on which version of GPS time the observatory is referenced to. @rossjjennings and I are having a discussion on Slack, and I think we have a solution of sorts, but it's very technical and unclear what the really right answer is. 

For this PR, it just popped up in a test I tried to write, but I think it makes sense to pull in a global `gps2utc.clk` and `tai2tt_bipmYY.clk` as well as observatory clock files. (We don't have `tai2tt_bipm21.clk` in PINT or TEMPO2.) Since they're just clock files, the machinery already implemented in this PR should work for those.

aarchiba:
> There are actually two different versions of that correction, depending on which version of GPS time the observatory is referenced to. @rossjjennings and I are having a discussion on Slack, and I think we have a solution of sorts, but it's very technical and unclear what the really right answer is.
> 
> For this PR, it just popped up in a test I tried to write, but I think it makes sense to pull in a global `gps2utc.clk` and `tai2tt_bipmYY.clk` as well as observatory clock files. (We don't have `tai2tt_bipm21.clk` in PINT or TEMPO2.) Since they're just clock files, the machinery already implemented in this PR should work for those.

Solved, I think: https://nanograv.github.io/pulsar-clock-corrections/T2runtime/clock/gps2utc.clk.html

aarchiba:
PINT's clock correction status tutorial now seems rather cheerier: https://nanograv-pint--1247.org.readthedocs.build/en/1247/examples/check_clock_corrections.html

`ncyobs2obspm.clk` and `obspm2gps.clk` are now in the repo so it should improve when it reruns.

aarchiba:
I think that we should support the following approach to finding any clock file:

- If `$PINT_CLOCK_OVERRIDE` is set, look for the file in there; if it is not present then
- Look for it in the global repository; if it is not present then
- Look in PINT's `data/runtime` directory.

Failing all that, no clock file is found and all TOAs are treated as needing corrections that aren't available (warning or exception based on `limits`).

aarchiba:
Notes from today's PINT meeting:

- It would be valuable to have some mechanism to do comparisons with specific versions of clock files. The combination of exporting and the environment variable, or using the git history of the repository and the environment variable, provides enough of a solution for now, but we may want to open an issue for wider discussion of how to do this.
- The above checklist is enough, when complete, for this to be merged - try not to expand the scope further.
- Ensure that if a clock file is pulled from the environment variable there is a clear and appropriate log message - we don't want surprise stale clock files. The severity of the message could be based on context: if there's no version of the clock file in the repository, not very severe at all. If there's a more recent version in the repository, maybe raise the issue with some more emphasis. 
- Creating custom TopoObs is something that people actually do, so supporting them well (and compatibly) is important. How can they specify their clock files? Putting them in the environment variable location should be an option, but so should constructing the object with an explicit path.

aarchiba:
I would be happier if we had testing of the "should we download a new version?" logic. This is doable in a sandbox by passing a repository URL that just points to a local directory and then manipulating the files there, the index, and the dates of files in the cache. Astropy even has a `tmp_cache` fixture that allows the test suite to avoid messing with, or depending on, what's in your astropy cache. 

Otherwise, though, all the code works.

aarchiba:
I think this is done; it works, it's reasonably well tested, and it lets people have more control over their clock files than they did before.

aarchiba:
Documentation: https://nanograv-pint--1247.org.readthedocs.build/en/1247/explanation.html#clock-corrections

aarchiba:
What happened to our CI?

dlakaplan:
Right now the docs say (correctly) that new observatories should be added to `observatories.py`, which is located in the source tree.  Maybe this is a topic for another PR but would it make more sense to put those in a runtime-accesible file (yaml or some other format) so that new installations are not needed?  This is similar to a comment @scottransom made a few weeks ago.

dlakaplan:
Yes, I think it's fine to move the clock file data to a separate issue and decouple it from this PR.

src/pint/observatory/__init__.py
```diff
@@ -1,11 +1,29 @@
...
+Observatories are added to the registry when the objects are created.
+For the observatories defined in PINT, these objects are created
+when the relevant module is imported.
+
+If you want to ensure that all PINT's observatories are available
+to your interpreter, you can run::
+
+    >>> import pint.observatory.observatories
+    >>> import pint.observatory.topo_obs
+    >>> import pint.observatory.special_locations
```
dlakaplan, line 13: I feel like this is a little clumsy.  It would be nice to offer a convenience function to list all loaded observatories after only a single import.  Like: `import pint.observatory;print(pint.observatory.names())`.  After the extra imports you can do `pint.observatory.Observatory.names()`: is there a better way?  A way that will capture satellites too?  Can we put those imports at the end of `__init__.py` without a circular import issue?

dlakaplan: maybe even `get_observatory()` without arguments will return the names.

aarchiba: Constructing and registering all those objects at import time is an uncomfortable hack for just these kinds of reasons. I'm hesitant to make importing `pint.observatory` pull in all satellite and topocentric observatories, and of course any the user constructs only appear after the user has constructed them so we can't really create a `ensure_all_observatories_are_defined()` function (that would just do the above imports).

src/pint/observatory/global_clock_corrections.py
```diff
@@ -0,0 +1,235 @@
...
+from pathlib import Path
+
+from astropy.utils.data import download_file
+from loguru import logger as log
+
+from pint.pulsar_mjd import Time
+
+global_clock_correction_url_base = (
+    "https://raw.githubusercontent.com/ipta/pulsar-clock-corrections/main/"
+)
```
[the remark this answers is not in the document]
aarchiba: Possibly? I think of that module as useful for finding things about how PINT is installed, like where the examples are. Moving this URL away from all the code that uses it is not necessarily a readability win?

src/pint/observatory/__init__.py
dlakaplan: I guess I am thinking of a user who wants to quickly figure out what a telescope is called (so including aliases could be helpful).  In astropy I can do:
```
import astropy.coordinates
astropy.coordinates.EarthLocation.get_site_names()
```
and that gives the names (although not aliases).  I do use that to e.g. figure out that I want `gemini_north` and not `gemini-north`.  

It looks like the function to update clock files does import all of the observatories (albeit in a different scope).  Is that not run regularly?

aarchiba: .... actually `pint.observatory.Observatory.names()` is a function, I might as well do the imports in there, surely anyone calling that wants all of them? That and `get_observatory()` should cover all reasonable cases where people need the imports?

aarchiba: This is a little annoying to test as all test cases are run in the same python session, so the imports will basically always have happened by the time the test is run. It's actually the notebooks that are more likely to catch this sort of thing.

paulray: Don't forget about `get_satellite_observatory()`

paulray: You can't know the names of satellite observatories in advance. They are defined at runtime based on an orbit file.

aarchiba: So these are observatories that don't become available until the user loads an orbit file? I can do the import, but they won't appear in `names()` until that happens.

dlakaplan: Maybe this could also wait until observatories -> data file change happens, at which point it wouldn't be a python import but reading a data-file.

paulray:
I'd like to merge this. @aarchiba can you resolve the conflicts?

PR #1249, opened 2022-05-20
PINTk improvements

dlakaplan:
* Add more explicit zoom status
* Make it so you can use `--version` without par/Tim files
* Print chi^2 and rms info about highlighted points (or all, if none are selected)

### DIFF
--- a/src/pint/pintk/paredit.py
+++ b/src/pint/pintk/paredit.py
@@ -9,6 +9,7 @@ import pint.models
 
 import pint.logging
 from loguru import logger as log
+import pint.pintk.plk
 
 
 class ParChoiceWidget(tk.Frame):
@@ -19,6 +20,7 @@ class ParChoiceWidget(tk.Frame):
 
     def __init__(self, master=None):
         tk.Frame.__init__(self, master)
+        self.configure(bg=pint.pintk.plk.background)
 
         self.choose_callback = None
         self.initLayout()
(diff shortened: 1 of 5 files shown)


dlakaplan:
Example output from 'x' key:
```
Selected TOAs:                 87
Selected Chi2:          2.0943136e+08
Selected Chi2/Ntoa:     2407257
Selected Weighted RMS:  1158217.9 us
------------------------------------
```

dlakaplan:
Some people were having trouble with the foreground text being hard to read.  This explicitly sets it.  But I don't know if that is good practice (not sure if there is any system-level override).

dlakaplan:
The last change is to fix this:
Screen Shot 2022-05-20 at 3 12 36 PM

scottransom:
This looks good to me.  Want me to merge?

dlakaplan:
I guess I still have a question whether the hardcoding of the foreground color is a good idea.  Maybe if I can figure out what was happening with Lulu's display, that would help.  I couldn't reproduce it.

Also, is the formatting for the chi^2 output as desired?  Should it maybe have a preceding divider?

dlakaplan:
OK, I checked with Lulu.  She has the apple "dark mode" on, which changes around some of the colors.  I can't fully reproduce everything that she sees (for instance, even with dark mode on I still see black text) but I think I can try to understand where those settings come from.

scottransom:
I think I'm fine with the chi^2 output.  Although a preceding divider might be good, since I think the other output has that.

scottransom:
OK, can I merge this now?

dlakaplan:
I think so, although if you wanted to double-check that the colors look OK on another system that could be helpful.

scottransom:
OK.  I'l check on a couple systems, and if they look good, will merge.

scottransom:
Tried it on three very different systems and they all looked good.  I'm merging.

PR #1253, opened 2022-05-25
Test cleanup

### LINKED ISSUE #1238: Test suite creates files in source code directories
After a full `pytest` run I find the following files created in the source code directory:

```
	tests/datafile/J0030+0451_priors.png
	tests/datafile/photonphase-test.evt
	tests/datafile/polycos-photonphase-test.evt
	tests/datafile/test.tim
```

These files should be created in a temporary directory instead.

aarchiba:
This fixes a few minor issues with test so that

- You can now run `pytest -n auto` to run the test suite in parallel (if you have `pytest-xdist` installed, which I recommend)
- The test suite should no longer write anything under `tests/` even if `$TEMPO` or `$TEMPO2` are set.
    - There used to be all kinds of stuff in `.gitignore` but now that junk just isn't created
- The testing variable `datadir` is a `pathlib.Path` now (Now you can do `datadir / "file.par"`!), which revealed a number of places we don't cope correctly with Path objects 

Closes #1238


aarchiba:
Okay that error is weird. Doesn't happen on my machine and the file definitely exists.

aarchiba:
The new test was using `J0030+0451_post.par` as input, but that file was only generated as output by some previous test (that I removed stray output from). This introduced order dependency between tests, but only if you'd never run the test that generated the output.

aarchiba:
`pathlib.Path` is *great*!

paulray:
Just curious, what is good about Pathlib, as compared to manipulating paths as string, e.g. with `os.path.join()`?

aarchiba:
Convenience (and portability):

```python
parfile = datadir / "file.par"
parfile.write_text(parfile_contents)
par_dir = parfile.parent
for p in par_dir.glob("*.par"):
    if p.is_file():
         shutil.copy(p, tmp_path / p.name)
```
The main incompatibilities I've noticed are:
- You can't just stick an extension on with `p + ".dat"`; there's `p.with_suffix(".dat")` but that replaces what's there
- It's not actually a string so while many functions work fine ones that use `isinstance`, particularly to implement "filename or open file" semantics, need a tweak

aarchiba:
Perhaps I should also say, for the purposes of this PR, I'm not introducing `pathlib` for the first time - `pytest`'s `tmp_path` fixture already uses `pathlib` (which is in the python standard library) and that fixture is by far the easiest way to get somewhere for a test to write things.

dlakaplan:
> Perhaps I should also say, for the purposes of this PR, I'm not introducing `pathlib` for the first time - `pytest`'s `tmp_path` fixture already uses `pathlib` (which is in the python standard library) and that fixture is by far the easiest way to get somewhere for a test to write things.

Unrelated to this PR: is there a point in changing the `config`/`runtimefile` functions to use `pathlib`?

paulray:
After reading this code, I definitely not convinced that pathlib is a net win. It's not clear to me that the benefits outweigh the annoyances of file names not being strings. But, I guess I'm willing to learn to live with it.

aarchiba:
> After reading this code, I definitely not convinced that pathlib is a net win. It's not clear to me that the benefits outweigh the annoyances of file names not being strings. But, I guess I'm willing to learn to live with it.

The changes in this PR to support Path objects on input are kind of not up to us - people outside PINT may well be passing us Path objects and we need to cope. In most settings they will behave like strings, it's really just `isinstance` that it matters. 

For a concrete example, think how many of our tests contain `os.chdir` to avoid a zillion `os.path.join(datadir, name)`. `datadir / name` is a much easier sell (and we need to never change the directory in a test!).

aarchiba:
suggestions from telecon:

- Add `pytest-xdist` to `requirements_dev.txt`
- Add changelog entry
- Mention `pytest-xdist` in docs (how to contribute section)
- Ensure `make test` uses it

PR #1254, opened 2022-05-25
Make ephemeris loading more robust (and work on our CI)

aarchiba:
I'm trying to identify and resolve network problems that appear to be causing CI failures in #1239

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -25,10 +25,13 @@ jobs:
         include:
           - os: ubuntu-latest
             python: '3.8'
-            tox_env: 'py38-test-cov'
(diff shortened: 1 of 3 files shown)


dlakaplan:
@paulray and I had a discussion over slack in March about various CI failures due to problems downloading the IERS tables.  It seemed they were discussed over some astropy threads as well (maybe discussed in https://github.com/astropy/astropy/issues/13007).  Does your fix address that too?

aarchiba:
So far this fix addresses only the fact that astropy will try to download the DE440 ephemeris from the official location, which won't talk to our CI machines. We have a mirroring system in place, but we try astropy first, and weren't catching the failure.

IERS tables are a different issue, I'll go see if there's anything I can do with that.

Issue #1252, opened 2022-05-24
Resorting TOAs does not change group data


dlakaplan:
Related to difficulties sorting out #1202:
* PINT groups TOAs by observatory
* If subsequent operations resort those TOAs, the group metadata do not change
* Some other PINT operations later assume that the groups are still accurate, so are incorrect

I think this is all OK if we don't resort, but so far there is nothing to prohibit/warn people from doing that.  I'm not sure what the best way to do this is:
* disable sorting (seems hard), or at least issue a warning?
* do not make assumptions about grouping (could require some re-writing, could slow down some calculations)
* just tell people to not do this and hope for the best?

You can demonstrate this with:
```
import pint.toa

t = pint.toa.get_TOAs("grouptest.tim")
print(f"Current group indices: {t.table.groups.indices}")
print(f"Current index column {t.table['index']}")
print("Resorting...")
t.table.sort("index")
print(f"Current group indices: {t.table.groups.indices}")
print(f"Current index column {t.table['index']}")
```
which gives:
```
Current group indices: [ 0  1 71]
Current index column index
-----
   66
    0
    1
    2
    3
    4
    5
    6
    7
    8
    9
   10
   11
   12
   13
  ...
   54
   55
   56
   57
   58
   59
   60
   61
   62
   63
   64
   65
   67
   68
   69
   70
Length = 71 rows
Resorting...
Current group indices: [ 0  1 71]
Current index column index
-----
    0
    1
    2
    3
    4
    5
    6
    7
    8
    9
   10
   11
   12
   13
   14
  ...
   55
   56
   57
   58
   59
   60
   61
   62
   63
   64
   65
   66
   67
   68
   69
   70
Length = 71 rows
```
on a test file.  In the first instance you can see that `index=66` is first (an Arecibo TOA), so that corresponds to the first group.  That then gets resorted but the groups stay the same.
[grouptest.tim.zip](https://github.com/nanograv/PINT/files/8764980/grouptest.tim.zip)

aarchiba:
I think PINT should allow TOAs to be in any order and group them internally, temporarily, only when they need to be grouped. It's only one place - where we convert the TOAs to TDB, astropy needs each call to have all Time objects at the same position - but I don't know what the performance implications might be. PINT's non-enforced requirement that TOAs be grouped by observatory is confusing to users (why aren't the TOAs in the order they are in the file?) and causes problems.

dlakaplan:
I think it's also at:
[...]

PR #1256, opened 2022-05-26
WIP: Remove requirement to keep TOAs grouped

dlakaplan:
Previously TOAs had to be grouped by observatory.  So any resorting could mess up later access (#1252). This just uses fancy indexing to eliminate that requirement.

### DIFF
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -832,7 +832,7 @@ def build_table(toas, filename=None):
             "delta_pulse_number",
         ),
         meta={"filename": filename},
-    ).group_by("obs")
+    )  # .group_by("obs")
 
 
 def _cluster_by_gaps(t, gap):
@@ -1124,18 +1124,14 @@ class TOAs:
     corrections and the resulting TOAs object may not be set up the way one
     would normally expect.
 
-    The contents are stored in an :class:`astropy.table.Table`; this can be used to
-    access the contained information but the data may not be in the order you
-    expect: internally it is grouped by observatory (sorted by the observatory
-    object). Not all columns of the table are computed automatically, as their
+    The contents are stored in an :class:`astropy.table.Table`. Not all columns of the table are computed automatically, as their
     computation can be expensive. Methods of this class are available to
     populate these additional columns. Methods that return data from the columns
     do so in the internal order.
 
     TOAs objects can accept indices that are boolean, list-of-integer, or
     slice, to produce a new TOAs object containing a subset of the TOAs in the
-    original.  Note that the result is still grouped by the ``obs`` column, so
-    slices cannot reverse the order. For example, to obtain a new TOAs object
+    original.  For example, to obtain a new TOAs object
     containing the entries above 1 GHz:
 
     >>> t[t.table['freq'] > 1*u.GHz]
@@ -1210,8 +1206,7 @@ class TOAs:
     Attributes
     ----------
     table : astropy.table.Table
-        The data for all the TOAs. It is grouped by ``obs``, that is, it is
-        not in the same order as the original TOAs.
+        The data for all the TOAs.
     commands : list of str
         "Commands" that were written in the file; these will have affected
         how some or all TOAs were interpreted during loading.
@@ -1300,8 +1295,7 @@ class TOAs:
         result in selecting those TOAs (though not necessarily in that order),
         a list/array of Booleans will result in selecting those TOAs for which
         the list/array has True, and a slice will select the corresponding
-        slice of TOAs (again, the required grouping by observatory may mean
-        that reordering TOAs will not work as expected).
+        slice of TOAs.
 
         Both a column and subset can be selected at once by forming a tuple, as in
         ``toas["fish", ::10]``; this will result in selecting that subset of the
@@ -1342,8 +1336,8 @@ class TOAs:
             if isinstance(index, np.ndarray) and index.dtype == bool:
                 r = copy.deepcopy(self)
                 r.table = r.table[index]
-                if len(r.table) > 0:
-                    r.table = r.table.group_by("obs")
+                # if len(r.table) > 0:
+                #    r.table = r.table.group_by("obs")
                 return r
             elif (
                 isinstance(index, np.ndarray)
@@ -1352,14 +1346,14 @@ class TOAs:
             ):
                 r = copy.deepcopy(self)
                 r.table = r.table[index]
-                if len(r.table) > 0:
-                    r.table = r.table.group_by("obs")
+                # if len(r.table) > 0:
+                #    r.table = r.table.group_by("obs")
                 return r
             elif isinstance(index, slice):
                 r = copy.deepcopy(self)
                 r.table = r.table[index]
-                if len(r.table) > 0:
-                    r.table = r.table.group_by("obs")
+                # if len(r.table) > 0:
+                #    r.table = r.table.group_by("obs")
                 return r
             elif isinstance(index, int):
                 raise ValueError("TOAs do not support extraction of TOA objects (yet?)")
@@ -1479,8 +1473,8 @@ class TOAs:
         # Normal unpickling behaviour
         self.__dict__.update(state)
         # Astropy tables lose their group_by
-        if self.table.groups.keys is None:
-            self.table = self.table.group_by("obs")
+        # if self.table.groups.keys is None:
+        #    self.table = self.table.group_by("obs")
         if not hasattr(self, "max_index"):
             self.max_index = np.maximum.reduce(self.table["index"])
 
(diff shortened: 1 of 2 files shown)


dlakaplan:
I have done some speed tests.  e.g., for 8000 TOAs (all Arecibo) it takes ~6.1s with the new implementation, and ~6.8 with the old.  Same for 2000 TOAs from Arecibo & GBT.  I have compared the MJDs, TDBs, and SSB positions/velocities to 20 digits and they seem identical at least for ground-based observatories.  

I would still like to check spacecraft more (@paulray ).

dlakaplan:
Also implemented:
```
for obs, grp in self.get_obs_groups():
...
```
where `obs` is the observatory name and `grp` is the indices corresponding to that observatory (need not be contiguous).  This uses `pint.utils.GroupIterator`

src/pint/utils.py
```diff
@@ -1595,3 +1596,38 @@ def print_color_examples():
...
+            print(f"{o} {i}")
+
+    """
+
+    def __init__(self, items):
+        self.items = items
+        self.unique_items = np.unique(items)
+        self.indices = []
+        for item in self.unique_items:
+            self.indices.append(np.where(self.items == item)[0])
```
aarchiba, line 1621: It might be preferable to use boolean indexing (just don't apply `np.where`) - among other things it enforces the length of the array you're indexing into. And you can always use `np.where` or `np.nonzero` later.

dlakaplan: I had done it like that earlier, but found that I consistently needed the extra `where` step.

src/pint/utils.py
```diff
@@ -1595,3 +1596,38 @@ def print_color_examples():
                     end="",
                 )
             print("")
+
+
+class GroupIterator:
```
aarchiba, line 1601: This might be easier to write as a generator:
```python
def group_iterator(items):
    for item in np.unique(items):
        yield self.items==item
```
Shorter, and it doesn't generate or keep the indexing arrays except when needed.

dlakaplan: good idea.  I've done that, although for now the interface is the same.

src/pint/toa.py
```diff
@@ -2072,34 +2063,33 @@ def apply_clock_corrections(
...
             site = get_observatory(
                 obs,
                 include_gps=include_gps,
                 include_bipm=include_bipm,
                 bipm_version=bipm_version,
             )
-            loind, hiind = self.table.groups.indices[ii : ii + 2]
             # First apply any TIME statements
-            for jj in range(loind, hiind):
+            for jj in grp:
```
aarchiba, line 2074: Is this loop necessary? can you do `self["mjd", grp] += time.TimeDelta(corr)`? (You might need both fancy and column indexing at once - `self["mjd", grp]`, or they might need to be separate - `self["mjd"][grp]`.)

src/pint/utils.py
aarchiba: Boolean indexing should work on TOAs objects, both for read and write, I think?

src/pint/toa.py
dlakaplan: I had tried this and got various non-implemented errors when I tried it with the clock corrections.  Here I figured that a loop of some sort was needed anyway to deal with the str->float conversion.  That could be more compact but something similar would be needed.

dlakaplan:
To validate this I have a separate version of `toa.py` and I'm comparing the following for a number of files:
```
    assert t.get_summary() == told.get_summary()
    assert (mjd == old_mjd).all()
    assert (tdb == old_tdb).all()
    assert (clkcorr == old_clkcorr).all()
    assert (ssb_obs_pos == old_ssb_obs_pos).all()
    assert (ssb_obs_vel_ecl == old_ssb_obs_vel_ecl).all()
```
The load times aren't significantly different (some are slightly lower for new vs. old) but the values are identical.  However, these are all ground-based.

src/pint/toa.py
dlakaplan: I looked at this again.  A few issues:
* I can't combine fancy & column indexing:
```
In [19]: t['mjd',:10]
<ipython-input-19-48bcf6d97e11> in <module>
----> 1 t['mjd',:10]

~/pythonpackages/PINT/src/pint/toa.py in __getitem__(self, index)
   1370                 return self.table[column]
   1371             else:
-> 1372                 return self.table[column, subset]
   1373         else:
   1374             r = []

~/opt/anaconda3/envs/pintdev3.6/lib/python3.6/site-packages/astropy/table/table.py in __getitem__(self, item)
   1669         else:
   1670             raise ValueError('Illegal type {} for table item access'
-> 1671                              .format(type(item)))
   1672 
   1673     def __setitem__(self, item, value):

ValueError: Illegal type <class 'tuple'> for table item access
```
* I can't add directly to the Column data:
```
In [25]: t['mjd'] + TimeDelta(10*u.ms*np.ones(len(t)))
TypeError                                 Traceback (most recent call last)
<ipython-input-25-ee04eeefd5bf> in <module>
----> 1 t['mjd'] + TimeDelta(10*u.ms*np.ones(len(t)))

TypeError: unsupported operand type(s) for +: 'Column' and 'TimeDelta'
```
* I can't add to the `data` either:
```
In [31]: t['mjd'].data + TimeDelta(10*u.ms*np.ones(len(t)))
TypeError                                 Traceback (most recent call last)
<ipython-input-31-5ae8f62b4d42> in <module>
----> 1 t['mjd'].data + TimeDelta(10*u.ms*np.ones(len(t)))

TypeError: unsupported operand type(s) for +: 'numpy.ndarray' and 'TimeDelta'
```
I haven't delved too far into this, so it could be user error.  But it also could be that our `Time` objects all have different locations so can't be a single `Time` array.

dlakaplan:
OK, I also did the same test on a combined NICER topocentric + Nancy data-set from Paul, and it looks the same.

src/pint/toa.py
aarchiba: - `:10` is slice indexing, not fancy, and the line error you're getting suggests that fancy indexing woudld work.
- Indeed Astropy `Column` objects don't support various things. 
- The last is nothing to do with any code we wrote, it's Astropy Time being fussy. You could convert the TimeDelta to days or make a Time object out of the Column.

aarchiba: In fact I didn't bother to add slice indexing because, with TOAs in some unfriendly order, I expected users to be surprised if they tried to use slices. I don't think it would be very difficult.

dlakaplan: I did try the `TimeDelta` as just a quantity - still wouldn't work with the column.  Making a Time object out of the Column would (I think) lose the individual location for each entry.   In any case I'm not sure it's worth doing that and going back?

dlakaplan:
Right now the clock corrections and TIME statements are done with 3 separate loops:
```
 First apply any TIME statements
            for jj in grp:
                if "to" in flags[jj]:
                    # TIME commands are in sec
                    # SUGGESTION(@paulray): These time correction units should
                    # be applied in the parser, not here. In the table the time
                    # correction should have units.
                    # @aarchiba: flags should store strings only
                    corr[jj] = float(flags[jj]["to"]) * u.s
                    self["mjd"][jj] += time.TimeDelta(corr[jj])

            gcorr = site.clock_corrections(time.Time(self["mjd"][grp]))
            # it would be preferable to do this with array operations but I get `NotImplemented` errors when I do so
            for jj, cc in zip(grp, gcorr):
                self["mjd"][jj] += time.TimeDelta(cc)

            corr[grp] += gcorr

            # Now update the flags with the clock correction used
            for jj in grp:
                if corr[jj] != 0:
                    flags[jj]["clkcorr"] = str(corr[jj].to_value(u.s))
```
It could be much more compact and done with only 1 explicit loop by doing:
```
time_statements = self.get_flag_value("to",0,float)*u.s
for obs, grp in self.get_obs_groups():
    site = ...
    correction[grp] = site.clock_corrections(time.Time(self["mjd"][grp])) + time_statements[grp]
    for jj in grp:
        ...
``` 
but this results in slightly different values, since the interpolation to get the clock corrections is being done on times prior to the "-to" flags, instead of after.  The values are therefore very slightly different (~1e-17 day).  I don't know if there is a specific logic either way, but for the moment I left it the way it was before so that the agreement with prior code is exact.

dlakaplan:
On the assumption that others agree with Anne, I have gone for the simpler implementation.  This leads to differences in clock corrections of ~1e-19 s and in TDBs of ~1e-15d.  I think both of those are acceptable.

dlakaplan:
Argh - I didn't mean to merge this into master - I was trying to merge nanograv:master into my branch and resolve some conflicts.  I appear to have done something wrong.  I think this is likely OK anyway, but feel free to undo that commit (I had hoped to discuss this a little more broadly).

PR #1257, opened 2022-05-27
Allow PINT to write clock correction files

aarchiba:
PINT is already able to read in clock correction files in either TEMPO or TEMPO2 format. This PR allows PINT to write out clock correction files in either format, optionally combining multiple clock correction files (for example ao2gps + gps2utc). This enables conversion of clock files between formats as well as conversion of other suitable information into clock files.

As it currently stands, the "merging" of clock files means combining them by adding the corrections from the included clock correction files. This is distinct from splicing together clock correction files that cover different intervals; that's a different problem.

When reading, writing, and merging clock correction files, comments are preserved. This allows comments at the beginning of the file, comments on clock correction lines, and comments following a clock correction. (Comments on their own line are associated with the previous clock correction.)

This allows more or less lossless conversion between TEMPO and TEMPO2 clock correction file formats, and is used by the global clock corrections [repository](https://nanograv.github.io/pulsar-clock-corrections/status.html) for this purpose.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - Added functions in pint.observatory to request the status of PINT's available clock corrections
 - Added the ability to query clock correction files or observatories for their last corrected MJD
 - Added an example showing how to check the status of your clock corrections
+- Added the ability to write clock files in TEMPO or TEMPO2 format
 ### Fixed
 - WLS fitters no longer ignore EFAC/EQUAD (bug #1226)
 ### Changed
--- a/src/pint/observatory/clock_file.py
+++ b/src/pint/observatory/clock_file.py
@@ -1,17 +1,12 @@
 """Routines for reading various formats of clock file."""
 
 import os
-import warnings
+import re
 from textwrap import dedent
 
 import astropy.units as u
 import numpy as np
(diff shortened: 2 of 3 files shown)


aarchiba:
> Does this keep comments? There is along history of clock correction files having comments about why things changed and then those getting lost in various conversions etc.

PINT does not currently record comments at all upon reading clock files. I had in mind that this would be used only for disposable temporary clock files with comments indicating that they were auto-generated. Some form of comment preservation could be added.

scottransom:
Probably not super necessary given that I think most of the comments in current clock files are for the archival corrections (meaning long ago).  Although I do see that they could be useful even now and in the future (i.e. the big step in the GBT corrections that I *think* is due to maser maintenance).

aarchiba:
Now preserves comments on read/write in both TEMPO and TEMPO2 formats. Needs more testing. Also merging discards comments (and it shouldn't! or at least not necessarily).

aarchiba:
Also worth thinking about: do we implement the TEMPO `f` flag (treat as a fixed measurement, do not interpolate)? Not fundamentally hard, just turn it into two corrections at the beginning and end of the constant time.

dlakaplan:
Is this ready to merge?  Were there any pending items from our meeting?

aarchiba:
I think we concluded this was ready to go.

PR #1259, opened 2022-05-31
Color TOAs by jump in pintk

dlakaplan:
Screen Shot 2022-05-31 at 2 22 19 PM
along with:

```
"Jump" mode selected
JUMP1 -state Coherence = red
JUMP2 -state Intensity = green
JUMP3 -gui_jump 3 = cyan
JUMP4 -gui_jump 4 = blue
JUMP5 -gui_jump 5 = burnt orange
JUMP6 -gui_jump 6 = brown
JUMP7 -gui_jump 7 = indigo
JUMP8 -gui_jump 8 = purple
  selected = orange
```
it doesn't really have a way to handle overlapping jumps, though.

### DIFF
--- a/src/pint/pintk/colormodes.py
+++ b/src/pint/pintk/colormodes.py
@@ -362,6 +362,7 @@ class JumpMode(ColorMode):
         return model.get_jump_param_objects()
 
     jump_colors = named_colors
+    selected_color = "orange"
 
     def displayInfo(self):
         outstr = '"Jump" mode selected\n'
@@ -375,7 +376,7 @@ class JumpMode(ColorMode):
             if jump.key_value is not None:
                 outstr += " " + " ".join(jump.key_value)
             outstr += f" = {color_name}\n"
-        outstr += "  selected = orange\n"
+        outstr += f"  selected = {selected_color}\n"
         print(outstr)
 
     def plotColorMode(self):
@@ -404,7 +405,9 @@ class JumpMode(ColorMode):
                 self.application.xvals[self.application.selected],
                 self.application.yvals[self.application.selected],
                 marker=".",
-                color="orange",
+                color=selected_color,
             )
         else:
-            self.application.plotErrorbar(self.application.selected, color="orange")
+            self.application.plotErrorbar(
+                self.application.selected, color=selected_color
+            )
--- a/src/pint/pintk/colormodes.py
+++ b/src/pint/pintk/colormodes.py
@@ -320,3 +320,85 @@ class ObsMode(ColorMode):
             )
         else:
             self.application.plotErrorbar(self.application.selected, color="orange")
+
+
+class JumpMode(ColorMode):
+    """
+    A class to manage the Jump color mode, where TOAs are colored
+    according to their jump.
+    """
+
+    def __init__(self, application):
+        super(JumpMode, self).__init__(application)
+        self.mode_name = "jump"
+
+    def get_jumps(self):
+        if self.application.psr.fitted:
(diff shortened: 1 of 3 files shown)


dlakaplan:
I also did some slight re-organization of how colors are defined/handled, to make it easier to implement other modes.  If needed I could undo that.

src/pint/pintk/colormodes.py
```diff
@@ -5,6 +5,28 @@
 import pint.logging
 from loguru import logger as log
 
+named_colors = {
```
aarchiba, line 8: Matplotlib has a list of named colors: https://matplotlib.org/stable/gallery/color/named_colors.html

src/pint/pintk/colormodes.py
```diff
@@ -320,3 +342,72 @@ def plotColorMode(self):
             )
         else:
             self.application.plotErrorbar(self.application.selected, color="orange")
+
+
+class JumpMode(ColorMode):
+    """
+    A class to manage the Jump color mode, where TOAs are colored
```
aarchiba, line 349: This is not docstring format.

src/pint/pintk/colormodes.py
```diff
@@ -320,3 +342,72 @@ def plotColorMode(self):
...
+    according to their jump.
+    """
+
+    def __init__(self, application):
+        super(JumpMode, self).__init__(application)
```
aarchiba, line 354: Just use `super().__init__(application)`.

src/pint/pintk/colormodes.py
```diff
@@ -320,3 +342,72 @@ def plotColorMode(self):
...
+        """
+        alltoas = self.application.psr.all_toas
+        for jumpnum, jump in enumerate(self.get_jumps()):
+            color_number = jumpnum % (len(self.jump_colors) - 1)
+            color_name = list(self.jump_colors)[color_number]
+            toas = jump.select_toa_mask(alltoas)
+            color = self.jump_colors[color_name]
+            # group toa indices by jump
+            if self.application.yerrs is None:
+                self.application.plkAxes.scatter(
```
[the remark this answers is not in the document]
dlakaplan: You can do this with `scatter`, but not clear to me that it works with `errorbar`.  I may be doing it wrong but quick tests & searches are failing.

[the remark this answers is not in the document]
aarchiba: I think with errorbar you need to specify whether you want to color the points or bars so the interface should be different.

src/pint/pintk/colormodes.py
scottransom: Yeah, that's a good point.  We probably shouldn't define our own.  If you do `import matplotlib.colors as mcolors` then `mcolors.cnames`, `mcolors.TABLEAU_COLORS`, and `mcolors.XKCD_COLORS` are all dictionaries just like you have defined.

The xkcd one is cute:  https://xkcd.com/color/rgb/

dlakaplan: Yes, I was trying to minimally change what was there.  There's also the question of an appropriate subset (usually I would use seaborn, but those aren't always nameable) but it wouldn't be too hard to identify the XKCD color most similar to the ones already chosen.

dlakaplan:
Argh - I didn't mean to merge this into master - I was trying to merge nanograv:master into my branch.  I appear to have done something wrong.  I think this is likely OK anyway, but @scottransom : feel free to under that commit.

scottransom:
I think it is fine.  We can tweak things as needed.

Issue #1260, opened 2022-05-31
Removing FD (and other?) parameters from a model doesn't change the number of terms; can't construct fitter


tcromartie:
Example: I read in some par (including FD1-4, unfrozen) and tim file with `get_model_and_toas`, then do `m.remove_param('FD4')`. It now shows that the model is missing FD4 from `m.components['FD']`, which is expected. However, constructing the fitter fails after that. David K found that `num_FD_terms` remains at the original number of model components. This should maybe throw a warning.

paulray:
Can you give the error message from fitter? And maybe give a simple example of triggering this bug (perhaps using a par file already in PINT, like `examples/B1855+09_NANOGrav_9yv1.gls.par`? That would make it easier to diagnose.

tcromartie:
```
parfile = "B1855+09_NANOGrav_9yv1.gls.par"
timfile = "B1855+09_NANOGrav_9yv1.tim"

m, t = get_model_and_toas(parfile, timfile)
display(m.components['FD']) # Shows FD1-3
print(m.components['FD'].num_FD_terms) # 3
m.remove_param('FD3')
print(m.components['FD'].num_FD_terms) # Shows FD1-2
display(m.components['FD']) # Still returns 3

m.validate()
f = pint.fitter.DownhillGLSFitter(t, m)
f.fit_toas()
```

Simply returns "KeyError: 3" from `frequency_dependent.py` line 89 (`FD_coeff` definition); the issue seems to be the number of terms not actually decreasing.

paulray:
This seems to work fine if you add `m.setup()` before your `m.validate()` call.

aarchiba:
> Yeah, I was wondering about exactly that last night. But I didn't have a chance to check it then. I think it is a safe (and sometimes required) thing to run m.setup() basically every time you change the model.

Can we abolish this? Like automatically trigger `m.setup()` on parameter addition and removal and on trying to create a Fitter?

paulray:
If there aren't any consequences of running it too many times, then for sure parameter additions and removals should trigger a `setup()`

tcromartie:
Addressed with #1264. Some outstanding questions about jumps and flags should be discussed (as well as the possibility of assigning a property to parameters to designate them as non-removeable), but the issues here are fixed.

PR #1262, opened 2022-06-01
Fix remove param

tcromartie:
Addresses #1260 by automatically running setup when removing parameters. Also adds test for parameter addition/removal.


aarchiba:
Looks like this accidentally pulled in changes you were working on in another branch. It's awkward when you want to use stuff from an un-merged PR, or even just to make changes to the same piece of code. Small PRs we can merge right away help, but sometimes you can't make a small standalone PR...

tcromartie:
@aarchiba that's why I closed it, I saw I didn't start from a clean branch. It should be standalone (doesn't rely on the old commits). Should I just start over, or can I cherry pick just this most recent commit?

PR #1264, opened 2022-06-03
Auto-setup when removing params

tcromartie:
Add auto-setup to remove_param; add test for addition/removal of parameters. Addresses #1260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 
 ## Unreleased
 ### Changed
-- model.phase() now defaults to abs_phase=True when TZR* params are in the model
+- `model.phase()` now defaults to `abs_phase=True` when TZR* params are in the model
 - TOAs no longer need to be grouped by observatory
 - removed explicit download of IERS and leapsecond data (handled now by astropy)
 - The default version of TT(BIPM) uses BIPM2021
-- model.setup() now gets called automatically after removing a parameter as part of remove_param
+- `model.setup()` now gets called automatically after removing a parameter as part of `remove_param`
 ### Added
 - logging now needs to be setup explicitly
 - Color-by-jump mode for pintk


tcromartie:
Note about failure: Adding setup() to remove_param breaks jump removal; in test_remove_jump_and_flags, when it tries to delete JUMP1 it also deletes JUMP2.

aarchiba:
> Notes about failure: Adding setup() to remove_param breaks jump removal; in test_remove_jump_and_flags, when it tries to delete JUMP1 it also deletes JUMP2. "PhaseJump" components are reflecting the deletion correctly (i.e. two mask parameters prior to first deletion, one remains after deletion). Flags also being removed properly. However, the param map shows that they're both missing after the first deletion. Doesn't appear to be related to the jump-and-flag deletion.

Do you mean that if you delete JUMP1, `setup()` then deletes JUMP2? That is a bug.

tcromartie:
Outstanding question because I'm not familiar with the `gui_jump` flag behavior / pintk: should `delete_jump_and_flags` control deletion for normal `-jump` flags and `-gui_jump` flags at the same time? 

Example: `test_remove_jump_and_flags` does the following:
1. adds a `PhaseJump` component
2. calls `add_jump_and_flags` to add `JUMP1` for one subset of TOAs and `JUMP2` for another
3. at this point, those TOAs have both `jump` and `gui_jump` flags (for either JUMP1 or JUMP2)
4. then, `delete_jump_and_flags` is called for the `JUMP1` TOAs, but that only deletes the `JUMP1` parameter from the model and the `jump = 1` flag. The `gui_jump = 1` flag remains.

It's totally possible that this is the expected behavior, but if it's not, I know @aarchiba said the flagging needs to be changed anyway, so maybe it's moot.

tests/test_parameters.py
```diff
@@ -587,3 +587,15 @@ def test_compare_key_value_list():
...
     pickle.dumps(p)
+
+
+def test_add_remove_param():
+    m = get_model(os.path.join(datadir, "B1855+09_NANOGrav_9yv1.gls.par"))
+    t = get_TOAs(os.path.join(datadir, "B1855+09_NANOGrav_9yv1.tim"))
+    FD4 = prefixParameter(parameter_type="float", name="FD4", value=0.0, units=u.s)
+    m.add_param_from_top(FD4, "FD")
+    assert len(m.components["FD"].params) == 4
+    f = pint.fitter.GLSFitter(toas=t, model=m)
```
aarchiba, line 598: What does creating a Fitter test? Can you write a comment here that explains?

tests/test_parameters.py
```diff
@@ -587,3 +587,15 @@ def test_compare_key_value_list():
 )
 def test_parameter_can_be_pickled(p):
     pickle.dumps(p)
+
+
+def test_add_remove_param():
```
aarchiba, line 592: Can you explain in the test name what is being tested? Like what is adding and removing parameters supposed to do to pass the test? Just be executable, or is there some property they are supposed to maintain?

tests/test_parameters.py
```diff
@@ -587,3 +587,17 @@ def test_compare_key_value_list():
...
+
+
+def test_add_remove_param_setup_success():
+    """Checks that add_param and remove_param don't require m.setup() to be run prior to constructing a fitter"""
+    m = get_model(os.path.join(datadir, "B1855+09_NANOGrav_9yv1.gls.par"))
+    t = get_TOAs(os.path.join(datadir, "B1855+09_NANOGrav_9yv1.tim"))
+    FD4 = prefixParameter(parameter_type="float", name="FD4", value=0.0, units=u.s)
+    m.add_param_from_top(FD4, "FD")
+    assert len(m.components["FD"].params) == 4
+    """Fitter construction used to fail after remove_param without m.setup(). Test this:"""
```
aarchiba, line 599: This would be good detail to put in the test title!

aarchiba: I'm just hoping the test title can tell me what property it's testing - `test_add_param_increases_number_of_params`, `test_fitter_can_be_constructed_after_remove_param` - so it's more obvious what feature the test is testing. I realize the test currently checks both of these, but it would be better if it could be two tests so we'd get separate answers on whether those two things worked. If the test is checking for a specific bug, it can make sense to include the bug number in the test name. (Nobody ever calls tests by hand so it's not particularly a problem if their names are long.)

My reasoning is, suppose I ran `make test` and your test here failed. What would I know went wrong? What was supposed to happen but isn't? How well does it help me find the place I broke something?

tcromartie: I'm going to modify what's now the second test again, so not quite ready for review.

aarchiba:
Does it make sense to add a test for what happens when you try to remove FD1 but there's FD0 and FD2 in the model? I'm not sure what's supposed to happen exactly.

tcromartie:
It doesn't seem to fail or throw a warning when you try to either add a non-consecutive FD param or remove an FD param from the middle (i.e. remove FD2 when you have FD1 and FD3). Should this be a separate issue? I feel like it could be a little messy, especially if this needs to apply to more params than just FD.

aarchiba:
> It doesn't seem to fail or throw a warning when you try to either add a non-consecutive FD param or remove an FD param from the middle (i.e. remove FD2 when you have FD1 and FD3). Should this be a separate issue? I feel like it could be a little messy, especially if this needs to apply to more params than just FD.

I think you're right, it's a bigger issue - on the one hand I don't think it should be a problem to remove one of a sequence of parameters, but I'm not sure what all code might assume that the numbers are contiguous.

aarchiba:
*sigh* You got bitten by the CHANGELOG conflict. Sorry. But it all looks good, feel free to hit merge when CI passes.

tcromartie:
> _sigh_ You got bitten by the CHANGELOG conflict. Sorry. But it all looks good, feel free to hit merge when CI passes.

Did I handle it correctly or should I have done something else? And will do, thanks for all your help!

PR #1267, opened 2022-06-03
Remove requirement to keep TOAs grouped

dlakaplan:
TOAs no longer need to be grouped, although they still can be.  Slight change in order of operations/interpolation in `apply_clock_correction` that results in ~3e-15s changes.

(3rd attempt at this, after messing up previous PRs)

### DIFF
--- a/src/pint/simulation.py
+++ b/src/pint/simulation.py
@@ -168,6 +168,7 @@ def make_fake_toas(ts, model, add_noise=False, name="fake"):
     for f in tsim.table["flags"]:
         f["name"] = name
 
+    tsim.table = tsim.table.group_by("obs")
     return tsim
 
 
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -290,10 +290,8 @@ def get_TOAs(
             log.info("Ephem changed, recalculation needed")
         recalc = True
         updatepickle = True
-    t.table = t.table.group_by("obs")
     if recalc or "tdb" not in t.table.colnames:
         t.compute_TDBs(method=tdb_method, ephem=ephem)
-
     if planets is None:
         planets = t.planets
     elif planets != t.planets:
@@ -306,6 +304,7 @@ def get_TOAs(
     if usepickle and updatepickle:
         log.info("Pickling TOAs.")
         save_pickle(t, picklefilename=picklefilename)
+    t.table = t.table.group_by("obs")
     return t
 
 
@@ -419,6 +418,7 @@ def get_TOAs_list(
         t.compute_TDBs(method=tdb_method, ephem=ephem)
     if "ssb_obs_pos" not in t.table.colnames:
         t.compute_posvels(ephem, planets)
+    t.table = t.table.group_by("obs")
     return t
 
 
@@ -838,7 +838,7 @@ def build_table(toas, filename=None):
             "delta_pulse_number",
         ),
         meta={"filename": filename},
-    ).group_by("obs")
+    )
 
 
 def _cluster_by_gaps(t, gap):
@@ -1130,18 +1130,14 @@ class TOAs:
     corrections and the resulting TOAs object may not be set up the way one
     would normally expect.
 
-    The contents are stored in an :class:`astropy.table.Table`; this can be used to
-    access the contained information but the data may not be in the order you
-    expect: internally it is grouped by observatory (sorted by the observatory
-    object). Not all columns of the table are computed automatically, as their
+    The contents are stored in an :class:`astropy.table.Table`. Not all columns of the table are computed automatically, as their
     computation can be expensive. Methods of this class are available to
     populate these additional columns. Methods that return data from the columns
     do so in the internal order.
 
     TOAs objects can accept indices that are boolean, list-of-integer, or
     slice, to produce a new TOAs object containing a subset of the TOAs in the
-    original.  Note that the result is still grouped by the ``obs`` column, so
-    slices cannot reverse the order. For example, to obtain a new TOAs object
+    original.  For example, to obtain a new TOAs object
     containing the entries above 1 GHz:
 
     >>> t[t.table['freq'] > 1*u.GHz]
@@ -1153,6 +1149,11 @@ class TOAs:
     TOAs objects also support assignment through these methods:
 
     >>> t['high', t['freq'] > 1*u.GHz] = "1"
+
+    TOAs are grouped by observatory by default, but this is just a convention.  To iterate over all observatory groups:
+
+    >>> for obs,idx in t.get_obs_groups():
+
 
     .. list-table:: Columns in ``.table``
        :widths: 15 85
@@ -1216,8 +1217,7 @@ class TOAs:
     Attributes
     ----------
     table : astropy.table.Table
-        The data for all the TOAs. It is grouped by ``obs``, that is, it is
-        not in the same order as the original TOAs.
+        The data for all the TOAs.
     commands : list of str
         "Commands" that were written in the file; these will have affected
         how some or all TOAs were interpreted during loading.
@@ -1306,8 +1306,7 @@ class TOAs:
         result in selecting those TOAs (though not necessarily in that order),
         a list/array of Booleans will result in selecting those TOAs for which
         the list/array has True, and a slice will select the corresponding
-        slice of TOAs (again, the required grouping by observatory may mean
-        that reordering TOAs will not work as expected).
+        slice of TOAs.
 
         Both a column and subset can be selected at once by forming a tuple, as in
         ``toas["fish", ::10]``; this will result in selecting that subset of the
@@ -1348,8 +1347,6 @@ class TOAs:
             if isinstance(index, np.ndarray) and index.dtype == bool:
                 r = copy.deepcopy(self)
                 r.table = r.table[index]
-                if len(r.table) > 0:
-                    r.table = r.table.group_by("obs")
                 return r
             elif (
                 isinstance(index, np.ndarray)
@@ -1358,14 +1355,10 @@ class TOAs:
             ):
                 r = copy.deepcopy(self)
                 r.table = r.table[index]
-                if len(r.table) > 0:
-                    r.table = r.table.group_by("obs")
                 return r
             elif isinstance(index, slice):
                 r = copy.deepcopy(self)
                 r.table = r.table[index]
-                if len(r.table) > 0:
-                    r.table = r.table.group_by("obs")
                 return r
             elif isinstance(index, int):
                 raise ValueError("TOAs do not support extraction of TOA objects (yet?)")
@@ -1562,6 +1555,10 @@ class TOAs:
         """Return a numpy array of the observatories for each TOA."""
         return self.table["obs"]
 
+    def get_obs_groups(self):
+        """Return an iterator over the different observatories"""
+        return pint.utils.group_iterator(self["obs"])
+
     def get_pulse_numbers(self):
         """Return a numpy array of the pulse numbers for each TOA if they exist."""
         # TODO: use a masked array?  Only some pulse numbers may be known
@@ -1660,7 +1657,7 @@ class TOAs:
         Parameters
         ----------
         gap_limit : astropy.units.Quantity, optional
-            The minimum size of gap to create a new group. Defaults to two hours.
+            The minimum size of gap to create a new cluster. Defaults to two hours.
         add_column : bool, optional
             Whether or not to add a ``clusters`` column to the TOA table (default: False)
         add_flag : str, optional
@@ -1763,10 +1760,7 @@ class TOAs:
             if not hasattr(self, "table_selects"):
                 self.table_selects = []
             self.table_selects.append(copy.deepcopy(self.table))
-            # Our TOA table must be grouped by observatory for phase calcs
             self.table = self.table[selectarray]
-            if len(self.table) > 0:
-                self.table = self.table.group_by("obs")
         else:
             raise ValueError("TOA selection not implemented for TOA lists.")
 
@@ -1798,14 +1792,13 @@ class TOAs:
         s += f"Number of observatories: {len(self.observatories)} {list(self.observatories)}\n"
         s += f"MJD span:  {self.first_MJD.mjd:.3f} to {self.last_MJD.mjd:.3f}\n"
         s += f"Date span: {self.first_MJD.iso} to {self.last_MJD.iso}\n"
-        for ii, key in enumerate(self.table.groups.keys):
-            grp = self.table.groups[ii]
-            s += f"{key['obs']} TOAs ({len(grp)}):\n"
-            s += f"  Min freq:      {np.min(grp['freq'].to(u.MHz)):.3f}\n"
-            s += f"  Max freq:      {np.max(grp['freq'].to(u.MHz)):.3f}\n"
-            s += f"  Min error:     {np.min(grp['error'].to(u.us)):.3g}\n"
-            s += f"  Max error:     {np.max(grp['error'].to(u.us)):.3g}\n"
-            s += f"  Median error:  {np.median(grp['error'].to(u.us)):.3g}\n"
+        for obs, grp in self.get_obs_groups():
+            s += f"{obs} TOAs ({len(grp)}):\n"
+            s += f"  Min freq:      {np.min(self['freq'][grp].to(u.MHz)):.3f}\n"
+            s += f"  Max freq:      {np.max(self['freq'][grp].to(u.MHz)):.3f}\n"
+            s += f"  Min error:     {np.min(self['error'][grp].to(u.us)):.3g}\n"
+            s += f"  Max error:     {np.max(self['error'][grp].to(u.us)):.3g}\n"
+            s += f"  Median error:  {np.median(self['error'][grp].to(u.us)):.3g}\n"
         return s
 
     def print_summary(self):
@@ -1907,7 +1900,7 @@ class TOAs:
         index_order : bool
             If True, preserve the order of the index column, but renumber so
             there are no gaps. If False, number according to the order TOAs
-            occur in the object (they will be grouped by observatory).
+            occur in the object.
 
         Returns
         =======
@@ -1955,9 +1948,7 @@ class TOAs:
         order_by_index : bool
             If True, write the TOAs in the order specified in the "index" column
             (which is usually the same as the original file);
-            if False, write them in the order they occur in the TOAs object
-            (which is usually the same as the original file except that all the
-            TOAs associated with each observatory have been grouped).
+            if False, write them in the order they occur in the TOAs object.
         include_info : bool, optional
             Include information string if True
         comment : str, optional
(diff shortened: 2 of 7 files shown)


dlakaplan:
Addresses #1252

src/pint/toa.py
```diff
@@ -306,6 +304,7 @@ def get_TOAs(
     if usepickle and updatepickle:
         log.info("Pickling TOAs.")
         save_pickle(t, picklefilename=picklefilename)
+    t.table = t.table.group_by("obs")
```
aarchiba, line 307: Why group the table rather than preserving the user's input order?

src/pint/simulation.py
```diff
@@ -168,6 +168,7 @@ def make_fake_toas(ts, model, add_noise=False, name="fake"):
     for f in tsim.table["flags"]:
         f["name"] = name
 
+    tsim.table = tsim.table.group_by("obs")
```
aarchiba, line 171: Why group the table now it is no longer necessary?

aarchiba, line 171: Why group the table now it is no longer necessary?

src/pint/toa.py
```diff
@@ -419,6 +418,7 @@ def get_TOAs_list(
         t.compute_TDBs(method=tdb_method, ephem=ephem)
     if "ssb_obs_pos" not in t.table.colnames:
         t.compute_posvels(ephem, planets)
+    t.table = t.table.group_by("obs")
```
aarchiba, line 421: Why group the table?

aarchiba, line 421: Why group the table?

src/pint/toa.py
```diff
@@ -1130,18 +1130,14 @@ class TOAs:
     corrections and the resulting TOAs object may not be set up the way one
     would normally expect.
 
-    The contents are stored in an :class:`astropy.table.Table`; this can be used to
-    access the contained information but the data may not be in the order you
-    expect: internally it is grouped by observatory (sorted by the observatory
-    object). Not all columns of the table are computed automatically, as their
+    The contents are stored in an :class:`astropy.table.Table`. Not all columns of the table are computed automatically, as their
```
aarchiba, line 1133: It would be good to reflow this.

src/pint/toa.py
```diff
@@ -1154,6 +1150,11 @@ class TOAs:
 
     >>> t['high', t['freq'] > 1*u.GHz] = "1"
 
+    TOAs are grouped by observatory by default, but this is just a convention.  To iterate over all observatory groups:
```
aarchiba, line 1153: I think TOAs should be left in the order they were in the tim files.

tests/test_toa.py
```diff
@@ -87,19 +87,17 @@ def test_mulit_obs(self):
...
 
-        # Table will be grouped by observatories, and will be sorted by the
-        # observatory so the TOA order will be different
-        assert toas.table["obs"][0] == site2.name
-        assert toas.table["mjd"][0] == t2.mjd
-        assert toas.table["obs"][1] == site3.name
-        assert toas.table["mjd"][1] == t3.mjd
-        assert toas.table["obs"][2] == site1.name
-        assert toas.table["mjd"][2] == t1.mjd
+        assert toas.table["obs"][0] == site1.name
```
aarchiba, line 90: What property is this test verifying? That the order of the result is just the concatenation of the originals? I don't understand what makes this the right behaviour.

aarchiba, line 90: What property is this test verifying? That the order of the result is just the concatenation of the originals? I don't understand what makes this the right behaviour.

aarchiba: Can you add some tests that check what order the TOAs returned from various operations are in? I think we should default to preserving the order in the tim file unless the user explicitly re-orders the file, and I think the tests should verify this.

aarchiba: Can you add some tests that check what order the TOAs returned from various operations are in? I think we should default to preserving the order in the tim file unless the user explicitly re-orders the file, and I think the tests should verify this.

aarchiba:
I think there is a discussion to be had: what order should TOAs be in, inside PINT? I suggest they stay in the order they were in `.tim` files unless users explicitly reorder them. Whatever the decision, we should have tests that verify that the order we claim TOAs will be in is the order they are actually in.

dlakaplan:
@aarchiba : some quick answers (can discuss further):
* Why group the table?  In some cases it could be useful, and I was trying to minimally change previous behavior.  If we discuss and decide that this isn't useful it's easy to now eliminate.  Or make it an option when loading TOAs.
* Seemingly trivial tests: most of these were more meaningful when grouping was more of a thing.  Again, I tried to just do minimal changes in case there was some other more subtle purpose that I didn't have time to figure out.  If we decide that they are deprecated we can change.

aarchiba:
Sorry for the doubled review comments, I don't know what happened with that.

dlakaplan:
@aarchiba / @scottransom : I am testing the result of sorting the TOAs.  To a large extent it seems fine, but when I compare the chi^2 they are not identical (differ at ~1e-14).  Some changes went away when I forced the first TOA to be the same, so it could be an absolute phase question, but not all.  Any reasons why the order should matter?

dlakaplan:
I added `tests/test_timing_model.py::test_resorting_toas` to test this.  There is one equality and one `isclose`, but I'd like to understand the reason that they are both not equal

aarchiba:
> @aarchiba / @scottransom : I am testing the result of sorting the TOAs. To a large extent it seems fine, but when I compare the chi^2 they are not identical (differ at ~1e-14). Some changes went away when I forced the first TOA to be the same, so it could be an absolute phase question, but not all. Any reasons why the order should matter?

The order definitely should not matter. Can you compare the residuals? Use argsort and compare residuals then reshuffle to reshuffle then residuals? Does `abs_phase=True` change anything?

tests/test_timing_model.py
```diff
@@ -388,6 +390,33 @@ def test_parfile_and_timfile_jumps(timfile_jumps):
...
+    tcopy = deepcopy(t)
+    # sort by frequency
+    i = np.argsort(t.table["freq"])
+    i0 = np.where(i == 0)[0][0]
+    # make sure the first index is the same
+    # otherwise there are very small differences
+    i[[0, i0]] = i[[i0, 0]]
+    tcopy.table = tcopy.table[i]
+    rsort = pint.residuals.Residuals(tcopy, m)
+    assert r.calc_chi2() == rsort.calc_chi2()
```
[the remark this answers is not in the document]
dlakaplan:
```
tcopy = copy.deepcopy(t)
# sort by mjd
i = np.argsort(t.table["mjd_float"])
tcopy.table = tcopy.table[i]
rsort = pint.residuals.Residuals(tcopy, m)
print((rsort.time_resids - r.time_resids[i]))
```
gives:
```
[-3.33413344e-19 -3.33519223e-19 -3.33413344e-19 -3.33519223e-19
 -3.33413344e-19 -3.33519223e-19 -3.33413344e-19 -3.33519223e-19
 -3.33519223e-19 -3.33413344e-19 -3.33519223e-19 -3.33413344e-19
 -3.33466283e-19 -3.33466283e-19 -3.33519223e-19 -3.33466283e-19
 -3.33466283e-19 -3.33466283e-19 -3.33519223e-19 -3.33492753e-19
 -3.33492753e-19 -3.33519223e-19 -3.33519223e-19 -3.33519223e-19
 -3.33519223e-19 -3.33519223e-19 -3.33519223e-19 -3.33519223e-19
 -3.33519223e-19 -3.33519223e-19 -3.33492753e-19 -3.33492753e-19
 -3.33519223e-19 -3.33519223e-19 -3.33519223e-19 -3.33519223e-19
 -3.33519223e-19 -3.33519223e-19 -3.33413344e-19 -3.33519223e-19
 -3.33519223e-19 -3.33519223e-19 -3.33492753e-19 -3.33492753e-19
 -3.33492753e-19 -3.33492753e-19 -3.33492753e-19 -3.33492753e-19
 -3.33492753e-19 -3.33492753e-19 -3.33492753e-19 -3.33492753e-19
 -3.33492753e-19 -3.33492753e-19 -3.33519223e-19 -3.33519223e-19
 -3.33519223e-19 -3.33519223e-19 -3.33519223e-19 -3.33519223e-19
 -3.33519223e-19 -3.33519223e-19] s
```

However, if I use `subtract_mean=False` in computing the residuals it gives 0.

tests/test_timing_model.py
```diff
@@ -388,6 +390,33 @@ def test_parfile_and_timfile_jumps(timfile_jumps):
...
+    # sort by MJD
+    tcopy = deepcopy(t)
```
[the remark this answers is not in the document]
dlakaplan: In fact, even when the residuals agree the chi^2 does not.  I can compare `residuals / uncertainties` for both Residuals objects and those agree, but chi^2 does not.  Still digging...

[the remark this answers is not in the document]
aarchiba: That sounds like there's some questionable mean(?) subtraction in the chi-squared calculation.

tests/test_timing_model.py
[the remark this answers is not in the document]
aarchiba: Looks like the mean subtraction is subtly order-dependent. To be fair, floating-point addition is subtly order-dependent, so that may be the cause.

tests/test_timing_model.py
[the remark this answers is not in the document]
dlakaplan: I don't think so.  I think it's just
```
return ((self.time_resids / toa_errors.to(u.s)) ** 2.0).sum().value
```

dlakaplan:
I tried this 100 times with random shuffles of 10 TOAs.  I see that the chi^2 values sometimes agree, sometimes disagree by +/- 5e-14.  So I suspect that this may be a floating point issue?

tests/test_toa_indexing.py
```diff
@@ -85,7 +77,6 @@ def test_getitem_slice(c):
     s = toas[c]
     assert set(s.get_mjds()) == set(m[c])
     if len(s) > 0:
-        assert np.all(s.table["mjd_float"] == s.table.group_by("obs")["mjd_float"])
```
aarchiba, line 88: This test can be greatly simplified now that we preserve order - you can just check that `toas[c].get_mjds() == toas.get_mjds()[c]` and drop the whole rest of the testing code. Other indexing tests can be similarly modified to be both simpler and stricter.

tests/test_toa_indexing.py
```diff
@@ -27,9 +27,6 @@ def test_get_TOAs(high_precision):
     toas = get_TOAs(StringIO(tim), ephem="de421")
     m = toas.get_mjds(high_precision=high_precision)
     assert isinstance(m, np.ndarray)
-    assert not np.all(
-        np.diff(m) > 0
-    ), "returned values should be grouped by observatory"
```
aarchiba, line 32: This should become a test that indeed all returned values are increasing (assuming that's what's in the input par).

tests/test_toa_indexing.py
```diff
@@ -57,7 +51,6 @@ def test_getitem_boolean(c):
     assert len(s) == np.sum(c)
     assert np.all(s.get_mjds() == m[c])
     if len(s) > 0:
-        assert np.all(s.table["mjd_float"] == s.table.group_by("obs")["mjd_float"])
```
aarchiba, line 60: I think this test differs from the previous in that it checks that goruping is preserved, so it can go.

tests/test_timing_model.py
```diff
@@ -385,6 +387,56 @@ def test_parfile_and_timfile_jumps(timfile_jumps):
...
+    assert np.isclose(r.calc_chi2(), rsort.calc_chi2(), atol=1e-14)
+
+
+def test_shuffle_toas_residuals_match(timfile_nojumps):
+    m = get_model(os.path.join(datadir, "NGC6440E.par"))
+    t = timfile_nojumps
+    r = pint.residuals.Residuals(t, m, subtract_mean=False)
+    tcopy = deepcopy(t)
+    i = np.arange(len(t))
+    np.random.shuffle(i)
```
aarchiba, line 421: Random shuffles are liable to cause a problem with tests that fail once and then not the next time. You can use `hypothesis`: `@given(permutations(np.arange(n)))`. This explores the space of permutations and finds the simplest one that fails the test.

dlakaplan: ok, good to know.  but I can't quite figure out the syntax to define n properly using a fixture or class-based method.  I just pushed up revised tests done in a single class but I had to define n using a global.  is there a better way?  when I tried to define it as a fixture there were various syntax errors.

aarchiba: The magic sauce is that `n` is supposed to be the length of the TOAs object, so you need to make or pick the TOAs object within hypothesis. I think `flatmap` is the right way to do this: https://hypothesis.readthedocs.io/en/latest/data.html#chaining-strategies-together bit I'd find it easier to use `@composite`:

```python
@composite
def toas_and_order(draw):
    toas_name = draw(select_from(["file1.tim", "file2.tim"]))
    toas = get_TOAs(toas_name)
    n = len(toas)
    ix = draw(permutations(np.arange(n)))
    return toas, ix
```
then you'd use `@given(toas_and_order())` to get that tuple inside your function.

I think there are examples of `@composite` doing this sort of thing elsewhere in our test suite.

dlakaplan: That is handy.  I reworked it a bit so that it loads the data initially through a data class (unofficial) to speed things up, but I think this now does what it should

dlakaplan:
@scottransom : I verified that I can delete observations correctly from the Ter5 data you sent.  that seemed to be a place where grouping was an issue. I also changed "group" -> "cluster" the documentation and code about the "v" key, and put in a default `get_cluster()` call to make that work.  I removed other `pintk` group discussion.  Are there other aspects to test?

dlakaplan:
I'm not sure how to test deleting stashed TOAs, since they aren't visible.  how does that work?

scottransom:
I was just talking about the "deleting" that happens in the TOA table that the user sees when they stash.  That is one of the only places where a group_by() is explicitly used in the pintk code:

```
                # Delete the selected points
                self.psr.stashed = copy.deepcopy(self.psr.all_toas)
                self.psr.all_toas.table = self.psr.all_toas.table[
                    ~self.selected
                ].group_by("obs")
```

And your code already gets rid of that `group_by`, so if stashing/unstashing works, you should be good.

dlakaplan:
OK, yes.  Can stash/unstash.

tests/test_timing_model.py
```diff
@@ -385,6 +394,60 @@ def test_parfile_and_timfile_jumps(timfile_jumps):
     assert fs[4] == "1"
 
 
+class TOAOrderSetup:
+    parfile = os.path.join(datadir, "NGC6440E.par")
+    timfile = os.path.join(datadir, "NGC6440E.tim")
```
aarchiba, line 399: Is this a single-telescope file? I think it's just GBT data? It should be fine but it would be nice to have a mixed-telescope data set to ensure we catch weirdness with treating things differently by telescope.

dlakaplan: I agree in principle, but the existing test data don't have lots of multi-telescope data and those that are there (1713) are very large.  I could possibly put in a new one.

dlakaplan: OK, have a new test `test_toa_shuffle.py` which does the same routines but with multi-telescope multi-frequency simulated data.

tests/test_timing_model.py
```diff
@@ -42,6 +47,9 @@ def timfile_nojumps():
     return get_TOAs(os.path.join(datadir, "NGC6440E.tim"))
 
 
+len_timfile_nojumps = len(get_TOAs(os.path.join(datadir, "NGC6440E.tim")))
```
aarchiba, line 50: Can we avoid reloading this file, or save time?

tests/test_toa_shuffle.py
```diff
@@ -0,0 +1,88 @@
...
+        simulation.make_fake_toas_uniform(
+            55020, 55500, 50, model=model, freq=2000 * u.MHz, obs="@"
+        ),
+    ]
+    f = io.StringIO()
+    for t in fakes:
+        t.write_TOA_file(f)
+    f.seek(0)
+    t = toa.get_TOAs(f)
+    r = pint.residuals.Residuals(t, model, subtract_mean=False)
```
aarchiba, line 42: Is this ever used?

aarchiba: Oh, I see that you can access this directly from anywhere.

tests/test_toa_shuffle.py
```diff
@@ -0,0 +1,88 @@
...
+        ),
+        simulation.make_fake_toas_uniform(
+            55020, 55500, 50, model=model, freq=2000 * u.MHz, obs="@"
+        ),
+    ]
+    f = io.StringIO()
+    for t in fakes:
+        t.write_TOA_file(f)
+    f.seek(0)
+    t = toa.get_TOAs(f)
```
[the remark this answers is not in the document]
dlakaplan: I didn't directly do this test, but I did extensively test the clock corrections in the old version and without the grouping.  Prior to a the reordering discussed above they were identical.  Now they are very close (just due to when the clock data are interpolated).  But I think I could put in another test

dlakaplan:
I added a test to shuffle the lines of a tim file before loading (perhaps in a clumsy way) and then verify the clock corrections compared to the non-shuffled version.

tests/test_toa_shuffle.py
```diff
@@ -0,0 +1,124 @@
...
+    timfile = os.path.join(datadir, "shuffletest.tim")
+    t = toa.get_TOAs(timfile)
+    with open(timfile) as f:
+        lines = f.readlines()
+    preamble = lines[0]
+    datalines = np.array(
+        [
+            x
+            for x in lines[1:]
+            if not (x.startswith("C") or x.startswith("#") or len(x.strip()) == 0)
```
aarchiba, line 101: Why do this? The shuffle should still work, shouldn't it?

aarchiba: Ah, I see. Makes the TOA permutation the same as the line permutation.

tests/datafile/shuffletest.tim
```diff
@@ -0,0 +1,123 @@
+FORMAT 1
+#J0523-7125_PSMC.rf.f4T 1382.000000 55415.829511131335920 4407.496  pks
```
aarchiba, line 2: Is there any reason this can't be used as an example file with multiple telescopes for other test code? The name suggests a more specific nature than we necessarily need. If it's really only good for the one use, maybe trim it to three or so TOAs from each telescope and put it in a triple-quoted string?

dlakaplan: It's from an unpublished project, and I didn't want to include the par file (which might have been helpful for other tests).  Ultimately for this test the TOAs don't really matter so I could just make up numbers and put it as a string - I was just exploring the testing possibilities.

dlakaplan: I changed it to fake data.

dlakaplan:
@aarchiba @scottransom : do you think this can be merged?

scottransom:
I certainly do.  Would like to hear from Anne first, though.

tests/test_toa_shuffle.py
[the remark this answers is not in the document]
dlakaplan: that is now done.

aarchiba:
Looks good to me.

PR #1272, opened 2022-06-06
Add demos of how to write a par file.

### LINKED ISSUE #1268: Not obvious enough how to write out a par file
A user reported:

> I decided to start from scratch in terms of trying to time a NANOGrav pulsar in PINT and while I've done a bunch of things before, I realized that the biggest hiccup was writing out a new par file. I got there through various digging through the code (knowing that pintk could do this too), and it's not long once you know what you're doing, but I recommend this be added somewhere in the tutorial! There is an example in the paper validation but it relies on tempo_utils which makes the whole thing moot.

We should ensure that the tutorials demonstrate writing out par files (possibly demonstrating our compatibility modes).

aarchiba:
Also added some testing that "foreign" formats can be correctly read.

Closes #1268


dlakaplan:
I think the new examples are good.  

But as smaller additions, would it make sense to add a little more explicit info in the docstring for `TimingModel` (it says you can use `as_parfile`, but maybe actually show how)  and or in the docstring for `as_parfile`?  Another possibility would be to add `to_parfile()` which would just wrap `as_parfile` with the required open/write commands: probably not needed, but could be a useful convenience for some.

dlakaplan:
Sure - whichever is easier.  I don't know a reason that one or the other should be the main one.

aarchiba:
Curses, the black version seems to have bit me.

dlakaplan:
In the docstring for `write_parfile`, is it worth saying that it uses `as_parfile`?  And vice-versa: in the docstring for `as_parfile` say that if you want to write to a file use `write_parfile`?

aarchiba:
> In the docstring for `write_parfile`, is it worth saying that it uses `as_parfile`? And vice-versa: in the docstring for `as_parfile` say that if you want to write to a file use `write_parfile`?

I think which one uses the other is an implementation detail, but it's probably worth each mentioning the other.

dlakaplan:
I can merge if you confirm that this is ready (I suspect it is).  Or feel free to merge yourself: I have looked at this enough

Issue #1277, opened 2022-06-09
Test suite emits uncaught IERS DEBUG message


aarchiba:
Normally `pytest`  should capture all warning messages, but (perhaps because of `loguru`) it emits a (colourful) message partway through the test suite. This message also emerges many other unexpected places (for example in a script that does no log configuration).
```
$ make test
pytest -n auto --ff --durations=20
===================================== test session starts ======================================
platform linux -- Python 3.8.0, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/naa280/projects/pint/PINT
plugins: forked-1.3.0, cov-3.0.0, xdist-2.3.0, hypothesis-6.46.2, anyio-3.2.1
gw0 [1462] / gw1 [1462] / gw2 [1462] / gw3 [1462] / gw4 ok / gw5 [1462] / gw6 [1462] / gw7 [1462gw0 [1462] / gw1 [1462] / gw2 [1462] / gw3 [1462] / gw4 [1462] / gw5 [1462] / gw6 [1462] / gw7 [1462]
..............s.....................s................................................... [  6%]
.....................s.....s.....s.s.................................................... [ 12%]
.........................X...................................s.......................... [ 18%]
................................s..........s.........ss.......x.xs..2022-06-09 14:54:23.189 | DEBUG    | pint.erfautils:<module>:71 - Running get_iers_up_to_date() to update IERS B table, and checking for updated leap seconds.
..x.s..Xs..s...X...X [ 24%]
```

aarchiba:
This appears in the tutorials even with the recommended logging setup: https://nanograv-pint.readthedocs.io/en/latest/examples/time_a_pulsar.html

dlakaplan:
I don't know why pytest doesn't capture this message.  But I think the reason it's seen elsewhere is that it's emitted when importing `erfautils`, which is before any user-specific log setup has been done.  So it just uses the default `loguru` setup.   Ways to fix that could include:
* changing the level from `DEBUG` to `TRACE` or something similar so that it's not seen with the default setup
* removing the logging statement altogether (is it actually useful?)
* moving the relevant lines to somewhere that doesn't always run on import but happens later, after logging setup.  It seems that having some functions called every time on import isn't optimal, but maybe there is a reason for that.

aarchiba:
I think if we can move away from code being run on import we should, it's asking for just this kind of trouble.

dlakaplan:
[...]

PR #1281, opened 2022-06-14
Remove explicit updating of IERSB/leapsecond tables

dlakaplan:
This moves `erfautils/get_iers_up_to_date()` and `astropy.time.update_leap_seconds()` which had been run on import of `erfautils` to `topo_obs`, when a new `TopoObs` is created.  They are only run once (this uses a global variable which mimics other behavior in there but could probably be better).  The debug statement about those calls no longer appears when running tests (#1277).


dlakaplan:
Currently these calls are not made for non-TopoObs locations.  It would be possible to add them, but are they needed (@paulray or others)?

paulray:
Well, the `GeocenterObs` uses UTC as the timescale, so it would need leap seconds. And it calls `EarthLocation.from_geocentric()` so it likely needs IERS (even though all the coords are 0.0 so no rotation is really needed). So maybe move the init to `Observatory` class rather than just the `TopoObs` subclass? They may be loaded for some observatories where they aren't strictly needed, but that was already happening.

dlakaplan:
I have done so.  e.g., 
```
In [1]: import pint.observatory

In [2]: pint.observatory.Observatory.get("Geocenter")
2022-06-14 12:35:16.727 | DEBUG    | pint.observatory:_load_iersb:322 - Running get_iers_up_to_date() to update IERS B table
2022-06-14 12:35:17.276 | DEBUG    | pint.observatory:_load_leapsecond:330 - Checking for updated leapseconds
```

aarchiba:
The call takes an MJD argument. The IERS table is at least potentially used any time you work with a Time object. Can we tidy this a little?

- If the function is cheap and does nothing when the IERS B table is up to date (as measured by its MJD argument), then we can call it anywhere it seems useful.
- We probably need it whenever someone loads in TOAs - but only for the latest MJD in the TOAs.
- We possibly also need it whenever an observatory does a clock correction (but only for the latest MJD in the correction) but catching TOAs on load (or on construction) will cover most use cases. 

I think if the function can be made harmless when the file is up to date (like, don't unconditionally zap the IERS_Auto thing) then it makes sense to ensure it is called whenever Time objects are made for TOAs. If you also want to call it when an observatory clock correction happens, that would be good too.

I would like to suggest *not* calling it without a meaningful MJD argument: if you have an IERS B table that covers your MJDs, it shouldn't go looking at the Net just because it's 45 days since it was downloaded. So a good guideline for where to call it from is, where do you have a relevant MJD?

paulray:
> would like to suggest _not_ calling it without a meaningful MJD argument: if you have an IERS B table that covers your MJDs, it shouldn't go looking at the Net just because it's 45 days since it was downloaded. So a good guideline for where to call it from is, where do you have a relevant MJD?

Concur. If you run on the same TOAs over and over, it shouldn't suddenly decide to update IERS tables. That way if you run a notebook once, then get offline, you should have no problems with it crashing trying to download.

dlakaplan:
* It does use astropy caching to store the file.  So it shouldn't re-download over and over for the same date.   
* it isn't as cheap (currently) as it should be.

dlakaplan:
I'm not actually sure if this is working how we want it to.  It says that we download IERS-B, and use it for future transformation.  I could understand why we want IERS-B (it's "final") but when we do:
```
# Now open IERS_Auto with no argument, so it should use the IERS_B that we just made sure was up to date
    iers_auto = IERS_Auto.open()

    if astropy.version.major >= 4:
        # Tell astropy to use this table for all future transformations
        earth_orientation_table.set(iers_auto)
```
I think it just uses the IERS-A.  Does anybody know for certain?  Do we actually need IERS-B?  If we only need IERS-A then could we just rely purely on the astropy caching (which already includes things like a max age for updating files, ...)?

dlakaplan:
If that isn't sufficient then I think we need to change it to do `earth_orientation_table.set(iers_b)`

dlakaplan:
OK, I'm pretty sure that the IERS-A isn't used, and we just rely on the standard astropy IERS_Auto download/caching.  So I've removed references to IERS-A download.  There are still some mentions of IERS-A in the `TOA` class docstrings, but I'm guessing some of that is deprecated or no longer works as intended.  I ran TOA creation with the example there that is supposed to raise an error but it doesn't:
```
Some random time in the past: 54567.8768768768768759:  4.500 us error at 'gbt' at 1400.0000 MHz {'backend': 'GUPPI'}
Now: 59746.5171607367356310:  4.500 us error at 'gbt' at 1400.0000 MHz {'backend': 'GUPPI'}
6 months from now: 59926.5171607412030690:  4.500 us error at 'gbt' at 1400.0000 MHz {'backend': 'GUPPI'}
18 months from now: 60286.5171607450374986:  4.500 us error at 'gbt' at 1400.0000 MHz {'backend': 'GUPPI'}
/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.9/site-packages/erfa/core.py:154: ErfaWarning: ERFA function "dtf2d" yielded 1 of "dubious year (Note 6)"
  warnings.warn('ERFA function "{}" yielded {}'.format(func_name, wmsg),
Random time in the future: 154567.8768768768768759:  4.500 us error at 'gbt' at 1400.0000 MHz {'backend': 'GUPPI'}
```
just an ERFA warning.

aarchiba:
As of Astropy 5.1, it looks like it does the right thing for IERS_Auto:

- A version is downloaded if nothing is available
- A version is downloaded if an MJD is requested from the predictive range (or past the end I assume?) *and* the file is more than 30 days old
- You can specify what to do for out-of-range values with `astropy.utils.iers.conf.iers_degraded_accuracy` - `warn`, `ignore`, or, presumably, raise an exception

https://docs.astropy.org/en/stable/utils/iers.html

Can we make Astropy behaviour reflect `limits=` with their temporary configuration setter `set_temp`? https://docs.astropy.org/en/stable/api/astropy.config.ConfigNamespace.html#astropy.config.ConfigNamespace.set_temp

If Astropy already does all this for us, what does the function here do? Do we still need this at all?

dlakaplan:
Yes, I suspect we can eliminate it

dlakaplan:
I am trying to test this.  With astropy 5.1 I am trying:
```
with astropy.utils.iers.conf.set_temp('iers_degraded_accuracy', 'error'):
    astropy.time.Time("2029-06-15",scale='utc').ut1
```
which I think should give an error since it's asking for UT->UT1 in 2029.  But I only get 2 ERFA warnings.  This is with nothing PINT-related installed.  Do others find this?

dlakaplan:
Note that I was able to trigger errors/warnings when I set `iers.conf.auto_download = False`.  That may be the expected behavior - I am finding the docs a little confusing, but the error/warning may only apply when using the bundled IERS-B and not the auto-downloaded versions.

aarchiba:
With Astropy 5.0.4 the parameter doesn't exist at all. Maybe you found a bug?

dlakaplan:
I upgraded to astropy 5.1 to check (I had been at 5.0.4).  I can try looking at their code later.

dlakaplan:
I think this is working as intended (at least by astropy):
https://github.com/astropy/astropy/blob/main/astropy/utils/iers/iers.py#L368
For the generic `IERS` class it issues a warning, but in the `IERS_Auto` that warning never comes:
https://github.com/astropy/astropy/blob/main/astropy/utils/iers/iers.py#L741
It may issue an error if the requested time is in the future *and* the latest download is old (>30 days by default), but as long as the download is current it just keeps on interpolating.

dlakaplan:
I think that since the Auto procedure will keep the file current and any extrapolation will only involve future (simulated) data, then warnings may not be needed here.  Which I think is also what you are saying.

src/pint/erfautils.py
```diff
@@ -88,114 +34,6 @@ def get_iers_b_up_to_date(mjd):
 asec2rad = 4.84813681109536e-06
 
 
-def old_gcrs_posvel_from_itrf(loc, toas, obsname="obs"):
```
aarchiba, line 91: This should also come out of `__all__` if it's going away.

src/pint/observatory/__init__.py
```diff
@@ -18,6 +19,8 @@
 from pint.pulsar_mjd import Time
 from pint.utils import interesting_lines
 
+# has the lead-second file been downloaded?
+_leapsecond_download = False
```
aarchiba, line 23: Is this needed?

src/pint/observatory/__init__.py
```diff
@@ -147,6 +150,9 @@ def get(cls, name):
         if name == "":
             raise KeyError("No observatory name or code provided")
 
+        # and download (if necessary) leapsecond files
+        cls._load_leapsecond()
```
aarchiba, line 154: Is this needed?

src/pint/observatory/__init__.py
```diff
@@ -305,6 +311,14 @@ def _get_TDB_ephem(self, t, ephem):
         """
         raise NotImplementedError
 
+    @classmethod
+    def _load_leapsecond(self):
```
aarchiba, line 315: Is this needed?

src/pint/observatory/__init__.py
dlakaplan: I don't know.  I suspect not, but it's pretty harmless (in terms of runtime) so I left it in.

dlakaplan:
I suspect that the leapsecond updating isn't needed either, although that's harder to test since the updates are so infrequent.  But I removed the explicit updates assuming that they will be handled by astropy

src/pint/observatory/topo_obs.py
```diff
@@ -5,6 +5,7 @@
 import astropy.units as u
 import numpy as np
 from astropy.coordinates import EarthLocation
+import astropy.time
```
aarchiba, line 8: Is this import used anywhere?

src/pint/observatory/__init__.py
```diff
@@ -10,6 +10,7 @@
 import astropy.constants as const
 import astropy.coordinates
 import astropy.units as u
+import astropy.time
```
aarchiba, line 13: Is this import used anywhere?

src/pint/observatory/topo_obs.py
dlakaplan: autoflake & remove

src/pint/observatory/__init__.py
dlakaplan: autoflake & remove

aarchiba:
Looks good to me! Feel free to merge when CI passes.

PR #1282, opened 2022-06-15
Bump required python to 3.8

### LINKED ISSUE #1278: PINT should update its minimum acceptable version of python to 3.8
Current versions of `numpy` and `astropy` require python 3.8, and perhaps we should as well. Python 3.8 was released 2019-10-14 and is receiving only security patches. Python 3.6 has reached end-of-life and the last security release was on 2021-09-04.

The primary advantage of changing the dependencies is that users will be able to use recent versions of our dependencies. Other minor language features we might benefit from:

- Dataclasses. These let you easily create classes without any of the boilerplate for `repr` and comparison and `__init__` - if those classes are primarily about storing data (like C structs). You can add behaviour as well, but the machinery doesn't help much with that.
- f-strings support the `=` format: `raise ValueError("Something went wrong for {file=}")` yields a message that includes the variable name, as `Something went wrong for file=/dev/null`.

That said, Ubuntu 18.04 comes with python3.6 as its default Python interpreter, and I know from personal experience that many scientific machines are still on this LTS release. It does have python 3.8 as an installable package, and `python3.8 -m venv` allows you to make virtualenvs in which this is the default python interpreter.

aarchiba:
I'm not sure this covers everything required, but:

- Add a python version requirement to `setup.cfg` (there wasn't one)
- Change CI to not run any pre-3.8 jobs (and add a 3.10 job)
- Change tox.ini to refer only to 3.8, 3.9, and 3.10
- Switch readthedocs to use 3.8
- Changelog
- PyPI picks up version requirements from setup.cfg

Also probably required but not done yet:

- Update Conda metadata (https://github.com/conda-forge/pint-pulsar-feedstock/blob/main/recipe/meta.yaml)
- Update versions of dependencies (astropy, numpy)? 

Closes #1278

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -24,31 +24,31 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            python: '3.8'
+            python: '3.10'
             tox_env: 'ephemeris_connection'
           - os: ubuntu-latest
-            python: '3.8'
+            python: '3.10'
             tox_env: 'black'
           - os: ubuntu-latest
-            python: '3.8'
-            tox_env: 'py38-test-cov'
+            python: '3.10'
+            tox_env: 'py310-test-cov'
           - os: ubuntu-latest
-            python: '3.8'
+            python: '3.10'
             tox_env: 'notebooks'
 #          - os: ubuntu-latest
 #            python: '3.8'
 #            tox_env: 'docs'
-          - os: ubuntu-latest
-            python: '3.8'
-            tox_env: 'py38-test-alldeps-cov'
+#          - os: ubuntu-latest
+#            python: '3.10'
+#            tox_env: 'py310-test-alldeps-cov'
           - os: macos-latest
-            python: '3.7'
-            tox_env: 'py37-test'
+            python: '3.10'
+            tox_env: 'py310-test'
 #          - os: windows-latest
 #            python: '3.8'
 #            tox_env: 'py38-test'
           - os: ubuntu-latest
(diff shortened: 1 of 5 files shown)


paulray:
PyPI seems to get all metadata from setup.cfg, so that should work fine.

aarchiba:
> Conda probably needs an update in the yaml file here: https://github.com/conda-forge/pint-pulsar-feedstock/blob/main/recipe/meta.yaml

Do I have access to that? It might need to be you to do that update...

aarchiba:
I think updating astropy or numpy versions aren't really necessary unless we're relying on something they use.

paulray:
> > Conda probably needs an update in the yaml file here: https://github.com/conda-forge/pint-pulsar-feedstock/blob/main/recipe/meta.yaml
> 
> Do I have access to that? It might need to be you to do that update...

At the moment I think only I (and Joe Glaser?) are project admins. I'd be happy to add you! But, I can also just do this update if I can figure out what needs changing.

paulray:
> I think updating astropy or numpy versions aren't really necessary unless we're relying on something they use.

Right, as long as our test on the minimum versions of all the packages keeps passing, we can keep them.

paulray:
This PR should fix the conda version: https://github.com/conda-forge/pint-pulsar-feedstock/pull/16

aarchiba:
> Let's merge this at telecon today

Sounds good to me. (I'd better fire off an email requesting python3.8 be installed on my work desktop. Why it's running 18.04 in 2022 I don't know.)

PR #1284, opened 2022-06-16
Multiprocessing functionality for event_optimize

devbhakt:
This sets up event_optimize to run on multiple cores using `pathos` and to use the blobs feature from `emcee` to keep track of the metadata (lnprior and lnlikelihood) within the sampler.


src/pint/scripts/event_optimize.py
```diff
@@ -519,6 +519,12 @@ def main(argv=None):
         help="Logging level",
         dest="loglevel",
     )
+    parser.add_argument(
+        "--numcores",
+        type=int,
+        default=1.0,
```
aarchiba, line 525: Python's `multiprocessing` module has a way to query the number of cores, it would be good if it were easy to tell this script "use all my cores, however many there are". Perhaps that should be the default?

src/pint/scripts/event_optimize.py
```diff
@@ -759,16 +766,17 @@ def main(argv=None):
...
+    if numcores != 1:
 
-        def unwrapped_lnpost(theta, ftr=ftr):
+        def unwrapped_lnpost(theta):
             return ftr.lnposterior(theta)
 
         import pathos.multiprocessing as mp
 
-        pool = mp.ProcessPool(nodes=8)
+        pool = mp.ProcessPool(nodes=numcores)
```
aarchiba, line 776: If possible it is better to use `ProcessPool` in a `with` statement to ensure that everything gets shut down correctly even if an exception occurs.

src/pint/scripts/event_optimize.py
devbhakt: I was not sure on whether it would be better to let the user specify the number or just run it on all cores by default. I can set it to query and use all cores. If so, should I remove the numcores parser argument?

aarchiba: Definitely not. But I would suggest either:

* Default to all cores, but let people specify a smaller number, or
* Default to one core but allow people to specify "all" or "auto" to get all cores.

devbhakt: Should be updated now to default to all cores.

src/pint/scripts/event_optimize.py
devbhakt: Should be updated now.

aarchiba: Uh, have you tried it? I don't think it will work - you need the actual chain generation to be within the `with` statement, as soon as you exit that all the cores shut down. Sorry if the actual computation happens somewhere different than I thought...

aarchiba:
I know we don't really have a good way to test this as part of our CI, so maybe it would be good to have a little test script you can run? There is a test that runs event_optimize in there somewhere, if you wanted to write something with a really tiny number of iterations...

src/pint/scripts/event_optimize.py
devbhakt: I tried it with the chain generation inside and outside the `with` statement and both appeared to work. I have changed it so that the chain generation is within the `with` statement now

aarchiba:
It looks like you have used a module, `pathos`, that is not currently a dependency of PINT. Can you explain why you want to add this dependency? How does it differ from Python's built-in `multiprocessing`?

devbhakt:
If we try and use Python's built-in `multiprocessing`, we run into the issue that our log posterior function is not pickle-able. `pathos` uses `dill` to commute between the processes and is able to transfer the log posterior function across the processes.

scottransom:
The pickling issue has been a very long-term one.  The `emcee_fitter` class, which contains both the timing mode, the TOAs, and custom methods that are needed for emcee, does not (or at least, did not) pickle.  But it does work with `dill`.  This has been a known issue with emcee (really with multiprocessing) and complex class structures for a long time (try googling "emcee multiprocessing pickle dill").  I'm not sure exactly (i.e. technically) *why* we can't pickle things -- and given all the changes in PINT, maybe it works now.  We should try it again.  But assuming this is still an issue, I don't see a real problem with requiring dill and maybe even pathos for this type of work.

scottransom:
I think the heart of the matter is that `emcee` can't pickle instance methods, which is how we compute the log posterior.  But this works with dill and pathos.  https://stackoverflow.com/questions/29320466/using-multiprocessing-in-emcee-library-inside-a-class

devbhakt:
I think that the weird showing the previous commits might be due to me having to do the forced push while using `git rebase`. https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch. 

However, comparing this branch to `nanograv master` shows only the 4 files that I have changed for the PR. Sorry for the mishap.

aarchiba:
Something is wrong - your PR is trying to make a whole lot of changes. Rebasing does not normally do this - I regularly rebase against master before merging. I'm not sure what happened or what the right way is to clean it up, but `git rebase -i` may well let you drop the changes that aren't actually meant to be part of this PR.

scottransom:
@aarchiba The strange thing is that it really *isn't* trying to make all those changes.  That's why Deven mentioned the strangeness above.  I used the command-line options to try merging this locally and even though github says that 50-some files need changing, it only changed the proper 3-4 of them that needed it since all those changes are already in the repo.  It is really strange.

aarchiba:
> @aarchiba The strange thing is that it really _isn't_ trying to make all those changes. That's why Deven mentioned the strangeness above. I used the command-line options to try merging this locally and even though github says that 50-some files need changing, it only changed the proper 3-4 of them that needed it since all those changes are already in the repo. It is really strange.

It includes many changes that are already in `master`, and fortunately `git` is smart enough when merging to recognize an already-applied diff. Perhaps some changes from master were merged in and then the whole thing was rebased, pulling those changes forward?

devbhakt:
@aarchiba I think that might be what happened. I think I've corrected the problem now. Apologies for the error.

aarchiba:
This script already has a conditional import of `corner` - if it's available you get a corner plot but it's not a dependency. So one could write this so you get multiprocessing if `pathos` is available and not if it isn't. Or we could make PINT depend on `pathos` (or `cloudpickle`?). 

Adding a dependency definitely calls for a discussion. Since there are many approaches to parallel processing, one approach is for packages like `emcee` to be as agnostic as possible by simply accepting a `pool` argument from the user's favourite multiprocessing tool and asking as little of it as possible. It might be worth pursuing something like this for most of PINT - currently the emcee fitter asks for instance methods to be pickleable, which is awkward - but that doesn't address things like this script.

devbhakt:
@aarchiba I have updated the code so that `event_optimize` is set to default single core with a `multicores` parser argument that sets to all core. The `pathos` implementation is now conditional.

aarchiba:
Is it possible to run this for a tiny number of steps with random seeds to verify that using the parallel version doesn't change the samples generated?

devbhakt:
@aarchiba I can modify the existing `test_event_optimize.py` to compare two runs, single core vs. multicore.

aarchiba:
I would recommend creating a new test in `test_event_optimize`, named after what property you are trying to test.

tests/test_event_optimize.py
```diff
@@ -36,3 +39,48 @@ def test_result(tmp_path):
...
+        cmd = f"{eventfile} {parfile} {temfile} --weightcol=PSRJ0030+0451 --minWeight=0.9 --nwalkers=10 --nsteps=50 --burnin 10 --multicores"
+        event_optimize.main(cmd.split())
+        with open("J0030+0451_samples.pickle", "rb") as f:
+            samples2 = pickle.load(f)
+        f.close()
+
+        for i in range(0, samples1.shape[1]):
+            assert stats.ks_2samp(samples1[:, i], samples2[:, i])[1] == 1.0
+    except ImportError:
+        pass
```
aarchiba, line 83: `pytest.skip` would be appropriate here - indicate that the test is being skipped because `pathos` is not available.

tests/test_event_optimize.py
```diff
@@ -36,3 +39,48 @@ def test_result(tmp_path):
...
+        event_optimize.maxpost = -9e99
+        event_optimize.numcalls = 0
+
+        cmd = f"{eventfile} {parfile} {temfile} --weightcol=PSRJ0030+0451 --minWeight=0.9 --nwalkers=10 --nsteps=50 --burnin 10 --multicores"
+        event_optimize.main(cmd.split())
+        with open("J0030+0451_samples.pickle", "rb") as f:
+            samples2 = pickle.load(f)
+        f.close()
+
+        for i in range(0, samples1.shape[1]):
```
aarchiba, line 80: `range(0, thing)` is the same as `range(thing)`

tests/test_event_optimize.py
```diff
@@ -36,3 +39,48 @@ def test_result(tmp_path):
...
+            assert stats.ks_2samp(samples1[:, i], samples2[:, i])[1] == 1.0
```
aarchiba, line 81: Why should this pass for correct code? Why should it fail for wrong code?

If the parallel version really follows the same algorithm, then providing the same random seed should ensure that you get exactly the same numbers. (Does `event_optimize` allow you to specify a random seed? It really should.)

aarchiba:
We should ensure that at least some of the CI runs have `pathos` installed, so that this new test gets run on CI machines. This should be fixable in `tox.ini`, I think?

tests/test_event_optimize.py
scottransom: I think this (whether multiprocessing would give the same answer or not) completely depends on how `emcee` handles multiprocessing.  This is one of the reasons why these tests are a bit tricky.

PR #1287, opened 2022-06-16
JSON observatories

dlakaplan:
Moved observatory definitions to `data/runtime/observatories.json`.  Still read & parsed in `observatory/observatories.py`.  Initial version: no way yet to override the file location


dlakaplan:
Now the reading/registering of observatories is handled by `observatories.read_observatories()`.  So far this is called within `observatory/observatories.py`, with the default data coming from the PINT data directory and `$PINT_OBS_OVERRIDE` able to override that.  It can be called subsequent times with `overwrite=True` to update existing observatories.  

@aarchiba : I would be fine moving the `read_observatories()` call out of this file and putting it somewhere else.  This whole module could also likely be merged with `topo_obs`.   What do you think?  Where/how should the file be read?  Should I wait to make changes until you have your clock correction PR merged?

dlakaplan:
A further possibility would be to move everything in `topo_obs.py` **and** `observatories.py` to `__init__.py`.  Do they need to be in sub-modules?  Wouldn't just `import pint.observatory` be OK?

aarchiba:
> A further possibility would be to move everything in `topo_obs.py` **and** `observatories.py` to `__init__.py`. Do they need to be in sub-modules? Wouldn't just `import pint.observatory` be OK?

The general advice is to keep code out of `__init__.py` files as much as possible - it's not a very informative file name, and it gets imported every time you import any of the submodules. But maybe ditching `observatories.py` and putting the function in `topo_obs.py` would be a good option? `topo_obs.py` has some real heft to it, all the machinery having to do with observatories that need clock corrections goes there. 

I should say you are liable to have tons of merge conflicts with the global clock corrections #1247 almost no matter how this one works. Life should get easier once it's merged.

dlakaplan:
Yes, and should also merge in #1281 first.  

I think from a naming perspective `observatories.py` is more obvious than `topo_obs.py`, but then less consistent with the spacecraft and special locations.  But I think we can definitely go with just one of them.

aarchiba:
Unfortunately both `topo_obs` and `TopoObs` are part of our public API and it's awkward to change them to something more intelligible. `pint.observatory.observatories` appears mostly in apologetic instructions that tell you to edit the file.

dlakaplan:
good point.

src/pint/observatory/observatories.py
```diff
@@ -1,970 +1,87 @@
-"""Definitions for standard observatories.
+"""Standard observatories read in from observatories.json
 
 These observatories are registered when this file is imported. As a result it
 cannot be imported until TopoObs has successfully been imported.
```
aarchiba, line 4: This warning isn't helpful - just `import topo_obs`.

src/pint/observatory/observatories.py
```diff
@@ -1,970 +1,87 @@
...
+from pint.observatory import bipm_default
+
+observatories_json = pint.config.runtimefile("observatories.json")
+
+pint_env_var = "PINT_OBS_OVERRIDE"
+
+__all__ = ["observatories_json", "read_observatories"]
+
+
+def get_default_value(entry, key, default=None):
```
aarchiba, line 21: `my_dict.get(key, default)` does this

dlakaplan: ok, great

src/pint/observatory/observatories.py
```diff
@@ -1,970 +1,87 @@
...
+
+    Returns
+    -------
+        value for assignment
+    """
+
+    return default if not key in entry else entry[key]
+
+
+# Default values for instantiating a new TopoObs object
```
aarchiba, line 38: Can you let the defaults in `TopoObs.__init__` handle this? Easiest is just not to pass unset values in `kwargs`.

src/pint/observatory/observatories.py
```diff
@@ -1,970 +1,87 @@
...
+def read_observatories(filename=observatories_json, overwrite=False):
+    """Read observatory definitions from JSON and create TopoObs objects, registering them
+
+    Set `overwrite` to ``True`` if you want to re-read a file with updated definitions
+
+    Parameters
+    ----------
+    filename : str, optional
+    overwrite : bool, optional
+        Whether a new instance of an existing observatory should overwrite the existing one.
```
aarchiba, line 65: What happens if `overwrite` is False?

src/pint/observatory/observatories.py
dlakaplan: probably.  I sort of went about it backwards, but that should be fine.

src/pint/observatory/observatories.py
```diff
@@ -1,970 +1,87 @@
...
+            )
+        if overwrite:
+            keywords["overwrite"] = True
+        # create the object, which will also register it
+        TopoObs(name=obsname, **keywords)
+
+
+# read the observaatories, with the potential to override the filename from the environment variable
+if pint_env_var in os.environ:
+    read_observatories(os.environ[pint_env_var])
```
aarchiba, line 85: I think it would be better (maybe?) to read the pint_env_var versions in after the default ones, with `overwrite=True`. Then someone can just create the one observatory they want to overwrite and not worry about replacing all the others.

src/pint/observatory/observatories.py
```diff
@@ -1,970 +1,87 @@
...
+    overwrite : bool, optional
+        Whether a new instance of an existing observatory should overwrite the existing one.
+
+
+    """
+    # read in the JSON file
+    observatories = json.load(open(filename))
+    for obsname in observatories:
+        keywords = {}
+        for keyword in default_keywords:
```
aarchiba, line 73: Can you skip this loop and just make `keywords` be (a copy of?) `observatories[obsname]`? Then the normal python call mechanism will handle unspecified entries by giving them the default arguments specified in the TopoObs constructor.

src/pint/observatory/observatories.py
```diff
@@ -1,970 +1,87 @@
...
+    ----------
+    filename : str, optional
+    overwrite : bool, optional
+        Whether a new instance of an existing observatory should overwrite the existing one.
+
+
+    """
+    # read in the JSON file
+    observatories = json.load(open(filename))
+    for obsname in observatories:
```
aarchiba, line 71: `for obsname, obsdict in observatories.items():` is more conventional and quicker.

src/pint/observatory/observatories.py
dlakaplan: An exception is raised (although not directly by this function).  Is it worth documenting that here or can it rely on the `TopoObs` documentation?

aarchiba: I think it's good to document it here, including the type of the exception (ValueError?) in  case someone wants to catch it and continue.

tests/test_observatory_envar.py
```diff
@@ -0,0 +1,55 @@
...
+        pass
+
+    o = Sandbox()
+    e = os.environ.copy()
+    try:
+        del os.environ["PINT_OBS_OVERRIDE"]
+    except KeyError:
+        pass
+
+    o.override_dir = tmp_path / "override"
```
aarchiba, line 18: You can probably just use o.override_dir=tmp_path, it's a fresh-generated empty directory.

tests/test_observatory_envar.py
```diff
@@ -0,0 +1,55 @@
...
+    finally:
+        os.environ = e
+
+
+def test_override_gbt(sandbox):
+    os.environ["PINT_OBS_OVERRIDE"] = str(sandbox.override_file)
+    import pint.observatory.observatories
+    from pint.observatory import get_observatory
+
+    newgbt = get_observatory("gbt")
```
aarchiba, line 54: Does this override the GBT coordinates for the rest of the test suite run? This is the downside of our using a global registry. It might be necessary to save and restore the registry in the sandbox fixture.

tests/test_observatory_envar.py
dlakaplan: I'm unsure of the right way to do this.  If I import `pint.observatory.observatories` in the sandbox creation (to save the registry), then when I run import again in the test it will recognize it as already imported and won't re-read it, correct?  On the other hand if I just import in the test function then I don't have a way to save the registry.

I guess I could ask if this particular test is necessary.  there are other tests that demonstrate overriding existing observatories.  the unique feature here is that it happens through the environment variable.  I agree it would be good to check that but it's causing me some trouble.

dlakaplan: OK, I think I figured it out.  I now have a test for the correct and incorrect (overridden) GBT in `test_observatory_envar.py`, and the correct GBT in the existing `test_observatory.py`.  The registry is preserved and there's also a method now to clear it.

aarchiba: This is why making stuff happen on import is a pain in the neck - the import actually happens once per test run, the first time the module is imported. It may make sense to have the module just call a function (no conditionals, nothing) and that function (which you can call from a test) do any checking of environment variables and multiple opening. Then testing that function is more feasible. It also opens a path to calling that function when the data is needed rather than on import, if we decide to do that.

tests/test_observatory.py
```diff
@@ -217,3 +220,38 @@ def test_observatories_registered():
...
+
+
+def test_is_gbt_still_ok():
+
+    gbt = get_observatory("gbt")
+    assert gbt._loc_itrf.y < 0
+
+
+@pytest.mark.parametrize("overwrite", [True, False])
+def test_observatory_override(overwrite):
```
aarchiba, line 232: This still appears to trash the GBT observatory coordinates. Perhaps a fixture that saves and restores the registry?

tests/test_observatory_envar.py
dlakaplan: Yes, I agree.  I was considering doing just that but wanted to wait for your clock PR to get merged first.

src/pint/observatory/observatories.py
```diff
@@ -1,970 +1,58 @@
...
-    name="northern_cross",
-    tempo_code="d",
-    itoa_code="BO",
-    itrf_xyz=[4461242.882451464, 919559.8351226494, 4449633.220012489],
-    origin="""Imported from TEMPO obsys.dat 2021 June 8.""",
-)
+# read the observatories
+read_observatories()
+# potentially override any defined here
+if pint_env_var in os.environ:
```
aarchiba, line 57: I think it makes sense to put this logic into a `read_observatories_from_usual_location()` function that (for the moment? or not?) gets called on import. Or, since we don't expect users to directly access the observatory registry without going through Observatory methods, just have them call the function without reference to importing anything? Same as the discussion for #1247, `Observatory.get` and `Observatory.names` should cover all the bases.

src/pint/observatory/__init__.py
```diff
@@ -110,11 +110,24 @@ def _add_aliases(cls, obs, aliases):
                     obs_aliases.append(alias)
             o._aliases = obs_aliases
 
+    @classmethod
+    def clear_registry(cls):
```
aarchiba, line 114: As long as the registry is populated on import, this is rather problematic as the user doesn't have an obvious way to repopulate the registry (using `importlib` to force a reload is decdedly non-obvious, particularly as you have to reload several modules). Maybe to build a test fixture it's better to just save the dictionary `cls._registry` and restore it?

dlakaplan: but it isn't populated only on import.  the `read_observatories()` function re-populates it.

src/pint/data/runtime/observatories.json
```diff
@@ -0,0 +1,1355 @@
...
+        "origin": "The Robert C. Byrd Green Bank Telescope.\nThis data was obtained by Joe Swiggum from Ryan Lynch in 2021 September.\n"
+    },
+    "gbt_pre_2021": {
+        "clock_file": "time_gbt.dat",
+        "itrf_xyz": [
+            882589.65,
+            -4924872.32,
+            3943729.348
+        ],
+        "origin": "The Robert C. Byrd Green Bank Telescope.\nThe origin of this data is unknown but as of 2021 June 8 it agrees exactly with\nthe values used by TEMPO and TEMPO2.\n"
```
aarchiba, line 20: This is meant to be a multiline string explaining the origin of the observatory position data. Sadly JSON appears to have no way to represent multiline strings other than one long string with `\n` in it many times. Does it make sense to write some code so that an array of strings can be supplied and then concatenated by the loading code? How human-readable/maintainable do we want this to be?

tests/test_observatory_envar.py
```diff
@@ -0,0 +1,84 @@
...
+import pytest
+import importlib
+
+import pint.observatory.observatories
+import pint.observatory.special_locations
+import pint.observatory
+
+
+@pytest.fixture
+def sandbox(tmp_path):
```
aarchiba, line 12: Why is this in a separate file? What is gained? Imports are still global across python invocations so this doesn't separate the registry or the imports from the non-envar version of this file.

src/pint/data/runtime/observatories.json
dlakaplan: I don't feel strongly either way.

tests/test_observatory_envar.py
dlakaplan: It was just for simplicity when I was testing.  if that violates some principle I can move it back

src/pint/data/runtime/observatories.json
dlakaplan: a related question is comments: JSON doesn't support them.  We could put in a `comment=` tag but in most cases those were all replicated in the `origin` tags so they didn't seem necessary.

dlakaplan:
I have now moved the parsing of the JSON and the functions needed to import it to `observatory/topo_obs.py` (I haven't deleted `observatories.py`, but that would be the intent.  And I made functions:
* `observatory/topo_obs/load_observatories_from_usual_locations()` 
* `observatory/special_locations/load_special_locations()` 
which do the loading explicitly (not just on import).  Not entirely convinced about the naming but I think this works better.

We can probably remove the "register-via-import" requirements and move to register-via-functions, but can discuss that later.

dlakaplan:
I guess the advantage to "register-via-import" is that it automatically avoids duplicating the calls.  So that the `import` call in `Observatory.names()` won't do anything extra and won't give warnings about existing observatories.

src/pint/observatory/__init__.py
```diff
@@ -415,7 +431,7 @@ def compare_t2_observatories_dat(t2dir=None):
     """Read a tempo2 observatories.dat file and compare with PINT
 
     Produces a report including lines that can be added to PINT's
-    observatories.py to add any observatories unknown to PINT.
+    observatories.json to add any observatories unknown to PINT.
```
aarchiba, line 434: Should this be tweaked to output JSON instead of python?

src/pint/observatory/special_locations.py
```diff
@@ -19,6 +25,8 @@
 
 from . import Observatory
 
+__all__ = ["load_special_locations"]
```
aarchiba, line 28: Should this include the classes? If they're meant for people to subclass... in an ideal world the doc generator would only make documentation for what's in `__all__` but I couldn't make ours do that.

src/pint/observatory/special_locations.py
```diff
@@ -2,6 +2,12 @@
 
 Special "site" locations (eg, barycenter) which do not need clock
 corrections or much else done.
+
+Can be loaded using :func:`pint.observatory.special_locations.load_special_locations`
```
aarchiba, line 6: Maybe clarify that this is only necessary if you've cleared the registry?

src/pint/observatory/observatories.py
```diff
@@ -1,977 +1,86 @@
-"""Definitions for standard observatories.
+"""Standard observatories read in from observatories.json
```
aarchiba, line 1: Should this whole file be gone?

tests/test_observatory.py
```diff
@@ -217,3 +220,38 @@ def test_observatories_registered():
 
 def test_gbt_registered():
     get_observatory("gbt")
+
+
+def test_is_gbt_still_ok():
```
aarchiba, line 225: The pytest plugins pytest-randomize and pytest-reverse could be handy to make this a tougher check (give it a chance to get run after any test that might cause a problem) - not that we should add them to `requirements_dev` but just for your own reassurance. Or if all the tests that might mess up the GBT are in one file, they normally get run in order, so this could be at the end.

src/pint/data/runtime/observatories.json
aarchiba: I think most of what one might want to record in a comment could go in the `origin` tag. The exception is when one might want to comment something out.

src/pint/observatory/observatories.py
aarchiba: If the functions are already in `topo_obs.py` we definitely don't want them here too. It's easy enough to get it back - this is git - if we decide we want the functions here instead.

aarchiba:
In work on #1305 I came across some messy tests that don't really work but should be rewritten using the tools of this PR. There they are flagged as FIXME. I don't think it necessarily makes sense to add complexity to this PR, but they are things like: what happens if you try to define an observatory but the clock files cannot be found? so I'm noting it here in hopes that we can add them back in after this PR is merged.

dlakaplan:
@aarchiba : I was trying to update this after merging in the recent PINT release, and I'm now getting clock-file related errors.  It's probably something I can fix but I didn't want to trample over what you had done.  What I see is that the `check_clock_corrections` notebook is failing because (I think) of telescopes with no clock file defined.  I can't find where in the code it checks that the clock file is non-zero.  When I run the `last_clock_correction_mjd` I get errors for telescopes like CHIME where it tries to read a non-existent file:

```
In [1]: import pint.observatory
In [2]: pint.observatory.get_observatory("jodrell_pre_2021").last_clock_correction_mjd()
2022-06-27 13:59:31.992 | DEBUG    | pint.observatory.clock_file:read_tempo_clock_file:601 - Loading TEMPO observatory (None) clock correction file /Users/kaplan/.astropy/cache/download/url/7bec4c689269ddc72b0caee20e97a75d/contents
/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.8/site-packages/erfa/core.py:154: ErfaWarning: ERFA function "dtf2d" yielded 1 of "dubious year (Note 6)"
  warnings.warn('ERFA function "{}" yielded {}'.format(func_name, wmsg),
2022-06-27 13:59:32.319 | INFO     | pint.observatory.topo_obs:find_clock_file:505 - Using global clock file for time_jb.dat
2022-06-27 13:59:32.319 | INFO     | pint.observatory.topo_obs:_load_gps_clock:222 - Loading global GPS clock file for jodrell_pre_2021
2022-06-27 13:59:32.321 | DEBUG    | pint.observatory.clock_file:read_tempo2_clock_file:463 - Loading TEMPO2-format observatory clock correction file /Users/kaplan/.astropy/cache/download/url/d3c81b5766f4bfb84e65504c8a453085/contents
2022-06-27 13:59:32.350 | INFO     | pint.observatory.topo_obs:find_clock_file:505 - Using global clock file for gps2utc.clk
2022-06-27 13:59:32.350 | INFO     | pint.observatory.topo_obs:_load_bipm_clock:237 - Loading BIPM clock version bipm2021 for jodrell_pre_2021
2022-06-27 13:59:32.353 | DEBUG    | pint.observatory.clock_file:read_tempo2_clock_file:463 - Loading TEMPO2-format observatory clock correction file /Users/kaplan/.astropy/cache/download/url/e00edeef4edde217d65207a9abeb6a8c/contents
2022-06-27 13:59:32.361 | INFO     | pint.observatory.topo_obs:find_clock_file:505 - Using global clock file for tai2tt_bipm2021.clk
Out[2]: 59729.0

In [3]: pint.observatory.get_observatory("chime").last_clock_correction_mjd()
2022-06-27 13:59:39.769 | DEBUG    | pint.observatory.clock_file:read_tempo_clock_file:601 - Loading TEMPO observatory (None) clock correction file /Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/data/runtime
2022-06-27 13:59:39.769 | ERROR    | pint.observatory.clock_file:read_tempo_clock_file:734 - TEMPO-style clock correction file /Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/data/runtime for site None not found
2022-06-27 13:59:39.770 | INFO     | pint.observatory.topo_obs:find_clock_file:508 - Using local clock file for 
Out[3]: -inf
```
What is the right way to specify that there is no clock file?  Did I somehow trample on some of your other changes?

aarchiba:
> @aarchiba : I was trying to update this after merging in the recent PINT release, and I'm now getting clock-file related errors. It's probably something I can fix but I didn't want to trample over what you had done. What I see is that the `check_clock_corrections` notebook is failing because (I think) of telescopes with no clock file defined. I can't find where in the code it checks that the clock file is non-zero. When I run the `last_clock_correction_mjd` I get errors for telescopes like CHIME where it tries to read a non-existent file:
> 
> ```
> In [1]: import pint.observatory
> In [2]: pint.observatory.get_observatory("jodrell_pre_2021").last_clock_correction_mjd()
> 2022-06-27 13:59:31.992 | DEBUG    | pint.observatory.clock_file:read_tempo_clock_file:601 - Loading TEMPO observatory (None) clock correction file /Users/kaplan/.astropy/cache/download/url/7bec4c689269ddc72b0caee20e97a75d/contents
> /Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.8/site-packages/erfa/core.py:154: ErfaWarning: ERFA function "dtf2d" yielded 1 of "dubious year (Note 6)"
>   warnings.warn('ERFA function "{}" yielded {}'.format(func_name, wmsg),
> 2022-06-27 13:59:32.319 | INFO     | pint.observatory.topo_obs:find_clock_file:505 - Using global clock file for time_jb.dat
> 2022-06-27 13:59:32.319 | INFO     | pint.observatory.topo_obs:_load_gps_clock:222 - Loading global GPS clock file for jodrell_pre_2021
> 2022-06-27 13:59:32.321 | DEBUG    | pint.observatory.clock_file:read_tempo2_clock_file:463 - Loading TEMPO2-format observatory clock correction file /Users/kaplan/.astropy/cache/download/url/d3c81b5766f4bfb84e65504c8a453085/contents
> 2022-06-27 13:59:32.350 | INFO     | pint.observatory.topo_obs:find_clock_file:505 - Using global clock file for gps2utc.clk
> 2022-06-27 13:59:32.350 | INFO     | pint.observatory.topo_obs:_load_bipm_clock:237 - Loading BIPM clock version bipm2021 for jodrell_pre_2021
> 2022-06-27 13:59:32.353 | DEBUG    | pint.observatory.clock_file:read_tempo2_clock_file:463 - Loading TEMPO2-format observatory clock correction file /Users/kaplan/.astropy/cache/download/url/e00edeef4edde217d65207a9abeb6a8c/contents
> 2022-06-27 13:59:32.361 | INFO     | pint.observatory.topo_obs:find_clock_file:505 - Using global clock file for tai2tt_bipm2021.clk
> Out[2]: 59729.0
> 
> In [3]: pint.observatory.get_observatory("chime").last_clock_correction_mjd()
> 2022-06-27 13:59:39.769 | DEBUG    | pint.observatory.clock_file:read_tempo_clock_file:601 - Loading TEMPO observatory (None) clock correction file /Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/data/runtime
> 2022-06-27 13:59:39.769 | ERROR    | pint.observatory.clock_file:read_tempo_clock_file:734 - TEMPO-style clock correction file /Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/data/runtime for site None not found
> 2022-06-27 13:59:39.770 | INFO     | pint.observatory.topo_obs:find_clock_file:508 - Using local clock file for 
> Out[3]: -inf
> ```
> 
> What is the right way to specify that there is no clock file? Did I somehow trample on some of your other changes?

So the approach I took in PINT was to specify `clock_file=[]` - that is, no clock file name at all - to indicate that no clock file is needed. But that happened in `observatories.py`, which you're deleting and replacing. I think that's the only change to that file you'd want to worry about, though I also made sure none of the observatories that come with PINT ask for any `clock_dir` other than the default (global, allowing overrides).

dlakaplan:
OK, I'm not sure what the error is.  If I run `examples/check_clock_corrections.py` even on the PINT master branch I still get error messages.  So maybe the error that I found was just due to a single misformatted clock file, and the rest are acceptable errors?

dlakaplan:
OK, the notebook test worked, so the errors must not have been important ones.  But is that behavior intended, when the clock is specified as empty (`""` or `[]`) that `last_clock_correction_mjd()` forces a read and shows an error message?

aarchiba:
Yep, definitely I messed it up. #1307 fixes handling of no-clock-files situations.

src/pint/observatory/__init__.py
aarchiba: Not really. This was a tool I used to compare with TEMPO and TEMPO2 so I could resolve any differences between PINT and the other two, so I just pasted the right things in. You could check that entries were correctly parsed by your reader, but you'd have to fiddle the registry first to produce an apparent discrepancy.

aarchiba:
> Do you mean #1309?

Er. Yes. Sorry! I wasn't handling empty `clock_file` entries very sensibly.

src/pint/observatory/special_locations.py
```diff
@@ -315,8 +330,16 @@ def posvel(self, t, ephem, group=None):
         return geo_posvel + stl_posvel
 
 
-# Need to initialize one of each so that it gets added to the list
-BarycenterObs("barycenter", aliases=["@", "ssb", "bary", "bat"])
-GeocenterObs("geocenter", aliases=["0", "o", "coe", "geo"])
-T2SpacecraftObs("stl_geo", aliases=["STL_GEO"])
-# TODO -- BIPM issue
+def load_special_locations():
```
aarchiba, line 333: Is it safe to rerun this? Like if nothing's been removed can you safely rerun this anyway? I assume so but it'd be good to say so in the docs, since the topo_obs version has the `overwrite` flag.

src/pint/observatory/special_locations.py
```diff
@@ -315,8 +330,16 @@ def posvel(self, t, ephem, group=None):
...
+    """Load :class:`~pint.observatory.special_locations.BarycenterObs`, :class:`~pint.observatory.special_locations.GeocenterObs`,
```
aarchiba, line 334: The first line of docstrings gets used in various summaries, so it's better to push longer stuff down and put a short punchy single-line summary as the first docstring line.

src/pint/observatory/special_locations.py
```diff
@@ -315,8 +330,16 @@ def posvel(self, t, ephem, group=None):
...
+    and :class:`~pint.observatory.special_locations.T2SpacecraftObs` into observatory registry.
+    """
+    # Need to initialize one of each so that it gets added to the list
+    BarycenterObs("barycenter", aliases=["@", "ssb", "bary", "bat"])
+    GeocenterObs("geocenter", aliases=["0", "o", "coe", "geo"])
+    T2SpacecraftObs("stl_geo", aliases=["STL_GEO"])
+    # TODO -- BIPM issue
```
aarchiba, line 341: Is this a concern? If they need BIPM TAI to TT data that's awkward because we allow different versions and we want users to be able to use a consistent version throughout PINT.

src/pint/observatory/topo_obs.py
```diff
@@ -509,3 +541,58 @@ def export_all_clock_files(directory):
...
+        if overwrite:
+            obsdict["overwrite"] = True
+        # create the object, which will also register it
+        TopoObs(name=obsname, **obsdict)
+
+
+def load_observatories_from_usual_locations(clear=False):
+    """Load observatories from the default JSON file as well as ``$PINT_OBS_OVERRIDE``, optionally clearing the registry
+
+    Parameters
```
aarchiba, line 584: Maybe worth clarifying that with `clear=True` this returns PINT to the state it normally is on import? Modulo changes in the environment variables.

tests/test_observatory.py
aarchiba: I think this still changes the GBT position

tests/test_observatory_envar.py
```diff
@@ -0,0 +1,101 @@
...
+    o.name = "gbt.json"
+    o.override_file = o.override_dir / o.name
+    with o.override_file.open("wt") as f:
+        f.write(wronggbt)
+
+    try:
+        yield o
+    finally:
+        os.environ = e
+        pint.observatory.Observatory.clear_registry()
```
aarchiba, line 53: This should probably use the load_from_usual_places

src/pint/observatory/special_locations.py
dlakaplan: I didn't put that in because the special locations (so far) never change, so if they are loaded they shouldn't need to be reloaded.  But it would throw an exception, so I put in `overwrite=True`

src/pint/observatory/special_locations.py
dlakaplan: OK, updated a bit

src/pint/observatory/special_locations.py
dlakaplan: No idea.  This predates me.

src/pint/observatory/topo_obs.py
dlakaplan: clarified.

tests/test_observatory_envar.py
dlakaplan: ok

tests/test_observatory.py
dlakaplan: Yes, you may be right (I hadn't realized that the environment persisted across different tests). For the moment I just added:
`    load_observatories_from_usual_locations(clear=True)` but I guess I can do a fixture if needed.

dlakaplan:
Probably easier to merge #1309 before this one?

src/pint/observatory/special_locations.py
aarchiba: It probably is a concern but I'm not sure what to do about it. Leave it I guess.

aarchiba:
#1312 is easily fixed with changes to our observatory list but  I'm sort of tempted to wait until this is merged to avoid needing to make the changes twice

dlakaplan:
@aarchiba : my attempt to merge with #1309 has resulted in some issues.  I now can't pass `tests/test_observatory.py::TestObservatory::test_clock_correction_file_not_available`.  But, strangely I can't pass that on the master branch either.  I'm assuming that test works for you?

dlakaplan:
OK, I think I found the issue.  The test assumes that there is no $TEMPO2 defined, but I had one.  So it passes on the CI machine but fails for me locally unless I unset that.  I can probably add a fixture to fix that in the test.

dlakaplan:
Actually, I suspect there is an issue with that test.  It sets:
```
                clock_dir="TEMPO2",
```
but it looks like the test is not supposed to be for a TEMPO2 directory (the next one is).  Is that an error?  And then for a clock file that isn't found, instead of raising the `NoClockCorrections` exception (like it does with TEMPO(2)), it does:
```
        log.info(f"No clock file for {name}")
        return ClockFile.null(friendly_name=name)
```
So it seems that this test was passing on the CI machine because there was no $TEMPO2, but that it wasn't actually testing correctly.  

Should this raise the exception instead?  Or should we modify the test?  Or do I have this all backwards?

dlakaplan:
Doing that test with:
```
                site.clock_corrections(self.test_time, limits="error")
```
like the other tests seems to work locally.

aarchiba:
Sorry this is all a bit of a mess because it intersects with #1305 . These are legacy tests, too. But there is some complexity in the behaviour we would like:

- If a clock file exists but data is past the end of it, this is potentially okay, so only raise an exception if`limits="error"` is set.
- If a clock file is expected but not found, we could immediately raise an exception, or we could treat this like the above points-out-of-range.
- If a clock file is not expected, that's fine, points are never out of range. 

We used to need missing files to be handled with some grace, because PINT included telescopes whose clock corrections could only be had from $TEMPO/$TEMPO2. Now missing clock files should only happen if PINT itself or the global repository are broken, or if the user specifically requested a custom clock file in a specific location ($TEMPO/$TEMPO2/specific path). 

I *think* that the correct response to a clock file being expected but not found is an immediate exception. Does this make sense?

Unfortunately, not all the tests necessarily agree on what the correct behaviour in this situation is, and the confusing issue of $TEMPO2 being set or not means I haven't necessarily caught all the particular cases where the tests might not agree with our idea of correct behaviour, or with each other.

Sorry about all this. I should probably reconcile this behaviour, and explicitly test for it, in #1305 .

dlakaplan:
Yes, I agree.  I couldn't tell how much of this was intended vs. legacy.  If it's OK, perhaps I will leave this as-is for the moment (tests largely passing, although I also had to reconcile some changes from `observatories.py` -> `observatories.json`) and then we can look at the other side later.

dlakaplan:
Yes, I'm seeing that also about the list of observatories.  I could put in a parenthetical note about running it locally.

dlakaplan:
Done.  and a further contribution in another PR about clocks would be great.

aarchiba:
> Yes, I'm seeing that also about the list of observatories. I could put in a parenthetical note about running it locally.

But why isn't it working on the readthedocs machine? That's weird. It really should work anywhere you run it. I guess `make notebooks` works right?

dlakaplan:
oddly, it seems that it needs to be run twice.  I'll figure it out

dlakaplan:
OK, the `names_and_aliases()` method needed the observatories to be imported

dlakaplan:
And it works now

aarchiba:
Looks like some telescopes include their primary name in their aliases and some don't. Perhaps we should standardize?

aarchiba:
> Yes, I agree. I couldn't tell how much of this was intended vs. legacy. If it's OK, perhaps I will leave this as-is for the moment (tests largely passing, although I also had to reconcile some changes from `observatories.py` -> `observatories.json`) and then we can look at the other side later.

There's a real can of worms here. Definitely not the job of this PR to fix it.

CHANGELOG.md
```diff
@@ -33,6 +39,10 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - jump_flags_to_params now works if some JUMPs are present, never modifies the TOAs, and is idempotent
 - jump_params_to_flags is now idempotent and unconditionally sets the -jump flag to a correct state
 ### Changed
+- TOAs no longer need to be grouped by observatory
+- model.phase() now defaults to abs_phase=True when TZR* params are in the model
+- removed explicit download of IERS and leapsecond data (handled now by astropy)
+- The default version of TT(BIPM) uses BIPM2021
```
aarchiba, line 45: This is weird, how did these entries wind up in this PR?

src/pint/observatory/special_locations.py
aarchiba: There is some complexity here, but discussion belongs on #1305 I think.

src/pint/observatory/topo_obs.py
```diff
@@ -177,6 +204,11 @@ def __init__(
         self.origin = origin
         super().__init__(name, aliases=aliases)
 
+    def __repr__(self):
+        aliases = [f"'{x}'" for x in self.aliases]
+        s = f"TopoObs '{self.name}' ({','.join(aliases)}) at [{self._loc_itrf.x}, {self._loc_itrf.y} {self._loc_itrf.z}]:\n{self.origin}"
```
aarchiba, line 209: `repr` is usually supposed to at least look  like python code that might reproduce the object. So `TopoObs(...)` might be more conventional.

CHANGELOG.md
dlakaplan: In a previous version, there were two instances of CHANGED in the changelog.  I had tried to consolidate them when this PR was in progress prior to the last release.  Now it's gotten messier.  

I also think it would be good if part of the release procedure was to add the Unreleased piece back into the changelog.

aarchiba: Indeed! Merges will still go wrong, though.

tests/test_observatory.py
```diff
@@ -215,5 +221,43 @@ def test_gbt_registered():
...
+    if not overwrite:
+        with pytest.raises(ValueError):
+            load_observatories(io.StringIO(wronggbt), overwrite=overwrite)
+    else:
+        load_observatories(io.StringIO(wronggbt), overwrite=overwrite)
+        newgbt = get_observatory("gbt")
+        assert newgbt._loc_itrf.y > 0
+        assert newgbt._loc_itrf.y != gbt_orig._loc_itrf.y
+    # put back in the original observatories
+    load_observatories_from_usual_locations(clear=True)
```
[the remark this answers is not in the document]
aarchiba: In fact merging the _envar file back in would make it easy to use the `sandbox` fixture, that automatically restores the registry.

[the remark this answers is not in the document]
dlakaplan: I added in a sandbox, not sure if the argument order is correct but it seems to work locally.

tests/test_observatory.py
```diff
@@ -1,18 +1,46 @@
...
+        pass
+
+    o = Sandbox()
+    e = os.environ.copy()
+
+    try:
+        del os.environ["PINT_OBS_OVERRIDE"]
+    except KeyError:
+        pass
+    reg = pint.observatory.Observatory._registry.copy()
```
aarchiba, line 35: If you've done this, you can just assign it back and avoid the reloading - potentially a help is a bug arises in the loading. Or you can ditch this line and rely on reloading to get the registry straightened out.

aarchiba:
I think that's all good apart from that little bit of Changelog weirdness.

CHANGELOG.md
```diff
@@ -33,6 +39,10 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - jump_flags_to_params now works if some JUMPs are present, never modifies the TOAs, and is idempotent
 - jump_params_to_flags is now idempotent and unconditionally sets the -jump flag to a correct state
 ### Changed
+- TOAs no longer need to be grouped by observatory
```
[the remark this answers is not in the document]
dlakaplan: I tried to undo whatever I had.  not sure if it worked.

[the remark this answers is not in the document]
aarchiba: The CHANGELOG.md as committed still seems to have two "Changed" sections, but no longer any duplicated entries. I think it makes sense to consolidate them.

PR #1292, opened 2022-06-21
Simulation changes

AstroJackson:
Marginally edited the make_fake_toas_uniform function in the simulation.py file. I added the ability to mask the toas before a for loop on line 257. This prevents lengthy runtimes if the user intended on masking these toas after running this function. The mask argument was added and defaults to None. If mask is None, the original function operates exactly the same as before this edit.


aarchiba:
Can you add to the docstring an explanation of what "mask" is supposed to do? I'm a little unclear on what the intended use of this is. Would it be more easily achieved by using the from_MJDs version of the fake TOAs maker? (do the linspace[mask] yourself and pass it to that)?

aarchiba:
> Your suggested method works great! The main difference being that adding "fuzz" would also have to be manually done by the user. That should not be an issue since that only comprises 2 lines of code in make_fake_toas_uniform. My original intent of adding a mask is that if I intend on masking the output of make_fake_toas_uniform, then it is best done within the function itself.

I guess my concern is how you'd figure out what the mask was supposed to be without essentially already constructing the list of MJDs, at which point the from_MJDs version seems more natural.

aarchiba:
If you wanted to add an easier way to add fuzz, that might make sense, or some suggestions in the documentation perhaps? If you don't think this is useful you can close it (nothing will be lost but it'll get cleared out of our list of PRs needing attention).

AstroJackson:
I think if anything, the `make_fake_toas_uniform` function seems redundant since it can be effectually the same as `make_fake_toas_fromMJDs` if the user inputs a uniform spacing of MJDs. Maybe something in the documentation can highlight that the for-loop on line 254 can cause avoidably slow runtimes depending on the end goal. I am not sure where to best include that detail.

PR #1293, opened 2022-06-21
Colormode typo

AstroJackson:
Fixed a typo that prevented pintk from color-coding based on JUMPed clusters.


aarchiba:
Looks good to me!

PR #1296, opened 2022-06-22
Clock error messages

aarchiba:
Includes considerable tidying of ClockFile.

Resolves an issue @dlakaplan had where out-of-range TOAs reported an unhelpful filename for where they came from.


aarchiba:
Docs: https://nanograv-pint--1296.org.readthedocs.build/en/1296/_autosummary/pint.observatory.clock_file.html

Issue #1290, opened 2022-06-20
pintk crashes when selecting the last TOA


paulray:
This I can replicate with NGC6440E.par/tim
Start pintk and switch X-axis to "index" (this is not necessary but for NGC6440E it makes it easier because the last two TOAs are very close to each other).
Select a random TOA with left click and hit "d" to delete it. Now try to select that last TOA. I get this:
```
Traceback (most recent call last):
  File "/Users/paulr/env/pint/lib/python3.9/site-packages/matplotlib/cbook/__init__.py", line 287, in process
    func(*args, **kwargs)
  File "/Users/paulr/src/PINT/src/pint/pintk/plk.py", line 1329, in canvasReleaseEvent
    self.stationaryClick(event)
  File "/Users/paulr/src/PINT/src/pint/pintk/plk.py", line 1358, in stationaryClick
    self.selected[ind] = not self.selected[ind]
IndexError: index 61 is out of bounds for axis 0 with size 61
```

aarchiba:
I haven't tracked it down yet, but both `pint.pintk.pulsar.Pulsar` and `pint.pintk.plk.PlkWidget` maintain a list of selected TOAs, and it seems likely that these are getting out of sync. I don't see why this should be - that list should live only in the GUI-free `pint.pintk.pulsar.Pulsar` - and abolishing the one in `pint.pintk.plk.PlkWidget` should fix that and probably other headaches.

aarchiba:
Good God, `pint.pintk.pulsar.Pulsar` has an internal list of selected TOAs but every function that can act on a subset of TOAs *also* accepts a `selected` argument; how that is supposed to relate to the `selected_toas` internally is unclear but I think they're supposed to always be the same.

scottransom:
For your last two comments, those two things are different, but similar.  And I think they are both required.  The `Pulsar` one is a subset of the full TOAs object, and the `plk` one is a boolean array that simply indicates if the user has drawn a box around (or clicked on) TOAs/residuals in the gui.  That latter one is passed to important routines in Pulsar that update things *for those gui-selected points*, that do not necessarily correspond to the set of selected TOAs that you (maybe) have just fitted.  For instance, you can fit a subset of TOAs, and in that process the two objects end up being the same.  But if you then gui-select some of those TOAs to have a JUMP added, they are not the same anymore.  Maybe there is a better way to do this, but there will always need to be at least some kind of temporary list/array that says which points the user has grabbed in the gui.

[...]

Issue #1289, opened 2022-06-20
pintk fails when .tim file is in a subdir


paulray:
The file exists and this runs fine if I copy the .tim file to my current dir.  Somehow it is losing the first part of the path. Probably a consequence of the change to pathlib?
```
% pintk pintecc.par QL1/split/split0000.tim
Traceback (most recent call last):
  File "/Users/paulr/env/pint/bin/pintk", line 33, in <module>
    sys.exit(load_entry_point('pint-pulsar', 'console_scripts', 'pintk')())
  File "/Users/paulr/src/PINT/src/pint/scripts/pintk.py", line 264, in main
    app = PINTk(
  File "/Users/paulr/src/PINT/src/pint/scripts/pintk.py", line 54, in __init__
    self.openPulsar(
  File "/Users/paulr/src/PINT/src/pint/scripts/pintk.py", line 156, in openPulsar
    self.psr = Pulsar(parfile, timfile, ephem, fitter=fitter)
  File "/Users/paulr/src/PINT/src/pint/pintk/pulsar.py", line 83, in __init__
    self.all_toas = get_TOAs(self.timfile, model=self.prefit_model, usepickle=True)
  File "/Users/paulr/src/PINT/src/pint/toa.py", line 231, in get_TOAs
    if not t.check_hashes():
  File "/Users/paulr/src/PINT/src/pint/toa.py", line 1765, in check_hashes
    if _compute_hash(t) != self.hashes[f]:
  File "/Users/paulr/src/PINT/src/pint/toa.py", line 101, in _compute_hash
    h.update(open(filename, "rb").read())
FileNotFoundError: [Errno 2] No such file or directory: 'split0000.tim'
```

aarchiba:
Surely pickles shouldn't rely on your current directory?

PR #1297, opened 2022-06-22
Fix random models

paulray:
The logic for picking the number of points used to compute random models breaks for spans of TOAs that are short (<1 day). This fixes that.


src/pint/pintk/pulsar.py
```diff
@@ -600,12 +600,14 @@ def random_models(self, selected):
             spanMJD = maxallMJD - minallMJD
```
aarchiba, line 600: Worth adding a few percent of extrapolation?

aarchiba: Wait, I see. Shouldn't `extra` be applied here rather than later? So it affects the number-of-TOAs calculation?

paulray: I decided not to, so that `spanMJD` is the actual span.

src/pint/pintk/pulsar.py
```diff
@@ -600,12 +600,14 @@ def random_models(self, selected):
...
             # want roughly 1 per day up to 3 years
```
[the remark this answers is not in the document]
paulray: Done

paulray:
Also fixes #1289

src/pint/pintk/pulsar.py
aarchiba: Well, it's the span of the data, not the span the `linspace` is supposed to cover. So if you're in the one-per-day regime, it allocates one per day and then spreads them farther than that by an amount `extra`. I suppose the intention is that `extra` should usually be small so the difference is subtle, but it seems like one should allocate points based on the span they are actually supposed to cover - that is, more points if you have more `extra`?

src/pint/toa.py
```diff
@@ -231,9 +231,15 @@ def get_TOAs(
             updatepickle = True
         else:
             if hasattr(t, "hashes"):
-                if not t.check_hashes():
+                try:
+                    if not t.check_hashes():
+                        updatepickle = True
+                        log.warning("Pickle file is based on files that have changed")
+                except FileNotFoundError:
```
aarchiba, line 238: This check is a good idea - we don't want crashes! maybe we should catch exceptions more aggressively, like if the file exists but isn't readable - but perhaps we should make the paths absolute before storing the hashes, so that their stored location doesn't depend on the current directory? `os.path.abspath()` or one of the methods of Path objects should do that. A test would be nice too - make a pickle in a temporary directory and then try to change directory and access it.

aarchiba: Maybe just accept that pickles have to be regenerated if you move around for now - but you shouldn't error out! - and file a bug describing the problem?

paulray:
Now this also fixes #1290

paulray:
OK, I think these are all good small changes. I'm going to merge it and file an issue about the pickle fragility for a more extensive fix at some point.

paulray:
I make issue #1301 for further discussion of the pickle issue.

src/pint/pintk/plk.py
```diff
@@ -1249,7 +1249,7 @@ def coordToPoint(self, cx, cy):
             if dist[ind] > clickDist:
                 log.warning("Not close enough to a point")
                 ind = None
-        return index
+        return ind
```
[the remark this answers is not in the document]
paulray: All I know is that @scottransom said this change would fix the problem and in my quick tests it did.

scottransom:
Index is the original TOA index.  And so that is a more reliable way of identifying a specific TOA.  In the case of the bug, that was being used as an index of the remaining TOAs, which is wrong.  So this might not be the correct solution to the problem.


On June 23, 2022 12:46:00 PM EDT, Anne Archibald ***@***.***> wrote:
***@***.*** commented on this pull request.
>
>
>
>> @@ -1249,7 +1249,7 @@ def coordToPoint(self, cx, cy):
>             if dist[ind] > clickDist:
>                 log.warning("Not close enough to a point")
>                 ind = None
>-        return index
>+        return ind
>
>Is it helpful to compute the original TOA index here? If it's not what is returned? Should that bit of the code just be removed?
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/nanograv/PINT/pull/1297#pullrequestreview-1017344572
>You are receiving this because you are subscribed to this thread.
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/pint/pintk/plk.py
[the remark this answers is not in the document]
aarchiba: It just looks risky - functions that call this could be expecting either the thing that was in the original TOA index column or the index into the current array of TOAs, and this change switches which this function returns.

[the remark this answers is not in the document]
paulray: OK, maybe we should handle this in a separate PR.  I'll remove that change.

Issue #1298, opened 2022-06-23
pint.toa.get_TOAs can't read from a file-like if usepickle=True


aarchiba:
The correct behaviour would be to fall back to not using a pickle as there is nowhere to store it.

I have some prototype code to fix this, mixed up with some pintk work.

abhisrkckl:
This is still an issue as of today.

scottransom:
I think raising an exception and exiting would be fine too.  This is the user screwing up.

PR #1299, opened 2022-06-23
WIP: Pintk testing

aarchiba:
This is a cleanup meant to make more of pintk testable and tested by pushing things into `pint.pintk.pulsar.Pulsar`. It includes attempts to replicate #1290 and a fix for #1298.

### DIFF
--- a/src/pint/pintk/plk.py
+++ b/src/pint/pintk/plk.py
@@ -1332,9 +1332,7 @@ class PlkWidget(tk.Frame):
         self.move = False
 
     def stationaryClick(self, event):
-        """
-        Call this function when the mouse is clicked but not moved
-        """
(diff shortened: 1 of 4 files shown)


aarchiba:
Some complexities that should be simplified, mostly because data is replicated in more than one place and can get out of sync:

- Pulsar has `selected_toas`, Pulsar functions often accept an argument `selected`, and the GUI side maintains a set of selected objects. These need to be clearly distinguished.
- TOAs can be refered to by their original index in the input file, or by their index into the set of TOAs currently available. There needs to be a clear rule about which one to use.
- PINTk tries to keep a flag on TOAs listing which JUMP numbers they refer to. This flag should be replaced by a convenience function that returns which JUMPs affect a TOA.

PR #1300, opened 2022-06-23
Fix jump_flags_to_params and jump_params_to_flags

### LINKED ISSUE #1294: PINT struggles reading the flag "tim_jump"
When a par file has the row "JUMP    -tim_jump 1         0 0 0" and the corresponding tim file has the flag "-tim_jump 1", pintk doubles the number of JUMPs. pintempo gives an error as well. When the flag renamed to anything else, in both the par and tim files, the issue goes away. 

I recommend renaming the JUMP flag to something other than "tim_jump". Some examples may be "time_jump", "jump_tim", or "jump". 

My operating system and platform is given by: Linux-2.6.32-754.35.1.el6.x86_64-x86_64-with-glibc2.12. 
pint.__version__ = 0.8.8+175.gb8b04f5
pint.__file__ = /users/jdtaylor/Jackson/PINT/src/pint/__init__.py
I am using a virtual environment though the issue persists on my local Mac as well.

aarchiba:
These two functions are used to relate JUMP parameters and TOA flags, and both had serious problems. This makes them do well-defined, documented things.

Closes #1294

### DIFF
--- a/src/pint/models/jump.py
+++ b/src/pint/models/jump.py
@@ -155,15 +155,21 @@ class PhaseJump(PhaseComponent):
     def jump_params_to_flags(self, toas):
         """Take jumps created from .par file and add appropriate flags to toa table.
 
-        This function was made specifically with pintk in mind for a way to properly
-        load jump flags at the same time a .par file with jumps is loaded (like how
-        jump_flags_to_params loads jumps from .tim files).
+        This function was made specifically with pintk in mind so as to maintain
+        a ``-jump`` flag on each TOA that is affected by any JUMP in the model,
+        listing the numbers of all JUMPs affecting that TOA.
+
+        This function wipes all ``-jump`` flags and reinitializes them to reflect
+        the actual situation.
 
         Parameters
         ----------
         toas: TOAs object
             The TOAs which contain the TOA table to be modified
         """
+        log.info(f"Initial TOAs {toas['jump']=}")
+        toas["jump"] = ""
+        log.info(f"Wiped TOAs {toas['jump']=}")
         # for every jump, set appropriate flag for TOAs it jumps
         for jump_par in self.get_jump_param_objects():
             # find TOAs jump applies to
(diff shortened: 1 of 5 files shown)


src/pint/models/jump.py
```diff
@@ -172,12 +178,12 @@ def jump_params_to_flags(self, toas):
             for d in toas.table["flags"][mask]:
                 if "jump" in d:
                     index_list = d["jump"].split(",")
-                    if str(jump_par.index) in index_list:
```
scottransom, line 175: What was this part doing that was problematic?

scottransom: Or is it just making sure that all TOAs that are affected by all JUMPs always have their flags re-written?  If so, I'm just wondering if the previous behavior caused an issue?

tests/test_find_clock_file.py
```diff
@@ -68,7 +68,7 @@ def test_can_find_in_repo(sandbox, temp_cache):
 
 
 def test_pint_env_overrides(sandbox, temp_cache):
-    os.environ["PINT_CLOCK_OVERRIDE"] = sandbox.override_dir
```
scottransom, line 71: Just making sure that this change was intended for this PR?

src/pint/models/jump.py
aarchiba: I had to rewrite this because a test was failing when I rewrote `jump_flags_to_params`. Essentially, this never removed a jump's number from the list - anythjng that was ever in the `-jump` flag always stayed there unless some other function took it out. Now this function can be (should be!) the only thing that ever modifies the `-jump` flag. 

(I still maintain that any function that wants to look at the `-jump` flag is better off just running essentially this function to find out what JUMPs they need to worry about. But since the `-jump` flag is only used in pintk I'm not yet confident I can make that change. It's still probably a good idea for any function that wants to look at the `-jump` flag to run this first to make sure it's up-to-date and no one has added a JUMP or a TOA or changed anything.)

The `if` statement here is harmless but since I'm starting from nothing and iterating over all the JUMPs, it's never true.

scottransom:
This looks good to me!  I will merge now.

PR #1302, opened 2022-06-23
MeerKAT can use the observatory clock file

aarchiba:
The global repository now has an auto-updated clock file for MeerKAT but it's not quite suitable for TEMPO2, so it has a different name. This PR makes PINT use it.


aarchiba:
Only drawback here is it's possible that the `mk2utc_observatory.clk` will go away if I can clarify what's really in that file and clean up its headers.

PR #1307, opened 2022-06-27
Fix pintk merge problem

aarchiba:
Pintk was crashing because fake TOAs were missing a column. This adds that column.


paulray:
Looks good to me!

PR #1308, opened 2022-06-27
Updated logging docs and example

dlakaplan:
Updated: 
https://nanograv-pint.readthedocs.io/en/latest/examples/PINT_walkthrough.html to use a slightly different logging setup, and put more explicit info at:
https://nanograv-pint.readthedocs.io/en/latest/controlling-logging.html
(mostly drawn from `logging` docstring)


aarchiba:
Looks good to me.

PR #1309, opened 2022-06-27
Handle no clock files better

aarchiba:
Current master emits loud and unhelpful error messages when working with telescopes for which no clock file is needed. This PR fixes that and checks that these telescopes behave better.


dlakaplan:
This looks good to me.  I'm happy to merge, or to wait until we discuss (not that there is much to discuss here).

aarchiba:
I think it's uncontroversial, so if you're happy please go ahead and merge.

PR #1311, opened 2022-06-28
Add How-to questions from Wiki

aarchiba:
This adds a Howto page to our docs to cover what's currently on https://github.com/nanograv/PINT/wiki/How-To . Additional content from the wiki might should be migrated into the docs. 

The doc page added by the PR is here: https://nanograv-pint--1311.org.readthedocs.build/en/1311/user-questions.html

### DIFF
--- a/docs/common-timing-workflow.rst
+++ b/docs/common-timing-workflow.rst
@@ -1,0 +1,44 @@
+
+How to: Common timing workflow
+==============================
+
+(This was originally written by Alex McEwen on the PINT wiki.)
+
+When I'm working on new pulsar solutions, my work flow usually follows
+something like this. This uses some of the basic tools that are included in the
+:ref:`user questions <user-questions>`. Please send
+suggestions/comments/questions to aemcewen@uwm.edu.
+
+
+1. load pulsar model/TOAs via get_TOAs() and get_model()::
+
+     model = get_model('parfile.par')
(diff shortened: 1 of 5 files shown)


aarchiba:
> Perhaps we delete things from the wiki that have been migrated to the official docs, and add a link on the wiki to that docs page. Then we can still add content to the wiki as questions from users come up and are answered.

Yes, that seems quite sensible, and also helps people find where in the docs this kind of question is answered, both for looking things up and for migrating further wiki content in future.

There may be other things on the wiki that should be migrated, I haven't really looked in detail.

aarchiba:
Looking at the longer entries referenced there, https://github.com/nanograv/PINT/wiki/How-To:-Simple-Python-PINT-Tools looks like it should be turned into a tutorial notebook, but that would take work. https://github.com/nanograv/PINT/wiki/How-To:-Common-Timing-Workflow wouldn't be so easy to convert. I'm sort of inclined to migrate them to separate How-to documents; we can point to an example notebook (which would have the advantage of maybe letting us know if we break this stuff!) later.

aarchiba:
How-tos should point to `pint.fitter.Fitter.auto`.

aarchiba:
Edited the wiki to point to the how-tos in the docs.

dlakaplan:
All this looks fine.  Just to note that we should periodically do this (move things from the wiki to rtd).  I can merge if you are done.

PR #1313, opened 2022-06-29
Remove obsolete super() syntax

aarchiba:
The presence of `super(ClassName, self)` is a holdover from Python 2. Its presence in our code misleads new people into thinking that's how you do it. This PR removes all of those.

### DIFF
--- a/src/pint/mcmc_fitter.py
+++ b/src/pint/mcmc_fitter.py
@@ -139,9 +139,7 @@ class MCMCFitter(Fitter):
     """
 
     def __init__(self, toas, model, sampler, **kwargs):
-        super(MCMCFitter, self).__init__(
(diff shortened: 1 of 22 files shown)


dlakaplan:
So this is just `super(class, self)` -> `super()`, right?  I realize the former syntax is outdated but does it do anything wrong?

aarchiba:
> So this is just `super(class, self)` -> `super()`, right? I realize the former syntax is outdated but does it do anything wrong?

The old syntax is a python 2 holdover, like `class Cheese(object):` (which this also removes the vestiges of). It's more error-prone - you can put the wrong class name in there, particularly if you're cutting and pasting - and looks more complicated. I'm removing it for consistency and because many of our contributors learn python idiom from what's in our source code.

paulray:
Looks good to me!

PR #1315, opened 2022-06-29
Emcee Backend chains filesave

devbhakt:
This PR uses in the `backends` functionality built into `emcee` to save the mcmc chains into a h5 file if the `h5py` package is available. This can be implemented into the MCMCfitter as well.


src/pint/scripts/event_optimize.py
```diff
@@ -758,6 +758,13 @@ def main(argv=None):
 
     import emcee
 
+    # Setting up a backend to save the chains into an h5 file
+    try:
+        backend = emcee.backends.HDFBackend(ftr.model.PSR.value + "_chains.h5")
```
aarchiba, line 763: It would be better to allow the user to specify a location for the (potentially large) HDF5 file - putting it in the current directory might not be helpful.

devbhakt: I have added in a backend filepath argument that lets the user specify a file path for the HDF5 file

src/pint/scripts/event_optimize.py
```diff
@@ -758,6 +764,18 @@ def main(argv=None):
 
     import emcee
 
+    # Setting up a backend to save the chains into an h5 file
+    try:
+        if backendpath:
+            backend = emcee.backends.HDFBackend(
```
aarchiba, line 770: It's preferable to construct the filename in the `if` statement so as not to duplicate the HDF5 construction code.

aarchiba:
It would be good to create a tiny test case where one ran a handful of walkers for a handful of steps with and without the HDF5 backend, just to make sure using it doesn't break anything. You'll presumably need random seeds, but the goal is to run it as fast as possible but still check that the HDF5 backend works, and that you can recover the chains from it.

devbhakt:
The initial test already has the backend disabled so I have added in a test that enables the backend and loads in the chains from the chains file.

aarchiba:
What happens if you run `event_optimize` and the chains already exist? Does it pick up from where it left off? Please write a test to demonstrate this.

devbhakt:
Currently `event_optimize` is set up so that any generated files from previous runs are overwritten. So following that example, I set up the backend to reset  (`backend.reset(nwalkers, ndim)`) at the beginning of each MCMC run. A built-in workaround to write chains to a new file would be to specific a different file path with the `--backendpath` argument

abhisrkckl:
Can this be merged?

tests/test_event_optimize.py
```diff
@@ -27,10 +30,44 @@ def test_result(tmp_path):
...
+
+        samples = None
+        os.chdir(tmp_path)
+        # Running with backend
+        cmd = f"{eventfile} {parfile} {temfile} --weightcol=PSRJ0030+0451 --minWeight=0.9 --nwalkers=10 --nsteps=50 --burnin=10 --backend"
+        event_optimize.maxpost = -9e99
+        event_optimize.numcalls = 0
+        event_optimize.main(cmd.split())
+
+        reader = emcee.backends.HDFBackend("J0030+0451_chains.h5")
```
aarchiba, line 66: How should the user specify the file name of the HDF5 file?

abhisrkckl:
> This can be implemented into the MCMCfitter as well.

This will be a useful feature to have.

devbhakt:
This is ready for merge

PR #1318, opened 2022-06-30
Change update frequency for index.txt

aarchiba:
`index.txt` must be checked to see what global clock corrections are available, whether old versions have been flagged as invalid, and how often other files should be checked for updates. That happens only if it is older than a certain amount. This amount was set to just a couple of minutes in the early stages of testing, but this PR sets it to 1 day. Longer would probably be fine.

`index.txt` is a small file but if one has a flaky net connection getting it can be slow and annoying.


aarchiba:
That would be great.

Issue #1286, opened 2022-06-16
pint.models.model_builder.parse_parfile makes all param names uppercase


abhisrkckl:
This causes case-sensitive parameter names (eg. TNRedAmp, TNRedGam) to be ignored by ModelBuilder. For example, the red noise components present in the IPTA DR2 par files are not recognized by PINT. It works if I manually create a PLRedNoise object and add it to the timing model.

aarchiba:
The intention is that parfile parsing is case insensitive (but ideally output parfiles should preserve the input case by default). This would mean that TNRedAmp should be called TNREDAMP internally, but it should work. Can you post an example showing that it doesn't?

abhisrkckl:
Here is an example program showing the issue along with the data.
[rnparams_test.zip](https://github.com/nanograv/PINT/files/8928388/rnparams_test.zip)
 
The data is from IPTA DR2. 
I am using the version of PINT corresponding to commit 3e802b5d358caca3a8bb07c39d93d078492bb402

aarchiba:
Is it fair to say this is fixed? We still forcibly uppercase everything, but that now works fine?

abhisrkckl:
Yes. I opened this issue because of the PLRedNoise bug. Since that is fixed, this is a non-issue now IMO. Also, I don't think making all parameters uppercase is a bad thing now.

PR #1321, opened 2022-07-01
Making the PLRedNoise parameters all caps

abhisrkckl:
Fixing Issue #1286 : PINT could not automatically load PLRedNoise component from par file before.

### DIFF
--- a/src/pint/models/noise_model.py
+++ b/src/pint/models/noise_model.py
@@ -429,7 +429,7 @@ class PLRedNoise(NoiseComponent):
 
         self.add_param(
             floatParameter(
-                name="TNRedAmp",
+                name="TNREDAMP",
                 units="",
                 aliases=[],
                 description="Amplitude of powerlaw " "red noise in tempo2 format",
@@ -437,7 +437,7 @@ class PLRedNoise(NoiseComponent):
         )
         self.add_param(
             floatParameter(
-                name="TNRedGam",
+                name="TNREDGAM",
                 units="",
                 aliases=[],
                 description="Spectral index of powerlaw " "red noise in tempo2 format",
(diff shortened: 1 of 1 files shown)


aarchiba:
There are already Prior objects in PINT: https://github.com/nanograv/PINT/blob/master/src/pint/models/priors.py How does this code relate?

src/pint/bayesian.py
```diff
@@ -0,0 +1,184 @@
...
+import pint.models
+import pint.toa
+import pint.residuals
+
+class Prior:
+    def __init__(self, model : pint.models.TimingModel, dist : str, width=1, info=None):
+        self.params = model.free_params
+        self.dist = dist
+        if self.dist in ['uniform', 'normal']:
+            values = np.array([getattr(model,param).value for param in self.params])
```
aarchiba, line 13: Should this be `self.free_params`?

src/pint/bayesian.py
```diff
@@ -0,0 +1,184 @@
...
+            uncertainties = np.array([getattr(model,param).uncertainty.value for param in self.params])
+            
+            if np.any(uncertainties==0):
+                raise ValueError("All fitting parameters must have non-zero uncertainties for this method to work. Update uncertainties in the par file or use dist='custom'.")
```
aarchiba, line 17: Are the uncertainties meant to establish the priors? Uncertainties are normally set as the result of a fitting operation, so they are usually an estimate of the posterior.

src/pint/models/noise_model.py
```diff
@@ -429,23 +429,23 @@ def __init__(
...
+                name="TNREDAMP",
                 units="",
                 aliases=[],
                 description="Amplitude of powerlaw " "red noise in tempo2 format",
             )
         )
         self.add_param(
             floatParameter(
-                name="TNRedGam",
+                name="TNREDGAM",
```
aarchiba, line 440: Is this an unrelated bug fix? It'd be good to make this a separate PR so it can be merged sooner.

abhisrkckl:
Sorry.. This pull request is broken. I will create a separate pull request for this.

PR #1323, opened 2022-07-01
Save pulse number in tim files

### LINKED ISSUE #1322: PINT ignores delta_pulse_number column when writing TOAs
I don't know if this is intentional, but if you try to change the pulse numbering using the `delta_pulse_number` column (which is also used in `pintk` when adding phase wraps), I don't think that propagates to the output tim file.  The writer converts the `pulse_number` column to the `-pn` flag, but doesn't do `delta_pulse_number` to `-padd` (although it does read in those flags).  

It's an easy fix, but I want to make sure what the intended behavior is first.  

This is an issue that @almcewen0 identified.

dlakaplan:
Previously the `pulse_number` column was written as the `-pn` flag in a tim file, but the `delta_pulse_number` was ignored.  This is even though the `-padd` flag could be used to initialize that on read.

Now any `delta_pulse_number` column is saved as `-padd`.  In addition there are a couple of convenience functions for dealing with flags:
`t.set_flag_value()`
`t.delete_flag()`
the former was easy before if all of the values were the same, but harder if they were an array.  this just wraps it.

Tests for new functionality included.

Fixes #1322


dlakaplan:
If this behavior was intended, then I can just ignore this.  But it was confusing since `pintk` uses the `delta_pulse_number` column (@scottransom ).  Maybe when the TOAs are fit that column gets added to the pulse number so it doesn't usually need to be saved?  In this case @almcewen0 was trying to save pre-fit TOAs but including phase wraps, so that's where the issue came from.

scottransom:
Yes, at least in `pintk`, as soon as a fit is done, the `delta_pulse_number` column is zero'd out, and then new pulse numbers are calculated for the best-fit model.  So you really don't usually write out TOAs where the `delta_pulse_number` column is set at all.  (Although maybe that could happen if you changed a phase for some TOAs and then immediately wrote the TOAs before doing a fit).

almcewen0:
> Yes, at least in `pintk`, as soon as a fit is done, the `delta_pulse_number` column is zero'd out, and then new pulse numbers are calculated for the best-fit model. So you really don't usually write out TOAs where the `delta_pulse_number` column is set at all. (Although maybe that could happen if you changed a phase for some TOAs and then immediately wrote the TOAs before doing a fit).

yea, in this case i was adding a ton of phase wraps, but not fitting (until they were all added). but i wanted to save the TOAs with the modified pulse numbers as a checkpoint.

dlakaplan:
That was my guess.  So I think @almcewen0 's use was slightly unusual, which led to this.  In any case I don't think this update will hurt (I just made it so it will only save the column if it's non-zero).

scottransom:
Yeah, I think this is probably fine.  It *is* kind of confusing to have both `-pn` and `-padd` flags, but there is nothing intrinsically wrong with it I don't think....

aarchiba:
How did we wind up with `-pn` getting turned into `pulse_number` on load, and `-padd` getting turned into `delta_pulse_number` on load? Why does PINT have different internal and external names for the same thing?

src/pint/toa.py
```diff
@@ -1639,6 +1639,37 @@ def get_flag_value(self, flag, fill_value=None, as_type=None):
             result.append(val)
         return result, valid_index
 
+    def set_flag_values(self, flagname, flagvalues):
```
aarchiba, line 1642: Could this just be implemented by letting `__setitiem__` accept arrays?

src/pint/toa.py
```diff
@@ -1639,6 +1639,37 @@ def get_flag_value(self, flag, fill_value=None, as_type=None):
...
+        assert np.isscalar(flagvalues) or (len(flagvalues) == self.ntoas)
+        for ii in range(self.ntoas):
+            if np.isscalar(flagvalues):
+                if not np.isnan(flagvalues):
+                    self.table["flags"][ii][flagname] = str(flagvalues)
+            else:
+                if not np.isnan(flagvalues[ii]):
+                    self.table["flags"][ii][flagname] = str(flagvalues[ii])
+
+    def delete_flag(self, flagname):
```
aarchiba, line 1661: Can this already be done by assigning `None` or `""` to the flag?

src/pint/toa.py
```diff
@@ -2037,11 +2062,10 @@ def write_TOA_file(
...
         # If pulse numbers were added to flags, remove them again
         if pnChange:
-            for flags in self.table["flags"]:
-                try:
-                    del flags["pn"]
-                except KeyError:
-                    pass
+            self.delete_flag("pn")
+        # Same for delta pulse numbers
+        if dpnChange:
```
aarchiba, line 2067: Without a `try: ... finally:` the way this function works is highly prone to leaving TOAs objects in broken states when exceptions occur, including KeyboardInterrupt. Can all these changes be made on  a fresh copy of the TOAs object, which is written out and then discarded?

aarchiba: This is particularly problematic for `-padd` - what happens if you find yourself with a TOAs object with both a `delta_pulse_numbers` column and some `-padd` flags - are the `-padd` redundant or are they to be added into the `delta_pulse_numbers`?

tests/test_toa_flag.py
```diff
@@ -50,3 +52,112 @@ def test_flag_value_str(self):
...
+ uwl_191207_024252.0000_0000.rf.pTF 1102.00000000 58824.11892727732292840 25.10700 pks
+"""
+    t = toa.get_TOAs(io.StringIO(s))
+    t.set_flag_values("test", 1)
+    for i in range(t.ntoas):
+        assert t["flags"][i]["test"] == "1"
+
+
+def test_flag_set_array():
+    s = """FORMAT 1
```
aarchiba, line 78: Is this string any different than the one in the last test? Could it go in a global variable? For that matter is there a reason not to just use fake TOAs, where one could more clearly express the properties they need to have?

tests/test_toa_flag.py
```diff
@@ -50,3 +52,112 @@ def test_flag_value_str(self):
...
+ uwl_191012_013412.0000_0000.rf.pTF 1102.00000000 58768.08628513784793768 16.29100 pks
+ uwl_191013_013912.0000_0000.rf.pTF 1102.00000000 58769.08628494704621659 18.55400 pks
+ uwl_191014_011651.0000_0000.rf.pTF 1102.00000000 58770.06840417512788122 21.57900 pks
+ uwl_191015_063631.0000_0000.rf.pTF 1102.00000000 58771.28385332518061546 23.07400 pks
+ uwl_191029_231442.0000_0000.rf.pTF 1102.00000000 58785.98420216310909581 18.68600 pks
+ uwl_191111_020722.0000_0000.rf.pTF 1102.00000000 58798.10659743353399165 17.39200 pks
+ uwl_191128_000422.0000_0000.rf.pTF 1102.00000000 58815.01493200936268835 20.28800 pks
+ uwl_191207_024252.0000_0000.rf.pTF 1102.00000000 58824.11892727732292840 25.10700 pks
+"""
+    t = toa.get_TOAs(io.StringIO(s))
```
aarchiba, line 111: If all these TOAs objects can be identical, perhaps they could be created in a fixture?

src/pint/toa.py
dlakaplan: I don't think so.  It would actually be nice to just allow:
```
t['flags']['key']
```
but I don't think that works with the current data model.  But maybe I'm wrong.  There is also the need to turn the flags into strings, but avoid(?) string arrays.

src/pint/toa.py
dlakaplan: Will that actually delete it or just set it to a null value?  I think the later, and there may be cases where the former is desired.  Certainly that was done at several places in the code so I just provided a wrapper.

tests/test_toa_flag.py
dlakaplan: no reason, just incremental adding and then not cleaning up later.

tests/test_toa_flag.py
dlakaplan: yes, done.

dlakaplan:
I'm not sure why the tests above are failing.  It's largely in the `test_precision.py` file.  They run fine for me locally.  Things look like:
```
______________________ test_time_from_longdouble_utc[mjd] ______________________
[379](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:380)
[380](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:381)
format = 'mjd'
[381](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:382)
[382](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:383)
    @given(reasonable_mjd())
[383](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:384)
>   @example(format="pulsar_mjd", i_f=(40000, 0.7333333333333333))
[384](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:385)
[385](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:386)
/home/runner/work/PINT/PINT/tests/test_precision.py:332: 
[386](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:387)
[387](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:388)
[388](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:389)
state = <hypothesis.core.StateForActualGivenExecution object at 0x7fe135cbf610>
[389](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:390)
wrapped_test = <function accept.<locals>.test_time_from_longdouble_utc at 0x7fe144f29700>
[390](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:391)
arguments = (), kwargs = {'format': 'mjd'}
[391](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:392)
original_sig = <Signature (format, i_f)>
[392](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:393)
[393](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:394)
    def execute_explicit_examples(state, wrapped_test, arguments, kwargs, original_sig):
[394](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:395)
        posargs = [
[395](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:396)
            p.name
[396](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:397)
            for p in original_sig.parameters.values()
[397](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:398)
            if p.kind is p.POSITIONAL_OR_KEYWORD
[398](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:399)
        ]
[399](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:400)
[400](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:401)
        for example in reversed(getattr(wrapped_test, "hypothesis_explicit_examples", ())):
[401](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:402)
            if example.args:
[402](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:403)
                assert not example.kwargs
[403](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:404)
                if len(example.args) > len(posargs):
[404](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:405)
                    raise InvalidArgument(
[405](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:406)
                        "example has too many arguments for test. Expected at most "
[406](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:407)
                        f"{len(posargs)} but got {len(example.args)}"
[407](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:408)
                    )
[408](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:409)
                example_kwargs = dict(zip(posargs[::-1], example.args[::-1]))
[409](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:410)
            else:
[410](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:411)
                example_kwargs = dict(example.kwargs)
[411](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:412)
[412](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:413)
>           bound = original_sig.bind(*arguments, **example_kwargs, **kwargs)
[413](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:414)
E           TypeError: bind() got multiple values for keyword argument 'format'
[414](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:415)
[415](https://github.com/nanograv/PINT/runs/7201529767?check_suite_focus=true#step:8:416)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/hypothesis/core.py:373: TypeError
```

src/pint/toa.py
aarchiba: Don't we allow `t['-key']` as if a flag were a column? I agree string arrays are awkward.

src/pint/toa.py
aarchiba: I'm pretty sure I arranged for flags set to a null value to be deleted, since there is no way to represent that in a tim file.

src/pint/toa.py
aarchiba: My main concern with the second was that when the code was temporarily modifying the initial object, a control-C could leave it with both. Clock corrections are in the same boat. But if the object is not temporarily modified, we should be fine.

aarchiba:
This looks like a `hypothesis` change - specifically a change in how `@example` is supposed to work. https://github.com/HypothesisWorks/hypothesis/releases/tag/hypothesis-python-6.49.1 supports this idea.

aarchiba:
Issue filed upstream: https://github.com/HypothesisWorks/hypothesis/issues/3398

src/pint/toa.py
```diff
@@ -1994,54 +2012,57 @@ def write_TOA_file(
...
-                flags=flags,
-                format=format,
-                alias_translation=self.alias_translation,
-            )
-            outf.write(out_str)
+        try:
+            toacopy = copy.deepcopy(self)
+            # have to deepcopy the flags separately
+            for i in range(len(self)):
+                toacopy.table[i]["flags"] = copy.deepcopy(self.table[i]["flags"])
```
dlakaplan, line 2021: This would be addressed by #1329

src/pint/toa.py
```diff
@@ -1994,54 +2012,57 @@ def write_TOA_file(
...
+                    obs_obj,
+                    name=flags.pop("name", name),
+                    flags=flags,
+                    format=format,
+                    alias_translation=toacopy.alias_translation,
+                )
+                outf.write(out_str)
+
+        finally:
+            del toacopy
```
aarchiba, line 2065: Not needed, it'll be deleted as soon as it goes out of scope.

src/pint/toa.py
dlakaplan: I guess I didn't realize that `t['key']` would work (I had tried `t['flags']['key']`).  But given the issues around string arrays and other format conversions, I still think this function is useful.

aarchiba: Would it be equally effective to make `t["key"]` apply `str` to each element of its argument? Possibly except if they're None... I'm trying to avoid something like the mess on TimingModels where there are heaps of different ways to set things and no one is sure which ones work.

dlakaplan: Yes, that might work.  As I said I didn't know about this syntax: I use `get_flag_value()` to get results, and that gives slightly different results to just `t["key"]` (you get to specify what the missing values are, and what the resulting data-type is).  But if there were a way to set values more easily that would be useful.

aarchiba:
CI failures fixed in #1332, ready for merging if you're okay with it

src/pint/toa.py
aarchiba: Yes, `.get_flag_values()` is analogous to a dictionary's `.get()` function, that allows a default value instead of the exception produced by indexing. But I'm not sure we need two different ways of setting flag values.

src/pint/toa.py
aarchiba: Merged now.

src/pint/toa.py
```diff
@@ -1994,54 +2002,57 @@ def write_TOA_file(
...
+                    freq,
+                    obs_obj,
+                    name=flags.pop("name", name),
+                    flags=flags,
+                    format=format,
+                    alias_translation=toacopy.alias_translation,
+                )
+                outf.write(out_str)
+
+        finally:
```
aarchiba, line 2054: No need for the `try:...finally:`

tests/test_pulse_number.py
```diff
@@ -161,3 +161,29 @@ def test_partial_pulse_numbers(model, toas):
...
+
+
+def test_save_delta_pulse_number(model, toas):
+    toas.compute_pulse_numbers(model)
+    toas["delta_pulse_number"][:10] += 1
+    outtim = StringIO()
+    toas.write_TOA_file(outtim)
+    outtim.seek(0)
+    newtoas = get_TOAs(outtim)
+    assert (newtoas[:10]["delta_pulse_number"] == 1).all()
```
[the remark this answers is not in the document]
aarchiba: In that case it's easy to include arbitrary things in the `delta_pulse_number` column and I think some negative numbers would be a good idea - just to make absolutely sure the TOA parsing doesn't randomly decide `-1` is a flag.

aarchiba:
I'm still not convinced that `toas.set_flag_value{"pn", toas["pulse_number"])` is better syntax than `toas["pn"] = toas["pulse_number"]`.

aarchiba:
Short version: don't expect assigning to multiple indexing to work. Sometimes you'll get lucky, but usually it won't work. If you stick to a single indexing operation while setting, that should always either work or raise an exception, there won't be surprises:

`toas[anything] = values` -> `toas.__setitem__(anything, values)`

Long version:

In numpy arrays, some operations are able to return references to the original array, while others return new arrays. In numpy, the references happen with slice syntax and some instances of `reshape` and a handful of others, but fancy indexing has to make a copy. In practical terms, this means you can use slice syntax to create a new array and then use it to modify the original, but you can't do that with fancy indexing. So `a[::-1][::2] = 7` works but `a[condition][::2] = 7` can't - the indexing-and-assignment operation (`thing[::2] = 7`) only has access to the copy produced by the first operation.

`toas[::2] = thing` could work; it goes through `TOAs.__setitem__(toas, slice(None, None, 2), thing)`. `toas[fancy] = thing` also goes through `TOAs.__setitem__(toas, fancy, thing)` so it can work too. `toas[fancy][:10] = thing` can't work, because fancy indexing requires the TOAs object to be copied (and indeed the deepcopy bug affected this). Can `toas[::2][:10] = thing` work? It depends on having "slices" of TOAs objects that reference the original; we don't support this. 

`toas["error"]` returns (a reference to) the Astropy column, which does actually allow `__setitem__`, so doing this can affect the parent TOAs object. `toas[::2]` has to return a fresh TOAs object, so you can't affect the original through it now that a bug is fixed. 

In `toas[anything] = values` -> `toas.__setitem__(anything, values)`, note that `anything` can be a tuple, so if we've implemented it you should be able to do things like `toas["error", ::2] = values` - it may make sense for us to make TOAs objects behave more like multidimensional (column/flag, TOA number) arrays, precisely to avoid any need to do multiple indexing.

aarchiba:
I should also say: don't expect modifying indexed objects (`t[::2].mutating_function()`) to work either, unless the API of the thing you're indexing specifically says it should.

dlakaplan:
I'm finding that any sort of array argument fails:
`t['test'] = ['1','2','3',...]`
it looks like that part of the `__setitem__` code wasn't fully finished.  

So `t['test']=...` works fine if the value is a single value (and can replace `set_flag_values()`).  If you want to set multiple different values then right now neither works.

dlakaplan:
OK, I'm not sure if what I did is optimal, but I made setting flags from array values work.  Note that I also put in the `str` conversion as part of that, so you don't have to pass a list of strings or string arrays.  

I also made is so that all flag setting will enforce `str()` if a scalar or array is passed.  I am not sure if this works with all use cases.

And because of this I was able to get rid of the `set_flag_values()` method.

PR #1324, opened 2022-07-04
Fix bug #1316

### LINKED ISSUE #1316: Dimensionless unit error for ELL1 binaries
In `fitter.py`, using the `get_derived_params` function (line 458) on a binary with the ELL1 model gives the error message
> TypeError: Argument 'E' to function 'ELL1_check' has no 'unit' attribute. You may want to pass in an astropy Quantity instead. 

It looks like when that function calculates the eccentricity it isn't assigned dimensionless units. Changing lines 592 and 600 to `ecc.nominal_value*u.dimensionless_unscaled,` fixes the issue. I'm not sure if there are similar issues in other classes.

aarchiba:
Uses `*u.s/u.s` to fix a problem that arises with `@quantity_input` and Astropy 4.0. Also adds the ability to run custom commands with `tox -e oldestdeps`. 

Closes #1316

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project, at least loosely, adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Changed
+### Added
+### Fixed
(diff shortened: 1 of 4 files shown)


dlakaplan:
Looks good to me, although I don't know what the change to `tox.ini` does.

aarchiba:
> Looks good to me, although I don't know what the change to `tox.ini` does.

That change makes it possible to use `tox -e oldestdeps -- pytest tests/test_thing_i_want.py::test_might_break_on_old_python`. I'm not sure where to document that, or how to make sure all our normal tox environments work this way.

PR #1325, opened 2022-07-04
Split Jodrell observatory based on backend

aarchiba:
The different Jodrell Bank instruments have different clock files to accommodate instrumental changes. This is encoded by using observatories `jbroach` or `jbdfb`, and PINT needs to support this.


dlakaplan:
Looks straightforward to me

PR #1329, opened 2022-07-05
Explicit deep copy of flags

### LINKED ISSUE #1328: Deepcopy of TOAs leaves shallow copy of TOA flags
Running:
```
import copy
import pint.config

m, t = get_model_and_toas(
    pint.config.examplefile("NGC6440E.par"), pint.config.examplefile("NGC6440E.tim")
)

tc = copy.deepcopy(t)
print(f"TOAs: {t is tc}")
print(f"TOA Table: {t.table is tc.table}")
print(f"Flags: {t.table[0]['flags'] is tc.table[0]['flags']}")
```
Gives:
```
TOAs: False
TOA Table: False
Flags: True
```
This is confusing: I would expect the deepcopy to yield independent flag objects, but it isn't.  This can have unexpected behavior if you e.g., manipulate the flags on a copy, you end up manipulating both sets.  The TOA object and the TOA table both seem to copy fine.

dlakaplan:
Fixes #1328.  Explicitly implements deepcopy of flags since they were left as a shallow copy.  Adds test for that (and other minor updates to `test_copy.py`).

### DIFF
--- a/tests/test_copy.py
+++ b/tests/test_copy.py
@@ -11,6 +11,7 @@ import astropy.units as u
 from pint.models import get_model
 from pint.fitter import WidebandTOAFitter
 import pint.fitter
+import pint.residuals
 from pint.toa import get_TOAs
 from pinttestdata import datadir
 
@@ -34,28 +35,36 @@ def toas():
 def test_copy_toa_object(toas):
     toa_copy = copy.deepcopy(toas)
 
-    assert sys.getsizeof(toas) == sys.getsizeof(toa_copy)
-    assert id(toas) != id(toa_copy)
+    assert toas == toa_copy
+    assert toas is not toa_copy
     assert toas.ntoas == toa_copy.ntoas
     assert all(toas.table["mjd"] == toa_copy.table["mjd"])
-    assert not (toas.table is toa_copy.table)
-    assert not (toas.table[0]["flags"] is toa_copy.table[0]["flags"])
+    assert toas.table is not toa_copy.table
+    assert toas.table[0]["flags"] is not toa_copy.table[0]["flags"]
+    for c in toas.table.colnames:
+        assert toas.table[c] is not toa_copy.table[c]
 
 
 def test_copy_model_object(model):
     model_copy = copy.deepcopy(model)
 
-    assert sys.getsizeof(model) == sys.getsizeof(model_copy)
-    assert id(model) != id(model_copy)
+    assert model is not model_copy
     assert len(model.params) == len(model_copy.params)
-    assert list(model.components.keys()) == list(model_copy.components.keys())
+    assert (model.components.keys()) == (model_copy.components.keys())
+    for p in model.params:
+        assert getattr(model, p) is not getattr(model_copy, p)
+
+
+def test_copy_residuals(model, toas):
+    r = pint.residuals.Residuals(toas, model)
+    r_copy = copy.deepcopy(r)
+    assert r is not r_copy
 
 
 def test_copy_fitter_object(model, toas):
     fitter = pint.fitter.Fitter(toas, model)
     fitter_copy = copy.deepcopy(fitter)
-    assert sys.getsizeof(fitter) == sys.getsizeof(fitter_copy)
-    assert not (fitter is fitter_copy)
+    assert fitter is not fitter_copy
 
 
 def test_copy_wideband_fitter_object():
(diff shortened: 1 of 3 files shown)


tests/test_copy.py
```diff
@@ -10,46 +10,69 @@
...
+    # The scope="module" setting ensures the TOAs object will be created
+    # only once for the whole module, which will save time but might
+    # allow accidental modifications done in one test to affect other tests.
+    return get_TOAs(timfile)
+
+
+def test_copy_toa_object(toas):
+    toa_copy = copy.deepcopy(toas)
+
+    assert sys.getsizeof(toas) == sys.getsizeof(toa_copy)
```
aarchiba, line 37: This is not a good idea. Dictionaries or lists can be equal but have different `getsizeof` because they have different amounts of allocated space.

tests/test_copy.py
```diff
@@ -10,46 +10,69 @@
...
+    assert id(toas) != id(toa_copy)
```
aarchiba, line 38: I think `is not` would be better.

tests/test_copy.py
```diff
@@ -10,46 +10,69 @@
...
+    assert toas.ntoas == toa_copy.ntoas
+    assert all(toas.table["mjd"] == toa_copy.table["mjd"])
+    assert not (toas.table is toa_copy.table)
+    assert not (toas.table[0]["flags"] is toa_copy.table[0]["flags"])
```
aarchiba, line 42: Do we need to add a check that the Column objects are not shared between the tables?

tests/test_copy.py
```diff
@@ -10,46 +10,69 @@
...
+
+
+def test_copy_model_object(model):
+    model_copy = copy.deepcopy(model)
+
+    assert sys.getsizeof(model) == sys.getsizeof(model_copy)
+    assert id(model) != id(model_copy)
+    assert len(model.params) == len(model_copy.params)
+    assert list(model.components.keys()) == list(model_copy.components.keys())
```
aarchiba, line 51: Is the `list` necessary here?

tests/test_copy.py
```diff
@@ -10,46 +10,69 @@
...
+
+    assert sys.getsizeof(toas) == sys.getsizeof(toa_copy)
+    assert id(toas) != id(toa_copy)
+    assert toas.ntoas == toa_copy.ntoas
+    assert all(toas.table["mjd"] == toa_copy.table["mjd"])
+    assert not (toas.table is toa_copy.table)
+    assert not (toas.table[0]["flags"] is toa_copy.table[0]["flags"])
+
+
+def test_copy_model_object(model):
```
aarchiba, line 45: This test could (should) check that the Parameter objects aren't shared - deepcopy should handle this, but they are mutable and if we're checking that the copy works okay we should check that.

tests/test_copy.py
```diff
@@ -10,46 +10,69 @@
...
+    assert id(model) != id(model_copy)
+    assert len(model.params) == len(model_copy.params)
+    assert list(model.components.keys()) == list(model_copy.components.keys())
+
+
+def test_copy_fitter_object(model, toas):
+    fitter = pint.fitter.Fitter(toas, model)
+    fitter_copy = copy.deepcopy(fitter)
+    assert sys.getsizeof(fitter) == sys.getsizeof(fitter_copy)
+    assert not (fitter is fitter_copy)
```
aarchiba, line 58: If we're being distrustful, we should maybe check that the models are different?

tests/test_copy.py
```diff
@@ -10,46 +10,69 @@
...
+def test_copy_wideband_fitter_object():
+    model = get_model(os.path.join(datadir, "J1614-2230_NANOGrav_12yv3.wb.gls.par"))
+    toas = get_TOAs(os.path.join(datadir, "J1614-2230_NANOGrav_12yv3.wb.tim"))
+    fitter = WidebandTOAFitter([toas], model, additional_args={})
+    fitter_copy = copy.deepcopy(fitter)
+
+    assert sys.getsizeof(fitter) == sys.getsizeof(fitter_copy)
+    assert id(fitter) != id(fitter_copy)
+    assert len(fitter.designmatrix_makers) == len(fitter_copy.designmatrix_makers)
+    for ii in range(len(fitter.designmatrix_makers)):
```
aarchiba, line 70: Is this better done with `zip(fitter.designmatrix_makers, fitter_copy.designmatrix_makers)`?

dlakaplan:
@aarchiba : your general comments on the tests are good, but I'll just note that those largely precede me.  I'll try to update some of them but may not do them all.

aarchiba:
> @aarchiba : your general comments on the tests are good, but I'll just note that those largely precede me. I'll try to update some of them but may not do them all.

This PR is necessary because we've realized that `deepcopy` wasn't doing what we wanted and because our tests weren't catching it. So I think having tests that would have caught it is a good idea.

dlakaplan:
Yes, I agree.  I added a test that does just that.  All I'm saying is I may not update every test - I'll try, but I have finite time.

aarchiba:
CI failures fixed in #1332, ready for merging if you're okay with it

aarchiba:
Looks good to me.

Issue #1331, opened 2022-07-05
design matrix uses a fixed F0


scottransom:
I'm working on updating random models to use the design matrix and while testing that, I noticed small differences with the current "exact" calculation which subtracts residuals.

In `timing_model.py`, the `designmatrix()` method, which we use in our fitters, currently converts from phase to time simply by dividing by `m.F0.value`.  For MSPs, this is likely fine, but for young pulsars, as @paulray has mentioned in the past, this can cause fairly large differences.

Anyone know if there is a reason why the design matrix is this way?  Is this compatible with Enterprise's use of the design matrix?

I'm happy to fix this (and potentially give the user an option of return the matrix in term of phase or time) if we agree that it is not the behavior that we should have.

dlakaplan:
I seem to recall a number of issues like this.  I feel like I fixed one or two, but I definitely didn't look exhaustively.  There's the `residuals.get_PSR_freq()` routine, which has `modelF0` as the default.  Could we change that to something else?  And I know the same math is implemented in several places (also in `spindown.py`, and probably elsewhere?).

paulray:
There was some earlier discussion here: https://github.com/nanograv/PINT/issues/866

aarchiba:
Also some discussion in #713 where it was also suggested that all such conversions go through a single function (`get_PSR_freq`?) and in #415

dlakaplan:
But I think that function should be moved to `timing_model` rather than `residuals`

scottransom:
Just to let people know, I'm going to move and improve `get_PSR_freq()` (and change its name) along with fixing this design matrix issue with my random model improvements PR.  They are all linked.  Hopefully you will see a PR over the next couple days.

PR #1334, opened 2022-07-09
WIP: Improvements to designmatrix and fast random models

scottransom:
This provides a new, faster, default method of computing random model residuals that uses the design matrix.  It appears to be factors of 2-6 faster than the earlier "exact" method, depending on how many TOAs there are.

In doing so, I improved the `designmatrix()` method of the timing models so that it doesn't use just a fixed `F0` to convert from phases to times, but allows also Taylor expansion calculations, numerical derivatives, and a phase-only method.  This is at least part of what has been discussed in #1331

To make the above work, I moved the fitter `get_PSR_freq()` method to become a timing model method called `get_spin_freq()`, which also takes the same options (except phase-only) that are mentioned above.

I think all the current tests pass and `pintk` works, but I haven't yet written new tests.


paulray:
If the calctype is `numerical` it isn't really returning the spin frequency right? It would be modified by orbital Doppler shifts and such. In fact, isn't it really returning the observed topocentric frequency, not the barycentric frequency?

scottransom:
Yeah, I have not yet tested the detailed differences between the different methods.  All I know for sure right now is that they give reasonably close values for an MSP binary system.  I'm truthfully not sure exactly what delays should be removed for which use cases.  It seems like most use cases would either require the spin rate at the SSB, or maybe at the pulsar system center.  I think it should be possible to choose which delays get applied for the numerical method so that you can get either of those or for the topocenter.  Maybe that is a good addition.  And the default would be SSB-like?  (which would include binary doppler shifts?)  There are a lot of options.

aarchiba:
> This provides a new, faster, default method of computing random model residuals that uses the design matrix. It appears to be factors of 2-6 faster than the earlier "exact" method, depending on how many TOAs there are.

Just to check: "how many TOAs there are" means "how many places you want to sample the random models", right? Not how many input TOAs there are? (I know the current random models splices a few real TOAs at real frequencies in with the fake TOAs at infinite frequencies, which leads to weirdness when fitting DMs.) 

- Could one speed up the process by keeping design matrix entries for any real TOAs one wanted to include from when they were calculated during the fitting? The Downhill fitters already cache design matrices, though they probably don't expose that to the user.
- One doesn't actually need to do a fit, just to compute the parameter covariance matrix, before one can use random matrices. Should fitters have a public interface for computing the covariance matrix only? (Even just "fit with zero iterations"?) Should there be a way to record parameter covariance matrices in, or adjacent to, par files?

src/pint/models/timing_model.py
```diff
@@ -1502,6 +1496,57 @@ def delete_jump_and_flags(self, toa_table, jump_num):
...
         self.components["PhaseJump"].setup()
 
+    def get_spin_freq(self, times, calctype="modelF0"):
+        """Return barycentric pulsar rotational frequency in Hz at specific times
+
+        Parameters
+        ----------
+        times : float or long double MJD (can be array), astropy.Time object, TOAs
+            The times (barycentric, inf freq) at which to compute the spin frequency
+        calctype : {'modelF0', 'numerical', 'taylor'}
```
aarchiba, line 1506: Perhaps a fourth option, that allowed passing in a fixed frequency, might be helpful to avoid fitter shenanigans where they realize they can rescale their errors by changing the F0 in the model? Or is that better handled by using the prefit model to do this computation?

src/pint/models/timing_model.py
```diff
@@ -1502,6 +1496,57 @@ def delete_jump_and_flags(self, toa_table, jump_num):
...
+        -------
+        freq : astropy.units.Quantity (can be array)
+            The spin frequency in Hz at each barycentric time
+
+        Note
+        ----
+        "numerical" is significantly slower, but is much more exact, and
+        """
+        calc = calctype.lower()
+        assert calc in ["modelf0", "taylor", "numerical"]
```
aarchiba, line 1523: Please raise a ValueError with a useful message to signal user error (like passing a wrong value to a function). Assertions are for detecting bugs in PINT internal code, and for use in tests.

src/pint/models/timing_model.py
```diff
@@ -1502,6 +1496,57 @@ def delete_jump_and_flags(self, toa_table, jump_num):
...
+
+        # ToDo: How does this handle Glitch or Piecewise models?
+        if calc == "numerical":
+            assert isinstance(times, TOAs)
+            return self.d_phase_d_toa(times)
+
+        # Handle various types of input times for "taylor" calc
+        if isinstance(times, TOAs):
+            # If we pass the TOA table, then barycenter the TOAs
+            bts = self.get_barycentric_toas(times)  # have units of days
```
aarchiba, line 1536: Do we actually want to barycenter the TOAs (like, make this dependent on sky position)? If so, should there be a faster shortcut computation method (`fasttaylor`) that just uses the un-barycentered times?

src/pint/models/timing_model.py
```diff
@@ -1502,6 +1496,57 @@ def delete_jump_and_flags(self, toa_table, jump_num):
...
+        elif isinstance(times, time.Time):
+            if times.scale == "tdb":
+                bts = np.asarray(times.mjd_long) << u.d
+            else:
+                raise ValueError("times as Time instance in needs scale=='tdb'")
+        elif isinstance(times, MJDParameter):
+            bts = np.asarray(times.value) << u.d  # .value is always a MJD long double
```
aarchiba, line 1543: If it's a Parameter maybe better to use `.quantity`, which always returns a Quantity ready to go?

src/pint/models/timing_model.py
```diff
@@ -1502,6 +1496,57 @@ def delete_jump_and_flags(self, toa_table, jump_num):
...
+        else:
+            bts = np.asarray(times) << u.d
```
aarchiba, line 1545: What if it's already a Quantity but accidentally in seconds?

src/pint/models/timing_model.py
```diff
@@ -1502,6 +1496,57 @@ def delete_jump_and_flags(self, toa_table, jump_num):
...
+        dts = bts - (self.PEPOCH.value << u.d)
```
aarchiba, line 1546: Again using the `.quantity` member of a parameter is probably better than stripping the units and then reapplying days.

src/pint/models/timing_model.py
```diff
@@ -1736,55 +1783,63 @@ def designmatrix(self, toas, acc_delay=None, incfrozen=False, incoffset=True):
         ----------
         toas : pint.toa.TOAs
             The TOAs at which to compute the design matrix.
-        acc_delay
-            ???
         incfrozen : bool
             Whether to include frozen parameters in the design matrix
         incoffset : bool
             Whether to include the constant offset in the design matrix
+        calctype : {'modelF0', 'numerical', 'taylor'}
```
aarchiba, line 1790: Is there a way to make the possibilities for `calctype` the same here as in other functions that accept this parameter? I guess it's a bit trivial for `get_spin_freq` but it's nicer to have a more uniform calling convention.

src/pint/pintk/plk.py
```diff
@@ -1068,11 +1068,11 @@ def plotResiduals(self, keepAxes=False):
                     else self.psr.postfit_resids
                 )
                 if self.y_unit == u.us:
-                    f0 = r.get_PSR_freq().to(u.MHz).value
+                    f0 = r.model.F0.quantity.to(u.MHz).value
```
aarchiba, line 1071: Should this be `r.model.get_spin_freq()` with some appropriate options? At a minimum so we can easily identify this as a place we do that conversion?

src/pint/models/timing_model.py
```diff
@@ -1502,6 +1496,57 @@ def delete_jump_and_flags(self, toa_table, jump_num):
             return
         self.components["PhaseJump"].setup()
 
+    def get_spin_freq(self, times, calctype="modelF0"):
```
aarchiba, line 1499: Should this allow `times=None` when just using `modelF0` to avoid forcing users to make up a bogus time?

src/pint/simulation.py
```diff
@@ -8,8 +8,9 @@
 from loguru import logger as log
 from astropy import time
 
-import pint.residuals
+import pint.residuals as pr
```
aarchiba, line 11: Any reason not to just use `from pint.residuals import Residuals`?

src/pint/simulation.py
```diff
@@ -474,36 +435,175 @@ def calculate_random_models(
...
 
-    f_rand = deepcopy(fitter)
-
     # scale by fac
     mean_vector = mean_vector * fac
     scaled_cov_matrix = ((cov_matrix.matrix * fac).T * fac).T
     random_models = []
-    for imodel in range(Nmodels):
+    for _ in range(Nmodels):
+        f_rand = deepcopy(fitter)
```
aarchiba, line 443: Is there any reason to copy the Fitter and not just the model?

src/pint/simulation.py
```diff
@@ -474,36 +435,175 @@ def calculate_random_models(
...
+        # dividing by fac brings us back to real units
+        rparams = np.random.multivariate_normal(mean_vector, scaled_cov_matrix) / fac
+        f_rand.set_params(dict(zip(param_names, rparams)))
+        random_models.append(f_rand.model)
+    return random_models
+
+
+def compute_random_model_resids_exact(fitter, models, toas, return_time=False):
+    """
+    Calculate residuals from random models exactly
```
aarchiba, line 454: Docstring first lines should be on the first line, and for functions they should be full sentences. These serve as the short summary in function lists.

src/pint/simulation.py
```diff
@@ -474,36 +435,175 @@ def calculate_random_models(
...
+        rn = pr.Residuals(toas, model, subtract_mean=False)
+        resids[ii] = (
+            rn.time_resids - r0.time_resids
+            if return_time
+            else rn.phase_resids - r0.phase_resids
+        )
+    return resids << u.s if return_time else resids
+
+
+def compute_random_model_resids_fast(fitter, models, toas, return_time=False):
```
aarchiba, line 489: The design matrix can also be directly used to compute one-sigma "error bands", with a small additional amount of linear algebra. Such error bands don't fully capture the shape of deviations but can be a very tidy way to summarize the uncertainty in a model. Would it be good to have a function to compute these and a plotting mode in pintk to show them?

scottransom:
> Just to check: "how many TOAs there are" means "how many places you want to sample the random models", right? Not how many input TOAs there are? (I know the current random models splices a few real TOAs at real frequencies in with the fake TOAs at infinite frequencies, which leads to weirdness when fitting DMs.)

Yes, that is correct.  I'll edit that in the description.

> * Could one speed up the process by keeping design matrix entries for any real TOAs one wanted to include from when they were calculated during the fitting? The Downhill fitters already cache design matrices, though they probably don't expose that to the user.

I think cached design matrix calculations would be super useful.

> * One doesn't actually need to do a fit, just to compute the parameter covariance matrix, before one can use random matrices. Should fitters have a public interface for computing the covariance matrix only? (Even just "fit with zero iterations"?) Should there be a way to record parameter covariance matrices in, or adjacent to, par files?

That would also be really useful, I think.

src/pint/models/timing_model.py
scottransom: Where would that fixed frequency come from?  Would that be perhaps passed as an optional argument?  Although, I guess, as you are saying, it could default to being the prefit F0....

src/pint/models/timing_model.py
scottransom: Yeah, I don't know the answer to the first question.  I think it depends on your use case and how precise you want things.  A `fasttaylor` method is probably a good idea no matter what, though.

src/pint/models/timing_model.py
scottransom: Because `.quantity` for an `MJDParameter` returns an astropy `Time` instance, which doesn't work for us here (at least not in an easy way).

src/pint/models/timing_model.py
scottransom: Good question.  Maybe a ValueError since the input times are supposed to be dates (i.e. MJDs, if an array)?

src/pint/models/timing_model.py
scottransom: `PEPOCH` is an `MJDParam`, so it returns a `Time` instance and not a day quantity.  So I think we have to do it something like this unless we want to work with astropy `TimeDelta`s, and we definitely don't want that.

src/pint/models/timing_model.py
scottransom: They are all identical now (including with the docstrings), except that `.designmatrix` also allows a `phase` option.

src/pint/pintk/plk.py
scottransom: Maybe?  But `get_spin_freq` requires times to be specified.  What time are we going to use for plotting?  (and that is all this is used for:  the 2nd axis of the residual plots)

src/pint/models/timing_model.py
scottransom: That's not a bad idea.  Although if you know you are not giving times *and* explicitly specifying "modelF0", why wouldn't you just use `m.F0.quantity` rather than calling a method?

src/pint/simulation.py
scottransom: Great idea!  And that sped up random model creation a bunch since the potentially large TOA objects aren't being copied, either.  Thanks for that.  I've pushed that up.

src/pint/simulation.py
scottransom: This is basically exactly how/why we use the random models now in `pintk`.  But I definitely think it might be useful to be able to do this computation, for instance, at the time of a non-phase-connected TOA in the future to see if it is likely phase-connected or not.  We use random models now for that, but this is probably a much more "surgical" option. Do you know of a ref for the linear algebra to do this?

src/pint/models/timing_model.py
aarchiba: My suggestion was that the caller be allowed to supply it, whether as an optional argument or actually as a floating-point number in `calctype`. But I'm not sure that's the right way to allow callers to avoid updating the frequency used mid-fit. Given the current implementation, Fitter instances could ensure that the model this was called on was the pre-fit model, not the model updated to reflect fitting.

PR #1335, opened 2022-07-09
Pl red noise bugfix

abhisrkckl:
Fixing Issue https://github.com/nanograv/PINT/issues/1286 : PINT could not automatically load PLRedNoise component from par file before.

### DIFF
--- a/tests/test_plrednoise.py
+++ b/tests/test_plrednoise.py
@@ -44,10 +44,10 @@ def modelJ0023p0923():
     return mb.get_model(io.StringIO(parfile_contents))
 
 def test_read_PLRedNoise_component(modelJ0023p0923):
-    assert ('PLRedNoise' in modelJ0023p0923.components)
+    assert 'PLRedNoise' in modelJ0023p0923.components
 
 def test_read_PLRedNoise_component_type(modelJ0023p0923):
-    assert (modelJ0023p0923.components['PLRedNoise'] in modelJ0023p0923.NoiseComponent_list)
+    assert modelJ0023p0923.components['PLRedNoise'] in modelJ0023p0923.NoiseComponent_list
 
 def test_read_PLRedNoise_params(modelJ0023p0923):
     params = ['TNREDAMP', 'TNREDGAM', 'TNREDC']
--- a/tests/test_plrednoise.py
+++ b/tests/test_plrednoise.py
@@ -44,16 +44,14 @@ def modelJ0023p0923():
     return mb.get_model(io.StringIO(parfile_contents))
 
 def test_read_PLRedNoise_component(modelJ0023p0923):
-    assert ('PLRedNoise' in modelJ0023p0923.components), \
-        "PLRedNoise component was not created while creating TimingModel."
+    assert ('PLRedNoise' in modelJ0023p0923.components)
 
 def test_read_PLRedNoise_component_type(modelJ0023p0923):
-    assert (modelJ0023p0923.components['PLRedNoise'] in modelJ0023p0923.NoiseComponent_list), \
-        "PLRedNoise was not correctly identified as a NoiseComponent."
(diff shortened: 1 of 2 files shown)


aarchiba:
Could you please add a test - just a simple par file that couldn't be read before but can now? Doesn't have to be a real par file, just a simple fake one with the necessary bits. You'll see we often use a StringIO to do this so we don't have to add new par files to the distribution.

abhisrkckl:
I have added a test case.

tests/test_plrednoise.py
```diff
@@ -0,0 +1,77 @@
...
+    TNRedGam 0.125393
+    TNRedC 14
+"""
+
+@pytest.fixture()
+def modelJ0023p0923():
+    return mb.get_model(io.StringIO(parfile_contents))
+
+def test_read_PLRedNoise(modelJ0023p0923):
+    assert ('PLRedNoise' in modelJ0023p0923.components
```
aarchiba, line 68: Separate assertions are preferable, you get easier to understand error messages.

tests/test_plrednoise.py
```diff
@@ -0,0 +1,77 @@
...
+                and modelJ0023p0923.components['PLRedNoise'] in modelJ0023p0923.NoiseComponent_list
+                and hasattr(modelJ0023p0923, 'TNREDAMP')
+                and hasattr(modelJ0023p0923, 'TNREDGAM')
+                and hasattr(modelJ0023p0923, 'TNREDC')
+            ), "PLRedNoise test failed."
```
aarchiba, line 73: `pytest` gives much more informative errors if you *don't* include an error message - it displays the expression that failed the assertion along with values for various things in it.

tests/test_plrednoise.py
```diff
@@ -0,0 +1,77 @@
...
+
+if __name__ == "__main__":
```
aarchiba, line 75: This isn't helpful.

tests/test_plrednoise.py
```diff
@@ -0,0 +1,77 @@
+"""Tests to ensure that ModelBuilder is able to read PLRedNoise 
+component properly."""
+
+import io
+import pytest
+
+import pint.models.model_builder as mb
+
+parfile_contents = """
```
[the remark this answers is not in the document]
abhisrkckl: I have removed a bunch of unnecessary parameters. I have left one each of TNEF TNEQ and TNECORR parameters just to make sure that the change does not affect other noise components.

tests/test_plrednoise.py
abhisrkckl: I have added better error messages.

aarchiba: Er, I meant this specific line - `if __name__=="__main__":` - it's not helpful with `pytest`.

tests/test_plrednoise.py
```diff
@@ -0,0 +1,63 @@
...
+    TNRedC 14
+"""
+
+@pytest.fixture()
+def modelJ0023p0923():
+    return mb.get_model(io.StringIO(parfile_contents))
+
+def test_read_PLRedNoise_component(modelJ0023p0923):
+    assert ('PLRedNoise' in modelJ0023p0923.components), \
+        "PLRedNoise component was not created while creating TimingModel."
```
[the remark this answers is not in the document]
aarchiba: If it helps you can try breaking something and see what pytest's messages look like. You can run the tests in a specific file with `pytest tests/test_specific.py`.

abhisrkckl:
> If you run `pre-commit install` in your usual PINT development directory/virtualenv, it should squawk at you if the code you're about to commit isn't blackened.
Ah! I was wondering why one of the tests was always failing. So it was a formatting issue. Fixed that now.

PR #1337, opened 2022-07-12
Allow clock files that contain only sporadic updates

### LINKED ISSUE #1326: PINT doesn't know how to handle clock files that are valid past their ends
The Jodrell Bank instruments DFB and Roach have clock files that just record time offsets due to instrumental changes. These can be up to date long past the last instrumental change. This is in contrast with the main Jodrell Bank clock file, which is *not* valid past the last entry because it is a measurement of the clock drift.

How can PINT distinguish between these two situations, to let users know appropriately when they have out of date clock files?

aarchiba:
Previously we assumed any clock file was only valid as far as its last (real) entry. This is not the case for files like `jbroach2jb.clk`, which contain entries only when the instrumental delay changes, and so users will get inappropriate warnings about out-of-date clock files. This PR fixes that.

As a side effect, it needs to rewrite `observatories.json` to reflect a new format with room for individual clock files to have metadata. 

Closes #1326

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - index.txt is only checked at most once a day
 - Moved observatories to JSON file.  Changed way observatories are loaded/overloaded
 - Split Jodrell Bank observatory based on backend to get correct clock files
+- Clock files can be marked as being valid past the end of the data they contain
 ### Added
 - delta_pulse_number column is now saved to -padd flag on TOA write
 ### Fixed
--- a/docs/examples/PINT_observatories.py
+++ b/docs/examples/PINT_observatories.py
@@ -44,7 +44,7 @@ print(gbt)
 # The observatory also includes info on things like the clock file:
 
 # %%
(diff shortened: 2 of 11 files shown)


dlakaplan:
I think this is good, but I'm wondering if we need better documentation of `observatories.json` (what entries are allowed/required) or if it's obvious.  This could also include explicit info about how to get the right IRTF coordinates from lat/lon/elevation.  Or maybe that is for a different PR.

aarchiba:
I think documentation of `observatories.json` would be good, possibly including tools to manipulate it (beyond the comparison with tempo/tempo2 function we have); lat/long/alt->xyz should be available from astropy so it would be an easy add to the tools. Or one could make the construction of TopoObs more flexible (for example permitting lat/long/alt) and then allow exporting constructed objects to `observatories.json` format.

That said, this PR makes fairly minimal changes to `observatories.json`, much less significant than I thought I would need. So I'm not sure that this should wait on that documentation or those tools.

dlakaplan:
If you don't want to do the docs in this PR, then that's reasonable.  So if you are otherwise done, I can merge it.

PR #1340, opened 2022-07-13
Compare parfiles

dlakaplan:
* Command-line utility to compare par files
* Use colors for output
* Bugfixes to reduce crashes

### DIFF
--- a/docs/examples/understanding_fitters.py
+++ b/docs/examples/understanding_fitters.py
@@ -123,7 +123,7 @@ powfit.print_summary()
 # There also a convenience function for pretty printing a comparison of two models with the differences measured in sigma.
 
 # %%
-display_markdown(wlsfit.model.compare(powfit.model, output="markdown"), raw=True)
+display_markdown(wlsfit.model.compare(powfit.model, format="markdown"), raw=True)
 
 # %% [markdown]
 # ## Generalized Least Squares fitter
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -1793,11 +1793,12 @@ class TimingModel:
         self,
         othermodel,
         nodmx=True,
+        convertcoordinates=True,
         threshold_sigma=3.0,
         unc_rat_threshold=1.05,
         verbosity="max",
         usecolor=True,
-        output="text",
+        format="text",
     ):
         """Print comparison with another model
 
@@ -1805,9 +1806,11 @@ class TimingModel:
         ----------
         othermodel
             TimingModel object to compare to
-        nodmx : bool
-            If True (which is the default), don't print the DMX parameters in
+        nodmx : bool, optional
+            If True, don't print the DMX parameters in
             the comparison
+        convertcoordinates : bool, optional
+            Convert coordinates from ICRS<->ECL to make models consistent
         threshold_sigma : float, optional
             Pulsar parameters for which diff_sigma > threshold will be printed
             with an exclamation point at the end of the line
@@ -1824,7 +1827,7 @@ class TimingModel:
                 "check"   - only print significant changes with logging.warning, not as string (note that all other modes will still print this)
         usecolor : bool, optional
             Use colors on the output to complement use of "!" and "*"
-        output : string, optional
+        format : string, optional
             One of "text" or "markdown"
 
         Returns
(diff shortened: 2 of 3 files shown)


dlakaplan:
Usage:
```
(pintdev) kaplan@plock[~/pythonpackages/PINT_dlakaplan] (comparepar) % compare_parfiles -h                         
usage: compare_parfiles [-h] [--dmx] [--no-dmx] [--sigma SIGMA] [--uncertainty_ratio UNCERTAINTY_RATIO] [--nocolor]
                        [--comparison {max,med,min,check}] [--log-level {TRACE,DEBUG,INFO,WARNING,ERROR}]
                        input1 input2

PINT tool to compare parfiles

positional arguments:
  input1                First input parfile
  input2                Second input parfile

optional arguments:
  -h, --help            show this help message and exit
  --dmx                 Print DMX parameters (default: False)
  --no-dmx              Do not print DMX parameters (default: False)
  --sigma SIGMA         Pulsar parameters for which diff_sigma > threshold will be printed with an exclamation point at the end of the line (default:
                        3)
  --uncertainty_ratio UNCERTAINTY_RATIO
                        Pulsar parameters for which the uncertainty has increased by a factor of unc_rat_threshold will be printed with an asterisk
                        at the end of the line (default: 1.05)
  --nocolor             Turn off colorized output (default: False)
  --comparison {max,med,min,check}
                        "max" - print all lines from both models whether they are fit or not (note that nodmx will override this); "med" - only print
                        lines for parameters that are fit; "min" - only print lines for fit parameters for which diff_sigma > threshold; "check" -
                        only print significant changes with logging.warning, not as string (note that all other modes will still print this)
                        (default: max)
  --log-level {TRACE,DEBUG,INFO,WARNING,ERROR}
                        Logging level (default: WARNING)
```

dlakaplan:
@scottransom : I am trying to colorize the output.  But I want to just adjust the background color while leaving the foreground static.  is there a way to do that?

scottransom:
Hmmm.  Good question.  I think to do that you would somehow have to query your terminal environment to know what the current foreground and background colors are.  I certainly don't know of a way to do that.

aarchiba:
Would it be possible to (optionally?) output a Markdown table, where markdown highlights (`*...*` or `/.../` say) were used to indicate things of interest? A Markdown table is quite a clear ASCII representation but also allows a web browser to render a nicer version. In particular there are libraries to handle displaying Markdown nicely (with colours!) on a terminal: https://python.plainenglish.io/dump-the-plain-text-help-in-your-command-line-interfaces-today-3282225274dd

dlakaplan:
I was part-way to turning it into more table-like output anyway, so if I finished that adding markdown would be trivial.

I did search on how to find the current foreground/background colors, but couldn't find a good answer.  This is mostly relevant since e.g., I have white text on black as the default, so just fg=black isn't right.

dlakaplan:
Here is what the markdown output looks like:
Screen Shot 2022-07-13 at 2 40 36 PM

almcewen0:
i think it looks great on the command line! it doesn't run in a jupyter notebook, but that might be better suited for the model.compare() function.

dlakaplan:
The command-line version wraps `model.compare()`.  it's changes to that function which do markdown in a notebook

swiggumj:
This looks great. Yeah, I think so!... I'd have to check, but I think it also might help clean up some of the code where things get compiled into per-pulsar summary files.

dlakaplan:
Now included in https://nanograv-pint--1340.org.readthedocs.build/en/1340/examples/understanding_fitters.html#Comparing-models

aarchiba:
> This looks great. Yeah, I think so!... I'd have to check, but I think it also might help clean up some of the code where things get compiled into per-pulsar summary files.

Indeed. This could (should!) be used directly in the timing_analysis notebooks so you can show the comparison. For the summaries, right now the table is shown as preformatted, but this would allow us to produce much nicer-looking tables as a drop-in replacement.

dlakaplan:
Were you thinking that the `fit.get_summary()` method should also return markdown?   Possible - I just didn't know if it would be useful.  But I can add it.

scottransom:
That looks sweet.  But looking at the readthedocs example, what is up with F1?!

dlakaplan:
@scottransom : actually wait a second.  I want to update the docs slightly.

scottransom:
Oops!  I saw that too late!

PR #1346, opened 2022-07-15
Typo fix

abhisrkckl:
This typo causes TimingModel.compare() function to fail when comparing models with ECL vs ICRS coordinate systems.


aarchiba:
Can you write a short test that triggers this bug (that is, fails before but works now)?

dlakaplan:
This looks like it just fixes a typo I introduced.  Which I probably should have tested for before - maybe @abhisrkckl can put in a quick test as @aarchiba requests.  J1713 has models with both ICRS and ECL coordinates that could be used.

abhisrkckl:
Created a test that triggers the bug.

dlakaplan:
I think your test is good, but could you add versions with `convertcoordinates=True` and `convertcoordinates=False`?

Also, using the parfiles the way you have is fine, but you could also:
```
parfileICRS = datadir / "tests/datafile/J1713+0747_NANOGrav_11yv0_short.gls.ICRS.par"
parfileECL = datadir / "tests/datafile/J1713+0747_NANOGrav_11yv0_short.gls.par"
```
or similar, although that might include other parameters you don't want to include (like DMXs) that make it slower.

dlakaplan:
P.S.  thanks for catching this.

abhisrkckl:
@dlakaplan  I have added the tests you suggested.

dlakaplan:
great.  I think if you retrigger the CI and get all green, this should be good to go

Issue #1310, opened 2022-06-28
Do derivatives wrt noise parameters make sense in design matrix?


abhisrkckl:
Context: I am trying to implement the fitting of noise parameters.

The designmatrix() function tries to compute derivatives w.r.t. noise parameters if they are unfrozen. Right now this throws an exception because the derivative functions are not implemented. Does this behavior make sense? 

I think the derivatives w.r.t. noise parameters should either be excluded from the design matrix or be set to zero (because timing residuals do not depend on noise parameters).

aarchiba:
I don't think it will be possible to fit for noise parameters using PINT's Fitter framework. Those are designed to operate as least-squares fitters, that is, their goal is to produce the smallest possible scaled sum of squares of residuals. (It's a little more complicated with correlated errors but essentially the same.) If you allow them to adjust EFAC (say) they will correctly conclude that they can produce really small scaled sums of residuals by shooting EFAC through the roof. 

"Fitting" noise parameters will require some kind of statistical analysis that sets some other, appropriate goal. The Bayesian fitters we used for the 15-year data release adjusted the noise parameters so that the distribution of residuals matched the idealized distribution of residuals as well as possible. (And they used a linearized version of the timing model to speed things up.)

There might be a frequentist analogue to this, but it's not clear how to formulate it - if all you have is a heap of non-overlapping EFACs or EQUADs, you might be able to adjust them so that the reduced chi-squared of the TOAs they affect was 1, but this won't be enough if you have both EFACs and EQUADs affecting the same TOAs, or if you have ECORRs or red noise parameters. I think Bayesian is the way to go here.

abhisrkckl:
I think the design matrix should not contain derivatives wrt noise parameters at all, since the timing residuals do not depend on them, and the derivatives of the phase or delay w.r.t. noise parameters don't make sense. I suggest putting a condition in the designmatrix(..) function to ensure that only the parameters associated with PhaseComponents and DelayComponents (and OFFSET) get designmatrix entries.

aarchiba:
If you're going to put that sort of condition in, I suggest that it instead raise an exception - if someone has set noise parameters to be free in their model, the Fitter is going to have a serious problem with that, especially if they are not included in the design matrix. Ideally parameter objects would individually be able to specify that they are not fittable, at which point trying to set them as free would raise an immediate exception.

abhisrkckl:
[...]

PR #1347, opened 2022-07-21
Option for excluding noise parameters in design matrix.

abhisrkckl:
Related to issue #1310. This PR does not fully address this issue as it is still under discussion. However, it does
(1) Raise a NotImplementedError if the user tries to compute the design matrix with a noise parameter unfrozen with a clearer error message. Earlier, it used to raise an AttributeError complaining about missing derivative functions.
(2) Provide an option to exclude noise parameters from the design matrix.

I have made the default behavior of the TimingModel.designmatrix() function the same as before to avoid breaking any existing code.


tests/test_designmatrix_noisepars.py
```diff
@@ -0,0 +1,46 @@
+import os
+import pint.models.model_builder as mb
+import pinttestdata
+import pytest
+
+
+@pytest.fixture()
+def model_and_toas():
+    datadir = pinttestdata.datadir
+    parfile = os.path.join(datadir, "J1713+0747_NANOGrav_11yv0_short.gls.par")
```
aarchiba, line 10: Does this par file have the noise parameters unfrozen? What makes it special? Any reason not to use a simpler one with a set of TOAs that is much faster to load?

tests/test_designmatrix_noisepars.py
```diff
@@ -0,0 +1,46 @@
...
+    timfile = os.path.join(datadir, "J1713+0747_NANOGrav_11yv0_short.tim")
+    return mb.get_model_and_toas(parfile, timfile)
+
+
+def _test_designmatrix(model, toas):
```
aarchiba, line 15: Better to directly express the test conditions in place rather than call this utility function - it makes it hard to tell from the test code what the expected behaviour is.

src/pint/models/timing_model.py
```diff
@@ -1744,6 +1756,8 @@ def designmatrix(self, toas, acc_delay=None, incfrozen=False, incoffset=True):
             Whether to include frozen parameters in the design matrix
         incoffset : bool
             Whether to include the constant offset in the design matrix
+        incnoise : bool
```
aarchiba, line 1759: Will it ever be a good idea to include noise parameters in the design matrix? Is there a reason to create this non-functional parameter?

tests/test_designmatrix_noisepars.py
```diff
@@ -0,0 +1,46 @@
...
+        # This should raise a NotImplementedError rather than an AttributeError.
+        M = model.designmatrix(toas)
+    except NotImplementedError:
+        pass
+
+    # This should work.
+    M = model.designmatrix(toas, incnoise=False)
+
+
+def test_designmatrix_noise_exception_falsepositive(model_and_toas):
```
aarchiba, line 26: This title is unclear. Perhaps in this situation no noise parameter is free and you want to make sure the design matrix can be created? Can the title say that?

src/pint/models/timing_model.py
```diff
@@ -552,6 +552,16 @@ def get_params_dict(self, which="free", kind="quantity"):
                 raise ValueError(f"Unknown kind {kind!r}")
         return c
 
+    def get_params_of_component_type(self, component_type):
```
aarchiba, line 555: This function needs documentation.

src/pint/models/timing_model.py
abhisrkckl: I did it this way to ensure that the default behavior of PINT remains unchanged so that nothing else breaks.

tests/test_designmatrix_noisepars.py
```diff
@@ -0,0 +1,46 @@
...
+    parfile = os.path.join(datadir, "J1713+0747_NANOGrav_11yv0_short.gls.par")
+    timfile = os.path.join(datadir, "J1713+0747_NANOGrav_11yv0_short.tim")
+    return mb.get_model_and_toas(parfile, timfile)
+
+
+def _test_designmatrix(model, toas):
+    try:
+        # This should raise a NotImplementedError rather than an AttributeError.
+        M = model.designmatrix(toas)
+    except NotImplementedError:
```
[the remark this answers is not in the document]
abhisrkckl: OK.. I'll change that.

tests/test_designmatrix_noisepars.py
abhisrkckl: I just picked one which had EFACs, EQUADs, and ECORRs.

tests/test_designmatrix_noisepars.py
abhisrkckl: OK.. I'll remove the utility function.

tests/test_designmatrix_noisepars.py
abhisrkckl: I have changed the function name so that it is clearer.

src/pint/models/timing_model.py
abhisrkckl: Added dosctring.

dlakaplan:
@aarchiba : should this be merged?

src/pint/models/timing_model.py
aarchiba: I think this parameter should be removed, and noise parameters not included. Any code that expects them to be included is currently broken, since none of them have correct derivatives.

tests/test_designmatrix_noisepars.py
aarchiba: Can you create a really simple one in-line? This test is likely to take unnecessarily long if you use this very large data set. It would also be best if it were obvious what is needed from the par file.

tests/test_designmatrix_noisepars.py
```diff
@@ -0,0 +1,62 @@
+import os
+import pint.models.model_builder as mb
+import pinttestdata
+import pytest
+
+
+@pytest.fixture()
+def model_and_toas():
+    datadir = pinttestdata.datadir
```
aarchiba, line 9: Better to just do `from pinttestdata import datadir`, above.

aarchiba: You can also use `datadir / "file.tim"` instead of `os.path.join`.

tests/test_designmatrix_noisepars.py
```diff
@@ -0,0 +1,62 @@
...
+    parfile = os.path.join(datadir, "J1713+0747_NANOGrav_11yv0_short.gls.par")
+    timfile = os.path.join(datadir, "J1713+0747_NANOGrav_11yv0_short.tim")
+    return mb.get_model_and_toas(parfile, timfile)
+
+
+def test_designmatrix_noise_exception_no_free_noise_param(model_and_toas):
```
aarchiba, line 15: This name seems to suggest that it expects an exception?

tests/test_designmatrix_noisepars.py
```diff
@@ -0,0 +1,62 @@
...
+    model.EFAC1.frozen = False
+
+    with pytest.raises(NotImplementedError):
+        # This should raise a NotImplementedError rather than an AttributeError.
+        M = model.designmatrix(toas)
+
+    # This should work.
+    M = model.designmatrix(toas, incnoise=False)
+
+    model.EFAC1.frozen = True
```
aarchiba, line 34: The model and TOAs are created anew for each test, so no need to worry about resetting it.

tests/test_designmatrix_noisepars.py
abhisrkckl: OK. I will create a smaller dataset for this.

tests/test_designmatrix_noisepars.py
abhisrkckl: I was not sure about this. I thought the same objects were used for all tests.

tests/test_designmatrix_noisepars.py
abhisrkckl: Yes.. That's confusing. I'll rename that.

src/pint/models/timing_model.py
abhisrkckl: I'll remove the incnoise parameter.

abhisrkckl:
I have made the following changes:
1. Created a new dataset with a simpler par file and less number of TOAs
removed the 'incnoise' parameter

abhisrkckl:
I have made the following changes:
1. Created a new dataset with a simpler par file and less number of TOAs
2. removed the 'incnoise' parameter
3. Simplified the unit test

tests/test_designmatrix_noisepars.py
aarchiba: The `pytest` documentation on fixtures has some detail on this, but the idea is to ensure that tests cannot contaminate each other's results by mangling the fixture objects. This does mean that fixture objects are re-created frequently, so it is important to ensure that they are as light-weight as possible.

tests/test_designmatrix_noisepars.py
```diff
@@ -0,0 +1,33 @@
...
+    parfile = datadir / "J1713+0747_small.gls.par"
+    timfile = datadir / "J1713+0747_small.tim"
+    return get_model_and_toas(parfile, timfile)
+
+
+def test_designmatrix_no_free_noise_param(model_and_toas):
+    model, toas = model_and_toas
+
+    # This should work.
+    M = model.designmatrix(toas)
```
aarchiba, line 18: Are there any properties M should have that you can `assert` here? Or if you just want the function call to not raise an exception, maybe say so in the function name? (Also you don't have to assign the return value to anything if you never use it.)

tests/test_designmatrix_noisepars.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_designmatrix_free_noise_params(model_and_toas):
```
aarchiba, line 21: Please use `pytest.mark.parametrize` to automatically turn this into three tests, one for each value of `noiseparam`. This will ensure separate pass/failure for each of the three, and also simplify the test and avoid the need to try to reset the model at the end of each one.

tests/test_designmatrix_noisepars.py
abhisrkckl: I have added new assertions to test for the shape of the design matrix.

abhisrkckl:
The failed CI test is due to a numpy/astropy version issue.

paulray:
For some reason the version specification in oldestdeps is `1.17.*` but the version being installed is `1.23.3`. Not sure why this is happening, but it is causing all of our oldestdeps CI runs to fail.

Issue #1349, opened 2022-07-22
Move all data files from docs/examples directory to src/data/pint/examples/


abhisrkckl:
The example data should all be collected in one place. Not having everything in one place causes some of the example scripts to break as pkg_resources can't find the correct data files. Issue #1345 is a symptom of this.

dlakaplan:
I think this is largely done, but possibly some files escaped because they had different extensions (like `.good`, as you mention).  So I agree that it would be good to move all of them.

abhisrkckl:
OK.. I'll move them to the correct place.

aarchiba:
Done?

Issue #1345, opened 2022-07-15
fit_NGC6440E_MCMC.py is broken


abhisrkckl:
Running this file throws a `FileNotFoundError` for `NGC6440E.par.good`. It works if I change the par file to `NGC6440E.par`.
The issue is that `NGC6440E.par.good` is not getting installed when I do `setup.py install`.

aarchiba:
Can you check that this works now?

scottransom:
This is working for me if I simply run the script from the command line while in the `docs/examples` directory

abhisrkckl:
It works now.

PR #1350, opened 2022-07-22
Moved example data to the correct location (src/pint/data/examples/)

abhisrkckl:
Fixes issue #1349 (and also issue #1345, which is a symptom of the more general issue #1349.)


abhisrkckl:
Based on running grep:
 `J0613-sim.par.good` is not used in any of the scripts.
`NGC6440E.par.good` is used in `MCMC_walkthrough.broken` and `fit_NGC6440E_MCMC.py`
`J0030+0451_P8_15.0deg_239557517_458611204_ft1weights_GEO_wt.gt.0.4.fits` is used in  `MCMC_walkthrough.broken` 
`templateJ0030.3gauss` is used in  `MCMC_walkthrough.broken` 

`fit_NGC6440E_MCMC.py` should work after moving these files as it accesses the example files via `pint.config.examplefile` function. (I have tested this.)

I don't know about `MCMC_walkthrough.broken`. Maybe it is planned to be fixed in the future?

dlakaplan:
Maybe it's better to remove those files which are not used by any example?

paulray:
Please keep the NGC6440E and J0030 files.

abhisrkckl:
I have also checked the test scripts. Only ` J0613-sim.par.good` can be safely deleted.

PR #1359, opened 2022-07-26
Raise exception for astropy <4 in __init__.py

### LINKED ISSUE #1354: `import pint` fails when astropy version is less than 4
Here is how to reproduce the issue:

```
>>> import astropy
>>> astropy.version.major=3
>>> import pint
   File "<stdin>", line 1, in <module>
   File "/home/abhimanyu/anaconda3/envs/pint-devel/lib/python3.9/site-packages/pint_pulsar-0.7+2686.gbfa227ef.dirty-py3.9.egg/pint/__init__.py", line 110, in <module>
     log.warning(
 NameError: name 'log' is not defined
```

dlakaplan:
Fixes #1354 

astropy >=4 is enforced by `requirements.txt`

### DIFF
--- a/src/pint/__init__.py
+++ b/src/pint/__init__.py
@@ -103,12 +103,3 @@ pint_units = {
     "pulse phase": u.dimensionless_unscaled,
     "hourangle_second": hourangle_second,
 }
-
-import astropy.version
-
-if astropy.version.major < 4:
-    log.warning(
-        "Using astropy version {}. To get most recent IERS data, upgrade to astropy >= 4.0".format(
-            astropy.__version__
-        )
-    )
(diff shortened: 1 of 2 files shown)


paulray:
I guess this is OK, but we've definitely had people install PINT, then upgrade via git pull and never need to run `pip` again so that requirements.txt doesn't get applied. This happened to many of us with loguru, for example.

dlakaplan:
If that's the case, then probably raising an exception is better than emitting a warning.

paulray:
Indeed, especially for something as old as astropy 3.

dlakaplan:
Now astropy < 4 raises an exception, and there is a test for that.

PR #1361, opened 2022-07-27
Update topo_obs to expose geodetic coords, allow geodetic input, allow JSON output

dlakaplan:
Properties: `o.x,o.y,o.z`, `o.lat,o.lon,o.alt` are now exposed.

Also properties `o.as_dict` and `o.as_json` to give a custom serialization.

Can input observatory coordinates as lat/lon/alt instead of x/y/z.

Appropriate tests included.


dlakaplan:
@aarchiba : this still doesn't do the documentation, but it does some of the other changes we were discussing.  You can input an observatory as geocentric or geodetic, and you can output as JSON (I'm not sure if they way I did it was ideal compared to the default serialization).

aarchiba:
Would it be tidier to allow and show EarthLocation objects? Permitting (x,y,z) vectors for backward compatibility? Rather than six new properties, I mean.

dlakaplan:
Yes, potentially.  I guess I was trying for minimal changes compared to what had been there, but it's probably fine to do more now.

aarchiba:
My thought was let's keep out interface as simple as possible and let Astropy's EarthLocation handle all the naming, lookups, and conversion.

dlakaplan:
EarthLocation doesn't directly serialize (I think), so if we want to go to/from JSON then we need some interface.

dlakaplan:
Not completely happy with the formatting for:
https://nanograv-pint--1361.org.readthedocs.build/en/1361/explanation.html#observatories
but I think the content is mostly OK.  Are there things that are missing?

[re format: why won't `.. code-block:: bash` or things like that work?]

scottransom:
@dlakaplan Looks like a couple minor conflicts after the latest merges. Once you fix and tests pass, I'm happy to merge.

dlakaplan:
Argh, I messed up the merge from master.  Trying to fix.  Don't merge back until I tell you (hopefully that latest commit did it).

dlakaplan:
OK, I think this is now good, but if you have a chance to double-check the code that would be great.  Anne had made changes which complicated the merge and I wasn't careful enough initially.

src/pint/observatory/topo_obs.py
```diff
@@ -197,22 +215,118 @@ def __init__(
...
         aliases = [f"'{x}'" for x in self.aliases]
-        s = f"TopoObs('{self.name}' ({','.join(aliases)}) at [{self._loc_itrf.x}, {self._loc_itrf.y} {self._loc_itrf.z}]:\n{self.origin})"
+        s = f"TopoObs('{self.name}' ({','.join(aliases)}) at [{self.x}, {self.y} {self.z}]:\n{self.origin})"
         return s
 
     @property
     def timescale(self):
         return "utc"
 
+    def __getattr__(self, name):
```
aarchiba, line 254: `__getattr__` breaks tab completion and generally makes code hard to understand. Can we do without it? `obs.location.lat` might be fine if that's how EarthLocations work?

src/pint/observatory/topo_obs.py
```diff
@@ -197,22 +215,118 @@ def __init__(
...
+        if name in self.__dict__:
```
aarchiba, line 255: `__getattr__` doesn't get called unless the name lookup has already failed, so this is not helpful.

src/pint/observatory/topo_obs.py
```diff
@@ -197,22 +215,118 @@ def __init__(
...
+            # this isn't necessary, but let's allow "alt" as an alias to "height"
+            if name != "alt":
+                return getattr(self.location, name)
+            else:
+                return getattr(self.location, "height")
+
+    @property
+    def as_dict(self):
+        """Return as a dict with limited/changed info"""
+        # is this better than the builtin __dict__ method and then updating some values?
```
aarchiba, line 267: Copying `__dict__` is more future-proof - less risk someone will forget to add a thing here.

tests/test_observatory.py
```diff
@@ -247,14 +254,67 @@ def test_observatory_override(sandbox, overwrite):
...
+    for p in gbt_orig.__dict__:
+        if not p in ["location", "_clock"]:
+            # everything else should be identical
+            assert getattr(gbt_orig, p) == getattr(gbt_reload, p)
+    # check distance separately to allow for precision
+    distance = gbt_orig.separation(gbt_reload)
+    assert distance < 1 * u.m
+
+
+def test_json_observatory_output_latlon_and_itrf(sandbox):
```
aarchiba, line 302: Am I right in understanding that this checks that an error is emitted if you have both xyz and lat/lon/alt or if you're missing lat but doesn't check that lat/lon/alt works? If so can the test name make that clear? Is specifying lat/lon/alt tested elsewhere?

src/pint/observatory/topo_obs.py
dlakaplan: I think going through another object is less than ideal.  I think it better to either go this way or go back to defining properties for at least the most common attributes.

tests/test_observatory.py
dlakaplan: I changed this name to `test_json_observatory_input_latlon_and_itrf_giveserror`
And yes, `test_json_observatory_input_latlon` checks that inputing an observatory via lat/lon works (and gives the same info)

src/pint/observatory/topo_obs.py
dlakaplan: Done

src/pint/observatory/topo_obs.py
dlakaplan: Removed.

scottransom:
I see there were several other comments.  Can we merge this now?  I was going to merge it last week and forgot!

dlakaplan:
let's discuss quickly tomorrow: I think there are some interface disagreements

src/pint/observatory/topo_obs.py
aarchiba: I would rather have a single attribute, `location`, that lets me access the location through the documented Astropy EarthLocation interface, than a heap of additional redundant attributes (or properties).

scottransom: Yeah, I think Anne's suggestion is a good one.  And it would likely simplify things in your code.

dlakaplan: OK, I've changed this as requested

dlakaplan:
If this is OK with @aarchiba and @scottransom , I think it's OK to merge.

src/pint/observatory/topo_obs.py
```diff
@@ -136,7 +153,11 @@ def __init__(
         tempo_code=None,
         itoa_code=None,
         aliases=None,
+        location=None,
         itrf_xyz=None,
+        lat=None,
+        lon=None,
+        height=None,
```
aarchiba, line 160: Er, don't the docs call fot `alt` rather than `height`?

src/pint/observatory/topo_obs.py
```diff
@@ -147,39 +168,39 @@ def __init__(
...
+        if sum(input) == 0:
             raise ValueError(
-                "Incorrect coordinate dimensions for observatory '%s'" % (name)
+                "EarthLocation, ITRF coordinates, or lat/lon/alt are required for observatory '%s'"
+                % name
             )
-
-        # Convert to astropy EarthLocation, ensuring use of ITRF geocentric coordinates
-        self._loc_itrf = EarthLocation.from_geocentric(*xyz)
+        if sum(input) > 1:
```
aarchiba, line 182: This doesn't catch the case where they specify `itrf_xyz` and `lon` (but not `lat`).

src/pint/observatory/topo_obs.py
```diff
@@ -197,22 +218,97 @@ def __init__(
...
-        s = f"TopoObs('{self.name}' ({','.join(aliases)}) at [{self._loc_itrf.x}, {self._loc_itrf.y} {self._loc_itrf.z}]:\n{self.origin})"
+        s = f"TopoObs('{self.name}' ({','.join(aliases)}) at [{self.location.x}, {self.location.y} {self.location.z}]:\n{self.origin})"
         return s
 
     @property
     def timescale(self):
         return "utc"
 
+    @property
+    def as_dict(self):
```
aarchiba, line 258: Should these `as_...` be properties or functions? Is this meant to be consistent with Astropy?

aarchiba, line 258: It would be good to have a round-trip test - ensure that sending something out to `as_dict` or `as_json` and then reading it back left everything unchanged. (Maybe Pyhont `==` is the right test?)

src/pint/observatory/topo_obs.py
dlakaplan: I switched back to `height` which is the `EarthLocation` standard after discussion. I just forgot to update the docs.  That is now done.

src/pint/observatory/topo_obs.py
dlakaplan: done.

src/pint/observatory/topo_obs.py
dlakaplan: I think that is done in `test_observatory`:
```
def test_json_observatory_output(sandbox):
    gbt_orig = get_observatory("gbt")
    load_observatories(io.StringIO(gbt_orig.as_json), overwrite=True)
    gbt_reload = get_observatory("gbt")

    for p in gbt_orig.__dict__:
        if not p in ["_clock"]:
            assert getattr(gbt_orig, p) == getattr(gbt_reload, p)
```
This tests round-trip through JSON.  I don't do a global `==` because some properties like `_clock` may be loaded from other tests.  So I check the properties separately.

```
def test_json_observatory_input_latlon(sandbox):
    gbt_orig = get_observatory("gbt")
    gbt_dict = gbt_orig.as_dict
    # remove ITRF
    del gbt_dict["gbt"]["itrf_xyz"]
    # add in geodetic
    gbt_dict["gbt"]["lat"] = gbt_orig.location.lat.value
    gbt_dict["gbt"]["lon"] = gbt_orig.location.lon.value
    gbt_dict["gbt"]["height"] = gbt_orig.location.height.value
    load_observatories(io.StringIO(json.dumps(gbt_dict)), overwrite=True)
    gbt_reload = get_observatory("gbt")

    for p in gbt_orig.__dict__:
        if not p in ["location", "_clock"]:
            # everything else should be identical
            assert getattr(gbt_orig, p) == getattr(gbt_reload, p)
    # check distance separately to allow for precision
    distance = gbt_orig.separation(gbt_reload)
    assert distance < 1 * u.m
```
Same here.  This dumps to `dict` and does the same thing.  Or did you have something else in mind.

dlakaplan: No idea.

aarchiba: I was thinking a simple unmodified round-trip, which your first test does. Sorry I missed it!

aarchiba: I think unless there's a solid reason for them to be properties, these should be functions - they construct a transformed version of the whole object, so it's a little weird that they look like attributes of it.

PR #1365, opened 2022-07-28
Make phaseogram plotting more robust

### LINKED ISSUE #1364: photonphase error
I'm running photonphase with the following command:
```
photonphase file.evt file.par --plot --polycos
```
and I get the following error:
```
  File "/Users/meo/opt/anaconda3/envs/py39/bin/photonphase", line 8, in <module>
    sys.exit(main())
  File "/Users/meo/opt/anaconda3/envs/py39/lib/python3.9/site-packages/pint/scripts/photonphase.py", line 256, in main
    phaseogram_binned(mjds, phases, bins=100, plotfile=args.plotfile)
  File "/Users/meo/opt/anaconda3/envs/py39/lib/python3.9/site-packages/pint/plot_utils.py", line 90, in phaseogram_binned
    years = (mjds.value - 51544.0) / 365.25 + 2000.0
AttributeError: 'numpy.ndarray' object has no attribute 'value'
```
I'm investigating the issue, but just in case someone else already found a solution...

paulray:
The double precision MJDs are passed to the plotter, it assumes the units are days.
Fixes #1364


aarchiba:
How does this cope with `atropy.time.Time` objects being passed? Those seem like another likely option that is easy to support (just take the `.mjd`).

paulray:
> Would it be easy to write some quick tests to make sure that it doesn't raise exceptions when passed the things it's supposed to accept? As long as you don't call `plt.show()` I think it should be safe to call matplotlib functions in the test suite even running headless.

I knew this was coming as soon as I submitted the PR :-)

paulray:
I added `Time` support and a test. @aarchiba could you see if this looks OK?
It writes the plots to a file, so hopefully it will work headless.

devbhakt:
Would it be possible to adjust the color array for the scatter plot so that if one is using highly weighted photons, then there is still variation within the color array?

paulray:
Hmm. Evidently for the astropy version in oldestdeps there is no `.value` property of a `Column`, so that test is failing. Not sure what the best solution is to that.

paulray:
> Would it be possible to adjust the color array for the scatter plot so that if one is using highly weighted photons, then there is still variation within the color array?

Right now the color array is just black with an alpha set by the weight. This could certainly be modified if you like. Once this PR is merged, feel free to add your own changes.   But, you may not want to change the plotting code as much as your weight array. You can map that however you like (linear, log, sqrt scaling, etc.), but it should go from 0.0 to 1.0.

Note that for large numbers of photons, the binned version of phaseogram may be preferable.

PR #1371, opened 2022-08-04
More flexible read/write for polycos

dlakaplan:
Changes:
* Can now read/write from/to a `Path` object or a stream (like `io.StringIO`).
* Registering formats done as a class method.  Done once on import for default formats (just tempo), can be done later if needed.
* Read polycos and generate from timing model now done as class methods that return new objects
* Documentation cleanup/improvements.
* implemented `len()`, `__getitem__` for polycos
* implemented `__getitem__` for Phase objects

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - Moved observatories to JSON file.  Changed way observatories are loaded/overloaded
 - Split Jodrell Bank observatory based on backend to get correct clock files
 - Clock files can be marked as being valid past the end of the data they contain
+- Polycos can be written/read from Path or Stream objects
+- Polyco format registration now done once as a class method
+- Polyco reading/generation from timing model done as class methods
 ### Added
 - delta_pulse_number column is now saved to -padd flag on TOA write
 - command-line utility to compare parfiles
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,7 +18,8 @@ import sys
 from packaging.version import parse
 
 import jupytext
-#import sphinx.ext.apidoc
+
+# import sphinx.ext.apidoc
 
 
 # If extensions (or modules to document with autodoc) are in another
@@ -76,7 +77,7 @@ autodoc_inherit_docstrings = True
 templates_path = ["_templates"]
 
 # The suffix of source filenames.
-#source_suffix = ".rst"
+# source_suffix = ".rst"
 
 # The encoding of source files.
 # source_encoding = 'utf-8-sig'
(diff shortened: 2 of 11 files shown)


dlakaplan:
P.S.  Polyco calculation is slow, because it relies on making a lot of TOAs and that is slow.  Which then makes testing slow (~5min just for `test_polcos.py`.  So if we make progress with any other speedups it will help a lot here.

dlakaplan:
Note that the change to the `Phase` object to allow indexing changes some assumed behavior.  Previously, with a `Phase` object `p`, `p[0]` would be the integer part and `p[1]` would be the fractional.  Now `p[0]` is the first item (integer & fractional) and you can still use `p.int` and `p.frac` to access the other attributes.  If this is an issue I can revert it but I found it useful for some testing.

aarchiba:
> P.S. Polyco calculation is slow, because it relies on making a lot of TOAs and that is slow. Which then makes testing slow (~5min just for `test_polcos.py`. So if we make progress with any other speedups it will help a lot here.

Some of the slow parts of working with TOAs are less with barycentric TOAs, so perhaps those could be the default for polyco tests?

More broadly, I don't think there's a lot we can do to speed up polyco generation, since those slow TOA calculations are exactly what is needed to get things as seen at specific observatories. (Incidentally, how are clock corrections handled?)

If we were willing to diverge from the TEMPO approach, we might find we got equally good or better approximations for the same or fewer TOAs if we distributed the TOAs according to the zeros of Chebyshev polynomials. Doing polynomial fitting on these points amounts to working in the Chebyshev basis, which is the right basis for limiting maximum absolute deviation.

We might be able to get a minor speedup by generating all the TOAs at once and reusing any that are needed for more than one polyco entry (if we do that). There's also an issue pointing out that for many satellite applications, polycos are only needed for specified Good Time Intervals, and we could skip generating any outside those.

Major speedups are going to require a careful analysis of where PINT spends its time when you're working with TOAs - I don't think that there is necessarily duplicated effort, but there may be many places where unnecessary precision is costing us a lot of time.

aarchiba:
> Note that the change to the `Phase` object to allow indexing changes some assumed behavior. Previously, with a `Phase` object `p`, `p[0]` would be the integer part and `p[1]` would be the fractional. Now `p[0]` is the first item (integer & fractional) and you can still use `p.int` and `p.frac` to access the other attributes. If this is an issue I can revert it but I found it useful for some testing.

Useful for testing frequently means more flexible for users.

dlakaplan:
Examples of new Phase interface:
```
from pint.phase import Phase
import numpy as np
p = Phase(np.arange(10),np.random.random(10))
print(p.int)
print(p.frac[:5])
print(p[::2])
```
all work

dlakaplan:
I removed the `__getitem__` method for the `Phase` objects so that we can potentially merge this without waiting for #1375 to converge.  I think the remaining changes (initialize polycos better, read/write from streams) are uncontroversial.

Issue #1372, opened 2022-08-05
[Feature Request] Single Command for Caching Remote Files for Runs on Machines with No Access to the Internet


JPGlaser:
Hey All,

As I said in last week's call, we should work towards adding a command that grabs the necessary files that normally it would attempt to grab from the web, and instead caches them locally. This would allow PINT to be used on compute nodes which usually have no outside access to the Internet.

Right now, I get around this by running `python -c "import pint.toa"` and that will nab some files from astropy and cache them for ~ a week.

It would also be nice to force PINT to tell astropy to use old cached files if the user requires it -or- if there is no internet connection, instead of killing the script.

~ Joe G.

dlakaplan:
This should accept as options a list of telescopes, or perhaps have a default list.

abhisrkckl:
I think it will be good to download new files (including metadata) only if it is necessary. In principle, we can just look at the latest TOA value and decide if the local files are sufficient.

dlakaplan:
https://nanograv-pint.readthedocs.io/en/latest/_autosummary/pint.observatory.topo_obs.export_all_clock_files.html#pint.observatory.topo_obs.export_all_clock_files does this **if** you have loaded in data pertaining to the relevant observatories.  So this could be run as the last step in the timing analysis.  But it should also be possible to pre-load various observatories.

dlakaplan:
@JPGlaser : does #1373 do what you want?  Is that the interface you want?

dlakaplan:
Note that this also doesn't deal explicitly with whatever files astropy uses on it's own.  that is discussed at https://docs.astropy.org/en/stable/utils/data.html#astropy-data-and-clusters: do you need something to do that as well?

aarchiba:
> This should accept as options a list of telescopes, or perhaps have a default list.

I think for Joe's application the easiest thing to do is just export all the telescope clock corrections PINT knows about.

aarchiba:
> I think it will be good to download new files (including metadata) only if it is necessary. In principle, we can just look at the latest TOA value and decide if the local files are sufficient.

The clock corrections code does this. You only get new clock corrections if you have a TOA past the end of the ones you currently have - or if the repository has flagged your current version of a clock correction file as needing updates (presumably because it contained errors).

aarchiba:
To make this happen, I think we need:

- [ ] A single command to download all clock corrections into the astropy cache.
[...]

PR #1373, opened 2022-08-08
Script to export clock files

dlakaplan:
I think this is what was requested in #1372 .  But I don't know if a better implementation is available.  

Usage:
```
export_clocks -o ./ gbt arecibo chime vla --log-level=DEBUG
```
So this runs as an executable script.  It could also be made into a standalone function if needed, although what is needed for that is very simple.

### DIFF
--- a/src/pint/scripts/export_clocks.py
+++ b/src/pint/scripts/export_clocks.py
@@ -47,4 +47,4 @@ def main(argv=None):
         obs = pint.observatory.get_observatory(obsname)
         obs._load_clock_corrections()
 
-    pint.observatory.topo_obs.export_all_clock_files(os.path.abspath(os.curdir))
+    pint.observatory.topo_obs.export_all_clock_files(args.directory)
--- a/src/pint/scripts/export_clocks.py
+++ b/src/pint/scripts/export_clocks.py
@@ -1,0 +1,50 @@
+import argparse
+import os
+import sys
+
+import pint.logging
+from loguru import logger as log
+
+pint.logging.setup(level=pint.logging.script_level)
+
+import pint.observatory
+import pint.observatory.topo_obs
+
+__all__ = ["main"]
+
+
+def main(argv=None):
+    parser = argparse.ArgumentParser(
+        description="PINT tool for export clock files",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
+    parser.add_argument(
+        "observatories", help="Observatories to export", type=str, nargs="+"
+    )
+    parser.add_argument(
+        "-o",
+        "--out",
+        default=os.path.abspath(os.curdir),
+        type=str,
+        dest="directory",
+        help="Destination directory",
+    )
+    parser.add_argument(
+        "--log-level",
+        type=str,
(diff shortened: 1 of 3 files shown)


dlakaplan:
It could also have a pre-determined set of observatories

src/pint/scripts/export_clocks.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+    parser.add_argument(
+        "-o",
+        "--out",
+        default=os.path.abspath(os.curdir),
+        type=str,
+        dest="directory",
+        help="Destination directory (set $PINT_CLOCK_OVERRIDE to this value to look here first for clock files)",
+    )
+    parser.add_argument(
```
aarchiba, line 33: More conventional would be a `--verbose` (possibly repeatable) and optionally a `--quiet` rather than explicitly referencing log levels. Since this is a specific application, one might be able to select sensible log levels to correspond to these.

dlakaplan: That's the type of interface I had gone for initially for applications, but @scottransom wanted this method

scottransom: And if the discussion is about the log levels, I was just thinking that we should be consistent with the other PINT command line programs.  Maybe `--verbose` could correspond to upping the log level by one, as well. (maybe by one for each verbose added).

aarchiba:
It's worth pointing out that I don't think this is relevant to #1372 - that is a request to be able to preload the astropy cache for use by machines without internet access (or containers that should minimize calls to the internet). It might have some use for non-PINT users who just want to pull out the clock files without having to face a line of PINT code?

src/pint/scripts/export_clocks.py
aarchiba: Yes, this is about log levels. I'm not sure users of a command line program will have any idea what log levels have to do with what output they want to see. Making `--verbose` bump up the log level, and `--quiet` bump it down by one might be a tidy way to present a more conventional command-line interface.

dlakaplan: Yes, this is what I had implemented months ago.  I agree it's more standard but I don't really care either way.

dlakaplan:
OK, if this doesn't do what #1372 wants (I thought it could, since it would then not have calls to check all of the clock files over the internet, but maybe I'm misunderstanding) then it's fine to close

src/pint/scripts/export_clocks.py
dlakaplan: See #1377 for discussion of this.

paulray:
Should we close this, or merge it?

dlakaplan:
Whatever @aarchiba thinks.  It may not be useful, but I don't think it's harmful.  However, merging with #1380 would be needed

aarchiba:
I don't know how useful this is - I'm not sure I have a use case in mind for it. It's not much code, though, so if it had tests to ensure we didn't break it then it wouldn't do any harm to include it.

abhisrkckl:
What is the status of this?

dlakaplan:
Still tabled.  I think it works for what it is, but it's not clear that it's what's needed.

PR #1374, opened 2022-08-11
WIP: single function call to preload Astropy cache

### LINKED ISSUE #1372: [Feature Request] Single Command for Caching Remote Files for Runs on Machines with No Access to the Internet

aarchiba:
This is intended to resolve #1372 by providing a single function call that preloads the Astropy cache. Documentation on how to forbid PINT/Astropy from connecting to the Internet should also be included.

### DIFF
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -52,6 +52,7 @@ import numpy as np
 import scipy.optimize.zeros as zeros
 from loguru import logger as log
 from scipy.special import fdtrc
+from astropy.utils.iers import IERS_Auto
 
 import pint
 import pint.pulsar_ecliptic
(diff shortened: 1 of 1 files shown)


aarchiba:
@JPGlaser I think that adjusting your script to call `pint.utils.preload_cache` should reliably ensure that all files are loaded to the Astropy cache even starting from nothing. You could call `pint.utils.set_no_internet()` at the beginning of all scripts, but almost certainly you are better off creating an Astropy config file with the same options in it. (You'll need to set them back in the updater script.)

aarchiba:
This one is a bit of a headache because I'm trying to get the "don't access the Internet, be okay with old data" to work with everything from Astropy 4.0 to 5.1, and the code to deal with out-of-range data has changed several times over the course of that time. I might need to create some specialized tox tasks and test all the versions. We particularly need 4.3 to work because that's what NANOGrav used for 15yr and they want to keep using it; 5.1 works fine and quite gracefully, but so far 4.3, 5.0, and 5.1 need different solutions.

JPGlaser:
@aarchiba We are okay with advancing to Astropy 5.1+ with the 15yr. This is because we have to advance our versions of other software for the wideband results. I think this feature would be a fantastic add as we get more folks in the IPTA using `pint`. ^_^

PR #1378, opened 2022-08-12
Piecewise orbit model

poneill129:
This is intended as an extension to the existing BT model currently implemented. BT Piecewise was designed to allow a user to fit for orbital phase (T0) with user-defined MJD blocks. 

Preliminary work was made to allow for other piecewise orbital parameters, but since this was a tool intended for a specific problem, the development of other piecewise orbital parameters was stopped. Though the foundations for future development in this area are present.

It feels like setting the piecewise parameters from binary_piecewise to BT_piecewise is currently a jury-rig which could do with improvements. Additional concerns surround this involves when a setup() function should be called. A better understanding of this would help performance time.


src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,325 @@
...
+                time_scale="utc",
+                value=group_end_mjd,
+            )
+        )
+        # print("going to binary_piecewise setup")
+        # self.setup()
+        # self.validate()
+
+    def update_binary_object(self, toas=None, acc_delay=None):
+        super().update_binary_object(toas=toas, acc_delay=acc_delay)
```
aarchiba, line 104: If all this method does is call the same method on `super()` can it not be omitted completely? Several other methods here seem to be in the same situation.

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,325 @@
...
+            )
+        for index in indices:
+            index_rf = f"{int(index):04d}"
+            for prefix in ["T0X_", "A1X_", "PLB_", "PUB_"]:
+                self.remove_param(prefix + index_rf)
+        self.setup()
+        self.validate()
+
+    def add_piecewise_param(self, param, param_unit, paramx, j):
+        if int(j) in self.get_prefix_mapping_component("X_"):
```
aarchiba, line 142: This should be after the `if j is None`, shouldn't it?

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,325 @@
...
+                "Index '%s' is already in use in this model. Please choose another." % j
+            )
+        if j is None:
+            dct = self.get_prefix_mapping_component(param + "X_")
+            j = np.max(list(dct.keys())) + 1
+        i = f"{int(j):04d}"
+        if param is "A1":
+            self.add_param(
+                prefixParameter(
+                    name=param + "X_{0}".format(i),
```
aarchiba, line 153: Nicer to use `f"X_{i}"`

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,325 @@
...
+                    units=param_unit,
+                    value=paramx,
+                    unit_template=lambda x: param_unit,
+                    description="Parameter" + param + "variation",
+                    description_template=lambda x: param,
+                    parameter_type="float",
+                    frozen=False,
+                )
+            )
+        elif param is "T0":
```
aarchiba, line 163: For strings it's better to use `==` than `is` - you could have two different strings that contain `T0` that were equal but not `is`.

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,325 @@
...
+                    parameter_type="float",
+                    frozen=False,
+                )
+            )
+
+        # self.setup()
+
+    def setup(self):
+        # print(self.PB)
+        super(BinaryBTPiecewise, self).setup()
```
aarchiba, line 181: Just use `super().setup()`.

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,325 @@
...
+        XR2_mapping = self.get_prefix_mapping_component("PUB_")
+        XR2s = {}
+        # print(f'XR2:{XR2_mapping.values()}')
+        for t0n in T0X_mapping.values():
+            # print("hello from T0 mapping")
+            T0Xs[t0n] = getattr(self, t0n).quantity
+        # if None in T0Xs.values():
+        # print("Group with non-defined T0X, applying default T0 to group")
+        # TODO set default T0 value
+        if None not in T0Xs.values():
```
aarchiba, line 205: Doesn't this skip defining all the parameters if any of the values is unset? That doesn't seem right?

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,325 @@
...
+        # print("Group with non-defined XR2, applying default A1 to group")
+        # TODO set default A1 value
+
+        if None not in XR2s.values():
+            # print(len(A1Xs.items()))
+            for xr2_name, xr2_value in XR2s.items():
+                self.binary_instance.add_binary_params(xr2_name, xr2_value)
+        self.update_binary_object()
+
+    def validate(self):
```
aarchiba, line 244: Are you sure there aren't any ways the new parameters could be set that would make the model unusable? `validate()` should check for invalid settings of the piecewise parameters. (Overlapping T0X intervals? T0X intervals with end before beginning? Any weirdness that won't work can be detected here and then you won't need to worry about it later.)

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+
+class BTpiecewise(BTmodel):
+    def __init__(self, axis_store_initial=None, t=None, input_params=None):
+        self.binary_name = "BT_piecewise"
+        super(BTpiecewise, self).__init__()
+        self.axis_store_initial = []
+        self.extended_group_range = []
+        self.d_binarydelay_d_par_funcs = [self.d_BTdelay_d_par]
+        if t is not None:
+            self._t = t
```
aarchiba, line 18: It's probably a good idea to set `selt._t` to None if `t` is None, rather than leaving it uninitialized.

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+        if input_params is not None:
+            if self.T0X is None:
+                self.update_input(input_params)
+            elif self.T0X is not None:
```
aarchiba, line 22: This is just `else` isn't it?

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+                self.update_input()
```
aarchiba, line 23: Isn't this doing exactly the same thing in both branches of the `if` statement?

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+        self.binary_params = list(self.param_default_value.keys())
+        # self.param_aliases.update({"T0X": ["T0X"], "A1X": ["A1X"]})
+
+    def set_param_values(self, valDict=None):
+        # print(valDict)
+        super().set_param_values(valDict=valDict)
+        self.piecewise_parameter_loader(valDict=valDict)
+
+    def piecewise_parameter_loader(self, valDict=None):
```
aarchiba, line 32: It's preferable if a function name is an imperative - `load_piecewise_parameters`, perhaps, or `setup_internal_structures` or something. This function should have a docstring so you can explain what it does and when to call it.

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+        )  # initialise arrays to store piecewise group boundaries
+        self.upper_group_edge = []
+        piecewise_parameter_information = (
+            []
+        )  # initialise array that will be 5 x n in shape. Where n is the number of pieces required by the model
+        # print(f"valDict:{valDict}")
+        if (
+            valDict is None
+        ):  # If there are no updates passed by binary_instance, sets default value (usually overwritten when reading from parfile)
+            self.T0X_arr = self.T0
```
aarchiba, line 47: Does this make `T0X_arr` into a scalar instead of an array? Won't this break things?

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+                0,
+                self.T0,
+                self.A1,
+                0 * u.d,
+                1e9 * u.d,
+            ]
+        else:
+            piece_index = (
+                []
+            )  # iniialise array used to count the number of pieces. Operates by seaching for "A1X_i/T0X_i" and appending i to the array. Can operate if pieces are given out of order.
```
aarchiba, line 61: Try to avoid comments on the same line as your code, it makes for nasty formatting.

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+                if key[0:4] == "T0X_" or key[0:4] == "A1X_":
+                    piece_index.append((key[4:8]))  # appends index to array
+            piece_index = np.unique(
+                piece_index
+            )  # makes sure only one instance of each index is present returns order indeces
+            for (
+                index
+            ) in (
+                piece_index
+            ):  # looping through each index in order they are given (0 -> n)
```
aarchiba, line 79: I'm not sure this is really the order you're getting. Doesn't `np.unique` give you the strings in sorted (lexicographical) order, regardless of the order that they were given?

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+                piecewise_parameter_information.append(param_pieces)
+            self.valDict = valDict
+            self.piecewise_parameter_information = sorted(
+                piecewise_parameter_information, key=lambda x: x[3]
+            )  # sorts the array chronologically by lower edge of each group,correctly works for unordered pieces (i.e. index 0 can correspond to an arbitrary group of data at any time)
+            # print(f"Filled information array: {self.piecewise_parameter_information}")
+            if len(self.piecewise_parameter_information) > 0:
+                # check = hasattr(self,"t")
+                # print(f"Piecewise parameter loader can see t: {check}")
+                if hasattr(self, "_t") is True:
```
aarchiba, line 123: Don't use `is True`; just say `if hasattr(...):`

aarchiba: Also if you set this to None in `__init__` you can avoid the need tor `hasattr`.

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+                    self.piecewise_parameter_information[i][3].value
+                )
+                upper_group_edge.append(
+                    self.piecewise_parameter_information[i][4].value
+                )
+            return [lower_group_edge, upper_group_edge]
+
+    def a1(self):
+        self.A1_val = self.A1X_arr * ls
+        if hasattr(self, "A1X_per_toa") is True:
```
aarchiba, line 209: What decides whether `self` has `A1X_per_toa`? When would it happen or not? Can you make sure it's always present?

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+            ret = self.A1X_per_toa + self.tt0 * self.A1DOT
+        else:
+            ret = self.A1 + self.tt0 * self.A1DOT
+        return ret
+
+    def get_tt0(self, barycentricTOA):
+        """tt0 = barycentricTOA - T0"""
+        if barycentricTOA is None or self.T0 is None:
+            tt0 = None
+            return tt0
```
aarchiba, line 219: Just write `return None`

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+        if len(barycentricTOA) > 1:
```
aarchiba, line 220: What happens if `len(barycentricTOA)==1`?

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+                )  # Defines object's index for each toa as an array of length = len(self._t)
+                (
+                    self.T0X_per_toa,
+                    self.A1X_per_toa,
+                ) = self.piecewise_parameter_from_information_array(
+                    self._t
+                )  # Uses the index for each toa array to create arrays where elements are the A1X/T0X to use with that toa
+        if len(barycentricTOA) > 1:
+            check = hasattr(self, "T0X_per_toa")
+            if hasattr(self, "T0X_per_toa") is True:
```
aarchiba, line 233: If you're going to create `check`, just use `if check:`; but again, when might `T0X_per_toa` not be present?

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+                if len(self.T0X_per_toa) == 1:
```
aarchiba, line 234: This `if` statement does the same thing on both branches.

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+            barycentricTOA = barycentricTOA * u.day
+        # print(f"Unique T0s being used in tt0 calculation: {np.unique(T0)}\n")
+        tt0 = (barycentricTOA - T0).to("second")
+        # print(f"piecewise tt0 being called, Calc'd with: {T0} ")
+        return tt0
+
+    def d_delayL1_d_par(self, par):
+        if par not in self.binary_params:
+            errorMesg = par + " is not in binary parameter list."
+            raise ValueError(errorMesg)
```
aarchiba, line 256: Just construct the error message in the ValueError construction: `raise ValueError(f"{par} is not in binary parameter list.")`

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+        par_obj = getattr(self, par)
```
aarchiba, line 257: Is this used anywhere?

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+                if par in self.orbits_cls.orbit_params:
+                    return self.d_delayL1_d_E() * self.d_E_d_par()
+                else:
+                    return np.zeros(len(self.t)) * u.second / par_obj.unit
+
+    def d_delayL2_d_par(self, par):
+        if par not in self.binary_params:
+            errorMesg = par + " is not in binary parameter list."
+            raise ValueError(errorMesg)
+        # print(par)
```
aarchiba, line 285: Please deal with debugging `print` statements - either delete them, or if you think they might be useful for debugging, use `log.debug`.

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,441 @@
...
+            if hasattr(self, "d_delayL2_d_" + par_temp):
+                func = getattr(self, "d_delayL2_d_" + par_temp)
+                return func() * index
+            else:
+                if par in self.orbits_cls.orbit_params:
+                    return self.d_delayL2_d_E() * self.d_E_d_par()
+                else:
+                    return np.zeros(len(self.t)) * u.second / par_obj.unit
+
+    def in_piece(self, par):
```
aarchiba, line 310: I think this function needs a docstring - what does it do?

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+import unittest
+from io import StringIO
+from pylab import *
+import pytest
+import pint.residuals
+import pint.simulation
+
+
+@pytest.fixture
+def build_model():
```
aarchiba, line 21: Because you use fixtures like `def test_thing(fixture):` ... `y = fixture.attribute`, it makes sense to use a noun-like name; here `model_no_pieces` would make sense, so you'd have `def test_something(model_no_pieces):`.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+    PMRA                                  0.0
+    PMDEC                                 0.0
+    PX                                    0.0
+    POSEPOCH           54995.0000000000000000
+    F0                   592.4214681620181649 0 
+    F1              -2.4181868552365951721e-15 0
+    PEPOCH             55352.6375960000000001
+    PLANET_SHAPIRO                          N
+    DM                  14.720549454038943568 1 
+    DM1                                   0.0
```
aarchiba, line 45: Can you omit some of the less important bits, like DM1 and PMRA? Keep this a little shorter and easier to glance over? Which bits are actually needed?

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+    # makes toas to go with the two non-overlapping, complete coverage model
+    m_piecewise = deepcopy(build_piecewise_model_with_two_pieces)
+    m_piecewise.setup()
+    toas = pint.simulation.make_fake_toas_uniform(
+        m_piecewise.START.value + 1, m_piecewise.FINISH.value - 1, 20, m_piecewise
+    )  # slightly within group edges to make toas unambiguously contained within groups
+    return toas
+
+
+def get_toa_group_indexes(model, toas):
```
aarchiba, line 76: This isn't recognized as a test because it doesn't start with `test_`. Wouldn't it be better to just use `model.which_group_is_toa_in(toas)` directly?

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+    # returns array of group indexes associated with each toa. (i.e. which group is each toa in)
+    model.setup()
+    index = model.which_group_is_toa_in(toas)
+    return index
+
+
+def get_number_of_groups(model):
+    # returns number of groups
+    model.setup()
```
aarchiba, line 85: Likewise it's better to use `model.get_number_of_groups()` directly.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+    number_of_groups = model.get_number_of_groups()
+    print(model.as_parfile())
+    print(model.get_number_of_groups())
+    return number_of_groups
+
+
+@pytest.fixture()
+def build_piecewise_model_with_two_pieces(build_model):
+    # takes the basic model frame and adds 2 non-ovelerlapping pieces to it
+    piecewise_model = deepcopy(build_model)
```
aarchiba, line 95: Fixtures always give you a fresh-created new object, so there's no need to copy it.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+        piecewise_model.add_piecewise_param(
+            "A1", "ls", piecewise_model.A1.value + i * 1e-5, i
+        )
+        piecewise_model.add_piecewise_param(
+            "T0", "d", piecewise_model.T0.value + i * 1e-5, i
+        )
+    return piecewise_model
+
+
+def test_add_piecewise_parameter(build_model):
```
aarchiba, line 114: I think this test is actually checking whether the model round-trips to a par file correctly? If so, maybe call the test something that says that?

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+            m_piecewise.START.value + 501 * i, m_piecewise.START.value + 1000 * i, j=i
+        )
+        m_piecewise.add_piecewise_param("A1", "ls", m_piecewise.A1.value + 1, i)
+        m_piecewise.add_piecewise_param("T0", "d", m_piecewise.T0.value, i)
+    m3 = get_model(StringIO(m_piecewise.as_parfile()))
+    param_dict = m_piecewise.get_params_dict(which="all")
+    copy_param_dict = m3.get_params_dict(which="all")
+    number_of_keys = 0
+    comparison = 0
+    for key, value in param_dict.items():
```
aarchiba, line 131: Can you just ask `assert param_dict == copy_param_dict`? If this fails you'll get a list of the mismatched keys/values.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+        number_of_keys = number_of_keys + 1  # iterates up to total number of keys
+        for copy_key, copy_value in param_dict.items():
```
aarchiba, line 133: Shouldn't this be `copy_param_dict`?

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+            if key == copy_key:  # search both pars for identical keys
+                if value.quantity == copy_value.quantity:
+                    comparison = (
+                        comparison + 1
+                    )  # iterates up to all matched keys, which should be all, hence total number of keys
+    assert comparison == number_of_keys  # assert theyre the same length
+
+
+def test_get_number_of_groups(build_piecewise_model_with_two_pieces):
+    # test to make sure number of groups matches with number of added piecewise parameters
```
aarchiba, line 143: Can you put this information in the function name?

tests/test_BT_piecewise.py
aarchiba: There are round-trip tests elsewhere as well, it would be good to add a BTpiecewise model to those, to benefit from that testing code as well.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+    index = get_toa_group_indexes(
+        build_piecewise_model_with_two_pieces, make_toas_to_go_with_two_piece_model
+    )
+    print(index)
+    should_be_ten_toas_in_each_group = [
+        np.unique(index, return_counts=True)[1][0],
+        np.unique(index, return_counts=True)[1][1],
+    ]
+    expected_toas_in_each_group = [10, 10]
+    is_there_ten_toas_per_group = np.array_equal(
```
aarchiba, line 164: it's better to use `numpy.testing.assert_array_equal`, it will tell you more about the mismatch

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+def test_paramX_per_toa_matches_corresponding_model_value(
+    param, build_piecewise_model_with_two_pieces, make_toas_to_go_with_two_piece_model
+):
+    # Testing the correct piecewise parameters are being assigned to each toa.
+    # Operates on the piecewise_parameter_from_information_array function. Requires group_index fn to be run so we have an array of length(ntoas), filled with information on which group a toa belongs to.
+    # Uses this array to apply T0X_i/A1X_i to corresponding indexes from group_index fn call. i.e. for T0X_i,T0X_j,T0X_k values and group_index return: [i,j,k] the output would be [T0X_i,T0X_j,T0X_k]
+    m_piecewise = build_piecewise_model_with_two_pieces
+    toa = make_toas_to_go_with_two_piece_model
+    m_piecewise.setup()
+    rs = pint.residuals.Residuals(toa, m_piecewise)
```
aarchiba, line 180: Is this ever used?

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+    do_arrays_match = np.array_equal(
+        do_toas_reference_piecewise_parameter, should_toa_reference_piecewise_parameter
+    )
+    assert do_arrays_match
+
+
+def test_problematic_group_indexes_and_ranges(build_model):
+    # Test to flag issues with problematic group indexes
+    m_piecewise = build_model
+    with pytest.raises(Exception):
```
aarchiba, line 216: Please specify what kind of exception is to be raised.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+        assert m_piecewise.add_group_range(
```
aarchiba, line 217: No point asserting anything if it's supposed to raise an exception. Just run the code.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+            m_piecewise.START.value, m_piecewise.FINISH.value, j=-1
+        )
+        assert m_piecewise.add_group_range(
```
aarchiba, line 220: This needs to be in a separate `with` statement - the first exception exits the `with`.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+            m_piecewise.FINISH.value, m_piecewise.START.value, j=1
+        )
+        assert m_piecewise.add_group_range(
+            m_piecewise.START.value, m_piecewise.FINISH.value, j=10000
+        )
+
+
+def add_groups_and_make_toas(build_model, build_piecewise_model_with_two_pieces, param):
```
aarchiba, line 228: This should be three separate functions, they don't share any appreciable code.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+
+
+def make_generic_toas(model):
+    # makes toas to go with the edge cases
+    return pint.simulation.make_fake_toas_uniform(
+        model.START.value, model.FINISH.value - 1, 20, model
+    )
+
+
+def convert_int_into_index(i):
```
aarchiba, line 272: Isn't there a utility function to do this in the `BTpiecewise` code?

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+    upper_edge_of_lower_group = boundaries[1][0]
+    lower_edge_of_upper_group = boundaries[0][1]
+    truth_array_comparison = [
+        [barycentric_toa.value <= upper_edge_of_lower_group],
+        [barycentric_toa.value >= lower_edge_of_upper_group],
+    ]
+    return truth_array_comparison
+
+
+@pytest.mark.parametrize(
```
aarchiba, line 300: You can pass a list of functions to `parametrize`.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+    "param",
+    [
+        "non-overlapping complete group coverage",
+        "overlapping groups",
+        "non-complete group coverage",
+    ],
+)
+def test_does_toa_lie_in_group(
```
aarchiba, line 308: The three branches here don't share any appreciable code so they should be three separate tests.

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+    ).resids_value
+    parameter_string = f"{param}_{convert_int_into_index(index)}"
+
+    output_param = []
+    m_piecewise_temp = add_offset_in_model_parameter(index, param, m_piecewise)
+
+    rs_temp = pint.residuals.Residuals(
+        toas, m_piecewise_temp, subtract_mean=False
+    ).resids_value
+    have_residuals_changed = np.invert(rs_temp == rs_value)
```
aarchiba, line 431: you can use `rs_temp != rs_value`

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+            parameter_string = param[0:2]
+            getattr(m_piecewise_temp, parameter_string).value = (
+                getattr(m_piecewise_temp, parameter_string).value - offset_size
+            )
+            m_piecewise_temp.setup()
+        return m_piecewise_temp
+
+
+def get_d_delay_d_xxxx(toas, model, parameter_string):
+    rs_temp = pint.residuals.Residuals(toas, model, subtract_mean=False)
```
aarchiba, line 516: Do you really need to construct Residuals here? Why?

tests/test_BT_piecewise.py
```diff
@@ -0,0 +1,560 @@
...
+    d_delay_temp = rs_temp.model.d_binary_delay_d_xxxx(
+        toas, parameter_string, acc_delay=None
+    )
+    return d_delay_temp
+
+
+# @pytest.mark.parametrize("param, index, offset_size",[("T0X",0, 1e-5*u.d),#("T0X",-1, 1e-5*u.d),("A1X",0, 1e-5*ls),("A1X",-1, 1e-5*ls)])
```
aarchiba, line 523: You can use `pytest.mark.skip` with a reason to indicate that a test should be skipped.

aarchiba:
This looks excellent! A little code cleanup would improve it a lot. I'd also like to suggest a few additional tests:

- Can you compare the residuals from BTPiecewise with several pieces to BT with the values appropriate for that piece? This is supposed to give exactly the same residuals but could go wrong. Checking derivatives would be good too.
- There are some tests elsewhere that use `numdifftools` to check analytical derivatives against numerical. Can you integrate a BTPiecewise model into that?
- Can you write some corner-case tests - what happens if the T0X intervals are out of order? If the end is before the beginning? If it's zero-width? If they're duplicated or overlap? Most of these should just raise exceptions, but tests can document what should happen if the user puts something wrong in.
- Can you write a test that creates two different BTpiecewise models and calls various functions of one interlaced with calls to the other? Just to make sure there's no global caching of anything (I'm sure you're fine but it's good to check).
- Can you write some tests that do some calculations, change some values, and do some new calculations; versus doing the second lot of calculations directly? This is to make sure that stuff doesn't linger in BTpiecewise objects from one call to the next.

dlakaplan:
Some discussion of the `T2` model came up, and I notice that `T2` implements binary parameter jumps as:
<div class="table-overflow" style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none; box-sizing: border-box; margin: 0px 0px 0.5em; padding: 0px; border: 0px; font-family: Merriweather, serif; font-size: inherit; font-stretch: inherit; line-height: inherit; vertical-align: baseline; overflow-x: auto; caret-color: rgb(42, 42, 42); color: rgb(42, 42, 42);">

BPJEP_k | Epoch of a step-jump in the binary parameters |  
-- | -- | --
BPJPH_k | Size of phase jump |  
BTJPB_k | Size of jump in orbital period |  
BTJA1_k | Size of jump in projected semimajor axis |  
BTJECC_k | Size of jump in orbital eccentricity |  
BTJOM_k | Size of jump in longitude of periastron |  
</div>
Is that in any way compatible with how things were implemented here?

scottransom:
Hi All:  What is the status of this code?  Is it working?  Or might it be soon?  I have a new student who will be working on timing Ter5A and would love to use this, if possible.

src/pint/models/binary_piecewise.py
poneill129: Reordered the statements, good catch

src/pint/models/binary_piecewise.py
poneill129: Changed string formatting as suggested (here and in other places).

src/pint/models/binary_piecewise.py
poneill129: changed all instances of is to == for strings

src/pint/models/binary_piecewise.py
poneill129: Changed to `super().setup()`

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: Changed as suggested

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: Forgot to add the docstring with the update, still to do...

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: Apparently it didn't, confused as to why since it's been like this through testing. Changed so it lives in a list so it matches the expected form of T0X_arr after self._t is declared

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: So it's populating the list `piece_index` with all the unique piecewise parameter keys (i.e. '0000', '0001' etc). Then looks up the corresponding values from valDict using that key. It's a matching process rather than using the iteration number of the loop.

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: removed instances of is True. There may still be some `hasattr`'s in other conditions that could follow the same treatment

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: Written `return None`!

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: Changed the condition such that `len(barycentricTOA)==1` will set correct `T0X_per_toa` array based on if the TOA lies in a group.

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: Testing variable, deleted.

poneill129: T0X_per_toa was not present on model initialisation, this was likely a hangover from testing when tt0 gets calculated

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: Duplicate branch removed

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: Addressed during string formatting corrections

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: Can't tell the exact line this is referencing, but par_obj is used flexibly get the unit we expect with an arbitrary piecewise parameter.

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: All gone now!

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
poneill129: Added doc string to in_piece.
It finds which TOAs reference a given piecewise binary parameter group par.
It looks up these values internally from the property group_array_index, which is generated when `self._t` is not `None`
Returns a boolean list (length = number of TOAs), True for the index where TOAs reference a given piecewise parameter, False otherwise. 

Just occurred to me that this may be an issue if the TOAs are reordered between calculations. Unless the TOAs always reach here in a controlled order

tests/test_BT_piecewise.py
poneill129: Addressed this only once (other functions like: build_with_two_pieces could receive a similar treatment)

tests/test_BT_piecewise.py
poneill129: Removed redundant functions like this and similar

tests/test_BT_piecewise.py
poneill129: Fixed as part of the previous comment

tests/test_BT_piecewise.py
poneill129: Unfortunately I couldn't find where some of these additional tests live when updating. Still needed.

Otherwise, renamed test to something appropriate. (Likely other instances of inappropriate test names)

tests/test_BT_piecewise.py
poneill129: Good spot, yes it should be

tests/test_BT_piecewise.py
poneill129: Unchanged yet, fell through the cracks while updating

tests/test_BT_piecewise.py
poneill129: Included the types of exceptions that should be raised (basically all value errors)

tests/test_BT_piecewise.py
poneill129: As suggested, just running the code. New function added to `binary_piecewise.py` called `lock_groups` which runs validate then setup. Avoids the need to run it after every piecewise parameter is added but must be run after adding the a set of/sets of piecewise parameters and date ranges for the boundaries.

tests/test_BT_piecewise.py
poneill129: Fixed to only include a single assertion in a `with` statement. I realise most of these have been removed during the update

tests/test_BT_piecewise.py
poneill129: Separated out into two functions since overlapping groups are now fail validation tests.

tests/test_BT_piecewise.py
poneill129: It's removed now, all it was doing is making 1->0001 so I just declare them where necessary now rather than using a function

tests/test_BT_piecewise.py
poneill129: Didn't know that, thanks!

tests/test_BT_piecewise.py
poneill129: Again separated into two tests, since overlapping groups are handled differently now

tests/test_BT_piecewise.py
poneill129: Talk about going the long way to get the same thing! Silly oversight on my part

tests/test_BT_piecewise.py
poneill129: It appears given the updates I don't (/perhaps never needed to), now constructing residuals is reserved for tests that actually use them.

tests/test_BT_piecewise.py
poneill129: Thanks for the info, but this test has been removed and replaced now, informed by some of the test suggestions you make in the next comment!

poneill129:
Adding some of my extra thoughts based on Anne's last block of suggestions.

> Can you compare the residuals from BTPiecewise with several pieces to BT with the values appropriate for that piece? This is supposed to give exactly the same residuals but could go wrong. Checking derivatives would be good too.
- Yes! Implemented as: `test_residuals_in_pieces_are_same_as_BT_piecewise` followed by the orbital parameter it is testing.
- Also `test_derivatives_in_pieces_are_same_as_BT_piecewise` followed by the orbital parameter is implemented to check the derivatives correctly being calculated inside/outside of a given piece. Warning: `test_derivatives_in_pieces_are_same_as_BT_piecewise_A1` is producing assertion errors if `assert_array_equal` is used, looks like its some form of round-off issue as the BT and BT_piecewise derivatives are effectively equal.  

>There are some tests elsewhere that use numdifftools to check analytical derivatives against numerical. Can you integrate a BTPiecewise model into that?
Haven't been able to spot where these live. Seems like a good idea to implement still

>Can you write some corner-case tests - what happens if the T0X intervals are out of order? If the end is before the beginning? If it's zero-width? If they're duplicated or overlap? Most of these should just raise exceptions, but tests can document what should happen if the user puts something wrong in.
Implemented as validation tests which are checked in the testing script.

>Can you write a test that creates two different BTpiecewise models and calls various functions of one interlaced with calls to the other? Just to make sure there's no global caching of anything (I'm sure you're fine but it's good to check).
>Can you write some tests that do some calculations, change some values, and do some new calculations; versus doing the second lot of calculations directly? This is to make sure that stuff doesn't linger in BTpiecewise objects from one call to the next.
I think I've made a start on testing some of these in the final test of the script. It performs some basic arithmetic and attempts to change piecewise parameter values in two independent models based on the other model. Currently passing but I doubt this is the exact form you were suggesting. Further discussion required.

poneill129:
Replying to @dlakaplan's comment. Yes there does seem to be a natural way these two can interface. I would need to think about how I would implement this. Either as a `adjust_piecewise_param(new_par_val,par_name)`. I'd need to check how exactly the T2 model adjusts orbital parameters but gut feeling tells me it can be made to work together.

dlakaplan:
@poneill129 : are you able to attend the weekly PINT call to discuss this?

poneill129:
> @poneill129 : are you able to attend the weekly PINT call to discuss this?

Yes I will be, will this be on Thursday? If so, what time does it start?

abhisrkckl:
A completed version of this (#1545) has already been merged. So I am closing this.

Issue #1377, opened 2022-08-11
Verbosity in scripts


dlakaplan:
Currently executable scripts change logging level via:
```
(pintdev) kaplan@plock[~/pythonpackages/PINT_dlakaplan] (polycoreadwrite) % pintk -h
usage: pintk [-h] [--ephem EPHEM] [--test]
             [-f {auto,downhill,WLSFitter,GLSFitter,WidebandTOAFitter,PowellFitter,DownhillWLSFitter,DownhillGLSFitter,WidebandDownhillFitter,WidebandLMFitter}]
             [-v] [--log-level {TRACE,DEBUG,INFO,WARNING,ERROR}]
             parfile timfile
...
  --log-level {TRACE,DEBUG,INFO,WARNING,ERROR}
                        Logging level
```
This isn't necessarily the most standard interface, but it is explicit.  On the other hand we could easily change them all to:
```
--verbose
```
where the number of `--verbose` or `-v` arguments increase the verbosity.  Do people have opinions on which way to go?

scottransom:
@dlakaplan We can have both, right?  If the default log level is, say, `INFO`, specifying `--log-level` gives you exactly what you might want.  But you could also specify `--verbose` and get the equivalent of one extra level of logging (i.e. `DEBUG`) whereas `--quiet` would give you `WARNING`.  And multiple uses of those just increment or decrement the level appropriately.  Basically they are all just different ways of changing the log level.

dlakaplan:
Yes, they could both be there, but I feel that it might be confusing to read.  But if that's the consensus then I can do it.

dlakaplan:
If you want both, is this what you are looking for?

```
(pintdev) kaplan@plock[~/pythonpackages/PINT_dlakaplan] (scriptverbose) % compare_parfiles -h                                                    
usage: compare_parfiles [-h] [--dmx] [--no-dmx] [--convertcoordinates] [--no-convertcoordinates] [--sigma SIGMA] [--uncertainty_ratio UNCERTAINTY_RATIO]
                        [--nocolor] [--comparison {max,med,min,check}] [--format {text,markdown}] [--log-level {TRACE,DEBUG,INFO,WARNING,ERROR,CRITICAL}] [-v] [-q]
                        input1 input2
...
  --log-level {TRACE,DEBUG,INFO,WARNING,ERROR,CRITICAL}
                        Logging level (default: WARNING)
  -v, --verbosity       Increase output verbosity (default: 0)
[...]

PR #1380, opened 2022-08-16
Making script verbosity easier to control

dlakaplan:
Addresses #1377

Separately, we might want to check which executables/scripts are still needed.  I didn't update all of them since some don't seem to be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - delta_pulse_number column is now saved to -padd flag on TOA write
 - command-line utility to compare parfiles
 - FD_delay_frequency function to easily access the FD model's excess delay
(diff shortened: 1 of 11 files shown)


paulray:
Looks good to me! Are the only excluded scripts the 2 alternatives to event_optimize.py? I'm not really sure what those are.

dlakaplan:
I updated the recipe for how to do this in:
https://nanograv-pint--1380.org.readthedocs.build/en/1380/_autosummary/pint.logging.html
and 
https://nanograv-pint--1380.org.readthedocs.build/en/1380/_autosummary/pint.logging.get_level.html#pint.logging.get_level

dlakaplan:
Both done.  `pintbary` has a number of print statements which is probably why I didn't update it before.

paulray:
Great! I think this can be merged once tests pass

Issue #1352, opened 2022-07-22
Why doesn't the extraction of single TOA objects with the indexing operator work?


abhisrkckl:
Doing something like `toa0=toas[0]` gives the following error:
>ValueError: TOAs do not support extraction of TOA objects (yet?)

aarchiba:
Single TOA objects are kind of broken - the way we handle clock corrections is built into TOAs objects and it's not really clear what to do when extracting a single TOA object. This is undesirable; ideally a TOAs object should look a lot like an array of individual TOA objects and those TOA objects would support versions of almost all the methods of a TOAs object. It's not so clear how to do this cleanup without duplicating lots of code, though.

abhisrkckl:
If that is the case, why do they exist? Anything we can do with a single TOA object can probably be done with a TOAs object with a row in its table. In fact, we can define a TOA object as a wrapper for a TOAs object with a single row.

scottransom:
Part of the reason is history.  Originally, we had individual TOA objects which were kept together as groups in lists.  But that was way to slow, so we brought in astropy tables.  But having an individual table for a single TOA seems like massive overkill.  So I'm not sure what the right answer is for individual TOAs.

PR #1381, opened 2022-08-17
Extract single TOA as table

dlakaplan:
Extracting a single TOA as a `TOA` object does not work.  But is there a reason we shouldn't enable a length=1 `TOAs` object?  This allows that, and updates the docs slightly.

Addresses #1352 

If this is a bad idea happy to shelve.


dlakaplan:
Currently includes a warning stating what it's doing.  That could be removed it it's too annoying.

dlakaplan:
Changed warning -> info

dlakaplan:
I think this is ready to merge as soon as the CI passes.

dlakaplan:
I'm not sure.  Possibly some of the `TOAs` construction could do that, but I think you can already construct TOAs from other TOAs.

Issue #1355, opened 2022-07-22
`pint.pint_matrix.combine_design_matrices_by_param` throws wrong exception when duplicate parameters are present.


abhisrkckl:
Here is how to reproduce the issue:
```
import pint, pint.models
import astropy.units as u
from pint.pint_matrix import DesignMatrixMaker, combine_design_matrices_by_param

parfile = pint.config.examplefile("B1855+09_NANOGrav_dfg+12_TAI.par")
timfile = pint.config.examplefile("B1855+09_NANOGrav_dfg+12.tim")
model, toas = pint.models.get_model_and_toas(parfile, timfile)

dmm = DesignMatrixMaker('phase', u.Quantity(1,''))
dm1 = dmm(toas, model, ['F0'], offset=True)
dm2 = dmm(toas, model, ['F1'], offset=True)

dm12 = combine_design_matrices_by_param(dm1, dm2)
```

The expected error message is "Input design matrix {} has duplicated parameters with matrix {}".
The actual error message I get is the following. There seems to be an undefined variable "ii" used in the raise statement.
> NameError                                 Traceback (most recent call last)
> Input In [31], in <cell line: 1>()
> ----> 1 dm12 = combine_design_matrices_by_param(dm1, dm2)
> 
> File ~/anaconda3/envs/pint-devel/lib/python3.9/site-packages/pint_pulsar-0.7+2686.gbfa227ef.dirty-py3.9.egg/pint/pint_matrix.py:611, in combine_design_matrices_by_param(matrix1, matrix2, padding)
>     607 for d_param in matrix2.derivative_params:
>     608     if d_param in base_params:
>     609         raise ValueError(
>     610             "Input design matrix {} has duplicated "
> --> 611             "parameters with matrix {}".format(ii, 0)
>     612         )
>     613 # check if input design matrix has same quantity and padding.
>     614 new_quantity_index = {}
> 
> NameError: name 'ii' is not defined

PR #1384, opened 2022-08-18
Fix bug #1355

abhisrkckl:
Fixes issue #1355


src/pint/pint_matrix.py
```diff
@@ -607,8 +607,7 @@ def combine_design_matrices_by_param(matrix1, matrix2, padding=0.0):
     for d_param in matrix2.derivative_params:
         if d_param in base_params:
             raise ValueError(
-                "Input design matrix {} has duplicated "
-                "parameters with matrix {}".format(ii, 0)
+                "Unable to combine the two design matrices as they have duplicated parameters."
```
aarchiba, line 610: Please ensure that the new message informs the user which parameters are duplicated like the old one did.

tests/test_combine_design_matrices_by_param.py
```diff
@@ -0,0 +1,28 @@
...
+
+@pytest.fixture()
+def model_and_toas():
+    parfile = examplefile("B1855+09_NANOGrav_dfg+12_TAI.par")
+    timfile = examplefile("B1855+09_NANOGrav_dfg+12.tim")
+    model, toas = get_model_and_toas(parfile, timfile)
+    return model, toas
+
+
+def test_combine_design_matrices_by_param(model_and_toas):
```
aarchiba, line 18: What property of the function does this test verify? If it's verifying more than one, please split it into separate tests so we can tell where failures occur.

src/pint/pint_matrix.py
abhisrkckl: I have changed this to mention `d_param` in the error message.

tests/test_combine_design_matrices_by_param.py
abhisrkckl: I have renamed this test and added a docstring to explain what is being tested.

src/pint/pint_matrix.py
```diff
@@ -607,8 +607,7 @@ def combine_design_matrices_by_param(matrix1, matrix2, padding=0.0):
     for d_param in matrix2.derivative_params:
         if d_param in base_params:
             raise ValueError(
-                "Input design matrix {} has duplicated "
-                "parameters with matrix {}".format(ii, 0)
+                f"Unable to combine the two design matrices as they have duplicated parameter(s). The duplicated parameter is {d_param}."
```
dlakaplan, line 610: Could I suggest a slight modification to make it easier to parse:

f"Unable to combine the two design matrices as they have duplicated parameter(s): '{d_param}'"

abhisrkckl: Done.

dlakaplan:
Also, can you add something to `CHANGELOG.md`?

Issue #1363, opened 2022-07-27
Split the computation of the correlated noise basis and weights?


abhisrkckl:
At present, the correlated noise basis matrix (F) and weights (W) for PLRedNoise and EcorrNoise are computed by a single function (ecorr_basis_weight_pair and pl_rn_basis_weight_pair). This is wasteful for the following reason: the weights depend on the noise parameter values whereas the basis matrix does not. This means that while sampling the parameter space in the context of something like an MCMC run, the basis matrix needs to be computed only once in principle for a given TimingModel and TOAs. This can be quite computationally expensive for PLRedNoise since its basis matrix contains sines and cosines. 

I propose that we split this computation into two functions (e.g., pl_rn_basis and pl_rn_weight) so that only the weight function needs to be called repeatedly.

aarchiba:
That seems quite reasonable. Is there any situation where the basis functions might need to be recomputed?

abhisrkckl:
There is only one scenario I can think of. By default, the fundamental frequency of the red noise is taken as 1/T where T is the data time span. However, if we want to take T as a model parameter, the basis matrix needs to be re-computed for every new value of T. 

AFAIK any value of T greater than or equal to the data time span gives similar results since fitting for F0 and F1 kills all low-frequency red noise components. So this is a moot point for most use cases, except if someone is specifically investigating the effect of T on their noise estimates.

aarchiba:
The optimization this is designed to enable would fail in that case, though. Not necessarily a serious problem but it is something to be aware of. (An alternative approach could cache the basis functions to avoid unnecessary recomputation.)

abhisrkckl:
The use case I mentioned is a really rare one and is not currently supported by PINT. i.e., the fundamental frequency of the Fourier basis is not a parameter of PLRedNoise. Currently, it is simply taken as the data time span. The `create_fourier_design_matrix` function has an option to set an arbitrary Tspan (reciprocal of the fundamental frequency) but is not actually used. If anyone wants to use it that way, they will have to extend PLRedNoise or create a separate model. BTW, tempo2 has a TNRedFLow parameter, which I think is supposed to be the fundamental frequency.

abhisrkckl:
The issue with the caching approach is that the toa objects are mutable. If the toa object changes between two function calls (e.g. changing the flag that ECORR is based on), then the cache must go stale.  I don't think there is any simple and reliable way to check this.

aarchiba:
[...]

PR #1385, opened 2022-08-19
Split correlated noise (ECORR and Red noise) basis and weight computation

abhisrkckl:
See issue #1363 for a discussion on this.


src/pint/models/noise_model.py
```diff
@@ -345,12 +345,10 @@ def get_ecorrs(self):
             ecorrs.append(getattr(self, ecorr))
         return ecorrs
 
-    def ecorr_basis_weight_pair(self, toas):
-        """Return a quantization matrix and ECORR weights.
+    def get_basis(self, toas):
```
aarchiba, line 348: Unfortunately, these functions will appear as methods of any TimingModel that has a noise component, so `get_basis` is a little confusing. Perhaps a mention of "noise" in the method name would help make it more obvious what basis you were talking about?

src/pint/models/noise_model.py
```diff
@@ -464,6 +486,27 @@ def get_pl_vals(self):
             amp, gam = self.RNAMP.value / fac, -1 * self.RNIDX.value
         return (amp, gam, nf)
 
+    def get_basis(self, toas):
+        """Return a Fourier design matrix. See the documentation for
```
aarchiba, line 490: The first line of a function docstring should be a simple sentence in the imperative voice that will serve as the summary in the table of methods.

src/pint/models/noise_model.py
```diff
@@ -474,20 +517,18 @@ def pl_rn_basis_weight_pair(self, toas):
...
 
     def pl_rn_cov_matrix(self, toas):
         Fmat, phi = self.pl_rn_basis_weight_pair(toas)
         return np.dot(Fmat * phi[None, :], Fmat.T)
 
 
-def create_quantization_matrix(toas_table, dt=1, nmin=2):
-    """Create quantization matrix mapping TOAs to observing epochs."""
+def get_ecorr_epochs(toas_table, dt=1, nmin=2):
+    """Find only epochs with more than 1 TOA for applying ECORR."""
```
aarchiba, line 528: Does this really not return epochs with only one TOA? I would expect ECORR to be applied to those too.

tests/test_split_corrnoise_basis_weights.py
```diff
@@ -0,0 +1,88 @@
...
+    create_fourier_design_matrix,
+    get_rednoise_freqs,
+    powerlaw,
+)
+
+
+@pytest.fixture()
+def model_and_toas():
+    parfile = examplefile("B1855+09_NANOGrav_9yv1.gls.par")
+    timfile = examplefile("B1855+09_NANOGrav_9yv1.tim")
```
aarchiba, line 21: This file has a lot of TOAs and takes a long time to load, every time the test suite is run. Is there a smaller example that would work?

tests/test_split_corrnoise_basis_weights.py
```diff
@@ -0,0 +1,88 @@
...
+    tbl = toas.table
+    t = (tbl["tdbld"].quantity * u.day).to(u.s).value
+    amp, gam, nf = plrncomponent.get_pl_vals()
+    Fmat = create_fourier_design_matrix(t, nf)
+    f = get_rednoise_freqs(t, nf)
+    weight = powerlaw(f, amp, gam) * f[0]
+    return (Fmat, weight)
+
+
+def test_ecorrnoise_basis_weights(model_and_toas):
```
aarchiba, line 63: Comparing against the old version is not an ideal test - if we make some improvement to the function and the test fails, we won't know whether we've fixed a bug or what. Can you write tests that check key properties of the functions you are using?

tests/test_split_corrnoise_basis_weights.py
```diff
@@ -0,0 +1,88 @@
...
+    model, toas = model_and_toas
+    ecorrcomponent = model.components["EcorrNoise"]
+
+    basis_old, weights_old = ecorr_basis_weight_pair_old(ecorrcomponent, toas)
+
+    basis, weights = ecorrcomponent.ecorr_basis_weight_pair(toas)
+
+    # The assertion expects an equality because the computations have not changed.
```
aarchiba, line 71: This is guaranteed to produce problems down the road, as soon as we do want to change how the functions work. Please come up with tests that check the properties we actually want from these functions.

src/pint/models/noise_model.py
abhisrkckl: I agree. I'll change this.

abhisrkckl: I'll change `get_basis ` -> `get_noise_basis`  and `get_weights` -> `get_noise_weights`.

src/pint/models/noise_model.py
abhisrkckl: Modified this.

src/pint/models/noise_model.py
abhisrkckl: The existing implementation does the same thing. One of the tests I've included (test_ecorrnoise_basis_weights) checks that the new code produces results identical to the existing code. 

The relevant code is essentially identical to the code in ENTERPRISE. I think someone copy-pasted it.
https://github.com/nanograv/enterprise/blob/2fb1bf3cdc086ebe10da4689a7312fdda54e47ab/enterprise/signals/utils.py#L744

abhisrkckl: There is a comment there that explicitly says "find only epochs with more than 1 TOA".

tests/test_split_corrnoise_basis_weights.py
abhisrkckl: I'll create a smaller version of this.

tests/test_split_corrnoise_basis_weights.py
abhisrkckl: I have changed the tests.

PR #1388, opened 2022-08-24
Delete jumped TOAs

AkashA98:
This can delete the TOAs that are jumped (not all of them, more below). And also the `stationaryClick` function repeatedly deletes closest TOAs even on multiple clicks at the same location (top vs bottom where I keep clicking on the same point). Is this the desirable behavior? If not, this is also fixed in this PR by working with TOA index rather than the in-plot index (if the closest TOA to click point is already deleted, then throw a warning and do nothing rather than deleting the closest non-deleted TOA)

Screenshot from 2022-08-24 11-58-05
Screenshot from 2022-08-24 11-58-16

Coming back to deleting the jumped TOAs, right now this allows for deleting a subset of TOAs that are jumped, not all of them. If we delete all of the jumped TOAs, then we need to update the jump flag, else fit throws an error
`pint.models.timing_model.MissingTOAs: Parameter 'JUMP1, -gui_jump, ['1']' does not correspond to any TOAs`. 

However, to resolve this, when I try to update the jump flag of the selected TOAs to "un-jump" them before deleting, I could only manage this when I select all the jumped TOAs. If I select a subset of TOAs and use the `add_jump` function to "un-jump" them, then I get a warning and a `loguru` error
`WARNING  (pint.models.jump              ): The selected toa(s) overlap an existing jump.Remove all interfering jumps before attempting to jump these toas.`

`log.error(
  File "/home/akash/softwares/miniconda3/envs/pint/lib/python3.10/site-packages/loguru/_logger.py", line 1989, in error
    __self._log("ERROR", None, False, __self._options, __message, args, kwargs)`

`File "/home/akash/softwares/miniconda3/envs/pint/lib/python3.10/site-packages/loguru/_logger.py", line 1954, in _log
    log_record["message"] = message.format(*args, **kwargs)`

`AttributeError: 'NoneType' object has no attribute 'format'`

Is changing the jump status of a subset of jumped TOAs, not a desirable behavior? Is there a specific reason for it? 

Wanted to ask this and get updates on this PR (I might have missed updating the code somewhere else that I might not be aware of) before I change anything else.


aarchiba:
There's some context for handling JUMPs. `-jump` flags are a mistake built into pintk; they attach model information to TOAs and are constantly getting out of date. I wrote a draft function to simply regenerate them all from scratch whenever needed, but I'm not sure whether it made it in. Ultimately I think they should go away entirely.

JUMPs that affect no TOAs are not actually a problem as long as those JUMPs are frozen; I had a draft function to automatically freeze all JUMPs that affect no TOAs, but that definitely didn't make it in.

If all TOAs are JUMPed, that's a problem too; my draft function froze one or more of the JUMPs in this situation.

If multiple JUMPs affect the same set of TOAs, you can have a degeneracy; this is more awkward to detect, but PINT's fitters should not be completely wrecked by the presence of such degeneracies.

To get to the purpose of your PR: 

- The mutliple clicking behaviour sounds like a bug that's worth fixing. Thanks!
- Deleting JUMPed TOAs - I'm not quite clear on what you want to do.

Are you trying to:

- Deleting all TOAs affected by a particular JUMP - It's not obvious to me why this is a useful operation, but if people are trying to do it interactively, this could be useful. 
- Make those TOAs not JUMPed but still part of the data set - This is tricky because you can have JUMPs that select by frequency (JUMP all TOAs between 1000 and 2000 MHz); how do you arrange for a TOA to be excluded from that set? pintk has "GUI jumps", which are indicated by (I believe) a flag like `-guijump 17` which you could just remove.

In any case, once you've removed some TOAs and/or changed their flags, you probably want to:

- Regenerate all the `-jump` flags to reflect the new realitty (there should be a function to do this)
- Freeze any JUMPs that do not affect any TOAs or are otherwise problematic

AkashA98:
The basic issue is that pintk in its current state throws a warning if you try to delete any TOAs that are jumped (This is a `guijump` using pintk; if the par file has pre-existing jumps, they can be deleted in pintk routinely). So if we add in a jump to a subset of TOAs and want to delete TOAs within that subset, that's when the warning pops up. We can re-generate the jump flag in this case, but the existing function that un-JUMPs the JUMPed TOAs throws this warning (when we select a subset of `guijump` JUMPed TOAs)

`WARNING  (pint.models.jump): The selected toa(s) overlap an existing jump. Remove all interfering jumps before attempting to jump these toas.`

and a subsequent logging error breaks the code. I wasn't sure before what was going on here, but I am looking at it now.

And regarding multiple clicks deleting closest non-deleted TOAs, a few people in the pint call last week suggested that this is desirable behavior.

AkashA98:
I think I addressed the issue with deleting the jumped TOAs. In the current state, the issue seems to be:

If we select, say 50 TOAs, add a `gui_jump` to it, then reselect **all** of them and hit "j" in pintk it deletes the jump flags from all the 50 TOAs and also removes the jump parameter. However, if we select the first 25 TOAs (say) and then hit "j" in pintk, instead of removing the `gui_jump` from these TOAs, it interprets as adding a second jump to these TOAs and hence it throws the warning

`WARNING (pint.models.jump): The selected toa(s) overlap an existing jump. Remove all interfering jumps before attempting to jump these toas.`

I changed this a bit by making it delete the jump flags from the selected TOAs if they already are already jumped (it can be any subset as opposed to all the TOAs with jump flag). Much of its functionality is a subset of an already existing function that deletes **all** the TOAs with a given jump number, but I was a bit hesitant to modify that function since it is used in some tests.

So if we want to delete TOAs that have a `gui_jump`, it first deletes the jump flags from them and then deletes the TOAs.

AkashA98:
There is an additional minor change that concerns how indexing is done for single clicks. The right single click nominally deletes the closest TOA to the click. The way it is done currently can have a mismatch:

Click on the plot gives the plot index of the TOA rather than its actual index. For example, if we have 1000 TOAs, the plot index and and the TOA index is same for all the TOAs. But if we delete the first 50, then the 500th TOA, with have a plot index of 450, but its index in the TOA table remains 500. So when we click on the 500th TOA, instead of deleting the TOA with index 500, it deletes the TOA with index 450, because plot index is passed to the `delete_TOAs` function rather than the actual TOA index. 

I believe that I fixed that in this same PR.

AkashA98:
I did modify info messages for logging when the TOAs are un-jumped saying something like "Removing existing jump JUMP2 from 20 TOAs" and jumped TOAs are deleted...

AkashA98:
> I did modify info messages for logging when the TOAs are un-jumped saying something like "Removing existing jump JUMP2 from 20 TOAs" and jumped TOAs are deleted...

The actual log messages look something like:
When a jump is added:
```
INFO     (pint.pintk.pulsar             ): PhaseJump component added
INFO     (pint.pintk.pulsar             ): New jump JUMP1 added for 70 toas.
INFO     (pint.pintk.plk                ): JUMP1 contains 162 TOAs for fit.
```

And when un-jumped:

`INFO     (pint.pintk.pulsar             ): Removing existing jump JUMP1 from 62 TOAs`

dlakaplan:
@scottransom : I think this is OK to merge but I am still a little confused about some of the JUMP info messages (which predate this PR).  So maybe good for you to check.

src/pint/pintk/pulsar.py
```diff
@@ -388,6 +388,8 @@ def add_jump(self, selected):
             )
             if self.fitted:
                 self.postfit_model.add_component(a)
+            log.info(f"New jump {retval} added for {selected.sum()} toas.")
```
scottransom, line 391: So I  just noticed that this is now a `log.info` *and* a `print`.  I think that we were suggesting that if anything changes with the JUMPs that you should give a `log.info` *rather than* a `print`.  I just noticed another `print` up above, as well, that should probably be a `log.info`.

src/pint/pintk/pulsar.py
```diff
@@ -412,12 +415,32 @@ def add_jump(self, selected):
...
+                jumped_selected = toas_jumped & selected
+                # Post fit model and prefit model share the same TOA table, so as long as we
+                # don't delete the jump altogether, modifying prefit model table flags is fine.
+                self.prefit_model.delete_not_all_jump_toas(
+                    self.all_toas.table["flags"][jumped_selected], num
+                )
+                log.info(
+                    f"Removing existing jump JUMP{str(num)} from {jumped_selected.astype(int).sum()} TOAs"
+                )
+                print(
```
scottransom, line 433: here's another example

AkashA98: Oh, I see, I misinterpreted this as having the option of both. I just left the `print` statement there to be a bit more informative. I can clean that up if required...

scottransom: Yeah, I think just the log.info's are enough.  Because the user is going to see the TOAs and/or JUMP disappear from the display as well.  So I think just zap the prints.

AkashA98: ok, thanks

AkashA98: @scottransom @dlakaplan I did change this to reflect jump changes as `log.info`. Can someone please approve the CI?

AkashA98:
This seems to be done except for the failed oldestdeps issue. If there aren't any further suggestions, this might be good to go... @scottransom @dlakaplan

dlakaplan:
Odd that the oldestdeps error seems to have gone away even without the fix from @paulray, so maybe something changed with GitHub, but this looks good.  I'll try to merge later.

src/pint/models/jump.py
```diff
@@ -242,3 +242,32 @@ def add_jump_and_flags(self, toa_table):
             dict1["jump"] = str(ind)
             dict1["gui_jump"] = str(ind)
         return name
+
+    def delete_not_all_jump_toas(self, toa_table, jump_num):
+        """Unjumps a subset of jumped TOAs when the selected subset  doesn't exactly
+        overlap with all the TOAs in a given jump.
+
+        Helper function for pintk. Mailny used when one wants to delete the jump from
```
dlakaplan, line 250: typo in this comment

src/pint/models/jump.py
```diff
@@ -242,3 +242,32 @@ def add_jump_and_flags(self, toa_table):
...
+        a subset of jumped TOAs.
+
+        Parameters
+        ----------
+        toa_table: list object
```
dlakaplan, line 255: is that the correct object type?  is it not a `astropy.table.Table`?

dlakaplan:
Can you also add a quick CHANGELOG entry?

AkashA98:
> Can you also add a quick CHANGELOG entry?

I can do that.... Thanks

src/pint/models/jump.py
AkashA98: That's a bad naming from my side. The object refers to the list of flags of the selected TOAs from the TOA table. I'll change the name...

src/pint/models/jump.py
```diff
@@ -242,3 +242,30 @@ def add_jump_and_flags(self, toa_table):
...
             dict1["gui_jump"] = str(ind)
         return name
+
+    def delete_not_all_jump_toas(self, toa_flags, jump_num):
+        """Unjumps the selected TOAs if they are already jumped. This is only called
+            when the selected TOAs are a subset of jumped TOAs with a particular jump.
+
+        Parameters
+        ----------
+        toa_flags: astropy.table.column.Column
```
dlakaplan, line 252: I thought it was a table, not a column?

Issue #1389, opened 2022-08-29
Residuals do not converge in tutorial/example script `Simulate_and_make_MassMass`


TigerHou2:
### Basic Information
Operating System: WSL Ubuntu-20.04
`pint.__version__`: `0.9.1+35.g31c3865d.dirty`
`pint.__file__`: `/mnt/c/Users/█/PINT/src/pint/__init__.py`
Detailed setup: 
- `pint` installed from fork of `master` per [documentation](https://nanograv-pint.readthedocs.io/en/latest/installation.html#install-from-source) for installing from source. The fork is on par with `master`. 
- Dependencies installed via `pip install -e .` in a `conda env` within `WSL 1`. 
- My native OS is Windows 10, Version 10.0.19044 Build 19044, if that is relevant. 

### Steps to Reproduce
1. Run the `Simulate_and_make_MassMass.ipynb` example in Jupyter Notebook.

### The Issue
When calling `pint.simulation.make_fake_toas_uniform`, two issues appear that are not in the [documented example](https://nanograv-pint.readthedocs.io/en/latest/examples/Simulate_and_make_MassMass.html):
1. `pint.simulation` issues the warning
```
Residual increasing at iteration <x> while attempting to simulate TOAs
```
2. `ValueError` is then raised by the `zero_residuals` function:
```
File /mnt/c/Users/█/PINT/src/pint/simulation.py:77, in zero_residuals(ts, model, maxiter, tolerance)
     75     ts.adjust_TOAs(-time.TimeDelta(resids))
     76 else:
---> 77     raise ValueError(
     78         "Unable to make fake residuals - left over errors are {}".format(
     79             abs(resids).max()
     80         )
     81     )

ValueError: Unable to make fake residuals - left over errors are 6.341263916438025e-07 s
```

It appears the residuals are bouncing back and forth without converging to zero, but I'm not familiar enough with `pint` to figure out the root cause. Calling `np.finfo(np.longdouble).eps` gives `1.084202172485504434e-19`, so I believe my setup satisfies the precision requirement for `pint`. Any ideas are appreciated, thanks!

paulray:
Do the tests pass if you do `make test`? Note that you may need to install the packages in requirements_dev.txt to get the tests to run.

TigerHou2:
After installing the packages required for development, `make test` produced numerous errors as follows (173 failed, so I'm uploading the test summary as a file):
[...]

PR #1392, opened 2022-08-30
Tools for dealing with limited precision

aarchiba:
As we see in #1389, users can miss or ignore the "limited precision" warning when trying to run PINT with 64-bit longdouble values. This PR tries to make PINT run less badly and signal more loudly in this setting.

- The test suite will not run and will emit an error message about precision.
- `pint.simulation` will make fake TOAs but only good to a few microseconds.
- Functions that require full precision to run can call `require_longdouble_precision()` to do the test and get an informative exception. I haven't identified any such functions yet (do Fitters work okay, for some values of "okay"?)

I would appreciate additional testing in a 64-bit-only environment (M1 macs, Windows MSVC).

### DIFF
--- a/src/pint/simulation.py
+++ b/src/pint/simulation.py
@@ -46,7 +46,7 @@ def _get_freq_array(base_frequencies, ntoas):
     return freq
 
 
-def zero_residuals(ts, model, maxiter=10, tolerance=1 * u.ns):
+def zero_residuals(ts, model, maxiter=10, tolerance=None):
     """Use a model to adjust a TOAs object, setting residuals to 0 iteratively.
 
     Parameters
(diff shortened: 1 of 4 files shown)


aarchiba:
In particular, the test suite should refuse to run at all, with a message about precision.

aarchiba:
Longer-term, we could go through and flag which tests require full precision and disable running of those tests on low-precision machines.

src/pint/simulation.py
```diff
@@ -75,14 +81,12 @@ def zero_residuals(ts, model, maxiter=10, tolerance=1 * u.ns):
...
-            "Unable to make fake residuals - left over errors are {}".format(
-                abs(resids).max()
-            )
+            f"Unable to make fake residuals - left over errors are {abs(resids).max()}"
         )
 
 
 def update_fake_toa_clock(ts, model, include_bipm=False, include_gps=True):
-    """Update the clock settings (corrections, etc) for fake TOAs
+    """Update the clock s:qettings (corrections, etc) for fake TOAs
```
paulray, line 89: Typo

paulray:
I tried this on my M1 Mac and it works as advertised. `make test` fails with a useful error message and the Simulate_and_make_MassMass.py script now runs fine. It warns of limited precision in the first cell, but otherwise seems to work great.  This seems good. We definitely do want to make sure that there are plenty of warnings when running dangerous things without extended precision, but much stuff will work just fine.

paulray (approved):

PR #1395, opened 2022-09-01
Fixes dmxparse save=True exception

### LINKED ISSUE #1394: Backward compatibility in dmxparse broken
Previously, if you passed `save=True` to `dmxparse`, it saved the results to `dmxparse.out` in the current directory. With change https://github.com/nanograv/PINT/commit/046881a813f25af4e11ea8fa24787067d2dc0ed5 by @dlakaplan you get exceptions in this case; you must supply a filename or an open file now.

I suggest detecting True (specifically) and treating that as `dmxparse.out`.

dlakaplan:
Fixes #1394 
Now if `save=True` will write to `dmxparse.out`.  

Also adds tests for writing to `StringIO` and default filename.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 ## Unreleased
 ### Changed
 - global clock files now emit a warning instead of an exception if expired and the download fails
+### Added
+### Fixed
+- dmxparse outputs to dmxparse.out if save=True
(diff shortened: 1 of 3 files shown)


dlakaplan:
The test writes to `dmxparse.out`.  It then deletes the file after.  I'm not sure if this is the best way to handle that - I'd prefer to use some auto-deleting option like a temp file or something that would be safer to use from parallel threads.  Is there a better way?

Issue #1327, opened 2022-07-05
PINT does not support DM red noise models


aarchiba:
An alternative to DMX is to treat DM as a red noise process that has a frequency dependence. PINT could support this the way it supports red noise processes, but it doesn't. It should, in a way as compatible with TEMPO2 as possible.

Parameters:

- TNDMAmp
- TNDMGam
- TNDMC
- TNsubtractDM

The implementation in tempo2 is at https://bitbucket.org/psrsoft/tempo2/src/9f4f29abe564a3f907f8b97cef79385011391a23/t2fit_nestlike.C#lines-28 and it specifiies various constants.

The TempoNest paper describes the model in more detail: https://ui.adsabs.harvard.edu/abs/2014MNRAS.437.3004L/abstract

It's not quite clear how best to compare PINT's implementation to TEMPO2's. Perhaps just do a fit and ensure that the best-fit value is the same (and thus the noise model is adequately similar)?

aarchiba:
There may be more interest in Keith et al.-style DM Gaussian Processes (#1192)

scottransom:
On that note, just to make sure you are aware, Even Keane's student Sai Susarla at Galway is working on implementing that for PINT over this summer.

aarchiba:
Yes, I know, but thanks for the reminder. I have offered support on the Fitter side - implementing the linear algebra to handle the penalization part - if that would be helpful. I know a fair bit about how our Fittters work.

abhisrkckl:
I would like to try my hand at this one.

aarchiba:
I *think* it's just a matter of cloning our red noise parameter model, adding a frequency dependence to the formula, and then chasing down whether our model agrees with that used by tempo2. It is the third part that is potentially rather tricky. My suggestion would be to start with a nearly-right modet, TOAs and par file, and do a fit with tempo2 and PINT both. If the fits take you to the same point, then the noise model should be weighting things the same. You can use a super-simple par file and a modest number of TOAs, but they need both frequency and time coverage, and some wiggliness for the noise parameters to fit.

aarchiba:
I'm also not sure how urgent it is - I had the impression that the EPTA/PPTA relied primarily on DM red noise models to handle DM variations, but I don't think that's actually the case. If no one is using this feature of tempo2, it's less essential that PINT support it.

PR #1396, opened 2022-09-01
Adding PLDMNoise model

blarsen10:
Adding a `PLDMNoise` model which models stochastic DM variations using a powerlaw spectrum. This model shares most of the same code as the `PLRedNoise` model. This new model uses the parameters `TNDMAMP`, `TNDMGAM`, and `TNDMC` as well as a reference frequency of 1400 MHz to match the conventions used in `enterprise` and `TempoNest`.

### DIFF
--- a/tests/test_pldmnoise.py
+++ b/tests/test_pldmnoise.py
@@ -110,6 +110,10 @@ def test_PLRedNoise_recovery():
     MJDs = np.linspace(57001, 58000, 150, dtype=np.longdouble) * u.d
     toas = make_fake_toas_fromMJDs(MJDs, model=model, error=1 * u.us, add_noise=True)
 
+    # get red noise basis and weights
+    rn_basis = model.components["PLRedNoise"].get_noise_basis(toas)
+    rn_weights = model.components["PLRedNoise"].get_noise_weights(toas)
+
     # fit model with red noise
     f1 = fitters.DownhillGLSFitter(toas, model)
     f1.fit_toas()
@@ -133,6 +137,14 @@ def test_PLRedNoise_recovery():
     model["TNDMC"].value = c
     model.validate()
 
+    # get DM noise basis and weights
+    dm_basis = model.components["PLDMNoise"].get_noise_basis(toas)
+    # DM basis will be off by a constant factor, depending on frequency
+    # of the test data
+    D = (1400 / toas.get_freqs().value) ** 2
+    dm_basis_scaled = dm_basis / D[:, None]
+    dm_weights = model.components["PLDMNoise"].get_noise_weights(toas)
+
     # refit model
     f2 = fitters.DownhillGLSFitter(toas, model)
     f2.fit_toas()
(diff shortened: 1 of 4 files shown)


dlakaplan:
Hi, @blarsen10 .  Do you think you could come to a PINT meeting to discuss this?  It looks good but it would help to chat.

Note that you seem to have committed `coverage.xml` and some updated tutorial files that shouldn't be part of this PR.  Can you remove them?

blarsen10:
Hi @dlakaplan . I'd be happy to discuss this at a PINT meeting! However, the meetings are on Tuesdays at 11 AM EST, right? Unfortunately these times conflict with my teaching schedule. Is there an alternative time I can come to discuss this? Let me know if not so I can look for someone to cover my section.

I removed the `coverage.xml` file and tutorial files from the PR.

dlakaplan:
Thanks.  That's too bad about the conflict.  This might also be something that could be discussed at a timing group busy time - @tcromartie organizes those.  Otherwise I think @abhisrkckl said he would take a peak at some of these results and make sure that there is agreement between PINT/temponest/enterprise.   If you have done some tests like that yourself and can share them here on in slack that would be helpful.  It looks like the tests you included mostly just verify that the model functions, not that it agrees with prior results.

abhisrkckl:
Issue #1327 is about this.

abhisrkckl:
BTW PR #1385 will break this.

dlakaplan:
Also, can you add an appropriate entry to `CHANGELOG.md`?

abhisrkckl:
I have checked the expressions. AFAIK they agree with what is implemented in ENTERPRISE.

abhisrkckl:
This PR will conflict with the recently merged PR #1385. That PR splits the noise basis and weight computation in PLRedNoise and ECorrNoise into two functions. Can you do the same for PLDMNoise also? See Issue #1363 also.

blarsen10:
@abhisrkckl Yes I can do that. Splitting the basis and weight computation sounds like a good improvement and it should be straightforward to fix this PR to match. I am a bit swamped at the moment so I can't get to it right away, but I should be able to within the next week or so.

I will also be adding a test in the test suite to ensure the results agree with `PLRedNoise` given monochromatic data.

tests/test_pldmnoise.py
```diff
@@ -0,0 +1,145 @@
...
+def test_read_PLDMNoise_params(modelJ0023p0923):
+    params = ["TNDMAMP", "TNDMGAM", "TNDMC"]
+    for param in params:
+        assert (
+            hasattr(modelJ0023p0923, param)
+            and getattr(modelJ0023p0923, param).quantity is not None
+        )
+
+
+def test_PLRedNoise_recovery():
```
abhisrkckl, line 85: I think it will be good to add another test that compares the basis matrix and weights separately for PLDMNoise and PLRedNoise for a single-frequency dataset. One would expect them to be equal or be off by some constant factor.

blarsen10: That's a good idea, thanks!

abhisrkckl:
@blarsen10 Is this ready to go?

abhisrkckl:
>  I'm guessing the procedure would be to open an issue?
Yes.

Issue #1397, opened 2022-09-14
merge_TOAs() fails with a useless traceback


gooddc:
This is coming from a mix of excised/not-excised tims, but the PINT problem twofold:
1. It would be great if merge_TOAs() could auto-fix this somehow, add a blank column or something and 
2. When merge_TOAs() fails, it doesn't provide anything useful for the user to fix the problem. 

The error message is just

> File /opt/conda/lib/python3.9/site-packages/pint/toa.py:263, in get_TOAs(timfile, ephem, include_bipm, bipm_version, include_gps, planets, model, usepickle, tdb_method, picklefilename, limits)
>     261     t = TOAs(timfile)
>     262 else:
> --> 263     t = merge_TOAs([TOAs(t) for t in timfile])
>     265 files = [t.filename] if isinstance(t.filename, (str, Path)) else t.filename
>     266 if files is not None:
> 
> File /opt/conda/lib/python3.9/site-packages/pint/toa.py:2420, in merge_TOAs(TOAs_list)
>    2418 num_cols = [len(tt.table.columns) for tt in TOAs_list]
>    2419 if len(set(num_cols)) > 1:
> -> 2420     raise TypeError(
>    2421         f"merge_TOAs() cannot merge. Inconsistent numbers of table columns: {num_cols}"
>    2422     )
>    2423 # Use a copy of the first TOAs instance as the base for the joined object
>    2424 nt = copy.deepcopy(TOAs_list[0])
> 
> TypeError: merge_TOAs() cannot merge. Inconsistent numbers of table columns: [9, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8, 8]

I'm not even sure what the 8 columns are in the tims that aren't causing the problem, so it's hard to track down what's different between the two.

dlakaplan:
Could you provide example files/command that cause this error?

paulray:
My table has more than 9 columns. Make sure to compute TDBs and posvels.
```
In [3]: ts = pint.toa.get_TOAs("NGC6440E.tim")
Downloading https://raw.githubusercontent.com/ipta/pulsar-clock-corrections/main/index.txt
|=====================================================================================================| 3.4k/3.4k (100.00%)         0s

In [4]: ts.table.colnames
Out[4]: 
['index',
 'mjd',
 'mjd_float',
 'error',
 'freq',
 'obs',
 'flags',
 'delta_pulse_number',
 'tdb',
[...]

PR #1398, opened 2022-09-14
Can ignore pulse_number column on TOA read/write

dlakaplan:
Reading or writing TOAs with `include_pn=False` will ignore the `pulse_number` column. It will issue a warning that this is happening.  This may make it easier to merge TOA files, but in general it may also be dangerous.  So I'm not convinced that this is overall a good idea but it will address #1397 

The error message when failing to merge is also a little more informative.

### DIFF
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -1228,6 +1228,7 @@ class TOAs:
            :func:`pint.toa.TOAs.compute_pulse_numbers` or extracted from the
            ``pn`` entry in ``flags`` with
            :func:`pint.toa.TOAs.phase_columns_from_flags`.
+           If it is present for some TOAs but not all, missing values are filled with ``NaN``.
        * - ``delta_pulse_number``
          - number of turns to adjust pulse number by, compared to the model;
            ``PHASE`` statements in the ``.tim`` file or the ``padd`` entry in
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -2478,8 +2478,8 @@ def merge_TOAs(TOAs_list):
     # but it should be more helpful
     columns = tables[0].colnames
     for i, tt in enumerate(tables[1:]):
-        extra_columns = [x for x in tt.table.colnames if not x in columns]
-        missing_columns = [x for x in columns if not x in tt.table.colnames]
+        extra_columns = [x for x in tt.colnames if not x in columns]
+        missing_columns = [x for x in columns if not x in tt.colnames]
         if len(extra_columns) > 0:
             log.warning(f"File {i+1} has extra column(s): {','.join(extra_columns)}")
         if len(missing_columns) > 0:
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -2436,11 +2436,20 @@ def merge_TOAs(TOAs_list):
         raise TypeError(f"merge_TOAs() cannot merge. Inconsistent planets: {planets}")
     num_cols = [len(tt.table.columns) for tt in TOAs_list]
     if len(set(num_cols)) > 1:
-        s = ""
-        for i, tt in enumerate(TOAs_list):
-            s += f"File {i} columns: {','.join(tt.table.colnames)}\n"
+        columns = TOAs_list[0].table.colnames
+        for i, tt in enumerate(TOAs_list[1:]):
+            extra_columns = [x for x in tt.table.colnames if not x in columns]
+            missing_columns = [x for x in columns if not x in tt.table.colnames]
+            if len(extra_columns) > 0:
+                log.warning(
+                    f"File {i+1} has extra column(s): {','.join(extra_columns)}"
+                )
+            if len(missing_columns) > 0:
+                log.warning(
+                    f"File {i+1} has missing column(s): {','.join(missing_columns)}"
+                )
         raise TypeError(
-            f"merge_TOAs() cannot merge. Inconsistent numbers of table columns: {num_cols}\n{s}"
+            f"merge_TOAs() cannot merge. Inconsistent numbers of table columns: {num_cols}"
         )
     # Use a copy of the first TOAs instance as the base for the joined object
     nt = copy.deepcopy(TOAs_list[0])
@@ -2462,6 +2471,21 @@ def merge_TOAs(TOAs_list):
         t["index"] += start_index
         start_index += tt.max_index + 1
         tables.append(t)
+    # the astropy vstack will enforce having the same columns throughout
+    # but the error messages aren't very useful.
+    # This isn't an exhaustive check
+    # (it just checks everything against the first observation)
+    # but it should be more helpful
+    columns = tables[0].colnames
+    for i, tt in enumerate(tables[1:]):
+        extra_columns = [x for x in tt.table.colnames if not x in columns]
+        missing_columns = [x for x in columns if not x in tt.table.colnames]
+        if len(extra_columns) > 0:
+            log.warning(f"File {i+1} has extra column(s): {','.join(extra_columns)}")
+        if len(missing_columns) > 0:
+            log.warning(
+                f"File {i+1} has missing column(s): {','.join(missing_columns)}"
+            )
     nt.table = table.vstack(tables, join_type="exact", metadata_conflicts="silent")
     # Fix the table meta data about filenames
     nt.table.meta["filename"] = nt.filename
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -310,7 +310,7 @@ def get_TOAs(
         save_pickle(t, picklefilename=picklefilename)
     if "pulse_number" in t.table.colnames and not include_pn:
         log.warning(f"'pulse_number' column exists but not being read in")
-        del t.table["pulse_number"]
+        t.remove_pulse_numbers()
     return t
 
 
@@ -1893,6 +1893,16 @@ class TOAs:
         phases = model.phase(self, abs_phase=True) + delta_pulse_numbers
         self.table["pulse_number"] = phases.int
         self.table["pulse_number"].unit = u.dimensionless_unscaled
+
+    def remove_pulse_numbers(self):
+        """Delete pulse numbers from TOA data"""
+
+        if "pulse_number" in self.table.colnames:
+            del self.table["pulse_number"]
+        else:
+            log.warning(
+                f"Requested deleting of pulse numbers, but they are not present"
+            )
 
     def adjust_TOAs(self, delta):
         """Apply a time delta to TOAs.
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -103,6 +103,7 @@ def get_TOAs(
     bipm_version=None,
     include_gps=None,
     planets=None,
+    include_pn=True,
     model=None,
     usepickle=False,
     tdb_method="default",
@@ -156,6 +157,8 @@ def get_TOAs(
         Whether to apply Shapiro delays based on planet positions. Note that a
         long-standing TEMPO2 bug in this feature went unnoticed for years.
         Defaults to False.
+    include_pn : bool, optional
+        Whether or not to read in the 'pn' column (``pulse_number``)
     model : pint.models.timing_model.TimingModel or None
         If a valid timing model is passed, model commands (such as BIPM version,
         planet shapiro delay, and solar system ephemeris) that affect TOA loading
@@ -305,6 +308,9 @@ def get_TOAs(
     if usepickle and updatepickle:
         log.info("Pickling TOAs.")
         save_pickle(t, picklefilename=picklefilename)
+    if "pulse_number" in t.table.colnames and not include_pn:
+        log.warning(f"'pulse_number' column exists but not being read in")
+        del t.table["pulse_number"]
     return t
 
 
@@ -1951,6 +1957,7 @@ class TOAs:
         commentflag=None,
         order_by_index=True,
         *,
+        include_pn=True,
         include_info=True,
         comment=None,
     ):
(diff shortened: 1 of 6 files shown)


dlakaplan:
#1400 may also be relevant.

dlakaplan:
Example error trace (now):
```
WARNING  (pint.toa                      ): File 5 has extra column(s): pulse_number
  File "/Users/kaplan/pythonpackages/PINT_dlakaplan/./test_merge.py", line 28, in <module>
    tall = pint.toa.merge_TOAs(t)
  File "/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/toa.py", line 2451, in merge_TOAs
    raise TypeError(
TypeError: merge_TOAs() cannot merge. Inconsistent numbers of table columns: [13, 13, 13, 13, 13, 14, 13]
```
is that reasonable?

scottransom:
I think that that TypeError looks pretty good.

aarchiba:
> I think that that TypeError looks pretty good.

Isn't that basically the error we were getting? That mentions inconsistent numbers of columns but gives no hint which columns are missing/extra/

aarchiba:
Perhaps a convenience function to drop the pulse number column if present would be a more natural way to address the problem of spurious pulse number columns? Plus, more directly, merge_TOAs can know about pulse number columns and add one that's all NaNs if necessary.

scottransom:
I personally don't even really think this is an "issue".  The real issue is that NANOGrav shouldn't have been writing pulse numbers into our exported TOA files.  That is putting model specific things in a .tim file (although hopefully they are *relatively* correct between proper models, at least).  I think I'd worry about fake pulse number columns with NaNs possibly even causing more problems.

aarchiba:
> I personally don't even really think this is an "issue". The real issue is that NANOGrav shouldn't have been writing pulse numbers into our exported TOA files. That is putting model specific things in a .tim file (although hopefully they are _relatively_ correct between proper models, at least). I think I'd worry about fake pulse number columns with NaNs possibly even causing more problems.

We have the limitation currently that unless all TOAs have pulse numbers, pulse numbers are ignored. So it won't cause any problems. But we should fix that (tied up in the way we compute absolute phases) for other reasons anyway. There are many situations where the pulse numbers can be regarded as known while we vary our models, including all the triple-system work, so it's perfectly reasonable to include pulse numbers when passing TOAs to a different optimizer.

The problem we do have - problem in the sense that our users are stuck and frustrated - is that they can't merge perfectly reasonable sets of TOAs and we don't tell them enough information for them to resolve the issue.

dlakaplan:
I am not sure of the best approach.  Right now this PR will
* provide warnings of incompatible numbers/names of columns (although the actual exception will still be the same), so users will hopefully understand better what is happening
* allow users to ignore pulse numbers on read and/or write, which can solve any incompatibility 

It does not include any NaN or other info for a dummy pulse number column, although that's easy.  But I didn't see any documentation of that.  if we decide that this is a better way to go I can add it (together with a warning) easily.

tests/test_toa_reader.py
```diff
@@ -234,6 +235,51 @@ def test_toa_merge_different_ephem():
...
+def test_toa_merge_different_columns():
+    filenames = [
+        datadir / "NGC6440E.tim",
+        datadir / "testtimes.tim",
+    ]
+    model = get_model(datadir / "NGC6440E.par")
+    # add a pulse_number column.  then the merge should fail
+    toas = [toa.get_TOAs(ff, model=model) for ff in filenames]
+    toas[0].compute_pulse_numbers(model)
+    with pytest.raises(TypeError):
```
aarchiba, line 247: Can you check that the exception message contains the name of the offending column? Pytest provides some tools for asserting about exception messages.

aarchiba:
I don't like adding even more arguments to our reading and writing functions. I'd prefer a function to drop pulse numbers (if present), right next to the function to add pulse numbers (based on a model). This achieves the same effect.

dlakaplan:
Now it will not write `pn=nan` to the tim file.  a more general solution may be useful, but it's hard since flags are strings.

src/pint/toa.py
```diff
@@ -305,6 +308,9 @@ def get_TOAs(
     if usepickle and updatepickle:
         log.info("Pickling TOAs.")
         save_pickle(t, picklefilename=picklefilename)
+    if "pulse_number" in t.table.colnames and not include_pn:
+        log.warning(f"'pulse_number' column exists but not being read in")
```
aarchiba, line 312: If the user has explicitly asked to ignore pulse numbers I don't think they should get a warning if pulse numbers are encountered. Nothing at all, or at most an info/debug-level note.

src/pint/toa.py
```diff
@@ -305,6 +308,9 @@ def get_TOAs(
     if usepickle and updatepickle:
         log.info("Pickling TOAs.")
         save_pickle(t, picklefilename=picklefilename)
+    if "pulse_number" in t.table.colnames and not include_pn:
```
aarchiba, line 311: Checking `include_pn` first makes it more obvious that this is controlled by an option, and also saves doing the `in` in the common (default) case where the option isn't set.

src/pint/toa.py
```diff
@@ -1888,6 +1895,16 @@ def compute_pulse_numbers(self, model):
         self.table["pulse_number"] = phases.int
         self.table["pulse_number"].unit = u.dimensionless_unscaled
 
+    def remove_pulse_numbers(self):
+        """Delete pulse numbers from TOA data"""
+
+        if "pulse_number" in self.table.colnames:
+            del self.table["pulse_number"]
+        else:
+            log.warning(
```
aarchiba, line 1904: I'd rather see this as "delete pulse numbers if present" and have it run smoothly with no warnings if they're not present - then this is a cheap way to make sure there are no pulse numbers (just before saving, say).

src/pint/toa.py
```diff
@@ -1951,6 +1968,7 @@ def write_TOA_file(
         commentflag=None,
         order_by_index=True,
         *,
+        include_pn=True,
```
aarchiba, line 1971: I think it's easier to just `remove_pulse_numbers` before saving rather than add yet another option to `write_TOA_file`.

src/pint/toa.py
```diff
@@ -2002,7 +2022,15 @@ def write_TOA_file(
         # FIXME: everywhere else the pulse number column is called pulse_number not pn
         toacopy = copy.deepcopy(self)
         if "pulse_number" in toacopy.table.colnames:
-            toacopy["pn"] = toacopy.table["pulse_number"]
+            if include_pn:
+                toacopy["pn"] = toacopy.table["pulse_number"]
+                # but delete those that are NaN
+                for i in np.where(np.isnan(self["pulse_number"]))[0]:
+                    del toacopy.table["flags"][i]["pn"]
+            else:
```
aarchiba, line 2030: Again, no warning about something explicitly requested.

src/pint/toa.py
```diff
@@ -2420,11 +2453,66 @@ def merge_TOAs(TOAs_list):
...
+    # check for the presence of various computable columns
+    #   pulse_number
+    #   ssb_obs_pos, ssb_obs_vel, obs_sun_pos
+    #   tdb, tdbld
+    # if one file has these and the others don't, try to add them if strict=False
+    all_colnames = [tt.table.colnames for tt in TOAs_list]
+    has_pulse_number = np.array(
+        ["pulse_number" in colnames for colnames in all_colnames]
+    )
+    has_posvel = np.array(
```
aarchiba, line 2465: This logic isn't quite right if some have a few columns but not all. (For example, some have ssb_obs_pos but not ssb_obs_vel.

src/pint/toa.py
```diff
@@ -2420,11 +2453,66 @@ def merge_TOAs(TOAs_list):
...
+            log.warning("Not all data have 'pulse_number' columns but strict=True")
+        else:
+            # some data have pulse_numbers but not all
+            # put in NaN
+            for i, tt in enumerate(TOAs_list):
+                if not "pulse_number" in tt.table.colnames:
+                    log.warning(
+                        f"'pulse_number' not present in data set {i}: inserting NaNs"
+                    )
+                    tt.table.add_column(np.ones(len(tt)) * np.nan, name="pulse_number")
```
aarchiba, line 2490: Doesn't this modify the input table? I don't think `merge_TOAs` should ever modify its inputs.

src/pint/toa.py
```diff
@@ -2420,11 +2453,66 @@ def merge_TOAs(TOAs_list):
...
+            for colnames in all_colnames
+        ]
+    )
+    has_tdb = np.array(
+        [("tdb" in colnames) and ("tdbld" in colnames) for colnames in all_colnames]
+    )
+    has_posvel_ecl = np.array(
+        ["ssb_obs_vel_ecl" in colnames for colnames in all_colnames]
+    )
+    if has_pulse_number.any() and not has_pulse_number.all():
```
aarchiba, line 2479: You don't need the all() check, any that has the required columns will be untouched.

tests/test_pulse_number.py
```diff
@@ -64,6 +64,12 @@ def test_pulse_number(model, toas):
     assert "pulse_number" in toas.table.colnames
 
 
+def test_remove_pulse_number(model, toas):
```
aarchiba, line 67: It'd be good to have a test to demonstrate what happens when pulse numbers aren't present and are removed - this is where one specifies whether or not there should be a warning.

tests/test_toa_reader.py
```diff
@@ -234,6 +236,121 @@ def test_toa_merge_different_ephem():
...
+    t = toa.get_TOAs(filenames[0], model=model)
+    t.compute_pulse_numbers(model)
+    f = StringIO()
+    t.write_TOA_file(f)
+    f.seek(0)
+    toas = [toa.get_TOAs(fname, include_pn=False) for fname in [f, filenames[1]]]
+    nt = toa.merge_TOAs(toas)
+
+
+def test_toa_merge_different_columns_ignorepn_onwrite():
```
aarchiba, line 338: Is there any need to test the combination of merging and `ignore_pn`? They seem to be unrelated features, easily tested separately.

src/pint/toa.py
```diff
@@ -103,6 +103,7 @@ def get_TOAs(
     bipm_version=None,
     include_gps=None,
     planets=None,
+    include_pn=True,
```
aarchiba, line 106: Is there any need for this? The column can easily be removed with `remove_pulse_numbers`.

tests/test_toa_writer.py
```diff
@@ -118,6 +118,23 @@ def test_roundtrip_ncyobs_toa_TEMPOformat(tmp_path):
...
+    model = get_model(datadir / "NGC6440E.par")
+    # read and add pulse numbers
+    t = toa.get_TOAs(datadir / "NGC6440E.tim", model=model)
+    t.compute_pulse_numbers(model)
+    t["pulse_number"][10] = np.nan
+    f = StringIO()
+    t.write_TOA_file(f)
+    f.seek(0)
+    contents = f.read()
+    assert "nan" not in contents
```
aarchiba, line 132: Best to convert contents to lowercase first to rule out writing NaN as well.

src/pint/toa.py
aarchiba, line 106: Is there any need for this? The column can easily be removed with `remove_pulse_numbers`.

tests/test_toa_writer.py
aarchiba, line 132: Best to convert contents to lowercase first to rule out writing NaN as well.

src/pint/toa.py
dlakaplan: ok

src/pint/toa.py
dlakaplan: ok, changed to debug

src/pint/toa.py
dlakaplan: ok

src/pint/toa.py
dlakaplan: ok

src/pint/toa.py
```diff
@@ -2420,11 +2453,66 @@ def merge_TOAs(TOAs_list):
...
+    )
+    has_tdb = np.array(
+        [("tdb" in colnames) and ("tdbld" in colnames) for colnames in all_colnames]
+    )
+    has_posvel_ecl = np.array(
+        ["ssb_obs_vel_ecl" in colnames for colnames in all_colnames]
+    )
+    if has_pulse_number.any() and not has_pulse_number.all():
+        if strict:
+            log.warning("Not all data have 'pulse_number' columns but strict=True")
```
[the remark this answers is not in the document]
dlakaplan: ok

src/pint/toa.py
dlakaplan: but the obsvel and TDB functions remove previous columns, so I thought that would take unnecessary time.

tests/test_pulse_number.py
dlakaplan: ok, added

tests/test_toa_writer.py
dlakaplan: good point

src/pint/toa.py
dlakaplan: especially for reading, I think it's needed.  this directly addresses the use case Deborah had where multiple files are being read in together so you can't remove pulse numbers individually.

tests/test_toa_reader.py
dlakaplan: I guess not.

src/pint/toa.py
scottransom: I think I'm with @aarchiba on this one.  That seems like a really specific and unneeded option.  Because even if you read the files together, you still need to merge them, and so can remove pulse numbers before you do that.  Unless I'm missing something...

dlakaplan: Merging is handled within `get_TOAs()`.  I guess with this now inserting pulse numbers by default then it's fine to remove this option, although it may be that instead the `strict` option should be used instead.

dlakaplan: Maybe a `strict` option isn't needed, but I personally prefer to be able to control if the data input will be additionally modified on read.

Issue #1303, opened 2022-06-24
PINT should hace a derivative-aware MCMC method


aarchiba:
When derivatives of the objective function are available, MCMC methods that use them, for example Hamiltonian Monte Carlo, can be very much faster than MCMC methods that aren't. PINT has gone to great lengths to make derivatives available, so we should do this. PyMC3 includes the supposedly tuning-free "NUTS" sampler, so it could probably be used for this, although there would probably be some plumbing work. This could probably work for `event_optimize` as well, where faster runtimes would be extremely welcome. I'm not sure how well it parallelizes.

aarchiba:
The NUTS sampler: https://arxiv.org/abs/1111.4246

aarchiba:
Parallel tempering to get more parallelism: http://auai.org/uai2017/proceedings/papers/289.pdf

dlakaplan:
@abhisrkckl: this may be relevant to you

abhisrkckl:
I think it is better to have an API that provides lnlikelihood, lnlikelihood_gradient, lnprior, and prior_transform than to have sampling methods baked into PINT itself. This will allow users to use the sampler of their choice rather than be tied to what PINT itself has.

I am working on an interface like this. I have not implemented lnlikelihood_gradient yet but it should be straightforward to do using the design matrix.

abhisrkckl:
Related discussion in issue #1310

PR #1407, opened 2022-09-27
Bayesian interface -- timing model + white noise

abhisrkckl:
A bayesian interface that allows the user to use PINT in combination with any sampler of their choice (both MCMC and nested samplers).

This PR implements the `BayesianTiming` class that provides the following methods:
1. Log likelihood
2. Log prior
3. Prior transform
4. Log posterior

At present, this only works with white noise models and does not support correlated noise models.

See the attachment for a sample posterior plot obtained using this interface and the NGC6440E sample data.
NGC6440E_efac_post
I have checked that the result is consistent with the ENTERPRISE result.

**Work in progress**
1. Support for correlated noise
2. The derivatives of log-likelihood, log prior, and log posterior (necessary for using derivative-aware samplers (issue #1303)).


src/pint/bayesian.py
```diff
@@ -0,0 +1,206 @@
+from copy import deepcopy
```
dlakaplan, line 1: I think if you include a top-level docstring, it will show up at https://nanograv-pint--1407.org.readthedocs.build/en/1407/_autosummary/pint.html in the module description

src/pint/bayesian.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class BayesianTiming:
+    """A wrapper around the PINT API that provides lnprior, prior_transform,
+    lnlikelihood, and lnposterior functions. This interface can be used to
+    draw posterior samples using the sampler of your choice.
+
+    Parameters
+    ----------
+    model : :class:`pint.models.TimingModel`
```
dlakaplan, line 18: I think this should be 
```
:class:`pint.models.timing_model.TimingModel`
```

dlakaplan:
I'll try to review this more closely, but with a quick look this looks really great & useful.  My main comment would be to fix some of the formatting to make:
https://nanograv-pint--1407.org.readthedocs.build/en/1407/_autosummary/pint.bayesian.BayesianTiming.html#pint.bayesian.BayesianTiming
a little easier to read.  Some of the code-block indenting and other things that are supposed to be bullet points aren't quite right.  This can be pretty annoying to debug but you can try `make docs` locally (although I don't always see the same results locally and on RTD).  

Also, your example notebook looks really good, but I assume it won't be included in the built tutorials, right?  Can you maybe then include some of the basic usage commands in the docstring since otherwise the notebook can be hard to find?  Those don't get run, so there is no penalty for a complex calculation.  Or maybe there's a better place to put that.

src/pint/bayesian.py
```diff
@@ -0,0 +1,206 @@
...
+        The best-fit values stored in this object are not used.
+    toas : a :class:`pint.toa.TOAs` instance. Contains the input toas.
```
dlakaplan, line 20: I think the right way to do this is:
```
    toas : :class:`pint.toa.TOAs`
        Contains the input toas.
```

docs/examples/bayesian-example-NGC6440E.py
```diff
@@ -0,0 +1,203 @@
...
+# I am cheating here by setting the priors around the maximum likelihood estimates.
+# This is a bad idea for real datasets and can bias the estimates. I am doing this
+# here just to make everything finish faster. In the real world, these priors should
+# be informed by, e.g. previous (independent) timing solutions, pulsar search results,
+# VLBI localization etc. Note that unbounded uniform priors don't work here.
+for par in model.free_params:
+    param = getattr(model, par)
+    param_min = float(param.value - 10 * param.uncertainty_value)
+    param_span = float(20 * param.uncertainty_value)
+    param.prior = Prior(uniform(param_min, param_span))
```
dlakaplan, line 42: So you are setting an attribute that persist in the model object?   Can you also include in the example how you set the priors with the separate dictionary?

src/pint/bayesian.py
```diff
@@ -0,0 +1,206 @@
...
+        self.params = [getattr(model, par) for par in self.param_labels]
+        self.nparams = len(self.param_labels)
+
+        if prior_info is not None:
+            for par in prior_info.keys():
+                distr = prior_info[par]["distr"]
+                if distr == "uniform":
+                    pmax, pmin = prior_info[par]["pmax"], prior_info[par]["pmin"]
+                    getattr(self.model, par).prior = Prior(uniform(pmin, pmax - pmin))
+                elif distr == "normal":
```
dlakaplan, line 71: should this raise an exception if `distr` is some other value?

src/pint/bayesian.py
```diff
@@ -0,0 +1,206 @@
...
+    def _validate_priors(self):
+        for param in self.params:
+            if not hasattr(param, "prior") or param.prior is None:
+                raise AttributeError(f"Prior is not set for parameter {param.name}.")
+            if isinstance(param.prior._rv, UniformUnboundedRV):
+                raise NotImplementedError(
+                    f"Unbounded uniform priors are not supported. (param : {param.name})"
+                )
+
+    def _decide_likelihood_method(self):
```
dlakaplan, line 90: Does this work for both narrowband and Wideband data?

src/pint/bayesian.py
```diff
@@ -0,0 +1,206 @@
...
+                    getattr(self.model, par).prior = Prior(norm(mu, sigma))
+
+        self._validate_priors()
+
+        self.likelihood_method = self._decide_likelihood_method()
+
+        self.track_mode = "use_pulse_numbers" if use_pulse_numbers else "nearest"
+
+    def _validate_priors(self):
+        for param in self.params:
```
dlakaplan, line 82: Shouldn't this be only for free params?

src/pint/bayesian.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+
+    def __init__(self, model, toas, use_pulse_numbers=False, prior_info=None):
+        # Make a deep copy to not mess up the original model.
+        self.model = deepcopy(model)
+        self.toas = toas
+
+        self.param_labels = self.model.free_params
+        self.params = [getattr(model, par) for par in self.param_labels]
+        self.nparams = len(self.param_labels)
```
dlakaplan, line 63: Again, might we want only free parameters?  Or is there a way to use a delta function prior?

dlakaplan:
Also (very minor) please include a CHANGELOG entry.

src/pint/bayesian.py
abhisrkckl: Corrected this.

src/pint/bayesian.py
abhisrkckl: Corrected this.

src/pint/bayesian.py
abhisrkckl: I have added a NotImplementedError in that case.

src/pint/bayesian.py
abhisrkckl: It works only works with narrow band now. I haven't tested with wideband data.

dlakaplan: OK, probably worth putting in the docs (& an issue?) and maybe checking when the toas are fed in?

docs/examples/bayesian-example-NGC6440E.py
abhisrkckl: > So you are setting an attribute that persist in the model object?

The model attribute in BayesianTiming is supposed to be a clone of the input model. So line 39 should have been `param = getattr(self.model, par)`. I have fixed this.

abhisrkckl: I have included the `prior_info` usage in `bayesian-example-NGC6440E.py` and a new test for this in `test_bayesian.py`.

src/pint/bayesian.py
abhisrkckl: > checking when the toas are fed in?

Yes... I think this is needed. I'll implement this. Maybe throw a NotImplementedError upon encountering wideband TOAs.

abhisrkckl: Implemented.

src/pint/bayesian.py
```diff
@@ -0,0 +1,209 @@
...
+       an instance of :class:`pint.models.priors.Prior`.
+
+    2. The parameters of BayesianTiming.model will change for every likelihood function call.
+       These parameters in general will not be the best-fit values. Hence, it is NOT a good
+       idea to save it as a par file.
+
+    3. Currently, only uniform and normal distributions are supported in prior_info. More
+       general priors should be set directly in the TimingModel object before creating the
+       BayesianTiming object. Here is an example prior_info object:
+       >>> prior_info = {
```
dlakaplan, line 44: This code block is still not formatting correctly for me at https://nanograv-pint--1407.org.readthedocs.build/en/1407/_autosummary/pint.bayesian.BayesianTiming.html.  Maybe add in a blank line and indent further?

dlakaplan:
I think the python code blocks don't work (or at least I've had issues with those in the past, especially when also trying other languages).  I would suggest just trying a blank line and extra indent.    But we can also probably defer some of the final formatting until later.

abhisrkckl:
I have just made that example into a single line and pointed the user to the bayesian example script in the docstring. If this doesn't work either let us just remove that one line and proceed.

abhisrkckl:
BTW is there a way to build the readthedocs locally?

dlakaplan:
Yes, look at https://nanograv-pint--1407.org.readthedocs.build/en/1407/contributing.html#write-documentation.  `make docs` should work. It will still take ~10 min.

Issue #1402, opened 2022-09-27
Add variable solar wind model


dlakaplan:
Update `SolarWindDispersion` model to have a different amplitude every year.

Hazboun6:
Below is a function for the solar wind dispersion that allows for a solar wind density profile of any power of 1/r^p, where r is the distance from the Sun. The function takes an impact parameter and solar angle (I called it the impact angle, because that is what made sense to me, but it is the solar elongation). This function allows you to build any of the models that are described in You, et al., 2007, which are written in Pulsar Astronomer-friendly notation in this [paper](https://iopscience.iop.org/article/10.3847/1538-4357/ac5829/pdf). You can also find a diagram that defines all of the variables in the 1/r^p function. 

https://github.com/nanograv/enterprise_extensions/blob/master/enterprise_extensions/chromatic/solar_wind.py#L279

Hazboun6:
In case it is useful... the link below links to the same page as above, but directly to a function that allows the user to either input a set of bins or a `t_init` and `t_final` along with a number of bins in order to bin the solar wind. It does seem reasonable that a single pulsar might not need a smaller bin than 1 year, but I don't think you should hard code that in. In the paper in my first comment we showed that a PTA can definitely resolve changes for bins that are 3 months long. 

https://github.com/nanograv/enterprise_extensions/blob/master/enterprise_extensions/chromatic/solar_wind.py#L21

mtlam:
Rather than generically every year, can we allow a piecewise-constant definition (SWX to suggest an equivalent though admittedly confusing notation)? Allowing the start and end times of the bins and then the electron density at 1 AU would be incredibly useful. I don't know if we would want to make the `p` parameter that @Hazboun6 suggested by time-dependent but at least the coefficient term `n_0(t) (r/r_0)^-p` be time-dependent over _any_ range would be really flexible.

Hazboun6:
@mtlam The model I linked above is effectively what you mentioned for the 1/r^2 case. It would be trivial to extend it to the 1/r^p case.

mtlam:
Oh yeah, exactly, I think we're in complete agreement. I'm just thinking about it in terms of practicality in the par file. If we were to follow the notation for DMX, then we would want at least three values: `SWX_NNNN`, `SWXR1_NNNN`, `SWXR2_NNNN`. I would also prefer it not be tied to a different amplitude every year but over arbitrary time spans (again, I think we both agree here).

dlakaplan:
[...]

PR #1410, opened 2022-09-28
More advanced solar wind models

dlakaplan:
This PR includes:
* A more generic solar wind model, based on Hazboun et al. (2022) and You et al. (2007, 2012): Spherically symmetric but with variable power-law index 
* A `SWX` model which is like `DMX` to have different segments with different solar wind properties
* A few utility functions
* Can fit for power-law index using approximate derivatives
* `SWX` model based on max DM, not `NE_SW`

I included tests that compare the new models to the existing model for the default `p=2` case, both in terms of all of the individual functions and in terms of fit results.  This led to me included a small part of a new data-set.


dlakaplan:
Addresses #1402

dlakaplan:
@Hazboun6 : an initial implementation.  I'm finding ~15% differences compared to the standard model for the p=2.  Is that expected?  I'm not sure how large the value in https://github.com/nanograv/enterprise_extensions/blob/master/enterprise_extensions/chromatic/solar_wind.py#L296 (first term) should be, or what units, so I guessed but haven't tested extensively.

dlakaplan:
Now a better implementation that just extends the existing `SolarWindDispersion` model for `SWM=1`

Hazboun6:
@dlakaplan I am officially done at UW on Friday, and my Mathematic license is dependent on being able to VPN into their network, so I decided I should just open the notebook and take the derivative. 

The derivative of this next expression with respect to p
image

is 
image

The term with the [0,1,0,0] is the derivative of the hypergeometric function WRT the second second element, so this isn't super useful, but a start. 

There is a Mathematic webpage with a few versions of this [Mathematica Derivatives](https://functions.wolfram.com/HypergeometricFunctions/Hypergeometric2F1/20/01/02/), but they would need some massaging. It seems like the 2nd and 3rd can be written as other special functions. Probably worth checking which is available in `scipy`.

dlakaplan:
I suspect the relations at https://en.wikipedia.org/wiki/Hypergeometric_function#Differentiation_formulas may make it possible - I will look later.

Unrelated: what are the units of the quantities in Enterprise?  When you have `_dm_p_int(b, 1e14, p)` presumably that first quantity is supposed to be a large number to account for the `z_p` term, but I want to make sure I get the right scale.

Hazboun6:
First answer [here](https://math.stackexchange.com/questions/2389169/derivative-of-standard-gaussian-hypergeometric-function-2f-1) is pretty clean for the derivative as well.

dlakaplan:
I think both of those are actually just derivatives wrt the last parameter.  The person is asking for derivative wrt a, but doesn't get it.  I'll have to look at the special functions books, or just leave out the derivative.

Hazboun6:
Everything is in seconds. [This](https://github.com/nanograv/enterprise_extensions/blob/master/enterprise_extensions/chromatic/solar_wind.py#L303) is the function that gives all of those quantities, and they are calculated directly from the planetSSB vectors. [Here](https://github.com/nanograv/enterprise/blob/master/enterprise/pulsar.py#L396) is where the conversion happens for PintPulsar objects.

Hazboun6:
> I think both of those are actually just derivatives wrt the last parameter. The person is asking for derivative wrt a, but doesn't get it. I'll have to look at the special functions books, or just leave out the derivative.

Oh, right!

Hazboun6:
> I'm not sure how large the value in https://github.com/nanograv/enterprise_extensions/blob/master/enterprise_extensions/chromatic/solar_wind.py#L296 (first term) should be, or what units, so I guessed but haven't tested extensively.

The `b` is in seconds. So something like 200 sec would be a reasonable test. 

When I searched for the n^(p) in the Bayesian analysis I had to use a log-uniform prior. That number varies wildly, especially if you were varying p as well.

dlakaplan:
OK, I fixed the upper limit and the agreement with the old model is a lot better now.

If we can't get an analytic derivative for p, then a MCMC or similar would still work (as you did).

dlakaplan:
A more complex usage with real data (J2145 from CHIME, analyzed by @LuluAgazie):
2145_swx
fitting for amplitude and offset for each year

dlakaplan:
Basic fitting works.  Here's a fit to 1 year of 2145 data with the DMX removed.  So I was fitting for DM and NE_SW.
Screen Shot 2022-09-29 at 4 57 01 PM

dlakaplan:
And I was able to run the same fit and get the same result with the SWX model and a single component.

dlakaplan:
Fitting two different SWX components simultaneously.
Screen Shot 2022-09-29 at 7 23 47 PM

dlakaplan:
Doing a grid search over the power-law index:
Screen Shot 2022-09-29 at 8 20 18 PM

dlakaplan:
Aside from the oldestdeps, this has the majority of what is needed.

dlakaplan:
I have a version mostly working to allow fitting for the power-law index.  I used the version of the Eqn. in Hazboun with z_p -> infinity.  So the term involving gamma functions was easy (derivative = polygama).  The other term I evaluated using a Pade expansion.   It's not perfect, but comparing the derivative I compute against a finite difference value comes pretty close for most elongations.  It deviates most at elongation=180, but that's OK.  I'm still trying to understand what's happening at 45 deg.

Screen Shot 2022-10-04 at 9 06 58 AM

dlakaplan:
With this I can fit for p.  I don't know yet how the errors in the derivative will affect things like convergence or estimation of uncertainties, but a fit to simulated data runs and give the right value.  @mtlam, @Hazboun6 and others: does this seem worth it?

dlakaplan:
OK, the issue near 40 deg is not actually 45.  it's from a singularity in the denominator of my expansion.  I can probably move that around by changing the expansion order but it's probably inevitable to have some such values.  Don't know if they are a big issue or not.  The actual value is dependent on p (for p=2 it's pi/2 - sqrt(21)/2) so it may be possible to have two series in the expansion and go between them if needed.

abhisrkckl:
The singularity may go away if you reduce the degree of the denominator.

dlakaplan:
yes, I was looking at that

Hazboun6:
@dlakaplan I think it is worth it. I'm not sure what the individual pulsars will see, but I know that even the 11-year NANOGrav dataset sees these higher order terms. The table from the paper is not super clear, but it shows that depending on the model chosen the extra terms can be pretty significant (look at the error bars). The terms with 95% in the exponent are not significant, just upper limits. 

In the advanced noise project I have been using Model 5 w/o the third term to mitigate the effect of a few pulsars' observations very close to the sun with fairly good results. 

To me the reason this is most important though, is that if we are going to start using DM GPs to model the DM variations, they aren't great at grabbing all of the solar wind variations. They do okay, but it's not really what they are meant to model. A pulsar with a few observations close to the sun can completely change the parameters of the DM GP if the higher order terms aren't included. 

image

dlakaplan:
I tested and made a change, so you can now have overlapping SWX regions.  I would be wary of fitting them simultaneously without many constraints, though, as those fits are very degenerate.  And the fits for the power-law index can fail if they go to p->1 (since we don't have a constrained fitter).

dlakaplan:
@Hazboun6 : do you have the data you used to make the table above?  I can make sure I get the same results

Hazboun6:
@dlakaplan It's the entire 11-year dataset. The fit was across all of the pulsars at once using `enterprise`. I'm not sure how you would do that in `Pint`.

dlakaplan:
OK, fitting the entire data-set within PINT is hard (I agree).  But here's how your model 5 would look within PINT if people then wanted to implement it to fit for other parameters:
```
 SolarWindDispersionX(
    floatParameter( SWX_0001            6.8               (1 / cm3) frozen=False),
    floatParameter( SWXP_0001           2.0               () frozen=False),
    MJDParameter(   SWXR1_0001          53005.0000000000000000 (d) frozen=True),
    MJDParameter(   SWXR2_0001          57387.0000000000000000 (d) frozen=True),
    floatParameter( SWX_0002            0.0013489628825916532 (1 / cm3) frozen=True),
    floatParameter( SWXP_0002           4.39              () frozen=True),
    MJDParameter(   SWXR1_0002          53005.0000000000000000 (d) frozen=True),
    MJDParameter(   SWXR2_0002          57387.0000000000000000 (d) frozen=True),
    floatParameter( SWX_0003            1.584893192461111e-25 (1 / cm3) frozen=True),
    floatParameter( SWXP_0003           16.25             () frozen=True),
    MJDParameter(   SWXR1_0003          53005.0000000000000000 (d) frozen=True),
    MJDParameter(   SWXR2_0003          57387.0000000000000000 (d) frozen=True)),
```
does that look reasonable?  

I also find that the 3 components contribute a max of:
```
<Quantity [0.00589969, 0.00872654, 0.00032351] pc / cm3>
```
to the DM (at conjunction for a source at latitude of 1 degree).  Do those values look plausible?

dlakaplan:
I suggest that, perhaps for the `SWX` model, we change the parameterization from `NE` to `DMMAX`.  A basic version of that is in the PR but should be reconciled with the existing models (or replace one) to simplify the naming.  But it makes the covariance much better for fitting (red vs. black):
dmmax

dlakaplan:
I've started getting some CI errors in the oldestdeps tests.  It's not actually anything that I wrote (as far as I can tell) but some basic `astropy` separation calculations.  I don't see this locally, even when I try to mimic `oldestdeps` (although I may not be doing that correctly.  Any ideas?  @paulray : I don't think this has anything to do with your other fixes, but maybe you can help diagnose?

```
_______________________________ test_conjunction _______________________________
[255](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:256)

[256](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:257)
    def test_conjunction():
[257](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:258)
        model = get_model(StringIO(par))
[258](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:259)
>       t0, elongation = pint.utils.get_conjunction(
[259](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:260)
            model.get_psr_coords(),
[260](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:261)
            Time(march_equinox, format="mjd") - 2 * u.d,
[267](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:268)
            precision="high",
[268](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:269)
        )
[269](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:270)

[270](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:271)
/home/runner/work/PINT/PINT/tests/test_solar_wind.py:67: 
[271](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:272)
[272](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:273)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/pint/utils.py:1860: in get_conjunction
[273](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:274)
    elongation = coords.get_sun(tt).separation(coord)
[274](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:275)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/astropy/coordinates/sky_coordinate.py:846: in separation
[275](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:276)
    lon2 = other.spherical.lon
[276](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:277)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/astropy/coordinates/sky_coordinate.py:596: in __getattr__
[277](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:278)
    if not attr.startswith('_') and hasattr(self._sky_coord_frame, attr):
[278](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:279)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/astropy/coordinates/baseframe.py:1701: in spherical
[279](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:280)
    return self.represent_as('spherical', in_frame_units=True)
[280](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:281)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/astropy/coordinates/baseframe.py:1128: in represent_as
[281](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:282)
    diffkwargs[comp] = diffkwargs[comp].to(new_attr_unit)
[282](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:283)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/astropy/units/quantity.py:693: in to
[283](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:284)
    return self._new_view(self._to_value(unit, equivalencies), unit)
[284](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:285)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/astropy/units/quantity.py:664: in _to_value
[285](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:286)
    return self.unit.to(unit, self.view(np.ndarray),
[286](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:287)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/astropy/units/core.py:987: in to
[287](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:288)
    return self._get_converter(other, equivalencies=equivalencies)(value)
[288](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:289)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/astropy/units/core.py:918: in _get_converter
[289](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:290)
    raise exc
[290](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:291)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/astropy/units/core.py:903: in _get_converter
[291](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:292)
    return self._apply_equivalencies(
[292](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:293)
[293](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:294)

[294](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:295)
self = Unit("1 / s"), unit = Unit("1 / s"), other = Unit("km / s")
[295](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:296)
equivalencies = []
[296](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:297)

[297](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:298)
    def _apply_equivalencies(self, unit, other, equivalencies):
[298](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:299)
        """
[299](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:300)
        Internal function (used from `_get_converter`) to apply
[300](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:301)
        equivalence pairs.
[301](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:302)
        """
[302](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:303)
        def make_converter(scale1, func, scale2):
[303](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:304)
            def convert(v):
[304](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:305)
                return func(_condition_arg(v) / scale1) * scale2
[305](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:306)
            return convert
[306](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:307)
    
[307](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:308)
        for funit, tunit, a, b in equivalencies:
[308](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:309)
            if tunit is None:
[309](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:310)
                try:
[310](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:311)
                    ratio_in_funit = (other.decompose() /
[311](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:312)
                                      unit.decompose()).decompose([funit])
[312](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:313)
                    return make_converter(ratio_in_funit.scale, a, 1.)
[313](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:314)
                except UnitsError:
[314](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:315)
                    pass
[315](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:316)
            else:
[316](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:317)
                try:
[317](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:318)
                    scale1 = funit._to(unit)
[318](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:319)
                    scale2 = tunit._to(other)
[319](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:320)
                    return make_converter(scale1, a, scale2)
[320](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:321)
                except UnitsError:
[321](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:322)
                    pass
[322](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:323)
                try:
[323](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:324)
                    scale1 = tunit._to(unit)
[324](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:325)
                    scale2 = funit._to(other)
[325](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:326)
                    return make_converter(scale1, b, scale2)
[326](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:327)
                except UnitsError:
[327](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:328)
                    pass
[328](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:329)
    
[329](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:330)
        def get_err_str(unit):
[330](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:331)
            unit_str = unit.to_string('unscaled')
[331](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:332)
            physical_type = unit.physical_type
[332](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:333)
            if physical_type != 'unknown':
[333](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:334)
                unit_str = "'{}' ({})".format(
[334](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:335)
                    unit_str, physical_type)
[335](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:336)
            else:
[336](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:337)
                unit_str = f"'{unit_str}'"
[337](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:338)
            return unit_str
[338](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:339)
    
[339](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:340)
        unit_str = get_err_str(unit)
[340](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:341)
        other_str = get_err_str(other)
[341](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:342)
    
[342](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:343)
>       raise UnitConversionError(
[343](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:344)
            "{} and {} are not convertible".format(
[344](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:345)
                unit_str, other_str))
[345](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:346)
E       astropy.units.core.UnitConversionError: '1 / s' (frequency) and 'km / s' (speed) are not convertible
[346](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:347)

[347](https://github.com/nanograv/PINT/actions/runs/3275871891/jobs/5391250242#step:8:348)
/home/runner/work/PINT/PINT/.tox/oldestdeps/lib/python3.8/site-packages/astropy/units/core.py:886: UnitConversionError
```

paulray:
Yeah, seems to be in `get_conjunction()` but I'm not sure why. Does it work with `tox -e oldestdeps` on your system?

dlakaplan:
The command that fails is:
```
astropy.coordinates.get_sun(tt).separation(coord)
```
where `tt` is an array of times and `coord` is some pulsar coordinate.  That seems pretty uncontroversial.

I've tried `tox -e oldestdeps` locally in a conda with python 3.8, and I thought it was working, but it looks like that was user error.  now I'm seeing failures so it will be easier to fix.  thanks!

paulray:
could `separation` be returning linear distance instead of angle in some cases?

dlakaplan:
Something weird was happening in the transformation to GCRS coordinates (which are the default frame of `get_sun`).  Seemed to be astropy version dependent. You got different results for `pulsar.separation(sun)` and `sun.separation(pulsar)`.  I might have a workaround that does GCRS->ICRS first

paulray:
I've had problems with `get_sun()` in the past. It can be very confusing since it returns coordinates in GCRS instead of ICRS. This produces some very non-intuitive behaviors.

dlakaplan:
I was very impressed at the speed of your response.  I guess it wasn't a response!

dlakaplan:
@blarsen10: this may be of interest

dlakaplan:
Screen Shot 2022-10-25 at 10 31 40 AM

dlakaplan:
@scottransom : was your suggestion to enforce continuity that I stop each SW section a little shy of opposition and add a linear (polynomial?) segment to join them?   I guess I could put in a `DM/DM1` model together with a mask parameter to make it only valid over a limited range, but that would effectively be like a DMX model.  Or am I misunderstanding?

scottransom:
I was thinking about adding a linear component from the solar max val to the min val on each side that "bends" the curve so that it meets at a common middle point.  But that was just one idea.

dlakaplan:
My worry with that is that it would change depending on what the neighboring segments are.  

What about just having the SW component go to 0 at opposition?  I realize that breaks from the `NE_SW` model, but if you don't have an explicit SW model then the average SW DM is just incorporated into the normal DM term, so it's not that different.  If the SW is just the delta DM then continuity is easy

scottransom:
Yeah, I think that's likely fine.  And 1) almost certainly easier to implement and 2) keeps the years independent.

dlakaplan:
I think this largely works with the model going to 0 at opposition.  This required a bit of fiddling to test against the previous model.  Basically:
* I wanted the new model to be "obvious" so that the maximum DM on a DM residual plot would correspond to the `SWXDM_?` value.
* So this means that the SWX model is shifted relative to the previous model.  
* But then the amplitude would need to be scaled for comparing DMs directly.  i.e., to go from the old style model to the new:
```
swxmodel.SWXDM_0001.quantity = model.get_max_dm() * swxmodel.get_swscalings()[0]
np.allclose(swxmodel.get_min_dms(),model.get_min_dm()*swxmodel.get_swscalings())
toas = pint.simulation.make_fake_toas_uniform(54000, 54000 + 2 * 365, 500, model=model, obs="gbt")
np.allclose((swxmodel.swx_dm(toas) + model.get_min_dm()), model.solar_wind_dm(toas))
```
* I tried to document this and it's also in the tests.

dlakaplan:
This plot hopefully illustrates the scaling/offset issues needed to match the new SWX model against the previous model.  I'll see if I can somehow include it in the docs.
swx_scaling

dlakaplan:
I'm not sure how useful this is for others, but if the plot above is useful, I'm also not sure of the best way to include it.  I could include a static image in the RTD build but I don't think we have any yet so I'd have to set that up.  Or I could make it a new example notebook (it runs pretty quickly).  Any thoughts?

dlakaplan:
I added a new notebook.  It is pretty straightforward, except that one it also does some chi^2 gridding so it takes about ~1 min.  If this is too slow then I can remove those pieces.

dlakaplan:
New notebook: https://nanograv-pint--1410.org.readthedocs.build/en/1410/examples/solar_wind.html

src/pint/models/solar_wind_dispersion.py
```diff
@@ -116,8 +376,775 @@ def d_delay_d_ne_sw(self, toas, param_name, acc_delay=None):
...
+
+    def get_theta0(self):
+        """Get elongation at conjunction
+
+        Returns
+        -------
+        astropy.quantity.Quantity
+        """
+        coord = self._parent.get_psr_coords()
+        if self._parent.POSEPOCH.value is not None:
```
scottransom, line 859: I notice that this code is replicated several times.  Is it worth adding a helper function to remove that replication?  And is the same code used in other modules?  If so, maybe we could have that in a util?

dlakaplan: Good idea.  I implemented a helper function in this module, but it's possible that it would be useful elsewhere.

scottransom:
Once the CI tests finish, I'm happy to merge.  Things look pretty good to me on my quick look through the code.  Are you happy with it, @dlakaplan?

dlakaplan:
Yes, I am happy.  I think it works as expected and  it has sufficient docs to get people using it

PR #1411, opened 2022-10-04
Fixed failing tests on oldestdeps

### LINKED ISSUE #1401: oldestdeps is using new numpy
In spite of the details in `tox.ini`, on the CI machines at least the `oldestdeps` is using the most recent numpy. This has presumably been going on for a while but it was brought to light by the removal of the deprecated `numpy.isscalar`.

paulray:
Attempt to fix #1401 

Required updating the oldest version of numpy and pinning the versions of scipy and matplotlib to keep it from automatically updating numpy to the most recent.

### DIFF
--- a/src/pint/scripts/photonphase.py
+++ b/src/pint/scripts/photonphase.py
@@ -227,7 +227,7 @@ def main(argv=None):
 
         # Calculate phases
         log.debug("Evaluating polycos")
-        phases = p.eval_phase(mjds)
+        phases = ptable.eval_phase(mjds)
         phases[phases < 0] += 1.0
         h = float(hm(phases))
         print("Htest : {0:.2f} ({1:.2f} sigma)".format(h, h2sig(h)))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 
 ## Unreleased
 ### Changed
+- Minimum supported versions updated to numpy 1.18.5, matplotlib 3.2.0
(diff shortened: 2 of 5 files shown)


paulray:
The reason the bug was not causing a problem before is that that test is skipped on headless machines (e.g. on the CI machines). Can we re-enable that test on CI?

paulray:
I have now removed all the skips if X11 is not available in `test_photonphase.py`. It seems to be passing fine, so I'm not sure why those skips were there. They allowed a bug to persist in the code.

src/pint/scripts/photonphase.py
```diff
@@ -227,7 +227,7 @@ def main(argv=None):
 
         # Calculate phases
         log.debug("Evaluating polycos")
-        phases = p.eval_phase(mjds)
+        phases = ptable.eval_phase(mjds)
```
paulray, line 230: @sashalvna can you look at this fix? Without this edit, it was crashing for me and the test was not passing. Do you agree it is correct now?

sashalvna: Yes, that is correct! ptable contains the polycos that it should be using to calculate the phases.

paulray: Thanks much! I can't figure out how it was working before, but it works now!

paulray:
Bizarre, the macos-latest tests ran super slow and got killed after 6 hours. What is going on?

paulray:
OK, this is ready to merge! It solves the CI failures we are getting and increases the test coverage in our CI runs, which revealed a bug.

PR #1415, opened 2022-10-07
Added try/except to get_summary() when PB not present

swiggumj:
Running `fitter.print_summary()` in cases where the timing model uses FBn rather than PB and the binary model is ELL1 produces an error because the ell1 check assumes PB is there. I've added a try/except to look for FB0 if PB fails. There are some other changes popping up because of blackening for some reason, which I did because tests weren't originally passing.

The main snippet that's changed is line 566.

### DIFF
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -492,7 +492,7 @@ class Fitter:
                     s += "Magnetic field at light cylinder = {:.4g}\n".format(
                         pint.derived_quantities.pulsar_B_lightcyl(F0, F1)
                     )
-                    I_NS = I = 1.0e45 * u.g * u.cm ** 2
+                    I_NS = I = 1.0e45 * u.g * u.cm**2
                     s += "Spindown Edot = {:.4g} (I={})\n".format(
                         pint.derived_quantities.pulsar_edot(F0, F1, I=I_NS), I_NS
                     )
@@ -553,10 +553,12 @@ class Fitter:
             if binary.startswith("BinaryELL1"):
                 ell1 = True
                 eps1 = ufloat(
-                    self.model.EPS1.quantity.value, self.model.EPS1.uncertainty.value,
+                    self.model.EPS1.quantity.value,
+                    self.model.EPS1.uncertainty.value,
                 )
                 eps2 = ufloat(
-                    self.model.EPS2.quantity.value, self.model.EPS2.uncertainty.value,
+                    self.model.EPS2.quantity.value,
+                    self.model.EPS2.uncertainty.value,
                 )
                 tasc = ufloat(
                     # This is a time in MJD
@@ -574,7 +576,7 @@ class Fitter:
                     )
                     pb = ufloat(p.to(u.d).value, perr.to(u.d).value)
                 s += "Conversion from ELL1 parameters:\n"
-                ecc = um.sqrt(eps1 ** 2 + eps2 ** 2)
+                ecc = um.sqrt(eps1**2 + eps2**2)
                 s += "ECC = {:P}\n".format(ecc)
                 om = um.atan2(eps1, eps2) * 180.0 / np.pi
                 if om < 0.0:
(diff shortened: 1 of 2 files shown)


swiggumj:
Apparently my version of black was out of date, so I've fixed that and the diffs look much better now. I'm still getting two test failures:
```
FAILED tests/test_fitter.py::test_fitter - ValueError: 'yerr' must not contain negative values
FAILED tests/test_pintk.py::TestPintk::test_result
```
...but they aren't very descriptive locally, so I'll see if they're still failing here and check the logs if so.

paulray:
I get the 'yerr' error locally on my machine as well. Not sure why.

paulray:
Could you stick in a quick test to show that it works with both types of par files?

swiggumj:
I'm very inexperienced when it comes to writing tests. Can you suggest (in words) what this might look like? It does look like there's already a par file I can use for the FB0 case (`tests/datafile/J0023+0923_ell1_simple.par`), so that's good.

paulray:
There is a `test_fbx.py` that you could just add one function to that tests that you can call `print_summary()` without getting an exception. Should be just a couple of lines.

tests/test_fbx.py
```diff
@@ -79,3 +81,19 @@ def test_derivative(modelJ0023, toasJ0023):
             assert np.nanmax(relative_diff) < tol, msg
         else:
             continue
+
+
+def test_summary_FB():
+    m = mb.get_model(os.path.join(datadir, parJ0023ell1))
```
aarchiba, line 87: Wouldn't this code be very much simpler if you just used `get_model_and_toas`? Like couldn't you drop all that stuff about posvels and planet ephem?

swiggumj: Yes. :)

src/pint/fitter.py
```diff
@@ -565,10 +565,16 @@ def get_derived_params(self):
                     self.model.TASC.quantity.mjd,
                     self.model.TASC.uncertainty.to(u.d).value,
                 )
-                pb = ufloat(
-                    self.model.PB.quantity.to(u.d).value,
-                    self.model.PB.uncertainty.to(u.d).value,
-                )
+                try:
```
dlakaplan, line 568: Rather than a try with a bare except, a better practice would be something like:
```
if hasattr(self.model, "PB"):
    pb = ...
elif hasattr(self.model, "FB0"):
    p, per = ...
    pb = ...

Comment (swiggumj) [reply src/pint/fitter.py:568]: Just made this change and encountered something weird. The test par file I'm using uses an ELL1 binary model with FBn params (PB is not present; same as the par file I was working with where I originally encountered this issue). The test I've written fails with the suggested code because it apparently thinks the model has PB?!

Comment (dlakaplan) [reply src/pint/fitter.py:568]: OK, it probably also needs `and self.model.PB.value is not None` and the same for `FB0`

Comment (swiggumj) [reply src/pint/fitter.py:568]: Yeah, that works (as expected). I'll push the updated version, but I'm still a bit confused why `hasattr(self.model, "PB")` evaluates to True here. (Doesn't really matter for this PR.)

Comment (dlakaplan) [reply src/pint/fitter.py:568]: depending on how the model is initialized, it might have properties that are there but unset.

File: tests/test_fbx.py [hunk 3]
```diff
@@ -79,3 +82,13 @@ def test_derivative(modelJ0023, toasJ0023):
...
+
+
+def test_summary_FB():
+    m, t = get_model_and_toas(
+        os.path.join(datadir, parJ0023ell1), os.path.join(datadir, timJ0023)
+    )
+    f = fitter.WLSFitter(toas=t, model=m)
+
+    # Ensure print_summary runs without an exception for an ELL1 model with FBX
+    f.print_summary()
```
aarchiba, line 94: Can you extract the summary to a string, and look in the string for appropriate text? Like, should it report a PB so you could look for PB in the string?

src/pint/fitter.py
aarchiba: `hasattr` is fine, but there's nothing wrong with catching exceptions - as long as you catch only the specific exception you are looking for. Just run the code on your test case, see what exception is emitted, and catch that, rather than a bare `except`.

tests/test_fbx.py
swiggumj: I'm not sure I completely understand, but my original issue was just that `print_summary()` threw an exception when using an ELL1 binary model that contains FBX params. That is now fixed.

swiggumj: (PB is reported in the summary.)

PR #1417, opened 2022-10-10
Clean up unnecessary imports

abhisrkckl:
There are no changes to the actual code. I have only removed unused imports.


dlakaplan:
Are you doing this manually or with some module?

abhisrkckl:
I am opening the whole PINT directory in VS code. It highlights unused imports, variables, etc in a slightly different color. 

I am not running any automated script because some of the imports that are unused in any given source file may be part of the interface of that module.

PR #1418, opened 2022-10-11
Fix test_fitter and test_pintk

paulray:


dlakaplan:
Yes, I think that exception is present in the latest matplotlib but not earlier (I can't seem to find when that was introduced, but it explains why I don't get that error locally).

paulray:
Interesting, so it was always buggy but matplotlib just started complaining. At least it is all gone now. Still, test_fitter could use additional improvements to be a better test. But at least this fixes the crashing bugs.

paulray:
Ok, this is ready to merge

dlakaplan:
Any CHANGELOG entry needed?  If not I can merge.

paulray:
I don't think a CHANGELOG is needed. The changes are only in tests, nothing user-facing.

PR #1424, opened 2022-10-19
`introduces_correlated_errors` as a class attribute of `NoiseComponent`s

### LINKED ISSUE #1421: `introduces_correlated_errors` should be a class attribute of `NoiseComponent`s
This will allow us to query these classes without instantiating them. See 0954f61eddf8d221e93f4424f4f80c6f731d5be5 in #1396 for a use case. This change will not break anything else as class attributes can be accessed just like instance attributes.

abhisrkckl:
Fixes #1421


abhisrkckl:

PR #1426, opened 2022-10-20
Bayesian interface for Wideband data (No correlated noise)

abhisrkckl:
I have implemented the Bayesian interface for analyzing wideband data. Correlated noise is still TODO..

I have created a new example script `bayesian-wideband-example.py` for this feature.

Here is a posterior corner plot from running this script.


dlakaplan:
The corner plot looks good.  Can you add the best-fit values from standard fitting to the plot (as truths)?  Or at least confirm that it agrees?

abhisrkckl:
Do you mean the pre-fit values present in the par file or post-fit values obtained using something like `WidebandTOAFitter`?

src/pint/bayesian.py
```diff
@@ -104,13 +104,17 @@ def _decide_likelihood_method(self):
...
+                method = "wls"
             else:
                 raise NotImplementedError(
                     "GLS likelihood for correlated noise is not yet implemented."
                 )
                 # return "gls"
 
+        toa_type = "wb" if self.is_wideband else "nb"
+
+        return f"{method}-{toa_type}"
```
dlakaplan, line 116: Is having this method return a string which you then rely on the best practice?  Maybe instead this function can set some a property like `self.lnlikelihood` to the correct value (`self._wls_nb_lnlikelihood`, `self._wls_wb_lnlikelihood`)?  You could also add a `log.debug()` statement with the choice it's made.  This way you don't have to decide in one method and then decode that in another.

dlakaplan:
post-fit

dlakaplan:
I think the docstring formatting is slightly different than what we've adopted elsewhere.  For instance, it should be:
```
"""Basic implementation of a factorized log prior.

         More complex priors must be separately implemented.
        Parameters
        ------------
            params : array-like 
                Parameters

        Returns:
        --------
            float: 
                Value of the log-prior at params
"""
```
so the data-type and description should go on different lines.  This will be easiest to see when the RTD build finishes, or if you build locally, but it's not doing all of the normal parsing as-is.

abhisrkckl:
Here is the plot with truths. (The EFAC1 and DMEFAC1 are taken from the par file.)
wb-corner

I am not sure why the EFAC1 and DMEFAC1 are not agreeing very well. I'll run an ENTERPRISE analysis and compare with this.

src/pint/bayesian.py
abhisrkckl: I agree. I'll change this.

abhisrkckl: done.

dlakaplan:
The RTD build failed, but that may have just been a timeout.  Otherwise this looks good but I agree the fit discrepancy should be sorted out.

abhisrkckl:
I have merged the master into this and did a few more simulations to make sure it is giving correct answers. This works as far as I can see.

See a simulation result below:
Figure_1

Timing model parameters (especially DM and DM1) have been recovered correctly. 
It looks like EFAC1 and DMEFAC1 have been overestimated, but this is an issue with the simulation and not with the analysis. I calculated the EFAC and DMEFAC by hand from the TOA and DM residuals, and they agree with the results above.

dlakaplan:
New notebooks look good.

abhisrkckl:
> New notebooks look good.

This is ready to go, then.

PR #1429, opened 2022-10-31
Simulate wideband TOAs using `zima`

abhisrkckl:
I have added wideband TOA simulation capability to `zima`. The underlying functionality is of course implemented in `pint.simulation`.

This is invoked using two new command line options `--wideband` and `--dmerror`).
Here is an example command:
`$ zima --startMJD=56600 --addnoise --wideband --dmerror=1e-4 J0100+8023.par J0100+8023.sim`

This feature computes the expected DM time series using the model (from the input par file), adds white noise based on the `--dmerror` option, and sets appropriate `-pp_dm` and `-pp_dme` flags. 

The injected error takes care of the DMEFACs and DMEQUADs as follows:  `-pp_dme` are set equal to the `--dmerror`, and the injected noise is computed with the help of `model.scaled_dm_uncertainty(toas)`.


abhisrkckl:
My use case for this is that I need to simulate WB TOAs to test PR #1426.

dlakaplan:
I haven't used `zima` extensively.  Have you checked this using the straight python interface in `pint.simulation`?  When I tested those for NB data they did respect `EFAC`/`EQUAD` in the model, but I don't know if I did WB data.

dlakaplan:
So can you confirm that the underlying tasks in `pint.simulation()` also work OK?

abhisrkckl:
I have added the test in test_fake_toas.

dlakaplan:
This looks good.  Do you want me to merge or wait until the meeting to discuss?

abhisrkckl:
We had discussed this a few times in the past meetings. I think this is OK to merge.

dlakaplan:
OK, I think you might just need a CHANGELOG entry

PR #1433, opened 2022-11-08
Add clock file for effix

### LINKED ISSUE #1432: PINT is missing the Effelsburg Asterix clock correction
When loading TOAs specifically for Effelsburg Asterix (a backend), PINT claims 
"2022-11-08 11:26:30.012 | INFO     | pint.observatory.topo_obs:clock_corrections:353 - Observatory effelsberg_asterix requires no clock corrections."

This isn't actually true. Clock correction info is available here: https://ipta.github.io/pulsar-clock-corrections/T2runtime/clock/effix2gps.clk.html.

When I plot residuals from ASTERIX data right now, I get something like this: 
image

But that's suspiciously similar to the clock correction file 
image

So I'm pretty sure this is what's messing me up! I'm supplying the par & tim I used in making that figure to @aarchiba and @dlakaplan via email, as it's not technically my data to splash around publicly.

aarchiba:
Needs testing.

Closes #1432

### DIFF
--- a/src/pint/data/runtime/observatories.json
+++ b/src/pint/data/runtime/observatories.json
@@ -556,6 +556,8 @@
             4900430.8
         ],
         "origin": "Imported from TEMPO2 observatories.dat 2021 June 7."
+        "clock_file": "effix2gps.clk",
+        "clock_fmt": "tempo2",
     },
     "leap": {
         "aliases": [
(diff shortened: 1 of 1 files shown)


aarchiba:
I've done a quick test and this seems to work.

PR #1443, opened 2022-11-11
Removed old/wip functions from gridutils

dlakaplan:
There were some old functions in there for reference, and a couple of attempts at methods to do faster gridding by grouping individual calls together (when the cost per fit is small).  The latter may still be worthwhile but having the code in there was confusing so I have removed it.


abhisrkckl:
> There were some old functions in there for reference
Are the features implemented in these functions available elsewhere?

dlakaplan:
The old ones are done in the current `gridutils` functions, and documented in a few examples on the wiki.

The WIP ones are not yet implemented but it's probably best to remove them until they work.

src/pint/gridutils.py
```diff
@@ -900,264 +900,3 @@ def tuple_chisq_derived(
...
-
-    # Restore saved model
-    ftr.model = savemod
-    return chi2
-
-
-##############################
-
-
-def plot_grid_chisq(
```
abhisrkckl, line 1104: This function is mentioned in `__all__` of this module. Shouldn't that be removed?

paulray: Is there a different grid plotting function that replaces this one?

dlakaplan: Is https://nanograv-pint--1443.org.readthedocs.build/en/1443/_autosummary/pint.gridutils.grid_chisq.html#pint.gridutils.grid_chisq a useful example?

abhisrkckl: I tried running the examples from the docs above. They work for me. The 2D example is missing a `plt.show()` at the end, although that's not a big deal.

I think the whole thing is slowed down significantly by logging. It would be nice to temporarily turn off logging for applications like this where some process is run in a loop.

abhisrkckl:
Is this ready to merge?

dlakaplan:
Yes, as long as the docs are OK.

Issue #1445, opened 2022-11-12
Redundant command line arguments in compare_parfiles.py


abhisrkckl:
There are two pairs of redundant parameters:
`--dmx`  and  `--no-dmx`
`--convertcoordinates` and `--no-convertcoordinates`

Only one of each pair is required to toggle these options.

PR #1444, opened 2022-11-11
Improve test coverage and minor bug fixes

abhisrkckl:
See #457

PR #1446, opened 2022-11-12
remove redundant cmd line args in compare_parfiles.py

abhisrkckl:
see #1445

### DIFF
--- a/src/pint/scripts/compare_parfiles.py
+++ b/src/pint/scripts/compare_parfiles.py
@@ -20,10 +20,11 @@ def main(argv=None):
         "--dmx", default=False, action="store_true", help="Print DMX parameters"
     )
     parser.add_argument(
-        "--convertcoordinates",
-        default=True,
-        action="store_true",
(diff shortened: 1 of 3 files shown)


dlakaplan:
I'm less sure that this is needed.  Without knowing what the default behavior is, it can be useful to be able to explicitly specify what you want to happen.

abhisrkckl:
If neither `--convertcoordinates` nor `--no-convertcoordinates` is provided, the behavior of the script is equivalent to `--convertcoordinates`.

abhisrkckl:
So aren't `--convertcoordinates` and `--no-dmx` redundant?

dlakaplan:
My point is that a user may not want to remember which behavior is default, so being able to specify explicitly on the command line can be useful if it doesn't otherwise hurt things.

abhisrkckl:
Ah... that makes sense. I'll close this issue and add the new tests to PR #1444.

PR #1453, opened 2022-11-18
Allow overlapping DMX ranges

### LINKED ISSUE #1452: Overlapping DMX ranges fails
If you add in any overlapping DMX ranges, the value is just from the last DMX that affects a given TOA rather than the sum.  My intuition is that we should allow overlapping ranges since we don't check to exclude that in the validation, and we don't offer tools to find non-overlapping values (such as adjusting limits).  It might not always behave great in fitting (if you have two ranges that are highly covariant) but shouldn't we allow it?

Quick example:
```
par = """
PSR J1234+5678
F0 1
DM 10
ELAT 10
ELONG 0
PEPOCH 54000
DMXR1_0001 54000
DMXR2_0001 55000
DMX_0001 1
DMXR1_0002 55000
DMXR2_0002 56000
DMX_0002 2
"""

model = get_model(StringIO(par))
toas = make_fake_toas_uniform(54000, 56000, 100, model=model, obs="gbt")
dmx = model.components["DispersionDMX"].dmx_dm(toas)

# add in new overlapping DMX
model.add_DMX_range(54500, 55000, dmx=0, frozen=True)
newdmx = model.components["DispersionDMX"].dmx_dm(toas)
```
So in the place I put the new range (54500 - 55000) changes, even though the new range has a value of 0.
dmx_ranges

dlakaplan:
Fixes #1452 

I _think_ that this is sufficient in terms of just summing the DMs from the different components (and that no change in the derivative is needed since that is done just one parameter at a time) but if that is wrong, let me know.


dlakaplan:
I also added `add_DMX_ranges()` since adding them one-by-one for many values is slow (it runs `validate`/`setup` after each).  I added some tests for this.  I think it's a lot faster.

dlakaplan:
As a test I added 400 DMX values to a simple model.  Doing them one-by-one took 42s.  Doing them as a single command took 0.5s.

dlakaplan:
You might want an annual DMX and then shorter-term DMXs to deal with other changes.  This isn't required, but it would then require some tools to help insert new ranges while not overlapping.  In my mind this is more complicated.

I found that I needed this because I had an annual DMX to go with annual SWX segments (since SWX is just Delta DM) and then shorter DMX to handle the residuals.  There are some degenerate parameters but it still works OK.

dlakaplan:
I added in checks for overlapping ranges, but I put it as `DEBUG` rather than `WARN` as I had a lot of them and wanted to be able to filter them out.

dlakaplan:
```
par = """
PSR J1234+5678
F0 1
DM 10
ELAT 10
ELONG 0
PEPOCH 54000
DMXR1_0001 54000
DMXR2_0001 55000
DMX_0001 1
DMXR1_0002 55000
DMXR2_0002 56000
DMX_0002 2
"""

model = get_model(StringIO(par))
# add in new overlapping DMX
model.add_DMX_range(54500, 55000, dmx=0, frozen=True)
```
now results in a valid model but:
```
WARNING  (pint.models.dispersion_model  ): Start of DMX_0003 (54500.0-55000.0) overlaps with DMX_0001 (54000.0-55000.0)
```
Does that look reasonable?  

I also added utility functions:
* `pint.utils.get_prefix_timeranges()` which will return a list of all DMX (or SWX or ...) indices, start and end times
* `pint.utils.find_prefix_bytime()` which will figure out which (if any) DMX (or SWX...) bin contains a given time
I don't think either of these existed before but let me know if I'm wrong.

dlakaplan:
I added another utility `merge_dmx()`.  One question that came up is if we require/assume that the DMXs are contiguous.  It makes sense for something like `F0`,`F1`... to all exist.  But is there a reason why you can't have `DMX_0002` and not `DMX_0001`?  It wouldn't happen automatically, but if you are deleting bins it might occur.   I don't know if there are assumptions that would be broken by this (e.g., `get_prefix_list` method).

abhisrkckl:
Is it ready to merge?

dlakaplan:
Yes, please do (or let me know if you find any issues)

PR #1454, opened 2022-11-18
Logging debugging and improvements

dlakaplan:
* Fixed `onlyonce` behavior (comparison of logging level had the wrong sign)
* Added better log message for glitch fitting to include glitch number
* Fixed some `onlyonce` messages to match better (variable number of spaces)
* Added `onlyonce` message for glitch
* `onlyonce` now only stores the match to the regex, so can change for things like the glitch number

### DIFF
--- a/src/pint/logging.py
+++ b/src/pint/logging.py
@@ -173,11 +173,12 @@ class LogFilter:
             r"Adding columns .*": False,
             r"Applying TT\(\S+\) to TT\(\S+\) clock correction \(\~27 us\)": False,
             r"No pulse number flags found in the TOAs": False,
-            r"SSB obs pos \[\S+ \S+ \S+\] m": False,
+            r"SSB obs pos \[\S+\s+\S+\s+\S+\] m": False,
             r"Column \S+ already exists. Removing...": False,
             r"Skipping Shapiro delay for Barycentric TOAs": False,
             r"Special observatory location. No clock corrections applied.": False,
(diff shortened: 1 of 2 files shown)


dlakaplan:
As an example:
```
pint.logging.setup(level="DEBUG")
m, t = get_model_and_toas("j0742.par", "j0742.tim")
f = pint.fitter.Fitter.auto(t, m)
f.fit_toas()
log.debug("Glitch phase for glitch 1: 0.00811988977788343 ")
log.debug("Glitch phase for glitch 1: 0.0081198897778834 ")
log.debug("Glitch phase for glitch 2: 0.0081198897778834 ")
```

you see:
```
...
DEBUG    (pint.models.glitch            ): Glitch phase for glitch 1: 0.00811988977788343 
...
DEBUG    (__main__                      ): Glitch phase for glitch 2: 0.0081198897778834 
```
so all of the message for glitch 1 are suppressed except the first.  But the message for glitch 2 comes out.

kerrm:
Hi David,

Thanks for fixing this.  I've reviewed the changes and I have one quick question:

`message_to_save = record["message"][: match.span()[1]]`

I don't imagine it's a common use case, but what if someone specifies a regular expression which matches the end of the string?  Would it be better just to use both bounds of the match?

dlakaplan:
I think that because we are using `re.match()` and not `re.search()` it will always match from the beginning, but it's a good idea to use a proper slice anyway.

kerrm:
Ah, good point!

I consider this PR ready to go.

abhisrkckl:
Oops closed by accident.

PR #1455, opened 2022-11-22
Edep templates

kerrm:
This PR only affects the code used for managing Fermi pulse profiles.  It

1. Primarily adds the capability to model and fit energy evolution of pulse profile components.
2. Re-works the older capability to do this to a cleaner interface.
3. Removes some kruft.
4. Substantially improves test coverage for old and new code.  (I hope!)


kerrm:
Indeed!

dlakaplan:
Do you think it would be possible to include some PINT tutorials on Fermi and/or NICER data (something that is pure photons)?  Maybe they exist elsewhere and just need links, but it could be helpful for others.

kerrm:
That's a good idea, but I don't have anything handy!  I could work up something basic (in a different PR).  If we want to include sample data sets, what's the best mechanism for that?  I have some very small ones in the test directory, so perhaps I could use that.

paulray (approved):

PR #1457, opened 2022-12-01
Added method to find mask parameters with no TOAs

dlakaplan:
```
bad_parameters = m.find_empty_masks(t)
```
will return a list of the mask parameters with no TOAs

```
bad_parameters = m.find_empty_masks(t, freeze=True)
```
will also freeze them so future validation/fitting work.


src/pint/models/timing_model.py
```diff
@@ -2534,6 +2534,32 @@ def validate_toas(self, toas):
...
+        """
+        bad_parameters = []
+        for maskpar in self.get_params_of_type_top("maskParameter"):
+            par = getattr(self, maskpar)
+            if "TNEQ" in str(par.name) or par.frozen:
+                continue
+            if len(par.select_toa_mask(toas)) == 0:
+                bad_parameters.append(maskpar)
+                if freeze:
+                    log.info(f"'{maskpar}' has no TOAs so freezing")
```
scottransom, line 2559: Since this is explicitly changing something that the user requested (accidentally or not), I think this should be a warning and not an info.

dlakaplan: Except the default is to not change.  Change only happens when the user sets `freeze=True`

scottransom: Ah.  I missed that.  So this isn't called by default anywhere?  The user has to explicitly call this before a fit?  If so, then I agree with your plan (and apologize!).

src/pint/models/timing_model.py
```diff
@@ -2534,6 +2534,32 @@ def validate_toas(self, toas):
...
+
+        Returns
+        -------
+        list
+            Parameters with no TOAs
+        """
+        bad_parameters = []
+        for maskpar in self.get_params_of_type_top("maskParameter"):
+            par = getattr(self, maskpar)
+            if "TNEQ" in str(par.name) or par.frozen:
```
abhisrkckl, line 2554: Why is there a special case for TNEQ?

dlakaplan: I don't actually know - I copied it from the `validate_toas()` method

dlakaplan:
The exception will now give a suggestion as to how to fix it:
```
pint.models.timing_model.MissingTOAs: Parameter 'JUMP2, freq, [<Quantity 3000. MHz>, <Quantity 3200. MHz>]' does not correspond to any TOAs: you might need to run `model.find_empty_masks(toas, freeze=True)`
```

I also deleted the `TNEQ` special case since we were not sure why it was in there (it came via https://github.com/nanograv/PINT/pull/735)

PR #1465, opened 2022-12-13
Update of #1165 (add site clock to simulation) to fix BIPM

### LINKED ISSUE #1464: Simulated TOAs are not correctly written into a tim file
The `pint.simulation.make_fake_toas_uniform` function generates uniformly spaced TOAs. When I plot the residuals of the simulated `TOAs` object, I get zero residuals up to some numerical error as expected. However, some modulations are seen when I write this `TOAs` object into a `tim` file and then plot the residuals from the generated tim file. 

Why does this happen? 

Here is some code to reproduce this. The difference is at the ~100 ns level.
```
import astropy.units as u
import matplotlib.pyplot as plt

parfile = examplefile("NGC6440E.par")
timfile = "fake.tim"

model = get_model(parfile)

toas = make_fake_toas_uniform(
        startMJD=56000,
        endMJD=56400,
        ntoas=100,
        model=model,
        obs="ao",
        error=1 * u.microsecond,
        freq=1400 * u.MHz
    )
res = Residuals(toas, model)

plt.plot(toas.get_mjds(), res.calc_time_resids(), label="From TOAs object (before writing)")

toas.write_TOA_file(timfile, name="fake", format="TEMPO2")

model1, toas1 = get_model_and_toas(parfile, timfile)
res1 = Residuals(toas1, model1)

plt.plot(toas1.get_mjds(), res1.calc_time_resids().to("us"), label="From tim file (after writing)")
plt.xlabel("MJD")
plt.ylabel("Residuals (us)")
plt.legend()
plt.show()
```

EDIT: Here is the residual plot.

dlakaplan:
This updates #1165 to fix #1464 and #1161.  That separately implemented a check as to whether the clock was OK, but didn't set `include_bipm` correctly when needed.  This updates that and adds a check of residuals roundtrip through a file at higher precision.

### DIFF
--- a/src/pint/simulation.py
+++ b/src/pint/simulation.py
@@ -64,10 +64,7 @@ def zero_residuals(ts, model, maxiter=10, tolerance=None):
     ts.compute_pulse_numbers(model)
     maxresid = None
     if tolerance is None:
-        if pint.utils.check_longdouble_precision():
-            tolerance = 1 * u.ns
-        else:
-            tolerance = 5 * u.us
+        tolerance = 1 * u.ns if pint.utils.check_longdouble_precision() else 5 * u.us
     for i in range(maxiter):
         r = pint.residuals.Residuals(ts, model, track_mode="use_pulse_numbers")
         resids = r.calc_time_resids(calctype="taylor")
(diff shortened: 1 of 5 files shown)


src/pint/simulation.py
```diff
@@ -85,13 +82,11 @@ def zero_residuals(ts, model, maxiter=10, tolerance=None):
         )
 
 
-def update_fake_toa_clock(ts, model, include_bipm=False, include_gps=True):
-    """Update the clock settings (corrections, etc) for fake TOAs
+def get_fake_toa_clock_versions(model, include_bipm=False, include_gps=True):
```
aarchiba, line 85: Can this share code with the version in `pint.toa`?

dlakaplan: That might be better - I was fixing what Jing had.  I'll see how straightforward it is.

Issue #742, opened 2020-07-07
Change dispersion constant?


mtlam:
Inspired by discussions of https://arxiv.org/abs/2007.02886, the dispersion constant is defined as <a href="https://www.codecogs.com/eqnedit.php?latex=\frac{c&space;r_e}{2\pi}" target="_blank"><img src="https://latex.codecogs.com/gif.latex?\frac{c&space;r_e}{2\pi}" title="\frac{c r_e}{2\pi}" /></a> (in the notation I more commonly use). This is equal to _K_ = 1.3445365918(9)-07 m^2/s, versus the value of 1.3447217e-07 m^2/s when running `DMconst.decompose()`, a difference of ~0.014%. This change would cause differences from previous timing packages but would be more fundamentally correct in terms of defintiions.

paulray:
This certainly would need some thought and discussion.  Can you get the precise values used by TEMPO and Tempo2, for comparison?

I think for sure we could imagine adding a dispersion model that gives the "Dispersion Slope" (as Shri calls it) and avoids the constant altogether.

mtlam:
It looks like both use the same 1/2.41 value, to some power of 10 depending on units.

I guess we could imagine adding in a different dispersion model but I was thinking at least for now just the notion of the fundamental value. The former falls into a similar category, at least in my brain, of being able to add in a gas temperature if someone so wanted, which I will say is less important for pulsars given the much lower plasma temperatures (he's concerned with ICM gas for FRBs). On the value of K, this definitely needs thought and discussion, no doubt, I don't think we should blindly implement it but given how other constants are used throughout the code versus hand-defined values, we should heavily consider it.

scottransom:
There is a nice discussion of this going on in the FRB2020 meeting slack.  Most pulsar-related codes (including PRESTO, TEMPO, Tempo2, and PINT) use 1/0.241 as the DM constant.  SIGPROC uses 4.15.  Lorimer & Kramer recommend 4.148808, and the public FRB codes use a variety of things around 4.15....  *sigh*

paulray:
https://github.com/nanograv/PINT/blob/d21a6cc9312031501585c12ae171d9076a2725bd/src/pint/models/dispersion_model.py#L14-L16

As evidenced by the big discussion happening among FRB2020 attendees, I think this comment should be expanded to explain the background for the choice of DMconst and the convention defined in Manchester and Taylor and how it is consistent among TEMPO, Tempo2 and PINT.

[...]

PR #1471, opened 2022-12-13
DM explanation in documentation and dispersion slope as a derived quantity

abhisrkckl:
See issue #742 for a detailed discussion.

### DIFF
--- a/docs/explanation.rst
+++ b/docs/explanation.rst
@@ -183,6 +183,28 @@ not (this is a particular concern when working with X-ray or gamma-ray data).
 .. _TAI: https://www.bipm.org/en/bipm-services/timescales/tai.html
 .. _wiki: https://github.com/nanograv/PINT/wiki/Clock-Corrections-and-Timescales-in-PINT
 
+Dispersion Measure
+------------------
+Radio waves emitted by the pulsar experience dispersion as they travel through the ionized 
+interstellar medium (ISM). The time delay due to the interstellar dispersion is given by
+:math:`\frac{K\times DM}{\nu^2}`, where :math:`\nu` is the frequency of the radio signal.
+The dominant source of this dispersion is the presence of free electrons in the ISM, and 
+to a first approximation, the DM can be interpreted as the electron column density along the 
+line of sight to the pulsar. :math:`K` is known as the DM constant, and should be equal to 
+:math:`\frac{e^2}{8 \pi ^2 c \epsilon _0 m_e} \approx 1.3445365918(9)E-07\; m^2/s` based
+on the latest measurements of the physical constants. However, pulsar astronomers have 
+traditionally used a fixed value :math:`1.3447217E-07\; m^2/s` for :math:`K` over the years.
+For example, the Handbook of Pulsar Astronomy by Lorimer & Kramer (Chapter 5) provides the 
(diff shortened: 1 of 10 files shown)


docs/explanation.rst
```diff
@@ -183,6 +183,28 @@ not (this is a particular concern when working with X-ray or gamma-ray data).
...
 
+Dispersion Measure
+------------------
+Radio waves emitted by the pulsar experience dispersion as they travel through the ionized 
+interstellar medium (ISM). The time delay due to the interstellar dispersion is given by
+:math:`\frac{K\times DM}{\nu^2}`, where :math:`\nu` is the frequency of the radio signal.
+The dominant source of this dispersion is the presence of free electrons in the ISM, and 
+to a first approximation, the DM can be interpreted as the electron column density along the 
+line of sight to the pulsar. :math:`K` is known as the DM constant, and should be equal to 
+:math:`\frac{e^2}{8 \pi ^2 c \epsilon _0 m_e} \approx 1.3445365918(9)E-07\; m^2/s` based
```
dlakaplan, line 194: Could you do `E-07` -< `\times 10^{-7}`?  The former doesn't format well.  And this is done below also.

dlakaplan:
Is it worth adding a method to the various dispersion models to convert the DM to that implied by a different value of K?  Or a function in `pint.utils`?  Might be too confusing, or might get into some of the messy definition issues.

docs/explanation.rst
abhisrkckl: Done.

abhisrkckl:
I was thinking of adding a new derived quantity dispersion_slope = DMconst * DM. We can also have a utility function convert_dispersion_measure(dm, dmconst_new) which changes the DM for a different value of DMconst.

dlakaplan:
I think that would be a good idea.

dlakaplan:
Would it be possible to give a quick example of using the new methods in the docs rather than just a mention?

Separately, how about making a `dispersion_slope` method within the `Dispersion` model and separately within the `TimingModel`?  So that `m.total_dm(t)` could also be accompanied by `m.total_dispersion_slope(t)`?  I think this would be pretty simple.

dlakaplan:
That may not be needed - I don't know if people will more easily find a method within the `TimingModel` or a separate function

abhisrkckl:
A function within TimingModel will be easier to find. `m.total_dispersion_slope(t)`, similar to `m.total_dm(t)` should be self-explanatory and will show up in code completion better than a similar function in `derived_quantities`.
I think it will be a good idea to implement it, especially since it's pretty simple.

dlakaplan:
Separately I was wondering if some of those methods should also be able to operate on `astropy.Time objects`, and not just TOAs. This might make for a simpler interface for some use cases, and the difference between a TOA and a more generic Time is almost never relevant for DM calculations.

This might be better left for a separate discussion/PR

abhisrkckl:
> Separately I was wondering if some of those methods should also be able to operate on astropy.Time objects, and not just TOAs. This might make for a simpler interface for some use cases, and the difference between a TOA and a more generic Time is almost never relevant for DM calculations.

This won't work if there are DM jumps since they need TOA metadata to work.

dlakaplan:
I think you need to update the docs to include your new methods

abhisrkckl:
I have updated the docs.

dlakaplan:
Looks good.  I'll merge if you are done.

abhisrkckl:
I am done.

PR #1472, opened 2022-12-13
Add import to fix pintk with matplotlib 3.6

rossjjennings:
@scottransom was seeing an error when trying to run `pintk` on Python 3.10:

```
pintk NGC6440E.par NGC6440E.tim                                                      12:09:19
  File "/home/sransom/python_venvs/py3/bin/pintk", line 33, in <module>
    sys.exit(load_entry_point('pint-pulsar', 'console_scripts', 'pintk')())
  File "/home/sransom/git/PINT/src/pint/scripts/pintk.py", line 267, in main
    app = PINTk(
  File "/home/sransom/git/PINT/src/pint/scripts/pintk.py", line 49, in __init__
    self.createWidgets()
  File "/home/sransom/git/PINT/src/pint/scripts/pintk.py", line 131, in createWidgets
    "plk": PlkWidget(master=self.mainFrame, loglevel=self.loglevel),
  File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 673, in __init__
    self.initPlk()
  File "/home/sransom/git/PINT/src/pint/pintk/plk.py", line 700, in initPlk
    self.plkFig = mpl.figure.Figure(dpi=self.plkDpi)
  File "/home/sransom/python_venvs/py3/lib/python3.10/site-packages/matplotlib/_api/__init__.py", line 224, in __getattr__
    raise AttributeError(
AttributeError: module 'matplotlib' has no attribute 'figure'
```

This seems to have been caused by a change in matplotlib 3.6, which makes it so that `mpl.figure` isn't accessible unless it is explicitly imported. This PR adds the explicit import and so avoids the error.

### DIFF
--- a/src/pint/pintk/plk.py
+++ b/src/pint/pintk/plk.py
@@ -9,6 +9,7 @@ from astropy.time import Time
 import astropy.units as u
 import matplotlib as mpl
 import numpy as np
+import matplotlib.figure
 from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg
 
 import pint.pintk.pulsar as pulsar


abhisrkckl:
Can you add a test for this (maybe in test_pintk.py)?

rossjjennings:
@abhisrkckl I think the test that's already there should have caught this, it's just being skipped in CI because there's no display for it to output to. But maybe I can make an actual unit test for the `openPulsar()` function, which is where this happened...

rossjjennings:
Hmm, nope, that's not working, for two reasons: (1) `openPulsar()` isn't a stand-alone function, it's a method attached to the `PINTk` class, and (2) this didn't happen in `openPulsar()`, it happened in the `PlkWidget` constructor, inside `PINTk.createWidgets()`, which seems even more closely tied to the GUI. I don't think I'm up for the challenge of trying to write tests for pintk that avoid the GUI right now.

rossjjennings:
So, to be clear, the main issue with creating tests for pintk seems to be that almost all the pintk code is written as methods on classes that take `tk.Tk` objects in their constructors, and (I assume) trying to create a `tk.Tk` object in an environment with no display will fail.

rossjjennings:
Hmm, maybe one could use `tkinter.Tcl()` instead of `tkinter.Tk()`...

abhisrkckl:
OK... This looks straightforward. Can it be merged?

rossjjennings:
Yeah, as far as I'm concerned it's ready to be merged.

PR #1475, opened 2022-12-14
Update of #1165 (add site clock to simulation) to fix BIPM

### LINKED ISSUE #1464: Simulated TOAs are not correctly written into a tim file

dlakaplan:
This updates #1165 to fix #1464 and #1161. That separately implemented a check as to whether the clock was OK, but didn't set `include_bipm` correctly when needed. This updates that and adds a check of residuals roundtrip through a file at higher precision.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 ## Unreleased
 ### Changed
 ### Added
-- method to identify mask parameters with no TOAs and optionally freeze them
 ### Fixed
+- Creating fake TOAs properly handles site clock corrections
 ### Removed
 
 ## [0.9.2] 2022-11-30
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 ## Unreleased
 ### Changed
 ### Added
+- method to identify mask parameters with no TOAs and optionally freeze them
 ### Fixed
 - Creating fake TOAs properly handles site clock corrections
 - corrected a precision issue with reading ASCII representations of pulse profiles
--- a/tests/test_fake_toas.py
+++ b/tests/test_fake_toas.py
@@ -10,6 +10,7 @@ import os
 import pint.config
 from pint.fitter import GLSFitter
 from pinttestdata import datadir
+import pytest
 
 
 def roundtrip(toas, model):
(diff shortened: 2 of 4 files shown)


dlakaplan:
So far this passes the RTD CI.  It doesn't have all of the updated tests that #1165 had: I will try to add those back.

dlakaplan:
Actually, I'm not sure those other tests are needed.   They mostly added roundtrip tests on top of others, but we have a separate roundtrip test here.  So I think this should be sufficient.

dlakaplan:
Separately I was wondering if some of those methods should also be able to operate on `astropy.Time` objects, and not just TOAs.  This might make for a simpler interface for some use cases, and the difference between a TOA and a more generic Time is almost never relevant for DM calculations.

CHANGELOG.md
```diff
@@ -7,8 +7,8 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 ## Unreleased
 ### Changed
 ### Added
-- method to identify mask parameters with no TOAs and optionally freeze them
```
paulray, line 10: Why is this line deleted?

tests/test_fake_toas.py
```diff
@@ -12,6 +12,34 @@
...
+    with tempfile.NamedTemporaryFile("wt") as f:
+        toas.write_TOA_file(f)
+        f.flush()
+        toas2 = get_TOAs(f.name, model=model)
+    return toas2
+
+
+def test_roundtrip():
+    # test for roundtrip accuracy at high precision
+    parfile = os.path.join(datadir, "NGC6440E.par")
```
paulray, line 25: This roundtrip test is only done with NGC6440E.par, which has CLK UTC(NIST), which isn't even a valid CLK entry for  PINT. We should do a roundtrip test with a few other par files, with different settings for CLK including TT(TAI) and TT(BIPM2019) and maybe different PLANET_SHAPIRO as well

CHANGELOG.md
dlakaplan: I don't know if that was intentional - changelog merges may have had issues

tests/test_fake_toas.py
dlakaplan: That's fair.  I think it failed before precisely because of the CLK value, but demonstrating that other clocks are OK is worthwhile.  I'll add that.

dlakaplan: ok, I think the latest push does this

dlakaplan: This does it for the clock/planet_shapiro combinations:
```
 ("UTC(NIST)", "Y"),
        ("UTC(NIST)", "N"),
        ("TT(BIPM2021)", "Y"),
        ("TT(BIPM2021)", "N"),
        ("TT(TAI)", "Y"),
        ("TT(TAI)", "N"),
```
happy to add others if wanted, but this already takes ~30s

paulray: That seems sufficient! Could even remove one of the UTC(NIST) ones.

dlakaplan: since the NIST one was the problematic one, I think it's best to keep it

PR #1476, opened 2022-12-15
Fix numerical slop reading in ASCII profiles, add tests.

kerrm:
Julia reported a problem making NICER TOAs after updates to this template fitting code.  It was because I had begun to enforce strict parameter bounds, but profiles saved to ASCII templates can lose enough precision to fail the bounds check.  This PR:

(1) will no longer silently fail when the likelihood is nonsense (because of a failed parameter check)
(2) will check ASCII profiles and, if they're close enough to parameter bounds, silently skootch them up or down to the bound
(3) adds tests for both changes


paulray:
Fixes bugs, AND improves coverage... I love it!

paulray:
Can you add a CHANGELOG entry?

PR #1479, opened 2022-12-19
Updated docs for DDK model to include DT92 convention

dlakaplan:


dlakaplan:
See https://nanograv-pint--1479.org.readthedocs.build/en/1479/_autosummary/pint.models.binary_ddk.BinaryDDK.html#pint.models.binary_ddk.BinaryDDK for the updated docs

Issue #1462, opened 2022-12-06
Implement DDS model


dlakaplan:
From tempo documentation:
```
DDS   Variation of DD which uses SHAPMAX = -ln(1-s) instead of s = sin i.  
      Also incorporates various higher-order effects for use with 0737.
      These are experimental and should not be trusted yet! (2016 Feb.)
      References: Kramer et al. Science 314:97 (2006)
                  Rafikov & Lai PRD 73:063003 (2006)
```

dlakaplan:
References are:
https://ui.adsabs.harvard.edu/abs/2006Sci...314...97K/abstract
https://ui.adsabs.harvard.edu/abs/2006PhRvD..73f3003R/abstract

PR #1490, opened 2022-12-30
ELL1+, DDS, & DDGR models, binary conversion

dlakaplan:
* Implements DDS model (#1462) but only the SINI -> SHAPMAX conversion
* Implements DDGR model (#1461)
* Implements ELL1+ model (like ELL1, but including second-order corrections in ECC)
* Separate module for binary conversion (#1491), also included in `convert_parfile`

Minor additions:
* method to return parameter as `uncertainties.ufloat`
* function to parse time from a variety of input formats
* method to return binary period and uncertainty (optionally evaluated at different times) regardless of binary model

Currently this largely works.  

This does not handle the aberration terms since they need access to the spin parameters.  

When using DDGR, PK parameters like `PBDOT` and `SINI` can be in the par file on reading but are ignored.  They cannot be set.  Those quantities can be accessed from the model through `m.PBDOT` etc. attributes.  They are included when printing as a par file (but commented out).  Same for `SINI` with DDS.


dlakaplan:
The test I included just compares the binary delay to the DD model.  I couldn't get the DD models in the test suite to work for fitting.  But I did try with some J1640 data I got from @tcromartie and that seems to work OK.

```
mport os
import pytest
import copy

import astropy.units as u
import numpy as np

import pint.toa as toa
import pint.fitter
import pint.logging

pint.logging.setup("WARNING")
m, t = get_model_and_toas(
    "J1640+2224_PINT_20220221.nb.par", "J1640+2224_PINT_20220203.nb.tim", usepickle=True
)
mDDS = copy.deepcopy(m)
mDDS.remove_component("BinaryDD")
mDDS.add_component(m.components["BinaryDD"].as_DDS())

f = pint.fitter.Fitter.auto(t, m)
fDDS = pint.fitter.Fitter.auto(t, mDDS)
assert np.allclose(f.resids.time_resids, fDDS.resids.time_resids)

f.fit_toas()
chi2 = f.resids.calc_chi2()
print(f"DD model: chi2={chi2} with SINI={f.model.SINI}")

fDDS.fit_toas()
chi2DDS = fDDS.resids.calc_chi2()
print(
    f"DDS model: chi2={chi2DDS} with SHAPMAX={fDDS.model.SHAPMAX} (SINI={1-np.exp(-fDDS.model.SHAPMAX.quantity)})"
)
assert np.isclose(1 - np.exp(-fDDS.model.SHAPMAX.value), f.model.SINI.value)
assert np.isclose(chi2, chi2DDS)

fDDS.model.SHAPMAX.quantity += 0.5
print(f"Reset SHAPMAX to {fDDS.model.SHAPMAX.quantity}")
fDDS.fit_toas()
chi2DDS = fDDS.resids.calc_chi2()
print(
    f"DDS model: chi2={chi2DDS} with SHAPMAX={fDDS.model.SHAPMAX} (SINI={1-np.exp(-fDDS.model.SHAPMAX.quantity)})"
)
assert np.isclose(1 - np.exp(-fDDS.model.SHAPMAX.value), f.model.SINI.value, rtol=1e-3)
assert np.isclose(chi2, chi2DDS)
```
I don't think this is right to include in the test suite, but if we can make a cut down version that could be OK.

dlakaplan:
So far this only does the alternate Shapiro delay.  I can't get the paper with the other terms.

dlakaplan:
I think the binary delays for DDGR are now working (compared to DD).  Note that the inferred PK parameters may be calculated too often - it's possible that this function should be cached or only called by the `setter` function.  But for right now it works.

Thinking ahead to the derivatives, as far as I understand we need derivatives for all of the removed parameters with respect to the new parameters.  And then PINT will construct the proper derivatives from that.  Is that correct?  So we need:

`d_PBDOT_d_MTOT`
`d_GAMMA_d_MTOT`
`d_k_d_MDOT`
`d_Dr_d_MTOT`
`d_Dth_d_MTOT`
`d_SINI_d_MTOT`

Do I understand this correctly?  I think I can calculate those, although including the implicit calculation of the relativistically-corrected semi-major axis in that might be harder.

dlakaplan:
Fitting for `MTOT` now seems OK.  There is an explicit test to see that the design matrix produces the expected change in chi^2.

abhisrkckl:
I think it makes more sense for the `binaryconvert` module to be in `pint.models` rather than in `pint`.

dlakaplan:
I don't think `pint.models`: that is for model definitions.  I could see it as a sub package in `pint.utils`.  So far we haven't been very consistent with such organization.

dlakaplan:
I tested the DDGR model in PINT vs. libstempo.  What I did was:
* generate DDGR model in PINT
* convert to the equivalent DD model
* use that to generate fake TOAs
* compare binary delays for the DD and DDGR models within PINT and between PINT & libstempo
The results vary a bit depending on the parameters, but for a 5d binary the PINT DD and DDGR delays agree to 8e-11s (rms).  The results involving libstempo (DD vs. DDGR, or libstempo DD[GR] vs. PINT DD[GR]) agreed to ~5e-8s.  Not sure why it's that different, but I think this is consistent with some tests that are already in PINT (e.g., testing DD for PINT vs. libstempo has better agreement, but that's for a wider binary).

PR #1482, opened 2022-12-20
Refactor `Dre` method in DD model for better readability (fix #1163)

abhisrkckl:
Also fixes some typos, expressions for ER and ETH, and an error in `Simulate_and_make_MassMass.ipynb` that was causing the notebooks build to fail (see discussion below).

#1163

Edit: The notebook fix was done in #1497

PR #1493, opened 2023-01-04
Update requirements for oldestdeps

paulray:
I updated the version of scipy used for the oldestdeps tests to a version released just before our numpy version 1.18.5. This constraint keeps it from installing a very new scipy, which pulls in a new numpy that causes crashes in the oldestdeps tests. So this bumps up the scipy minimum required version to 1.4.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project, at least loosely, adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Changed
+- Changed minimum supported version of scipy to 1.4.1
+### Added
+### Fixed
+### Removed
+
 ## [0.9.3] 2022-12-16
 ### Changed
 ### Added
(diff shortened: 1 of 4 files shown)


paulray:
Fixes #1401 (again).  Should solve issue seen in #1482

paulray:
Oldestdeps now passes. This can be merged as soon as someone else takes a look. The only significant change is updating the minimum version of scipy to 1.4.1.  Actually, I'll add a changelog entry and push that now.

paulray:
OK, I botched, then fixed, the CHANGELOG merge. It should be good to go now

dlakaplan:
Those are pretty big bumps to the `scipy` and `matplotlib` versions.  `scipy` 1.4.1 is from Dec 2019, and `matplotlib` 3.2.0 is from March 2020.  But I guess in both cases those are consistent dates with our `numpy` requirement so they make sense.  So I can merge this unless anybody objects (or I can wait a day or two).

dlakaplan:
And does adding the oldestdeps to the tox environments mean it will get tested more explicitly?

paulray:
I'm not sure the effects of adding oldestdeps to the environments list. I don't know why it wasn't there before. I just noticed that `tox -l` didn't show the oldestdeps environment, which seemed wrong, so I added it.

Issue #1495, opened 2023-01-06
Implement read-only linked parameters?


dlakaplan:
Many models might have various derived quantities that a user would want to access but should not write.  An example of this is in the DDGR model, where PK parameters like `PBDOT` and `SINI` are in the par file but are ignored on reading, and then calculated from `MTOT` and `M2` on writing.  

If we didn't need to keep those on read then we could just delete those parameters from the model and make read-only properties.  However, those would not be included in the par files when written.

My suggestion is to make a read-only parameters (`funcParameter`?) that is based on a function call with values computed as needed.  So it could have:

`funcParameter(name="M1",value=(lambda mt,m2 : mt-m2, ("MTOT","M2"),...)`

so that when a user wants `model.M1` it runs the function with `(getattr(model,"MTOT").quantity,getattr(model,"M2").quantity))` as inputs.  This would be visible like a normal parameters and be written to the par file but would be read-only and dynamic (we have read-only parameters, but they take fixed values).  

If there is a better way to do this I'm open to suggestions.  Right now I am kludging it by:

* keeping the parameters like `SINI` initially
* deleting them in `setup()` so they cannot be written
* adding additional parameters to `print_par()` so that they get computed at that time and written
Not great, but it does the minimal amount.

dlakaplan:
Actually, I'm not sure how to do this.  I don't know how one parameter can access others without some notion of a parent model component.

PR #1496, opened 2023-01-07
Function parameters

dlakaplan:
You can now define parameters that are read-only functions of other parameters:
```

base_par = """
PSR J1234+5678
ELAT 0
ELONG 0
F0 1
F1 3e-10
DM 10
PEPOCH 57000
"""

m = get_model(io.StringIO(base_par))

p = pint.models.parameter.funcParameter(
    name="AGE",
    description="Spindown age",
    pars=("F0", "F1"),
    func=lambda f0, f1: f0 / 2 / f1,
    units="yr",
)
m.components["Spindown"].add_param(p)
```

Then for example:
```
In [2]: print(m)
# Created: 2023-01-06T20:19:14.321710
# PINT_version: 0.7+3368.g11b481c5
# User: David Kaplan (kaplan)
# Host: murgle
# Format: pint
PSR                            J1234+5678
DILATEFREQ                              N
DMDATA                                  N
NTOA                                    0
CHI2                                  0.0
ELONG                   0.000000000000000
ELAT                    0.000000000000000
PMELONG                               0.0
PMELAT                                0.0
PX                                    0.0
ECL                              IERS2010
F0                                    1.0
F1                                  3e-10
PEPOCH             57000.0000000000000000
# AGE                  52.81347969004824541
PLANET_SHAPIRO                          N
DM                                   10.0
```
Note that `AGE` is commented out on output, so it will not be read.
or:
```
In [3]: m
Out[3]: 
TimingModel(
  AstrometryEcliptic(
    MJDParameter(   POSEPOCH            UNSET,
    floatParameter( PX                  0.0               (mas) frozen=True),
    AngleParameter( ELONG               0.000000000000000 (deg) frozen=True),
    AngleParameter( ELAT                0.000000000000000 (deg) frozen=True),
    floatParameter( PMELONG             0.0               (mas / yr) frozen=True),
    floatParameter( PMELAT              0.0               (mas / yr) frozen=True),
    strParameter(   ECL                 IERS2010          frozen=True)),
  DispersionDM(
    floatParameter( DM                  10.0              (pc / cm3) frozen=True),
    floatParameter( DM1                 UNSET,
    MJDParameter(   DMEPOCH             UNSET),
  SolarSystemShapiro(
    boolParameter(  PLANET_SHAPIRO      N                 frozen=True)),
  Spindown(
    floatParameter( F0                  1.0               (Hz) frozen=True),
    MJDParameter(   PEPOCH              57000.0000000000000000 (d) frozen=True),
    floatParameter( F1                  3e-10             (Hz / s) frozen=True),
    funcParameter(  AGE                 52.81347969004824541 (yr) frozen=True))
)
```
As the values of (in this case) `F0` and `F1` change, `m.AGE` will change with them.

Addresses #1495 

Aside from the notebook build I think this is ready for review.


dlakaplan:
I added some `funcParameters` to allow the DDK model to also report values in the IAU convention:
```
# Created: 2023-01-12T09:18:53.644382
# PINT_version: 0.7+3381.g0df68f97.dirty
# User: David Kaplan (kaplan)
# Host: plock
# Format: pint
PSRJ                           J1713+0747
EPHEM                               DE430
CLK                          TT(BIPM2015)
UNITS                                 TDB
RM                                    8.7 0 0.5
DILATEFREQ                              N
DMDATA                                  N
NTOA                                    0
CHI2                                  0.0
ELONG                 256.668695241000023 0 0.00000000200000000000
ELAT                   30.700360494000002 0 0.00000000400000000000
PMELONG                             5.267 0 0.002
PMELAT                             -3.443 0 0.004
PX                                   0.82 0 0.03
ECL                              IERS2010
...
KIN                                  71.4 0 0.7
KOM                                  83.0 0 3.0
KINIAU                              108.6
KOMIAU                                7.0
"""

**Comment by dlakaplan** [comment]
Currently, if you define a `funcParamter` using `lambda` functions it will work fine, but fail on pickle tests.  I think we could solve this with https://github.com/uqfoundation/dill.  Is there a reason to/not to allow that?

**Comment by dlakaplan** [comment]
Right now, these functions only return a single value with no uncertainty.  It would be possible to include uncertainties using the `uncertainty` package, but that would require all of the functions to use those data-types and math variants, and might need to drop some `Quantity` support.  That might be worthwhile but I think would be separate.

**Comment by abhisrkckl** [comment]
I think these parameters should be commented out while writing to a par file. Otherwise PINT may give warnings about unknown parameters when the par file is read in again (same thing with tempo and tempo2).

**Comment by abhisrkckl** [comment]
> Currently, if you define a funcParamter using lambda functions it will work fine, but fail on pickle tests.

Does this issue happen with normal functions?

I think the most convenient use case for this feature will be with lambdas, and they should be allowed. Is there a way to check if the `func` argument is a lambda? If yes, we can emit a warning about the pickling issue, but I think it should work as usual otherwise.

**Comment by dlakaplan** [comment]
> I think these parameters should be commented out (#) while writing to a par file. Otherwise PINT may give warnings about unknown parameters when the par file is read in again (same thing with tempo and tempo2).

Well, in the tempo(2) DDGR model it prints out things like PBDOT, which PINT needs to know to ignore. That was actually what started this.  But you are right that in other cases it may cause issues.  So I put in a flag to determine whether or not those are included in the par file - False by default.  That could be changed to commenting out instead.

**Comment by dlakaplan** [comment]
> > Currently, if you define a funcParamter using lambda functions it will work fine, but fail on pickle tests.
> 
> Does this issue happen with normal functions?
> 
> I think the most convenient use case for this feature will be with lambdas, and they should be allowed. Is there a way to check if the `func` argument is a lambda? If yes, we can emit a warning about the pickling issue, but I think it should work as usual otherwise.

Added a warning.

**Comment by dlakaplan** [comment]
I changed the behavior so a parameter is either listed in the par file or commented out

**Comment by aarchiba** [comment]
Would it perhaps make more sense for PINT to do something sensible when encountering these on read? At a minimum, ignore them? And squawk if the par file requests they be fit? Maybe even check whether the value in the par file is close to the value computed from other parameters (so we can tell users what's happening if they try changing these values)?

**Comment by dlakaplan** [comment]
It does explicitly ignore them on read

**Comment by aarchiba** [comment]
How does this interact with alternative parametrizations? For example, PINT can accept PB/PBDOT or FB0, FB1, FB2, ... interchangeably in any (?) binary model. It would be helpful to be able to easily compute either parametrization from the other, can that fit into this framework?

**Comment by dlakaplan** [comment]
That's a good question, and it's relevant as well to #1490 .  In that case I think most DDGR models use just PB, since having explicit time derivatives on top of GR gets very complicated, but it should probably be allowed.

Right now I **think** I could include that using this, but it would take a little thought.  That's because by default both `PB` and `FB0` are in there are parameters and one is typically unset.  So there are two issues:
* If both are available, would that confuse things down the line?
* If both are there to start, then it would have to see which is used, delete the other, and replace it.

**Comment by dlakaplan** [comment]
So far, this does implement making `FB0` a function of `PB` and vice versa, although I think that is doable (it requires some checks to make sure the choice of the native parameterization is correct).  But in other cases I don't think it is harmful. e.g., an `ELL1` model can show `ECC` and `OM` (not `T0` yet, since this doesn't yet support `MJDParameters`: possibly a separate PR).

e.g.,

```
In [3]: m = get_model("tests/datafile/J0023+0923_ell1_simple.par")

In [4]: m
Out[4]: 
TimingModel(
  AstrometryEcliptic(
    MJDParameter(   POSEPOCH            56567.0000000000000000 (d) frozen=True),
    floatParameter( PX                  0.9               (mas) +/- 0.2 mas frozen=True),
    AngleParameter( ELONG               9.070393800000000 (deg) +/- 0d00m00.000036s frozen=True),
    AngleParameter( ELAT                6.309108530000000 (deg) +/- 0d00m00.000324s frozen=True),
    floatParameter( PMELONG             -13.9             (mas / yr) +/- 0.03 mas / yr frozen=True),
    floatParameter( PMELAT              -0.3              (mas / yr) +/- 0.4 mas / yr frozen=True),
    strParameter(   ECL                 IERS2010          frozen=True)),
  BinaryELL1(
    floatParameter( PB                  UNSET,
    floatParameter( PBDOT               UNSET,
    floatParameter( A1                  0.03484142        (ls) +/- 7e-08 ls frozen=True),
    floatParameter( A1DOT               UNSET,
    floatParameter( EDOT                UNSET,
    floatParameter( OMDOT               UNSET,
    floatParameter( M2                  UNSET,
    floatParameter( SINI                UNSET,
    floatParameter( FB0                 8.338721549e-05   (1 / s) +/- 3e-14 1 / s frozen=True),
    MJDParameter(   TASC                56567.0260936200000000 (d) +/- 8e-08 d frozen=True),
    floatParameter( EPS1                7.2e-06           () +/- 3.1e-06 frozen=True),
    floatParameter( EPS2                -4e-06            () +/- 3.2e-06 frozen=True),
    floatParameter( EPS1DOT             UNSET,
    floatParameter( EPS2DOT             UNSET,
    funcParameter(  ECC                 8.236504112789600253e-06 () frozen=True),
    funcParameter(  OM                  119.05460409907714933 (deg) frozen=True),
    floatParameter( FB1                 -1.87e-20         (1 / s2) +/- 1.4e-21 1 / s2 frozen=True),
    floatParameter( FB2                 2.4e-28           (1 / s3) +/- 5e-29 1 / s3 frozen=True)),
  DispersionDM(
    floatParameter( DM                  14.3281           (pc / cm3) +/- 4e-05 pc / cm3 frozen=True),
    floatParameter( DM1                 UNSET,
    MJDParameter(   DMEPOCH             56538.0000000000000000 (d) frozen=True)),
  SolarSystemShapiro(
    boolParameter(  PLANET_SHAPIRO      N                 frozen=True)),
  Spindown(
    floatParameter( F0                  327.8470205906107 (Hz) +/- 7e-13 Hz frozen=True),
    MJDParameter(   PEPOCH              56567.0000000000000000 (d) frozen=True),
    floatParameter( F1                  -1.22783e-15      (Hz / s) +/- 4e-20 Hz / s frozen=True))
)
```
You cannot set those parameters:
```
In [5]: m.ECC.value = 2
<ipython-input-5-c5ad1861da3c> in <cell line: 1>()
----> 1 m.ECC.value = 2

~/pythonpackages/PINT_dlakaplan/src/pint/models/parameter.py in value(self, value)
   2328     @value.setter
   2329     def value(self, value):
-> 2330         raise AttributeError("Cannot set funcParameter")
   2331 
   2332     @property

AttributeError: Cannot set funcParameter
```
and on print-out they are commented out, so they won't be read subsequently:
```
In [6]: print(m)
# Created: 2023-01-17T11:04:51.467698
# PINT_version: 0.7+3413.g9c87528e.dirty
# User: David Kaplan (kaplan)
# Host: plock
# Format: pint
PSRJ                           J0023+0923
EPHEM                               DE436
CLK                          TT(BIPM2015)
UNITS                                 TDB
DILATEFREQ                              N
DMDATA                                  N
NTOA                                    0
CHI2                                  0.0
ELONG                   9.070393800000000 0 0.00000001000000000000
ELAT                    6.309108530000000 0 0.00000009000000000000
PMELONG                             -13.9 0 0.03
PMELAT                               -0.3 0 0.4
PX                                    0.9 0 0.2
ECL                              IERS2010
POSEPOCH           56567.0000000000000000
F0                      327.8470205906107 0 7e-13
F1                           -1.22783e-15 0 4e-20
PEPOCH             56567.0000000000000000
PLANET_SHAPIRO                          N
DM                                14.3281 0 4e-05
DMEPOCH            56538.0000000000000000
BINARY                               ELL1
A1                             0.03484142 0 7e-08
FB0                       8.338721549e-05 0 3e-14
TASC               56567.0260936200000000 0 8e-08
EPS1                              7.2e-06 0 3.1e-06
EPS2                               -4e-06 0 3.2e-06
# ECC              8.236504112789600253e-06
# OM                  119.05460409907714933
FB1                             -1.87e-20 0 1.4e-21
FB2                               2.4e-28 0 5e-29
```

Issue #1483, opened 2022-12-20
PINT example files should load without warnings


aarchiba:
When I load `NGC6440E.par` and the corresponding `.tim` file, I get the following warnings:
```
WARNING  (pint.logging                  ): /home/archibald/projects/pint/PINT/src/pint/models/timing_model.py:373 UserWarning: PINT only supports 'T2CMETHOD IAU2000B'
WARNING  (pint.toa                      ): CLOCK = UTC(NIST) is not implemented. Using TT(BIPM2021) instead.
WARNING  (pint.toa                      ): Unknown command MODE in line MODE 1
```

We should remove the inputs that cause these.

When PINT's example files and test files were not separate, we kept around some files with unsupported settings for testing purposes. Now that the example files are in a separate location, we should remove the settings that cause PINT to emit unnecessary warnings.

paulray:
I agree. The "standard" example of NGC6440E that we use in our examples and walkthrough should have a fully supported parfile.

aarchiba:
I think this logic applies to everything in `pint.config.examplefile(...)`. Should be an easy fix I think?

PR #1487, opened 2022-12-21
Fix notebooks

aarchiba:
Currently I can't run `make notebooks`; see bug #1486. This PR is meant to fix this.

Currently outstanding problems:

- [ ] `Simulate_and_make_MassMass.ipynb` - `ValueError: UTC times during a leap second cannot be represented in pulsar_mjd format`
- [ ] `Wideband_TOA_walkthrough.ipynb` - `fitter.get_noise_covariancematrix()` doesn't work with a downhill fitter
- [ ] `understanding_fitters.ipynb` - `pwfit.plot()` produces a plot showing zero residuals
- [ ] `check_clock_corrections.ipynb` - doesn't actually list clock correction ending MJDs

PR #1499, opened 2023-01-14
fix NGC6440E.par

abhisrkckl:
The example files `NGC6440E.par` and `NGC6440E.par.good` had unsupported `UNITS UTC(NIST)` and `T2CMETHOD TEMPO`. `PINT` was ignoring these lines with warnings and using `TT(BIPM2021)` and `IAU2000B` instead. I have changed these par files so that the warnings go away. This is the first step in addressing issue #1483. 

EDIT: I have also changed the warning for a "MODE" statement in the TOA file to be more informative.

### DIFF
--- a/src/pint/data/examples/NGC6440E.par
+++ b/src/pint/data/examples/NGC6440E.par
@@ -8,10 +8,10 @@ POSEPOCH      53750.000000
 DM              223.9  1 0.3
 SOLARN0               0.00
 EPHEM               DE421
-CLK              UTC(NIST)
+CLK            	  TT(BIPM2021)
 UNITS               TDB
 TIMEEPH             FB90
-T2CMETHOD           TEMPO
+T2CMETHOD           IAU2000B
 CORRECT_TROPOSPHERE N
 PLANET_SHAPIRO      N
 DILATEFREQ          N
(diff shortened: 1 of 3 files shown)


abhisrkckl:
There is also another warning about `MODE 1` in `NGC6440E.tim`. I have not removed it as I am not sure if that is required to use this file with `tempo` and `tempo2`.

paulray:
I don't know if TEMPO or Tempo2 still default to not using TOA errors if MODE 1 is omitted. 
PINT could be changed to silently ignore MODE 1 with no no warning. If MODE 1 is omitted, PINT acts as if it was present.
I'm not sure what should happen if there is MODE 0 (which I've never seen). Does PINT even have the ability to ignore TOA errors?

abhisrkckl:
> I don't know if TEMPO or Tempo2 still default to not using TOA errors if MODE 1 is omitted.

`MODE 1` is required by `tempo2` to use weighted least-squares. `tempo2` will use ordinary least-squares if `MODE 0` is given or if no `MODE` is specified.

paulray:
Looks good. This can be merged as far as I am concerned.

abhisrkckl:
This is ready to go, I think. The failing CI test is due to a connection issue at the ephemeris download server.

abhisrkckl:
Closing this as this duplicates #1487

Issue #1171, opened 2022-01-28
Add  HAWC and MAGIC in observatories.py


parthasarathipal:
Hi,
   I am trying to do pulsar analysis with HAWC data. Please add HAWC and MAGIC location infotmation in the Observatories.py file.

paulray:
This is best done via a pull request by someone with knowledge of those observatories, and it would be useful if it included information about the source of the location information, as is done for other observatories in observatories.py.

aarchiba:
If someone has exact position information, preferably with provenance as @paulray says, I'd be happy to write the pull request.

paulray:
As an example, HESS was just added via #1503

abhisrkckl:
I see that MAGIC is already included in observatories.dat. HAWC is not there though.

dlakaplan:
I emailed the HAWC spokesperson a while ago.  Just tried again.

PR #1503, opened 2023-01-18
Adding H.E.S.S. telescope to the observatory.json file

MRegeard:
Hello, 

I am starting as a PhD student at APC, Paris under the supervision of Arache Djannati-Ataï (@djannati), studying pulsars with HESS telescope. Therefore, I am using PINT to compute phases of pulsars. 

This pull request is to add HESS to the list of observatories in the observatories.json file.
As you can see I already added HESS to the file. I have tested the file on HESS data and it seems to work properly. 

Tell me if you have any comments or anything. I am kind of new on GitHub stuff so I may have not done things the right way.

Cheers.
Signed-off-by: Maxime Regeard <regeard@apc.in2p3.fr>


paulray:
Thanks for the PR! I'm surprised that HESS wasn't already included, but happy to hear that PINT is being used on HESS data. I guess the times HESS records are already in UTC(GPS) time, so no special clock correction file is needed?

abhisrkckl:
Can you add a CHANGELOG entry?

djannati:
Hi Paul,
We have been developing a pipeline for HESS based PINT since quite a while (3 yrs or so). Till now we have mainly used it to x-check the Tempo2 results, but the aim is to use PINT for next publications. HESS has indeed no clock correction files. The HESS timestamps are based on a GPS receiver on site, so there are no clock correction files. They are recorded as Unix timestamps (a decision made by computing people I guess at the beginning of HESS) with leap seconds corrected for --in principle!.  In the newer gammapy/PINT based pipeline, timestamps are converted and stored as TT. This conversion is still to be double-checked.

paulray:
By any chance do you have the coordinates to add HAWC and MAGIC as well?  That would close #1171 as well

dlakaplan:
> By any chance do you have the coordinates to add HAWC and MAGIC as well? That would close #1171 as well

That would also require knowledge of their clock systems, right?  Or are those available?

paulray:
As far as I'm concerned, this can be merged.

PR #1511, opened 2023-01-19
Create TOAs from numpy array or Time object directly

dlakaplan:
This allows something like:
```
t = pulsar_mjd.Time(np.array([55000, 56000]), scale="utc", format="pulsar_mjd")
errors = np.array([1.3, 2.1]) * u.us
freqs = np.array([1400, 3400]) * u.MHz
obs = "gbt"
flags = [{"type": "good"}, {"type": "bad"}]

toas = toa.get_TOAs_array(t, obs, errors=errors, freqs=freqs, flags=flags, name="data")
```
I think it uses the same machinery as normal TOA creation to do the clock corrections, set scales, etc.  I added some basic tests that compare against standard creation.

I also overloaded the addition operator to allow merging of TOAs:
```
toas3 = toas + toas2
```
TOAs can be merged in-place:
``` 
toas.merge(toas2)
```
and this is implemented by `+=`:
```
toas += toas2
```
keeps the same object


dlakaplan:
I haven't done any speed tests

dlakaplan:
@MRegeard : I think this effectively does what you were after.  Is that the sort of interface you'd like?

If you want to test the branch and provide any feedback, that would be welcome.

scottransom:
@dlakaplan I went to try to run some speed tests based on your testing code and I think there is an issue.  It seems like the TOAs from your new method all have the same clock correction.  You can see that with the following test code:

```
import numpy as np
from pint import pulsar_mjd
from pint import toa

obs = "gbt"

for nn in [3]:  # , 100, 1000, 10000]:
    ts = pulsar_mjd.Time(
        np.linspace(55000, 56000, nn), scale="utc", format="pulsar_mjd"
    )
    es = (np.random.standard_normal(nn) + 10.0) << u.us
    fs = (np.ones(nn) * 1400) << u.MHz
    toas = toa.get_TOAs_array(ts, obs, errors=es, freqs=fs)  # , flags=flags, **kwargs)
    toalist = [toa.TOA(t, obs=obs, error=e, freq=f) for t, e, f in zip(ts, es, fs)]
    toas_fromlist = toa.get_TOAs_list(toalist)
    # assert np.all(toas.table == toas_fromlist.table)

print(toas.table["flags"])
print(toas_fromlist.table["flags"])
```
that gives
```
                flags                
-------------------------------------
{'clkcorr': '2.7611500002795458e-05'}
{'clkcorr': '2.7611500002795458e-05'}
{'clkcorr': '2.7611500002795458e-05'}
                flags                
-------------------------------------
{'clkcorr': '2.7221880003809927e-05'}
{'clkcorr': '2.7412030001401897e-05'}
{'clkcorr': '2.7611500002795458e-05'}
```
So I'm not sure how your `assert` is passing in the test!  Unless me omitting the flags or kwargs is somehow affecting that....

dlakaplan:
OK, I think I know what happened.  I had given explicit flags, so they were all initialized to different objects.  You hadn't, so they were all the same (empty) object.  I added an deep-copy of the flag objects and I think it works.  Subtle - glad you caught it.

dlakaplan:
P.S.  When I run the speed test with crude timing I get:
```
100 List: 0.14369511604309082 Array 0.08748579025268555
1000 List: 0.8177459239959717 Array 0.7311990261077881
10000 List: 9.074469804763794 Array 6.562474012374878
```
so it's better but not amazing.  Note that I had to add in a dummy call before to get the various clock files loaded & initialized: otherwise the first call to either took a lot longer.  I can look tomorrow at profiling.

MRegeard:
@dlakaplan Thanks for doing that this quickly. 
I'm going to try this and give you my feed back.

MRegeard:
Ok so I have done a few tests and it seems to be exactly what I needed for my analysis. Even if the improvement in time is not amazing, I think that it is really straight forward and user friendly for computing phases with gamma-ray data !

I take advantage of the opportunity to ask two "naive" questions:

1. In order to compute phases from a .par file, is it always necessary to compute TDBs and posvels ? I am actually doing it when computing phases for HESS but I was wondering if there is a use case where we don't want them to be computed. 

2. When I was computing my phases, I use this code : 
```
include_bipm = False
include_gps = False

toa_list = list(toa.TOA(m, obs='hess', ephem='DE421') for m in mjd)

ts = toa.TOAs(toalist=toa_list)
ts.clock_corr_info={'include_bipm': str(include_bipm), 'include_gps': str(include_gps)}
ts.compute_TDBs(ephem='DE421')
ts.compute_posvels(ephem='DE421',planets=True)

phss = model.phase(ts,abs_phase=True)[1]
phases_not_clock = np.where(phss < 0.0 , phss + 1.0 , phss)
``` 
So you can see that I am not applying `apply_clock_corrections`because I assumed that it wasn't necessary since in HESS we don't have clock corrections.
However, while doing the tests with `get_TOAs_array`, I get phases which have difference at 1e-4 from those I get with my code above. And I see in the code of `get_TOAs_array`that `apply_clock_corrections` in applied (of course I have put `include_bipm` and `include_gps` to `False` in `get_TOAs_array` to compare. 
So I get that even if `include_bipm` and `include_gps`are false, `apply_clock_corrections`does something that imply this difference. If you could explain why there is this behaviour.

dlakaplan:
Doing some profiling, the majority of the time is doing the clock corrections which is common to both methods. So I think changes there would have the biggest impact.

paulray:
What are the differences between the two methods in microseconds (rather than pulse phase)?

paulray:
> ts = toa.TOAs(toalist=toa_list)
> ts.clock_corr_info={'include_bipm': str(include_bipm), 'include_gps': str(include_gps)}
> ts.compute_TDBs(ephem='DE421')
> ts.compute_posvels(ephem='DE421',planets=True)

Instead of this, I recommend `toa.get_TOAs_list(toa_list, ephem='DE421',include_bipm=False, include_gps=False)`

paulray:
>> What are the differences between the two methods in microseconds (rather than pulse phase)?

> is 0

I'm confused. How is 1E-4 in phase equal to 0 in microseconds?

MRegeard:
> > > What are the differences between the two methods in microseconds (rather than pulse phase)?
> 
> > is 0
> 
> I'm confused. How is 1E-4 in phase equal to 0 in microseconds?

I called `ts.table[mjd]`. is that what you meant by microseconds ?

paulray:
No I just meant how many microseconds is 1E-4 phase for your pulsar?

MRegeard:
Oh ok my bad. 

It's around 0.0275 us.

paulray:
OK, that is the size of the BIPM correction. It must be getting applied in one method and not the other.

paulray:
Agreed.  As I said above, I recommend `toa.get_TOAs_list()` instead of doing all the steps manually.

MRegeard:
Ok I see. Thank you very much for these responses.

paulray:
Sure! Definitely let us know if you still can't get the two methods to agree.

MRegeard:
If I use `get_TOAs_list`and `get_TOAs_array`with the same input parameters, the two methods are in total agreement !

dlakaplan:
Not sure if this was an issue, but it seemed that the clock corrections got applied to the `mjd` column, not the `mjd_float` column, so that could get out of sync.  Is that intentional?  Did that not matter?  I think this fixes that too (for other TOA creation as well).

dlakaplan:
With that fix a test was failing because `get_mjds()` could return the `mjd` column or the `mjd_float` column depending on an argument.  The test was written so that it compared mjds (in this case from `mjd_float`) with explicit BIPM setting and with the defaults.  Once I had changed it so that the `mjd_float` column had clock corrections the test failed.  Now the BIPM settings are the same on both ends so the test passes.

src/pint/toa.py
```diff
@@ -2584,3 +2581,250 @@ def merge_TOAs(TOAs_list, strict=False):
...
+    commands=None,
+    hashes=None,
+    limits="warn",
+    **kwargs,
+):
+    """Load and prepare TOAs for PINT use from an array of times.
+
+    Creates TOAs from a an array of times, applies clock corrections, computes
+    key values (like TDB), computes the observatory position and velocity
+    vectors.
```
paulray, line 2609: Should we say that this requires that all of the TOAs be at the same observatory? Also, does this work properly for an observatory whose position is a function of time (e.g. a spacecraft)?

docs/user-questions.rst
```diff
@@ -112,12 +111,13 @@ and then convert them to a :class:`pint.toa.TOAs` object::
     t = pulsar_mjd.Time(np.array([55000, 56000]), scale="utc", format="pulsar_mjd")
     obs = "gbt"
 
-    toalist = [toa.TOA(tt, obs=obs, error=1 * u.us) for tt in t]
-    toas = toa.get_TOAs_list(toalist)
+    toas = toa.get_TOAs_array(t, obs)
 
 Note that we also use ``Time`` from :mod:`pint.pulsar_mjd` rather than :class:`astropy.time.Time` directly to allow the 
 ``pulsar_mjd`` format, designed to avoid leap seconds.  But this would work with standard :class:`astropy.time.Time`
```
paulray, line 117: I wouldn't say that pulsar_mjd "avoids" leap seconds, but rather that is handles leap seconds according to standard conventions used in pulsar work. Also, I don't think you really need to use `pulsar_mjd.Time`. Once you have imported the code that defines `pulsar_mjd`, I think normal `Time` should work fine.

scottransom:
The `mjd` and `mjd_float` columns getting out of sync is almost certainly *not* intentional.  However, since `mjd_float` is only a 64-bit float, it is only intended for non-precision things (like plotting).  So it probably isn't a big deal (but I'm glad you are fixing it!).  However, if `tdb` and `tdbld` got out of sync like that, I think it *would* be a problem....

src/pint/toa.py
dlakaplan: The latest docstring does say that:
```
    Note that this requires all TOAs to be measured at a single observatory.  If you
    wish to combine multiple data-sets use :func:`pint.toa.merge_TOAs`
```
But I will test for spacecraft.

dlakaplan:
> The `mjd` and `mjd_float` columns getting out of sync is almost certainly _not_ intentional. However, since `mjd_float` is only a 64-bit float, it is only intended for non-precision things (like plotting). So it probably isn't a big deal (but I'm glad you are fixing it!). However, if `tdb` and `tdbld` got out of sync like that, I think it _would_ be a problem....

Yes, the `mjd_float` is mostly just for convenience/output with `get_mjds()`.  But we should try to keep them in sync anyway.  We might think later if adding some syntactical sugar to make modifying the `mjd` column automatically propagate to the `tdb` would be possible/useful.

docs/user-questions.rst
dlakaplan: OK, I have clarified that.

src/pint/toa.py
dlakaplan: Spacecraft work fine. I put in a test and added it to the docs.

dlakaplan: Actually, since this works I simplified the Fermi workflow:
```
t, obs, energies, weights = load_Fermi_Times(eventfile, weightcolumn="PSRJ0030+0451")
toas = toa.get_TOAs_array(t, obs, include_gps=False,
                include_bipm=False, planets=False, ephem="DE405",
                flags=[{"energy": str(e), "weight": str(w)} for e, w in zip(energies.to_value(u.MeV), weights)],
                )
```
It's not hugely different but could be slightly more efficient.  If you think this is useful I can replicate this for the other functions in `event_toas.py` (should the Fermi functions maybe be in there?)

dlakaplan:
@paulray : is there a reason why the various event tools return lists of `TOA` objects rather than `TOAs`?

dlakaplan:
I made an updated function even simpler for handling Fermi data.  I don't know that 2 steps are needed (although please correct me if I'm wrong), especially since filtering the TOA table is relatively easy now.  So you can just do:
```

    toas2 = get_Fermi_TOAs(
        eventfileraw,
        weightcolumn="PSRJ0030+0451",
        planets=False,
        ephem="DE405",
    )
```
I assumed that `it was OK to hard-code `include_gps` and `include_bipm` since these wouldn't be referenced to terrestrial clocks.  If that's incorrect it's easy to make those options.

paulray:
There are a few things I'm worried about with the Time array initialization.

What happens to the `location` attribute of the `Time`? Do it get overwritten?  Should we check to make sure it is `None`?

In a quick test, I made a Fermi observatory, and I see this:
```
In [24]: site.get_gcrs(times[0])
Out[24]: <Quantity [-6698526.87266797,  1018499.15531248,  1343133.70272975] m>

In [25]: site.earth_location_itrf(times[0])
Out[25]: <EarthLocation (0., 0., 0.) m>
```
So, in the code for `get_TOAs_array`, which calls `earth_location_itrf`, how does it get the right locations for each of the times?

Maybe it all works, but a look at the code has me confused.

dlakaplan:
What I see is that the location of the TOAs for a spacecraft doesn't matter.  When I construct a list of individual TOA objects using the old method I see:
```
In [22]: print(tl[0].mjd.location)
(0., 0., 0.) m
```
so that's what I copied.  It looks like (in `satellite_obs.py`) that the `get_gcrs()` and `get_TDB()` method get overloaded and ignore the TOA locations.  Instead they just interpolate the satellite position to do the corrections. e.g., `get_gcrs()` does:
```
return (
            np.array([self.X(t.tt.mjd), self.Y(t.tt.mjd), self.Z(t.tt.mjd)])
            * self.FT2["X"].unit
        )
```
I made a new `Time` object with the same MJD but a `location` that differed by 1e8 m.  The GCRS and TDB were the same:
```
In [85]: print(f"{tl[0].mjd.location} {fermi.get_gcrs(tl[0].mjd)} {fermi.get_TDBs(tl[0].mjd,ephem='DE440')}")
(0., 0., 0.) m [5360463.5788544  3636868.92423346 2393578.14363066] m [56876.47375201]

In [86]: print(f"{t2.location} {fermi.get_gcrs(t2)} {fermi.get_TDBs(t2,ephem='DE440')}")
(1.e+08, 1.e+08, 1.e+08) m [5360463.5788544  3636868.92423346 2393578.14363066] m [2456876.97375201]
```
Does this make sense?  I didn't write any of this - I was just copying what I could infer.

dlakaplan:
I was looking a bit more at the `compute_TDBs()` method.  I think there were some legacy work-arounds in there dealing with `astropy` `Time`'s, and hopefully they are not all necessary now.  I tried some local tests and they were OK, but we'll see how the rest of the test-suite does

paulray:
I'm sorry to have been slow to review this. I want to run some tests with satellite data but I'm swamped and going on travel for the rest of this week so I'll miss the telecon tomorrow.

dlakaplan:
> I'm sorry to have been slow to review this. I want to run some tests with satellite data but I'm swamped and going on travel for the rest of this week so I'll miss the telecon tomorrow.

@paulray : let me know if there are any tests that I can help with.  Or maybe @kerrm can provide some test data.

kerrm:
I wanted to point out that there is a test with Fermi data already: test_fermiphase.py.  It will try to compute the phase of a small set of photons in an FT1 file, and compares it to a column of pre-computed phases from tempo2.  Is that helpful?

dlakaplan:
Yes, it already passes that test.  I also added a test that compares the new methods with those used there:
`tests/test_toa_create.py::def test_toas_fermi`

paulray:
Are those Fermi data barycentered or geocentered? If so, then it doesn't test the moving observatory situation

dlakaplan:
Neither, I think, since it loads an observatory position file (I think):
```
    eventfileraw = datadir / "J0030+0451_w323_ft1weights.fits"
    ft2file = datadir / "lat_spacecraft_weekly_w323_p202_v001.fits"
    get_satellite_observatory("Fermi", ft2file, overwrite=True)
```

paulray:
Oh that is good!

dlakaplan:
And `tests/test_toa_create.py::test_toas_fermi_notoalist` does the same test but using the new wrapper I made (`get_Fermi_TOAs()`).

Is that actually testing the barycentering?

paulray:
OK, I looked over the code and I think it is good. That Fermi test shows that it is working correctly. I won't be on the call tomorrow, but I'm ok with this being merged.  Some external codes may stop working if they called the old `load_Fermi_TOAs()` but that is a small change and will result in some speedup.

PR #1460, opened 2022-12-02
ELL1k timing model

abhisrkckl:
ELL1k is a modified version of the ELL1 model applicable to short-orbital period binaries where the periastron advance timescale is comparable to the data span. 
Ref: https://arxiv.org/abs/1808.02508
tempo2 implementation: https://bitbucket.org/psrsoft/tempo2/src/master/ELL1kmodel.C 

EPS1 and EPS2 are the rectangular components of the Laplace-Runge-Lenz vector. The effect of periastron advance is to rotate this vector as a function of time. When the periastron advance timescale is much longer than the data span, this rotation can be expressed as a linear approximation using EPS1DOT and EPS2DOT (this is valid for the vast majority of binary pulsars using ELL1). However, when the periastron advance timescale is of the order of or shorter than the data span, this linear approximation breaks down. ELL1k parametrizes this exact evolution of EPS1 and EPS2 using OMDOT and LNEDOT := EDOT/ECC. 

Also includes some sourcery refactoring and typo fixes.

PR #1514, opened 2023-01-25
WIP: Standardize citations in code

abhisrkckl:
The citation conventions are all over the place right now. This PR tries to standardize them like

```
Lorimer and Kramer [lk2004_]

.. _lk2004: https://ui.adsabs.harvard.edu/abs/2004hpa..book.....L/abstract
```

This follows #1460 and #1490.


dlakaplan:
I think you need to check the RTD: not all of the links are rendering correctly.

dlakaplan:
Personally I like to be a little more explicit about the references:
```
- Taylor and Weisberg (1989), ApJ, 345, 434 [tw89]_

    .. [tw89] https://ui.adsabs.harvard.edu/abs/1989ApJ...345..434T/abstract
```
so that it gives both the actual info and a link to ADS (in case somebody has a local copy etc, they shouldn't be forced to go to ADS to see what the paper is).

dlakaplan:
We could also consider https://www.sphinx-doc.org/en/master/usage/extensions/extlinks.html

abhisrkckl:
This PR is stale. So I am closing it.

Issue #1480, opened 2022-12-20
Tag-specific FD parameters


abhisrkckl:
Since different instruments may be using different templates, we may need multiple FD parameters for different telescopes, backends etc. 

One way to implement this will be as a PrefixMaskParameter which has properties of both prefix and mask parameters.

rossjjennings:
As discussed at the PINT meeting today, a similar, but not necessarily directly related, issue is #1059 (Mask parameters have unhelpful names).

PR #1515, opened 2023-01-26
WIP: System-dependent FD parameters (FDJUMP)

abhisrkckl:
This implements #1480. I have not really implemented anything yet. This WIP PR is to call dibs and invite discussion.

System-specific FD parameters are required for some telescopes (e.g. MeerKat) in the IPTA DR3. This will work like normal FD parameters, but with masks. These are called FDJUMPs in tempo2.
```
FDJUMP1 tel gbt 1e-6 1
FDJUMP1 tel pks 1.1e-6 1
FDJUMP2 tel gbt 1e-7 1
```

The tempo2 implementation of FDJUMP is subtly different from FD in the following way:
```
FD_delay = np.polyval(FD_coeff, np.log(freq_in_GHz))
FDJUMP_delay = mask * np.polyval(FD_coeff, freq_in_GHz)
```
i.e., FDJUMP is computed using frequency rather than log frequency.

I'll also implement a boolean parameter FDJUMPLOG (the name is up for discussion) which will switch between using frequency or log-frequency. The default will be to use frequency to be compatible with tempo2.

I hope to finish this in time to be useful for DR3.


abhisrkckl:
@blarsen10 Does the above address your requirement?

blarsen10:
Thanks for volunteering for this @abhisrkckl! This seems to me like it should cover what we need. Happy to test this once you have it ready.

PR #1517, opened 2023-02-01
Simplify pyerfa use 

aarchiba:
This ensures that `pyerfa` is actually installed, and it drops the `astropy._erfa` (which has disappeared in the most recent version).


aarchiba:
Already fixed in master; I was using an old copy.

PR #1519, opened 2023-02-02
Allow docs build without examples (much faster)

dlakaplan:
Standard RTD builds take a long time.  About 6 min on my system.  Much of that time is spent running the example notebooks.  This gives a new build target for the docs that excludes them.  It takes about 90s on my system:
```
(pintdev) kaplan@plock[~/pythonpackages/PINT_dlakaplan/docs] (docsnoexamples) % make cleaner; time make html >& ../html.out
rm -rf _build/*
rm -rf _autosummary/*.rst
make html >&../html.out  367.42s user 15.96s system 96% cpu 6:35.46 total
```
vs.
```
(pintdev) kaplan@plock[~/pythonpackages/PINT_dlakaplan/docs] (docsnoexamples) % make cleaner; time make htmlnoexamples >& ../htmlnoexamples.out
rm -rf _build/*
rm -rf _autosummary/*.rst
make htmlnoexamples >&../htmlnoexamples.out  78.49s user 10.44s system 89% cpu 1:39.79 total
```
So you would run `make htmlnoexamples`

I also added `cleaner` which removes not only the files in `_build/` (done by standard `make clean`) but `_autosummary/`, since those may need to be regenerated when modules are added/removed.  Note that this directory is hard-coded since I couldn't find that it was set at the top level.

Finally, I updated the version of `sphinx-rtd-theme` so that search seems to work again,  and added the logo (back?) to the RTD page.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,7 +16,7 @@ formats:
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.8
+  version: 3.11
   install:
     - requirements: requirements_dev.txt
     - requirements: requirements.txt
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -16,7 +16,7 @@ formats:
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.11
+  version: 3.8
   install:
     - requirements: requirements_dev.txt
     - requirements: requirements.txt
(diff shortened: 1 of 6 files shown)


aarchiba:
What does the documentation look like with this option? Does the tutorials section just show as blank? Dangling links? It would be nice to put in a message so we know what happened if we end up with a version generated with this flag.

dlakaplan:
No broken links, just none of the tutorials that need building (although if you have the `.ipynb` files still lying around it won't delete them, so then they will show up):
Screen Shot 2023-02-02 at 2 00 52 PM

abhisrkckl:
The PR looks good to me. Is it ready to go?

dlakaplan:
Yes

PR #1521, opened 2023-02-02
Increase PLDMNoise test tolerance

abhisrkckl:
This fixes the issue where `test_PLRedNoise_recovery` in `test_pldmnoise.py` keeps failing in oldestdeps.
`test_PLRedNoise_recovery` tests the equivalency of `PLRedNoise` and `PLDMNoise` for a monochromatic dataset.
I am just increasing the tolerance of this test.

@blarsen10 Is there any reason why we should not do this?


blarsen10:
No problem with me - I think I just set it to an arbitrary small number. The change you put here seems reasonable. If tests still fail with a tolerance of 1e-6 though I might need to investigate...

Issue #1520, opened 2023-02-02
Black version changed


dlakaplan:
The release of `black` is now 23.1.0, which is causing our CI to fail. It looks like we only require `black>=22.0`.  Should we update that to `23.0`?  Or pin it to 22?

PR #1522, opened 2023-02-02
Update black version

abhisrkckl:
Update black version from 22.x to 23.x and re-blacken the package.

#1520

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,6 +6,6 @@ repos:
     -   id: check-merge-conflict
     -   id: check-symlinks
 -   repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 23.1.0
     hooks:
     -   id: black
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v2.3.0
+    rev: v4.4.0
     hooks:
     -   id: check-yaml
     -   id: check-merge-conflict
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -30,7 +30,7 @@ tox
 pre-commit
 typed-ast>=1.5.0
 #black>=19.0a0,<20.0a0
-black>=23.0
+black~=23.0
 pygments
 ipython
 pathlib2
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -30,7 +30,7 @@ tox
 pre-commit
 typed-ast>=1.5.0
 #black>=19.0a0,<20.0a0
-black>=22.0
+black>=23.0
 pygments
 ipython
 pathlib2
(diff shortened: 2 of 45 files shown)


.pre-commit-config.yaml
```diff
@@ -6,6 +6,6 @@ repos:
     -   id: check-merge-conflict
     -   id: check-symlinks
 -   repo: https://github.com/psf/black
-    rev: 22.3.0
+    rev: 23.1.0
```
aarchiba, line 9: According to the black stability policy, any version 23.*.* should be fine, and this allows bug fixes: https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy

requirements_dev.txt
```diff
@@ -30,7 +30,7 @@ tox
 pre-commit
 typed-ast>=1.5.0
 #black>=19.0a0,<20.0a0
-black>=22.0
+black>=23.0
```
aarchiba, line 33: This should probably be black 23.*.* - not 24.* or everything will break when that version comes out (again).

src/pint/fitter.py
```diff
@@ -135,7 +135,9 @@ def __get__(self, instance, owner=None):
                 )
             try:
                 cache = instance.__dict__
-            except AttributeError:  # not all objects have __dict__ (e.g. class defines slots)
+            except (
+                AttributeError
+            ):  # not all objects have __dict__ (e.g. class defines slots)
```
aarchiba, line 140: Would it help to put the comment on a new line to avoid the ugly bracketing?

.pre-commit-config.yaml
abhisrkckl: Should this be changed to "rev: 23"?

src/pint/fitter.py
abhisrkckl: Yes. I'll do that.

tox.ini
```diff
@@ -125,7 +125,7 @@ skip_install = true
 changedir = .
 description = use black
 deps =
-    black>=22.0
+    black>=23.0
```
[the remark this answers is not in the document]
abhisrkckl: changed to black ~= 23.0

requirements_dev.txt
abhisrkckl: changed to black ~= 23.0

.pre-commit-config.yaml
abhisrkckl: It doesn't work without the full version number.

aarchiba:
Does this need a Changelog entry?

It might be helpful to add a section at the end of CONTRIBUTING.rst listing known, regular maintenance/development issues. I'd suggest a note about fixing this every year, and maybe also something about adding new observatories or clock files for existing observatories. Maybe also some notes about how to release a new version - I seem to recall there was some nuance there? Updating the Changelog, perhaps? I seem to recall discussing that with @paulray .

abhisrkckl:
> Does this need a Changelog entry?
Added entry.

> It might be helpful to add a section at the end of CONTRIBUTING.rst listing known, regular maintenance/development issues.
I think this can be a separate PR and it needs some discussion, IMO.

PR #1527, opened 2023-02-06
Fix PLDM test error 

dlakaplan:
Adding try/except to fit so that if the fit doesn't move, it doesn't raise exception.

Not sure why this only seems to happen in oldestdeps.


abhisrkckl:
This looks ready to go.

dlakaplan:
Yes, I think merge when you are happy.

PR #1531, opened 2023-02-13
TCB to TDB conversion

abhisrkckl:
This PR provides a TCB to TDB conversion tool (`tcb2tdb`) which works like the `transform` plugin of `tempo2`.

### DIFF
--- a/src/pint/models/model_builder.py
+++ b/src/pint/models/model_builder.py
@@ -116,7 +116,7 @@ class ModelBuilder:
             p_line = " ".join([k] + v)
             warnings.warn(f"Unrecognized parfile line '{p_line}'", UserWarning)
             # log.warning(f"Unrecognized parfile line '{p_line}'")
-        
+
         return tm
 
     def _validate_components(self):
--- a/src/pint/models/model_builder.py
+++ b/src/pint/models/model_builder.py
@@ -73,7 +73,7 @@ class ModelBuilder:
         self._validate_components()
         self.default_components = ["SolarSystemShapiro"]
 
-    def __call__(self, parfile, allow_name_mixing=False):
+    def __call__(self, parfile, allow_name_mixing=False, allow_tcb=False):
         """Callable object for making a timing model from .par file.
 
         Parameters
@@ -103,7 +103,14 @@ class ModelBuilder:
         # Make timing model
         cps = [self.all_components.components[c] for c in selected]
         tm = TimingModel(components=cps)
-        self._setup_model(tm, pint_param_dict, original_name, setup=True, validate=True)
+        self._setup_model(
+            tm,
+            pint_param_dict,
+            original_name,
+            setup=True,
+            validate=True,
+            allow_tcb=allow_tcb,
+        )
         # Report unknown line
         for k, v in unknown_param.items():
             p_line = " ".join([k] + v)
@@ -418,6 +425,7 @@ class ModelBuilder:
         original_name=None,
         setup=True,
         validate=True,
+        allow_tcb=False,
     ):
         """Fill up a timing model with parameter values and then setup the model.
 
@@ -529,7 +537,7 @@ class ModelBuilder:
         if setup:
             timing_model.setup()
         if validate:
-            timing_model.validate()
+            timing_model.validate(allow_tcb=allow_tcb)
         return timing_model
 
     def _report_conflict(self, conflict_graph):
@@ -543,7 +551,7 @@ class ModelBuilder:
             )
 
 
-def get_model(parfile, allow_name_mixing=False):
+def get_model(parfile, allow_name_mixing=False, allow_tcb=False):
     """A one step function to build model from a parfile.
 
     Parameters
@@ -570,11 +578,11 @@ def get_model(parfile, allow_name_mixing=False):
         # # parfile is a filename and can be handled by ModelBuilder
         # if _model_builder is None:
         #     _model_builder = ModelBuilder()
-        model = model_builder(parfile, allow_name_mixing)
+        model = model_builder(parfile, allow_name_mixing, allow_tcb=allow_tcb)
         model.name = parfile
         return model
     else:
-        tm = model_builder(StringIO(contents), allow_name_mixing)
+        tm = model_builder(StringIO(contents), allow_name_mixing, allow_tcb=allow_tcb)
         return tm
 
 
--- a/src/pint/models/model_builder.py
+++ b/src/pint/models/model_builder.py
@@ -116,6 +116,7 @@ class ModelBuilder:
             p_line = " ".join([k] + v)
             warnings.warn(f"Unrecognized parfile line '{p_line}'", UserWarning)
             # log.warning(f"Unrecognized parfile line '{p_line}'")
+        
         return tm
 
     def _validate_components(self):
@@ -600,6 +601,7 @@ def get_model_and_toas(
     picklefilename=None,
     allow_name_mixing=False,
     limits="warn",
+    allow_tcb=False,
 ):
     """Load a timing model and a related TOAs, using model commands as needed
 
@@ -643,7 +645,7 @@ def get_model_and_toas(
     -------
     A tuple with (model instance, TOAs instance)
     """
-    mm = get_model(parfile, allow_name_mixing)
+    mm = get_model(parfile, allow_name_mixing, allow_tcb=allow_tcb)
     tt = get_TOAs(
         timfile,
         include_pn=include_pn,
--- a/src/pint/models/tcb_conversion.py
+++ b/src/pint/models/tcb_conversion.py
@@ -1,0 +1,8 @@
+import numpy as np
+
+# These constants are taken from Irwin & Fukushima 1999.
+# These are the same as the constants used in tempo2 as of 10 Feb 2023.
+IFTE_MJD0 = np.longdouble("43144.0003725")
+IFTE_KM1 = np.longdouble("1.55051979176e-8")
+IFTE_K = 1 + IFTE_KM1
+IFTE_F = 1/(1 + IFTE_KM1)
--- a/src/pint/models/tcb_conversion.py
+++ b/src/pint/models/tcb_conversion.py
@@ -9,7 +9,7 @@ IFTE_K = 1 + IFTE_KM1
 
 def scale_parameter(model, param, n):
     factor = IFTE_K**n
-    
+
     if hasattr(model, param) and getattr(model, param).quantity is not None:
         par = getattr(model, param)
         par.value *= factor
@@ -18,14 +18,15 @@ def scale_parameter(model, param, n):
 
 
 def transform_mjd_parameter(model, param):
-    factor = 1/IFTE_K
+    factor = 1 / IFTE_K
     tref = IFTE_MJD0
-    
+
     if hasattr(model, param) and getattr(model, param).quantity is not None:
         par = getattr(model, param)
-        par.value = (par.value - tref)*factor + tref
+        par.value = (par.value - tref) * factor + tref
         if par.uncertainty_value is not None:
             par.uncertainty_value *= factor
+
 
 def convert_tcb_to_tdb(model):
     """This function performs a partial conversion of a model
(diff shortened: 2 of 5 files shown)


dlakaplan:
Which of your list of extra parameters is done by the tempo2 transform?

Is a reverse transform needed?  I know PINT stays in TDB, but producing TCB could be useful for comparison elsewhere.

Is a test possible comparing the results to the tempo2 transformation result?

abhisrkckl:
> Which of your list of extra parameters is done by the tempo2 transform?

Tempo2 transform converts DM and DM derivatives. The reason why I did not implement this is that I was confused by what the transform plugin was doing. Namely, tempo2 was using a conversion factor IFTE_K^-3 instead of IFTE_K for DM as one would expect based on the dimensions of DMconst*DM. I discussed this with Mike Keith and he agrees that this is a mistake, and he has pushed a correction to the tempo2 repo. I'll implement the same in this PR.

There is also a chromatic measure (CM) that gets converted, but that is not implemented in PINT.

Tempo2 also converts *some* of the post-Keplerian parameters, but not others. I am not sure what the logic is there. I don't think this will make much of a difference either way, since we are only intending the conversion to be approximate. I'll do some simulations and confirm this.

> Is a reverse transform needed? I know PINT stays in TDB, but producing TCB could be useful for comparison elsewhere.

I am not sure about this. I'll pose this question in the meeting. In any case, this is pretty easy to implement.

> Is a test possible comparing the results to the tempo2 transformation result?

I think so. Is `tempo2Test` the right place to put such a test?

src/pint/scripts/tcb2tdb.py
```diff
@@ -0,0 +1,55 @@
...
+        9. Red noise parameters including FITWAVES, powerlaw red noise and 
+           powerlaw DM noise parameters
+    """
+    log.info(welcome_message)
+
+    model = get_model(args.input_par, allow_tcb=True)
+    convert_tcb_to_tdb(model)
+    model.write_parfile(args.output_par)
+
+    log.info(f"Output written to {args.output_par}.")
```
paulray, line 55: I suggest changing this to `log.warning` and including words about how the output is only approximately correct. I know that is in the help, but I think a user should see that every time it runs.

abhisrkckl: There is a "welcome_message" that is shown every time the script runs, and has this information. I have changed that to a warning.

dlakaplan:
I don't think the full welcome message should be a warning.  But there should be a warning when just using the `convert_tcb_to_tdb` function.

src/pint/scripts/tcb2tdb.py
```diff
@@ -0,0 +1,56 @@
...
+        4. Solar wind parameters
+        5. Binary post-Keplerian parameters including Shapiro delay 
+           parameters
+        6. Jumps and DM Jumps
+        7. FD parameters
+        8. EQUADs
+        9. Red noise parameters including FITWAVES, powerlaw red noise and 
+           powerlaw DM noise parameters
+    """
+    log.warning(welcome_message)
```
dlakaplan, line 50: I think this should only be shown when run with `--help` or similar.  But a separate warning about the accuracy should be part of the conversion method, not the script.

src/pint/models/tcb_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+           powerlaw DM noise parameters
+
+    Parameters
+    ----------
+    model : pint.models.timing_model.TimingModel
+       Timing model to be converted.
+    """
+
+    if model.UNITS in ["TDB", None]:
+        # ~ issue warning here ~
```
dlakaplan, line 81: Where is the warning?

src/pint/models/tcb_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+        scale_parameter(model, "PMELONG", IFTE_K)
+        transform_mjd_parameter(model, "POSEPOCH")
+
+    # Although DM has the unit pc/cm^3, the quantity that enters
+    # the timing model is DMconst*DM, which has dimensions
+    # of frequency. Hence, DM and its derivatives will be
+    # scaled by IFTE_K**(i+1).
+    if "DispersionDM" in model.components:
+        scale_parameter(model, "DM", IFTE_K)
+        for i in range(1, len(model.get_DM_terms())):
```
dlakaplan, line 106: This could also be done with `get_prefix_mapping`

src/pint/models/tcb_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+]
+
+# These constants are taken from Irwin & Fukushima 1999.
+# These are the same as the constants used in tempo2 as of 10 Feb 2023.
+IFTE_MJD0 = np.longdouble("43144.0003725")
+IFTE_KM1 = np.longdouble("1.55051979176e-8")
+IFTE_K = 1 + IFTE_KM1
+
+
+def scale_parameter(model, param, n):
```
dlakaplan, line 21: Could you add a docstring?  It should also explain what this will do if the parameter is `None`

src/pint/models/model_builder.py
```diff
@@ -1,3 +1,5 @@
+"""Building a timing model from a par file."""
```
dlakaplan, line 1: Any functional changes here or just sourcery?

abhisrkckl: `ModelBuilder.__call__`, `get_model`, and `get_model_and_toas` functions now have an `allow_tcb` option. When this option is True, it will not throw an error when it encounters a TCB par file and silently reads the par file. The default behavior remains to throw an error for TCB par files.

src/pint/models/tcb_conversion.py
abhisrkckl: Oops forgot to add that.

src/pint/models/tcb_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+    factor = IFTE_K**n
+
+    if hasattr(model, param) and getattr(model, param).quantity is not None:
+        par = getattr(model, param)
+        par.value *= factor
+        if par.uncertainty_value is not None:
+            par.uncertainty_value *= factor
+
+
+def transform_mjd_parameter(model, param):
```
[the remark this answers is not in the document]
abhisrkckl: I have added a more detailed docstring.

This function is for scaling a single MJD parameter.  The logic of which parameters are/aren't changed is not implemented in this function. I intended this function and `scale_parameter` to be utility functions to be called by `convert_tcb_to_tdb`.

src/pint/models/tcb_conversion.py
abhisrkckl: Added more detailed docstring. `param` is supposed to be a string. It will throw an error if it is not a string.

src/pint/models/tcb_conversion.py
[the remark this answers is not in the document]
dlakaplan: Even if you didn't intend it to be used in some way, users might.  So I think having the info here is important.

src/pint/scripts/tcb2tdb.py
abhisrkckl: I have changed this to a help message. 

A brief warning will be emitted from `convert_tcb_to_tdb` function every time it is called.

src/pint/models/tcb_conversion.py
abhisrkckl: Added warning.

src/pint/models/tcb_conversion.py
```diff
@@ -0,0 +1,119 @@
...
+    ----------
+    model : pint.models.timing_model.TimingModel
+       Timing model to be converted.
+    """
+
+    if model.UNITS in ["TDB", None]:
+        # ~ issue warning here ~
+        return
+
+    if "Spindown" in model.components:
```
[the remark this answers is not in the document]
abhisrkckl: Changed.

src/pint/models/model_builder.py
abhisrkckl: `ModelBuilder.__call__` will emit a warning about this,

aarchiba: Wait, if you set `allow_tcb` you get a wrong model? Shouldn't it do the conversion?

abhisrkckl:
Based on today's discussion, I have implemented TCB->TDB conversion on read. Both `get_model` and `get_model_and_toas` functions not take an `allow_tcb` option. This is False by default and will throw an error upon encountering a TCB par file. If True, this will read the TCB par file and convert it to TDB. If "raw", this will return an unconverted malformed TimingModel object containing TCB parameters.

src/pint/models/tcb_conversion.py
```diff
@@ -0,0 +1,168 @@
...
+# These constants are taken from Irwin & Fukushima 1999.
+# These are the same as the constants used in tempo2 as of 10 Feb 2023.
+IFTE_MJD0 = np.longdouble("43144.0003725")
+IFTE_KM1 = np.longdouble("1.55051979176e-8")
+IFTE_K = 1 + IFTE_KM1
+
+
+def scale_parameter(model, param, n, backwards):
+    """Scale a parameter x by a power of IFTE_K
+        x_tdb = x_tdb * IFTE_K**n
```
bwmeyers, line 24: I'm quite possibly misunderstanding, but should one of the `x_tdb` instead be `x_tcb`?

src/pint/models/tcb_conversion.py
```diff
@@ -0,0 +1,168 @@
...
+        transform_mjd_parameter(model, "PEPOCH", backwards)
+
+    if "AstrometryEquatorial" in model.components:
+        scale_parameter(model, "PMRA", 1, backwards)
+        scale_parameter(model, "PMDEC", 1, backwards)
+        transform_mjd_parameter(model, "POSEPOCH", backwards)
+    elif "AstrometryEcliptic" in model.components:
+        scale_parameter(model, "PMELAT", 1, backwards)
+        scale_parameter(model, "PMELONG", 1, backwards)
+        transform_mjd_parameter(model, "POSEPOCH", backwards)
```
bwmeyers, line 146: Is there a reason to have the `transform_mjd_parameter(model, "POSEPOCH", backwards)` independently in each clause, rather than just run it after the `if/elif` block finishes? Just reduces repeated code (I think?).

src/pint/models/tcb_conversion.py
bwmeyers: Resolved.

src/pint/models/tcb_conversion.py
bwmeyers: Resolved.

paulray:
Just FYI, I just pulled this and tried it on a TCB par file and it seemed to work. However, the last warning was a little odd because it is not true. The input par file was TCB:
```
% tcb2tdb J0659+1414_NRT_20190322.par J0659+1414_NRT_20190322_TDB.par          

WARNING  (pint.logging                  ): /Users/paulr/src/PINT/src/pint/models/timing_model.py:367 UserWarning: PINT does not support 'DILATEFREQ Y'
WARNING  (pint.logging                  ): /Users/paulr/src/PINT/src/pint/models/timing_model.py:370 UserWarning: PINT only supports 'TIMEEPH FB90'
WARNING  (pint.models.timing_model      ): PINT does not support 'UNITS TCB' internally. Reading this par file neverthelessbecause the `allow_tcb` option was given. This `TimingModel` object should not beused for anything except converting to TDB.
WARNING  (pint.logging                  ): /Users/paulr/src/PINT/src/pint/models/timing_model.py:399 UserWarning: START cannot be unfrozen... Setting START.frozen to True
WARNING  (pint.logging                  ): /Users/paulr/src/PINT/src/pint/models/timing_model.py:402 UserWarning: FINISH cannot be unfrozen... Setting FINISH.frozen to True
WARNING  (pint.logging                  ): /Users/paulr/src/PINT/src/pint/models/model_builder.py:130 UserWarning: Unrecognized parfile line 'EPHVER 5'
WARNING  (pint.models.tcb_conversion    ): The TCB to TDB conversion is only approximate. The resulting timing model should be re-fit to get reliable results.
WARNING  (pint.models.tcb_conversion    ): The input par file is already in the target units. Doing nothing.
INFO     (pint.scripts.tcb2tdb          ): Output written to J0659+1414_NRT_20190322_TDB.par.
```

paulray:
I notice that the roundtrip test is failing.

dlakaplan:
@abhisrkckl : can you share the log messages like the ones @paulray showed above?  It would be good to see what they look like if they have changed at all.

abhisrkckl:
@dlakaplan I have fixed this. The error message now looks like this:

> $ tcb2tdb test.par  test_tdb.par
> WARNING  (pint.models.timing_model      ): PINT does not support 'UNITS TCB' internally. Reading this par file neverthelessbecause the `allow_tcb` option was given. This `TimingModel` object should not beused for anything except converting to TDB.
> WARNING  (pint.models.tcb_conversion    ): The TCB to TDB conversion is only approximate. The resulting timing model should be re-fit to get reliable results.
> INFO     (pint.scripts.tcb2tdb          ): Output written to test_tdb.par.
>

abhisrkckl:
Improved log messages from `tcb2tdb`:

> $ tcb2tdb test.par  test_tdb.par
> WARNING  (pint.models.timing_model      ): PINT does not support 'UNITS TCB' internally. Reading this par file nevertheless because the `allow_tcb` option was given. This `TimingModel` object should not be used for anything except converting to TDB.
> WARNING  (pint.models.tcb_conversion    ): Converting this timing model from TCB to TDB. Please note that the TCB to TDB conversion is only approximate and the resulting timing model should be re-fit to get reliable results.
> INFO     (pint.scripts.tcb2tdb          ): Output written to test_tdb.par.

abhisrkckl:
I have also updated the Explanation page to reflect these changes.

paulray:
I'd like to merge this now. Is that fine?

Issue #1430, opened 2022-11-02
Visualize DM model and/or DM residuals for wideband data (using pintk?)


abhisrkckl:
This will be a very useful tool to have when one is fitting wideband data.

PR #1534, opened 2023-02-14
Plot wideband DMs and DM residuals in `pintk`

abhisrkckl:
Implements #1430

Three new plotting options: WB DM, WB DM res, WB DM err. These are disabled for narrowband data.

Also fixes a bug introduced in #1663, and fixes a broken import.


abhisrkckl:
DM Measurements:
image

DM Residuals:
image

DM Errors:
image

These new radio buttons are disabled for narrowband data.
image

src/pint/residuals.py
```diff
@@ -370,11 +370,17 @@ def calc_phase_resids(
             # and delta_pulse_numbers (from PHASE lines or adding phase jumps in GUI)
             i = pulse_num.copy()
             f = np.zeros_like(pulse_num)
-            if np.any(np.isnan(pulse_num)):
+
+            c = np.isnan(pulse_num)
+            if np.any(c):
```
dlakaplan, line 375: What is the purpose of this?  Won't the first exception mean you never get to the later code?

abhisrkckl: I presume you are referring to the `if pulse_num is None` clause.
That clause checks if the pulse number column is missing altogether from the TOAs object, and raises an exception immediately if it is not found.

The `if np.any(c)` clause is checking if any of the pulse numbers are `nan` when the column itself is present.

dlakaplan: right.  but then you have another `if np.any(c)` down below (L381) but it would raise an exception here.

abhisrkckl: Ah! that's not required. I have removed it.

PR #1539, opened 2023-02-23
Clean up test_erfautils

aarchiba:
This is meant to address some bugs that used to be in Astropy that might have affected us; these tests can probably all be deleted once our oldest supported version includes fixes for all of these. (Which may be now; the CI on this PR should show us.)


aarchiba:
Looks like one of the tests that passes now fails on oldestdeps (probably Astropy <5).

PR #1542, opened 2023-03-10
Apply DMJUMP time delay for narrowband timing

blarsen10:
I was interested in correcting some DM offsets particular to certain datasets within the IPTA DR3 data combination, which can apparently result if a different initial DM value was used when fitting ToAs using frequency-resolved pulse portraits. It seemed that the `DispersionJump` class would be perfect for correcting this, except it looks like it's only set up to apply jumps to just the DM value used in wideband timing, but not to the time delay/residuals.

The following changes will apply the time delay from DMJUMPs given narrowband toas, but not apply any delay to wideband toas as originally set up. Let me know if this seems sensible or if anyone has suggestions for a better way to do this.

Here's also a comparison of some of the affected residuals from the DR3 data combination before and after the inclusion of a DMJUMP fit into the model to show the delay is being applied correctly:

Before:
image

After:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project, at least loosely, adheres to [Semantic Versioning](https://sem
 - Moved `DMconst` from `pint.models.dispersion_model` to `pint` to avoid circular imports
 - Removed references to `astropy._erfa` (removed since `astropy` 4.2)
 - Refactor `Dre` method, fix expressions for Einstein delay and post-Keplerian parameters in DD model
+- Apply time delays from `DispersionJump` class when using narrowband ToAs
 - Updated contributor list (AUTHORS.rst)
 - Emit an informative warning for "MODE" statement in TOA file; Ignore "MODE 1" silently
 - Version of `sphinx-rtd-theme` updated in `requirements_dev.txt` 
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -757,7 +757,7 @@ class DispersionJump(Dispersion):
             )
         )
 
-        self.delay_funcs_component += [self.dmjump_dispersion_delay]
+        # self.delay_funcs_component += [self.dmjump_dispersion_delay]
 
     def setup(self):
         super().setup()
@@ -805,7 +805,7 @@ class DispersionJump(Dispersion):
         No delay for wideband timing.
         """
         if toas.is_wideband():
-            return []
+            return np.zeros(toas.ntoas) * u.s
         else:
             return self.dispersion_type_delay(toas)
 
--- a/src/pint/models/dispersion_model.py
+++ b/src/pint/models/dispersion_model.py
@@ -736,8 +736,9 @@ class DispersionJump(Dispersion):
 
     Notes
     -----
-    This DM jump is only for modeling the DM values, and will not apply to the
-    dispersion time delay.
+    For wideband timing, this DM jump is only for modeling the DM values, and
+    will not apply to the dispersion time delay. For narrowband timing, this
+    will apply the usual dispersion time delay.
     """
 
     register = True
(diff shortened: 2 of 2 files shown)


abhisrkckl:
Is this behavior of DMJUMP the same in tempo and tempo2?

abhisrkckl:
Will you be able to come to today's PINT meeting to discuss this?

blarsen10:
Yes

blarsen10:
I plan on changing this following @scottransom's suggestion to add a flag to set whether or not a delay is applied.

pennucci:
Hey.  So as far as I know, DMJUMP never made it into tempo2.  I am aware 
of its status in tempo, enterprise, and to a lesser degree pint.  I do 
not think DMJUMP is used in calculating any delay to TOAs in enterprise 
but it has been a long while since I looked at this.  Its effect can 
change residuals, of course.

blarsen10:
Hi @pennucci thank you for the input! Based on your comment on slack, do you think a system-dependent FD parameter should instead be used for this type of case? If so then there's no need to change anything in the `DispersionJump` class?

dlakaplan:
It sounds like based on #1716 that this is still needed, and provides the majority of the functionality.  @rossjjennings : is that correct?  If so, it might only need a test to make it mergable.

rossjjennings:
@dlakaplan: As far as I can tell, that's correct, but I haven't successfully run a fit with the changes in this PR yet. I tried it on an IPTA pulsar at one point and it didn't seem to be fitting correctly, but that might have been another issue with the model I was using.

abhisrkckl:
The solution implemented here is the right one for narrowband TOAs, but I suspect it will break the wideband analysis. The DMJUMPs as they are implemented in PINT currently have been used for wideband analysis (most notably NG12.5), and changing its behavior to produce a delay will break the existing wideband analysis. 

Since this is breaking backwards compatibility, I am strongly against merging this.

So my suggestion would be to leave DMJump alone and implement a similar, but separate and independent component that does the right thing for narrowband TOAs. This way, DMJump remains as it is, and the existing wideband analyses won't break. My suggestion for a name would be something like "DMHop".

rossjjennings:
@abhisrkckl I have to disagree here. First of all, I will point out that Bjorn's current implementation does an explicit check for wideband TOAs and preserves the current behavior of DMJUMP in that case, so it should not actually break backwards compatibility. 

But also, there is no real conceptual difference between the intended wideband and narrowband behavior here that would warrant splitting them up, and having both "DMJump" and "DMHop" with similar, but slightly different, meanings would just be confusing, and serve no real purpose. I would even argue that the current wideband behavior, which never includes delays, is broken -- although the difference shouldn't matter except in a case that AFAIK no one has tried, where a single DMJUMP parameter is applied to a group of wideband TOAs that includes TOAs from two or more different frequency bands. In that case, the DMJUMP should apply a delay between the bands, even in the wideband case.

scottransom:
We just chatted about this a bit on the abbreviated PINT call.  I think it might be worth bringing this to the wider timing group (@tcromartie) to think about -- there are a lot of subtle issues here. I find it hard to even think about whether or which DM JUMPs and/or time delays (fixed or freq dependent) would be needed if two PTAs model the freq-dependent pulse profile distributions differently. This might need some good diagrams and/or simulations. @blarsen10 have you thought about any of that in detail?

blarsen10:
> We just chatted about this a bit on the abbreviated PINT call. I think it might be worth bringing this to the wider timing group (@tcromartie) to think about -- there are a lot of subtle issues here. I find it hard to even think about whether or which DM JUMPs and/or time delays (fixed or freq dependent) would be needed if two PTAs model the freq-dependent pulse profile distributions differently. This might need some good diagrams and/or simulations. @blarsen10 have you thought about any of that in detail?

As you mention, there are a lot of subtleties here, and I don't have a very good understanding of them. I'm not sure how much I can weigh in there. My reasoning for setting this up in the first place was largely based on a comment in [(Miles et al 2023)](https://ui.adsabs.harvard.edu/abs/2023MNRAS.519.3976M/abstract) (the MPTA DR1 paper) which mentioned a fiducial DM was chosen to create the frequency-resolved pulse portraits, and therefore that PTA-specific DMs (and FDs) may be required if combining MeerKAT TOAs with other PTAs. I then tried out the DMJump to correct a frequency-trend in the EDR3 J1909-3744 combination (which was affecting only the MeerKAT residuals), and the DMJump corrected the trend. I don't know whether it's the _right_ thing to do, but I can at least attest that it works in this particular case.

abhisrkckl:
Closing this since the same functionality was implemented in #1731 .

Issue #1541, opened 2023-03-09
Repeated code in fermi_toas.py


paulray:
The functions `load_Fermi_TOAs()` and `get_Fermi_TOAs()` are almost identical, except that `load` returns a list of individual `TOA` objects, while `get` returns a `TOAs` object that is constructed from an array. At minimum, the repeated code should be merged, and perhaps only one of these needs to exist?

dlakaplan:
I think I added `get_Fermi_TOAs()` but wasn't comfortable removing or modifying the existing code in case there was some specific use-case or other aspect that I missed, or in case people had explicit code that depended on it.  It may be easy to just make `load_Fermi_TOAs()` call the other function and then just repackage the output.  But I would not want to delete that function without further consultation.

PR #1543, opened 2023-03-29
Convert TOAs to list of TOA objects; simplify fermi methods

dlakaplan:
`get_TOAs_list` converts a list of `TOA` objects to a `TOAs` object.  Now there is a method `to_TOA_list` that undoes that.  By default it also undoes clock corrections to restore the `TOA` objects to their native state.  A test for this is included in `test_toa_create`

This enables `load_Fermi_TOAs` to not duplicate code (#1541).


dlakaplan:
Older methods are still there, but docs now indicate that you should convert to `TOAs` and that you might look at an alternate method:
https://nanograv-pint--1543.org.readthedocs.build/en/1543/_autosummary/pint.event_toas.load_fits_TOAs.html

PR #1545, opened 2023-03-31
Updated piecewise orbital model

poneill129:
Tried merging these changes into the existing PR but looks like its opened as a new one, apologies!
Tagged reviewers and commenters so you don't miss it. (@scottransom @aarchiba @dlakaplan)

Had a chance to address Anne's comments on the piecewise orbital model. 

I will add the responses to the comments in the original PR #1378 as well as here in a follow-up comment.


dlakaplan:
Can you share in the docstring and/or here how such a model is defined?

poneill129:
> This looks great, @poneill129. Once the docstring is added to as requested by David K, and the minor conflict in `src/pint/models/__init__.py` is fixed, do you think this is basically ready to merge? My student is wanting to test/use it and it would be better to do that (I think) if it is merged than trying to work in a side branch.

Hi, the docstring turned into a more involved problem than anticipated. Trying to replicate the convention for other standalone binary models I found myself in a rabbit hole.

 As part of the model declaration an arbitrary piece and associated range are added. Then attempting to add a piece directly to the model (not via the pint facing binary_piecewise module) is bugged in that the dict that is used in the update_binary_object() causes an error when adding a piece. 

I will need to find a way of producing the bug to report it, I will emphasise that this does not happen when piecwise model is declared via the wrapper module. Effectively it’s a timing model that I can only make work via the pint interface.

scottransom:
@poneill129 Coming back to this again...  Is the current status still accurate based on your last comment above with an error when you try to add a piece?  If so, might you be able to email me with some example code as to how/when that is happening?  And also, what is your "wrapper" module that allows things to work?  Is there an example usage of that?  I'm going to start playing with this this week as two of my students really do need it starting now-ish.

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
+"""The BT (Blandford & Teukolsky) model.
```
dlakaplan, line 1: This should state that this is a piecewise version

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+from pint import GMsun, Tsun, ls
+from pint.models.parameter import floatParameter
+from pint.models.pulsar_binary import PulsarBinary
+from pint.models.stand_alone_psr_binaries.BT_model import BTmodel
+from pint.models.timing_model import MissingParameter, TimingModel
+from pint.models.stand_alone_psr_binaries.BT_piecewise import BTpiecewise
+
+
+class BinaryBTPiecewise(PulsarBinary):
+    """Model implemenring the BT model.
```
dlakaplan, line 24: And this docstring should also be updated

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+
+    This is a PINT pulsar binary BT model class a subclass of PulsarBinary.
+    It is a wrapper for stand alone BTmodel class defined in
+    ./stand_alone_psr_binary/BT_model.py
+    All the detailed calculations are in the stand alone BTmodel.
+    The aim for this class is to connect the stand alone binary model with PINT platform
+    BTmodel special parameters:
+    GAMMA Binary Einsten delay coeeficient
```
dlakaplan, line 32: typo

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+                description="Time dilation & gravitational redshift",
+            )
+        )
+        self.A1_value_funcs = []
+        self.T0_value_funcs = []
+        self.remove_param("M2")
+        self.remove_param("SINI")
+        self.T0.value = 1
+        self.A1.value = 1
+        self.add_group_range(50000, 51000, frozen=False, j=0)
```
dlakaplan, line 55: Can this be initialized with `None` instead of random MJDs?

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+        self.add_piecewise_param("T0", 0)
+        self.add_piecewise_param("A1", 0)
+
+    def add_group_range(self, group_start_mjd, group_end_mjd, frozen=True, j=None):
+        """Add an orbital piecewise parameter group range.
+        Parameters
+        ----------
+        group_start_mjd : np.float128
```
dlakaplan, line 63: it would be good to also accept quantities or Times here.  Maybe look at how `DMX` and `SWX` are defined?

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+                raise ValueError(
+                    f"Invalid index for group. Cannot index beyond 9999 (yet?)"
+                )
+
+        i = f"{int(j):04d}"
+        self.add_param(
+            prefixParameter(
+                name="PLB_{0}".format(i),
+                units="MJD",
+                unit_template=lambda x: "MJD",
```
dlakaplan, line 88: I don't think a `lambda` is needed here

dlakaplan:
Hi, @poneill129 .  Great that you made progress.  I started to look at your code but I realize you may have some updates waiting to go.  Should I keep reviewing things now or wait?

poneill129:
Hi @dlakaplan, The proposed change is only very small, affecting only the "toa_belongs_in_group" function in BT_piecewise.py. I'm still trying to pull things together. Thank you for the comments you've already given, if you want to keep going (avoiding toa_belongs_in_groups) I'll pull the changes in while I go!

src/pint/models/__init__.py
```diff
@@ -21,6 +21,7 @@
 # Import all standard model components here
 from pint.models.astrometry import AstrometryEcliptic, AstrometryEquatorial
 from pint.models.binary_bt import BinaryBT
+from pint.models.binary_piecewise import BinaryBTPiecewise
```
dlakaplan, line 24: This is fine, but you could also consider defining `BinaryBTPiecewise` in `binary_bt.py` (like how `dispersion_model.py` has `DispersionDM` and `DispersionDMX`

dlakaplan: (the standalone model piece should be in a separate file)

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+                f"index must be a float, int, list, or array - not {type(index)}"
+            )
+        for index in indices:
+            index_rf = f"{int(index):04d}"
+            for prefix in ["T0X_", "A1X_", "PLB_", "PUB_"]:
+                self.remove_param(prefix + index_rf)
+        self.validate()
+        self.setup()
+
+    def add_piecewise_param(self, param, j, param_unit=None, paramx=None):
```
dlakaplan, line 134: Since there is a small, defined list for this, could you have it work like:
`add_piecewise_params(j, **kwargs)`
and then parse kwargs for `T0=...` and `A1=...`?  And make sure they can accept floats or quantities (using the default units for the given parameters when only floats are passed)?

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+                    value=paramx,
+                    unit_template=lambda x: param_unit,
+                    description="Parameter" + param + "variation",
+                    description_template=lambda x: param,
+                    parameter_type="float",
+                    frozen=False,
+                )
+            )
+
+    def lock_groups(self):
```
dlakaplan, line 198: Is this needed?  Is it used?

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+        XR1_mapping = self.get_prefix_mapping_component("PLB_")
+        XR1s = {}
+        # piecewise parameter ranges PUB-piece upper bound
+        XR2_mapping = self.get_prefix_mapping_component("PUB_")
+        XR2s = {}
+
+        for t0n in T0X_mapping.values():
+            T0Xs[t0n] = getattr(self, t0n).quantity
+
+        for t0_name, t0_value in T0Xs.items():
```
dlakaplan, line 230: would it make more sense to do a single loop over the PLB/PUB mappings?  And then for each index check whether there was a A1X or T1X value and set them if so?

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+            ls_A1X = list(dct_A1X.items())
+
+            j_plb = [((tup[1]).split("_"))[1] for tup in ls_plb]
+            j_pub = [((tup[1]).split("_"))[1] for tup in ls_pub]
+            j_T0X = [((tup[1]).split("_"))[1] for tup in ls_T0X]
+            j_A1X = [((tup[1]).split("_"))[1] for tup in ls_A1X]
+
+            if j_plb != j_pub:
+                raise ValueError(
+                    f"Group boundary mismatch error. Number of detected upper bounds: {j_plb}. Number of detected upper bounds:{j_pub}"
```
dlakaplan, line 319: probably include a space before the values

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+                (length: number of piecewise groups) List of piecewise parameter labels e.g with pieces T0X_0000, T0X_0001, T0X_0003, returns [0,1,3]
+        """
+        group_indexes = []
+        for i in range(0, len(self.binary_instance.piecewise_parameter_information)):
+            group_indexes.append(
+                self.binary_instance.piecewise_parameter_information[i][0]
+            )
+        return group_indexes
+
+    def get_group_indexes_in_four_digit_format(self):
```
dlakaplan, line 392: Is this needed?  If you have the method above to return integers can't you just use:
```
[f"{int(...):04d" for x in ...]
```

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+                (length: number of piecewise groups) List of piecewise parameter labels in 4 digit format e.g with pieces T0X_0000, T0X_0001, T0X_0003, returns [0000,0001,0003]
+        """
+        group_indexes = []
+        for i in range(0, len(self.binary_instance.piecewise_parameter_information)):
+            group_indexes.append(
+                f"{int(self.binary_instance.piecewise_parameter_information[i][0]):04d}"
+            )
+        return group_indexes
+
+    def get_T0Xs_associated_with_toas(self, toas):
```
dlakaplan, line 406: maybe instead of having specialized methods for `TOX` and `A1X`, you can have 1 method to see which group(s) TOAs are associated with?  Then the `TOX` and `A1X` can just be gotten using the existing mappings.

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+        )[1]
+        if temporary_storage is not None:
+            self.binary_instance.group_index_array = temporary_storage
+        return A1X_per_toa
+
+    def does_toa_reference_piecewise_parameter(self, toas, param):
+        """Query whether a TOA/list of TOAs belong(s) to a specific group
+        Parameters
+        ----------
+        toas :
```
dlakaplan, line 454: try to be explicit with the types/class references.  So this would be of type `pint.toa.TOAs`.

dlakaplan:
Hi, @poneill129 .  Overall this looks good, although I think you could simplify some of the smaller functions.  But I notice that you don't have any tests yet.  It would be great to have some tests like:
* make a model with upper/lower ranges for the whole data-set, and compare the binary delays with existing BT
* make a models divided into two ranges.  make 2 BT models and compare appropriately
* test the various setup/adding/removing functions.  Make sure that errors are raised appropriately for the conditions you give (like overlapping ranges)
You might look at some of the tests for `DMX`, `WaveX` or similar to see various examples.

poneill129:
> Hi, @poneill129 . Overall this looks good, although I think you could simplify some of the smaller functions. 
Cheers  again for these, and I'm working my way through them now

>But I notice that you don't have any tests yet. It would be great to have some tests like:
I don't know if it's been missed or I'm misunderstanding but I should have a test file in the form of "test_BT_piecewise.py". It contains tests that might be similar to:
> * make a models divided into two ranges.  make 2 BT models and compare appropriately
> * make a model with upper/lower ranges for the whole data-set, and compare the binary delays with existing BT


I'm fairly sure there's a test in there for testing the error raised when overlapping groups are declared
> * test the various setup/adding/removing functions.  Make sure that errors are raised appropriately for the conditions you give (like overlapping ranges)


Regardless, I'm going to look into the test file again and identify which tests closely resemble what you've outlined, and I'll adapt what is there if I can't find an equivalent test.


>   You might look at some of the tests for `DMX`, `WaveX` or similar to see various examples.
Will do, thanks for the pointer!
Cheers for these again!

dlakaplan:
Great - I missed the tests since the browser didn't show me the code initially.

src/pint/models/binary_piecewise.py
poneill129: Corrected

src/pint/models/binary_piecewise.py
poneill129: Corrected

src/pint/models/binary_piecewise.py
poneill129: corrected

src/pint/models/binary_piecewise.py
poneill129: Changed to accept times and quantities, updated documentation to match

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+
+    def add_group_range(self, group_start_mjd, group_end_mjd, frozen=True, j=None):
+        """Add an orbital piecewise parameter group range.
+        Parameters
+        ----------
+        group_start_mjd : np.float128
+                Lower group edge
+        group_end_mjd : np.float128
+                Upper group edge
+        j : int
```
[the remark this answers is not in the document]
poneill129: Now it will! Documentation updated

src/pint/models/binary_piecewise.py
poneill129: Removed entire unit template line

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+                )
+
+        i = f"{int(j):04d}"
+        self.add_param(
+            prefixParameter(
+                name="PLB_{0}".format(i),
+                units="MJD",
+                unit_template=lambda x: "MJD",
+                description="Beginning of paramX interval",
+                description_template=lambda x: "Beginning of paramX interval",
```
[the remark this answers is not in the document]
poneill129: Same again

src/pint/models/__init__.py
poneill129: I've shifted the binary_piecewise class into BTmodel now, made changes in __init__ to accommodate

src/pint/models/binary_piecewise.py
poneill129: Now operates as described

src/pint/models/binary_piecewise.py
poneill129: So I found myself needing this function as validates/setups cannot be run before all pieces and corresponding group boundaries are declared. Otherwise there are floating piecewise parameters that don't have a group to go with them (which validate does not like). The function is intended to be be used after all add_group_range and add_piecewise_param have been called

src/pint/models/binary_piecewise.py
poneill129: Implemented the sorting algorithm described

src/pint/models/binary_piecewise.py
poneill129: Added extra space

src/pint/models/binary_piecewise.py
```diff
@@ -0,0 +1,465 @@
...
+        ----------
+        toa : toa
+           TOA/TOAs to check which group they're in
+        Returns
+        -------
+        np.array
+           str elements, look like ['0000','0001'] for two TOAs where one refences T0X_0000 or T0X_0001.
+        """
+        # asks the model what group a single TOA/list of TOAs
+        barycentric_toa = self._parent.get_barycentric_toas(toa)
```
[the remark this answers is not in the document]
poneill129: reduced to barycentric_toa = toa (this is one of a few convenience functions which are going to change with testing)

src/pint/models/binary_piecewise.py
poneill129: Yeah, this was an one of the functions I would use in the testing script. Does nothing essential, scrapped

src/pint/models/binary_piecewise.py
poneill129: implemented new function paramX_per_toa('T0',toa) which does as was described returns same as get_T0Xs_associated_with_toas. This function changes the structure of the testing suite since now a lot of convenience functions can be consolidated into this

src/pint/models/stand_alone_psr_binaries/BT_piecewise.py
```diff
@@ -0,0 +1,416 @@
+import astropy.constants as c
+import astropy.units as u
+import numpy as np
+import pint.toa
+from pint import GMsun, Tsun, ls
+from pint.models.stand_alone_psr_binaries.BT_model import BTmodel
+from .binary_generic import PSR_BINARY
+
+
+def extract(lst):
```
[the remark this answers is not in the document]
poneill129: Nope, and its gone now

src/pint/models/binary_piecewise.py
poneill129: I left some of these unchanged in the interest of time, since some of these functions are being removed

poneill129:
I've folded in a lot of the changes discussed, though in doing so, some of the functions in what was binary_piecewise.py have been made redundant. Those functions were used in the test suite which I am now working on fixing.

poneill129:
This PR is a little untested. This should be functional, and I'm trialing some demo notebooks which I will circulate for you to play around with while I address the testing suite. 

Currently dealing a bug due to my current setup that means binary_piecewise still exists and get_model() calls are trying to use it rather than going via binary_bt like it should now.

poneill129:
Hi all, more updates on the piecewise model (this time on the testing suite side of things). The tests appear to pass in my local version of this, it is only in the CI tests that tests for other functions begin failing. I haven't found a way to narrow what it is that's causing it, but I have a feeling there is something lingering after the piecewise tests are done. 

I don't know if its possible for this to be the reason, but if its not I cannot see why the test is failing since it shouldn't be affected by any of the changes I've made. 

And to ease nervous minds, I've painstakingly looked through each of the tests, and done some quick fitting using pulsar data, to verify the model and tests are still behaving as intended (i.e. everything is not broken!)

dlakaplan:
Hi, Patrick.  The CI failure that I saw was not something you did.  The test was for something else and was close to passing, but occasionally it fails because the tolerances may be too tight.  I just tried to re-run it.

abhisrkckl:
Hi @poneill129, from reading the comments above, it seems that this PR is pretty much ready. Are you planning to do anything more in this PR?

poneill129:
Not at the moment, happy with it. But if anything else needs clearing up I'm able to make changes.

Issue #1548, opened 2023-04-10
New CI failure in test_dds.py


paulray:
I got a CI failure in my automatic CI runs.  I tried to replicate it on my local machine but it ran fine. Then I used pip to upgrade all my packages and now it fails.  Here is the new failure:
```
____________________________________________________________________ TestDDSFit.test_ddsfit ____________________________________________________________________

self = <test_dds.TestDDSFit object at 0x12e1d3670>

    def test_ddsfit(self):
        f = pint.fitter.Fitter.auto(self.t, self.m)
        f.fit_toas()
        chi2 = f.resids.calc_chi2()
        fDDS = pint.fitter.Fitter.auto(self.t, self.mDDS)
    
        fDDS.fit_toas()
        chi2DDS = fDDS.resids.calc_chi2()
>       assert np.isclose(
            1 - np.exp(-fDDS.model.SHAPMAX.value), f.model.SINI.value, rtol=1e-3
        )
E       AssertionError: assert False
E        +  where False = <function isclose at 0x11a168700>((1 - 0.09734670487036771), 0.9043287915482681, rtol=0.001)
E        +    where <function isclose at 0x11a168700> = np.isclose
E        +    and   0.09734670487036771 = <ufunc 'exp'>(-2.329476396012421)
E        +      where <ufunc 'exp'> = np.exp
E        +      and   2.329476396012421 = floatParameter( SHAPMAX             2.329476396012421 () +/- 0.12020910636129484 frozen=False).value
E        +        where floatParameter( SHAPMAX             2.329476396012421 () +/- 0.12020910636129484 frozen=False) = TimingModel(\n  AbsPhase(\n    MJDParameter(   TZRMJD              56252.5252525262476375 (d) frozen=True),\n    strParam...-21 Hz / s frozen=True)),\n  TroposphereDelay(\n    boolParameter(  CORRECT_TROPOSPHERE N                 frozen=True))\n).SHAPMAX
E        +          where TimingModel(\n  AbsPhase(\n    MJDParameter(   TZRMJD              56252.5252525262476375 (d) frozen=True),\n    strParam...-21 Hz / s frozen=True)),\n  TroposphereDelay(\n    boolParameter(  CORRECT_TROPOSPHERE N                 frozen=True))\n) = <pint.fitter.DownhillWLSFitter object at 0x136caa340>.model
[...]

PR #1549, opened 2023-04-10
Fix DDS CI failures

dlakaplan:
Address #1548


dlakaplan:
@paulray : this passed CI.  It uses a single random seed so hopefully if it passes once given a version it should keep passing.  does this pass for you locally?

paulray:
This passes on my local Mac

paulray:
Was the issue just randomness? Or did the some version change in astropy/scipy/numpy cause it to be more likely to fail?

dlakaplan:
I don't really know - I wasn't able to reproduce it.

PR #1552, opened 2023-04-14
Include SolarSystemShapiro only if Astrometry is present.

abhisrkckl:


paulray:
Looks good to me!

abhisrkckl:
THis is ready to go I think. It's pretty small.

PR #1553, opened 2023-04-14
Validate TimingModel components

abhisrkckl:


dlakaplan:
I think the code here mostly makes sense. It would be nice if it were a bit more general, but there are so many special cases that would be hard.  If we can get agreement on the criteria then this can be merged.

dlakaplan:
It would help though to document this clearly.  So in the docstring for the validate method please write out the criteria that will raise an exception.  And maybe also in some of the higher-level docs on the `TimingModel`

abhisrkckl:
> It would help though to document this clearly. So in the docstring for the validate method please write out the criteria that will raise an exception. And maybe also in some of the higher-level docs on the TimingModel

Added docstring for `validate_component_types` method.

abhisrkckl:
@dlakaplan We discussed the validation criteria in today's meeting and people agreed that they make sense.

PR #1555, opened 2023-04-18
Sourcery refactors I : Tests, Examples, and Profiling

abhisrkckl:
I am finding the sourcery squiggles a bit annoying/distracting. So I am going through the entire codebase and applying the suggested refactorings that make sense to me. I have done my best to preserve comments.

In this process, I have found and removed some ugly hacks that are no longer required. I have also commented out some unreachable code (I have added comments saying that I have done this).

I am doing this in multiple PRs so that each it doesn't end up as one huge Frankenstein's monster.

This PR does not change any source code and only changes tests, examples etc.


docs/_ext/paramtable.py
```diff
@@ -55,17 +54,14 @@ def run(self):
                     entry += para
                 elif c == "name":
                     text = d[c]
-                    alias_list = d.get("aliases", [])
-                    if alias_list:
+                    if alias_list := d.get("aliases", []):
```
dlakaplan, line 57: I guess as long as our min version is 3.8 this operator is probably OK, but I don't think we use it widely elsewhere.

dlakaplan:
I took a quick look through everything and as you say it all looks fine.  My only question was that you introduce the `:=` operator which we don't use widely.  But we do use it and as long as our min python version is >= 3.8 that should be fine.  So I can merge if you are done.

docs/_ext/paramtable.py
abhisrkckl: I think this is a useful shorthand.

Since oldestdeps tests are succeeding, I don't think this will create any issue.

PR #1550, opened 2023-04-13
WIP: OFFSET as a fittable parameter 

abhisrkckl:
See #38

PR #443, opened 2018-11-12
Fit tzrmjd

luojing1211:
This is a pull request for fixing issue #431.

This branch is still under development.

Issue #38, opened 2015-07-02
Explicit Absolute Phase


paulray:
Traditionally, there is a hidden overall phase offset in TEMPO timing model fitting. It would be better in PINT to make this an explicit parameter that can be fitted for, or not, as desired.

One possibility is to have TZRMJD, TZRSITE, TZRFRQ define the phase 0 and make TZRMJD explicitly a fittable parameter.  In this case, the site and frequency could be chosen by the user and would not be chosen according to the strange rules that govern TEMPO and Tempo2's handling of these parameters.

An alternative, suggested by @demorest is to make a new parameter (with a different name!) that is an MJD(TDB) specifying phase 0.0 at infinite frequency at the barycenter.  This is simpler from a conceptual standpoint, since the phase model for the pulsar spin is defined in TDB units (modulo issue #37). This new parameter could be fitted or not, as desired by the user.  This is somewhat the way it is currently done in PINT, but the parameter name is TZRMJD, which is confusing.

So, we need to make a choice for PINT's implementation and put this functionality in.

paulray:
Here is a summary of the discussion on the PINT telecon 2016 Dec 5:

What we discussed was not trying to overload the current usage of the TZR* parameters or trying to make them fittable, but rather adding a new parameter (tentatively called PHASE0EPOCH) that is included in the par file and is fittable.  This parameter would specify the MJD(TDB or TCB) of phase 0.0, similar to the current PEPOCH, DMEPOCH, and POSEPOCH that specify the 0 points for expanding those values as polynomials.

We mentioned that it should be straightforward to write out the TZR* parameters as well, since the delays can be computed to get the TOA arrival time at a particular site and frequency.  We thought that probably PHASE0EPOCH would be identical to TZRMJD as long as TZRSITE was SSB and TZRFREQ was 0.0 (infinity), but we weren't 100% certain of this.  Still we kept PHASE0EPOCH separate from the TZR* parameters because we want it treated differently as a fittable parameter and the origins of the TZR names are obscure.

With PHASE0EPOCH as part of the model, there is no need for sticking an extra column in the design matrix to account for an arbitrary phase. And the residuals can be computed directly from the model, with no need to subtract any kind of mean (weighted or unweighted).

paulray:
[...]

PR #1557, opened 2023-04-19
Explicit phase offset as a fittable parameter

abhisrkckl:
This PR implements a `PhaseOffset` component that applies an overall phase offset (`PHOFF`) between physical TOAs and the TZR TOA. This works by using a new attribute `tzr` in the `TOAs` class that distinguishes between physical and TZR TOAs.

The sign of `PHOFF` is set to be the same as that of the implicit `Offset`. The design matrix entries for `PHOFF` and `Offset` should be identical.

If the `PhaseOffset` component is present in the timing model, implicit offset (`Offset`) will not be included in the design matrix and the correlation matrix, and the mean will not be subtracted from the residuals.

See issues #38 and #431 and PR #443. 
#1550 was an attempt to do this that didn't work.


abhisrkckl:
I think this works and is ready to go.

dlakaplan:
Can you give some examples of how this would work?  Have you added discussion to the docs?

dlakaplan:
Can you share the links to the RTD sections you updated?

dlakaplan:
That's good.  Maybe also include a snippet showing how to add just a parameter?

abhisrkckl:
I am working on an example.

src/pint/residuals.py
```diff
@@ -107,7 +108,7 @@ def __init__(
         self.toas = toas
         self.model = model
         self.residual_type = residual_type
-        self.subtract_mean = subtract_mean
+        self.subtract_mean = subtract_mean and "PhaseOffset" not in model.components
```
dlakaplan, line 111: Maybe a `debug` message when the presence of a PhaseOffset overrides `subtract_mean`?

src/pint/toa.py
```diff
@@ -1255,6 +1255,8 @@ class TOAs:
         The TOA objects this TOAs should contain.
     toatable : astropy.table.Table, optional
         An existing TOA table
+    tzr : bool
+        Whether TZR TOA
```
dlakaplan, line 1259: seems like this is missing some text

src/pint/toa.py
```diff
@@ -1291,9 +1293,11 @@ class TOAs:
         available to use names as compatible with TEMPO as possible.
     wideband : bool
         Whether the TOAs also have wideband DM information
+    tzr : bool
+        Whether TZR TOA
```
dlakaplan, line 1297: same.

dlakaplan:
I think the tests should:
* show what happens if you try to fit for both TZRMJD and PHOFF
* show how the residuals change when you change PHOFF
* show that the residuals are not mean subtracted

paulray:
> * show what happens if you try to fit for both TZRMJD and PHOFF

I didn't think TZRMJD was ever a fittable parameter. Is it?

paulray:
Maybe in the doc paragraph you can say that once you have a non-zero PHOFF, you can make it zero by computing d_phase/d_TOA at the time of the TZR TOA and turning that into a time difference required to zero out the PHOFF?  I assume that is how TEMPO and Tempo2 output their updated TZR TOAs after a fit?

src/pint/toa.py
abhisrkckl: fixed

src/pint/toa.py
abhisrkckl: fixed

src/pint/residuals.py
abhisrkckl: added debug message

abhisrkckl:
@dlakaplan Added the example `phase_offset_example.py`

dlakaplan:
The example looks great.  Probably that's sufficient.  I don't know if we also want a few lines in the docstring for `PhaseOffset` showing how to add one. Maybe just a link to the example?

abhisrkckl:
The process of adding the explicit offset is pretty much the same as any other component/parameter. i.e., add `PHOFF` to the par file or add `PhaseOffset` to the model using the `add_component` method. Do we need separate instructions for that?

dlakaplan:
Probably not.  so maybe just a link to the example in the docstring.

dlakaplan:
OK, great.  I can merge as soon as it passes (note it seems like there are GitHub errors)

Issue #1546, opened 2023-04-06
No T2 binary model


dlakaplan:
Do we need to implement such a model for compatibility?

dlakaplan:
For the moment let's leave this as-is.  But I can add something to the docs telling people to convert to the underlying model which should work OK in most cases.

PR #1558, opened 2023-04-20
Documented binary conversion

dlakaplan:
Doesn't implement a fix for #1546 but documents how to get around it


dlakaplan:
Only a doc change

PR #1561, opened 2023-04-21
Allow separate calculation of residual means

dlakaplan:
Previously you would have to make a whole new `Residuals` object with a different value of `subtract_mean`.  Now that can be computed on the fly with the defaults overridden.

Also allow `adjust_TOAs` to be used with a scalar and/or Quantity input


dlakaplan:
I encountered the need for this when trying to simulate new TOAs that are appended onto existing TOAs.  You need to remove the difference in the means but this was a pain since you had to make many new residuals objects with different options.  You still need to create the objects but the options used don't matter.

Example:
```
m = get_model(io.StringIO(par))
t = pint.simulation.make_fake_toas_uniform(59990, 61000, 20, m, add_noise=True)
t2 = pint.simulation.make_fake_toas_uniform(61005, 61100, 20, m, add_noise=False)
mean0 = pint.residuals.Residuals(t, m).calc_time_mean()
mean = pint.residuals.Residuals(t2, m).calc_time_mean()
t2.adjust_TOAs(-(mean - mean0))
t2.adjust_TOAs((np.random.normal(size=len(t2)) * t2.get_errors()))
tc = pint.toa.merge_TOAs([t, t2])
```
the residuals from `tc` will be as flat as desired.  Maybe there is a better way to do this.

dlakaplan:
This is an issue even when the model has a `TZRMJD`

abhisrkckl:

dlakaplan:
Yes, please do.

PR #1562, opened 2023-04-25
Changed hypothesis version for oldestdeps

dlakaplan:
CI failure due to `hypothesis` updates.  Pin it to an older version instead.


dlakaplan:
This passes `oldestdeps`.  If it passes the other tests I think it can be merged.

PR #1564, opened 2023-04-27
WIP: Sourcery refactors IV

abhisrkckl:


src/pint/pintk/timedit.py
```diff
@@ -184,12 +184,11 @@ def applyChanges(self):
...
-            fout.close()
-            log.info("Saved timfile to %s" % filename)
-        except:
-            if filename == () or filename == "":
+            with open(filename, "w") as fout:
+                fout.write(self.editor.get("1.0", "end-1c"))
+            log.info(f"Saved timfile to {filename}")
+        except Exception:
+            if filename in [(), ""]:
                 log.warning("Write Tim cancelled.")
```
dlakaplan, line 192: Maybe this should be "Writing timfile cancelled"?

src/pint/scripts/event_optimize_multiple.py
```diff
@@ -339,13 +331,13 @@ def main(argv=None):
         if tname[-6:] == "pickle" or tname == "analytic":
             # Analytic template
             try:
-                gtemplate = cPickle.load(file(tname))
-            except:
+                gtemplate = pickle.load(file(tname))
```
dlakaplan, line 334: Why are you switching from `cPickle` to `pickle`?

abhisrkckl: cPickle was discontinued in python 3. Since we are no longer supporting python2, this part will not work.
https://stackoverflow.com/questions/49579282/cant-find-module-cpickle-using-python-3-5-and-anaconda

abhisrkckl: Quote from StackOverflow:
>There is no cPickle in Python 3. Just import pickle. pickle will automatically use the C accelerator.

src/pint/pintk/timedit.py
abhisrkckl: Changed.

PR #1568, opened 2023-05-03
WIP: Override par file when building models

dlakaplan:
I don't know if this is valuable, but there are times when it is useful to add parameters when building a model and/or override values in the par file.  So far this just does the former.  So you can do:
```
get_model(..., PEPOCH=50000, F0=10 / u.d)
```
and those values will be set, but they have to already exist.  What might be more useful is to allow the user to set parameters that don't yet exist in the same way (rather than by appending to the parfile).  That requires a bit more machinery to get the parsing right, but I thought I'd check first to see if others thought this was useful.

### DIFF
--- a/src/pint/models/model_builder.py
+++ b/src/pint/models/model_builder.py
@@ -5,8 +5,10 @@ import warnings
 from io import StringIO
 from collections import Counter, defaultdict
 from pathlib import Path
-from pint.models.astrometry import Astrometry
+from astropy import units as u
+from loguru import logger as log
 
+from pint.models.astrometry import Astrometry
 from pint.models.parameter import maskParameter
 from pint.models.timing_model import (
     DEFAULT_ORDER,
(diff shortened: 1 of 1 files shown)


abhisrkckl:
I think this is a useful feature. However, I think the parameters should be quantities (i.e., `F2=u.Quantity(-7e-17, "Hz^2")` instead of just the number). This way, the user doesn't have to remember what arbitrary unit PINT is using for any given parameter.

The MJDs can be astropy.time.Time or time.Time objects I guess.

dlakaplan:
I agree that would be nice, but that gets a lot more complicated since then it has to move some of the parameter parsing around to before the parameter is created.  This isn't about what unit PINT assumes: it's what unit the par file assumes.

I'll see if there is an easy way to do it.

dlakaplan:
OK, there is now a way to supply `Quantity` or `Time` parameters even if they don't exist yet. It's a little hacky, but  I think writing something else would be a lot more complicated.

abhisrkckl:
Will this create a new parameter if it is not already there in the par file? Eg: adding F2 when the par file only has F0 and F1.

Also, does this handle mask parameters in some way?

abhisrkckl:
I think this looks good. I'll merge it if it's ready.

abhisrkckl:
Is this ready to go? (except for the changelog conflict)

dlakaplan:
Yes, I think this is ready.

PR #1569, opened 2023-05-04
Wishlist for PINT development

abhisrkckl:
As discussed in a previous meeting, I have collected a wish list of features and improvements for PINT.

### DIFF
--- a/WISHLIST.md
+++ b/WISHLIST.md
@@ -1,0 +1,71 @@
+# PINT Development Wishlist
+
+Items with an active/work in progress pull request are marked with (*).
+
+## New features
+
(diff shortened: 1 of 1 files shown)


dlakaplan:
Is there supposed to be a way to claim or discuss individual items?

abhisrkckl:
> Is there supposed to be a way to claim or discuss individual items?

This is a very important point. We should discuss this in the next meeting, I think.

dlakaplan:
Would it be better to just include the wishlist in an issue or PR so that we can comment on it directly, rather than as a separate file?

abhisrkckl:
As we have decided in the meeting, I have made an issue #1576 for this. I have pinned the issue so  that it appears on top.

PR #1573, opened 2023-05-10
Implemented Roemer delay terms up to O(e^3) in ELL1 model

wcfiore:
Extended the approximation for the Roemer delay in the ELL1 model to third order in eccentricity. 

This plot shows, for x=1 and e=0.001 the difference between the O(e^3) expression and the no-approximations expression used in the BT/DD models, modulo a constant offset. The residuals are of order e^4.

roemer-residuals


dlakaplan:
Can you redo that plot for a few values of ECC just to check?  And maybe explicitly include the terms here so we can see them?

dlakaplan:
I see that you put in the partial derivatives.  Did you test fitting?  Does it converge OK?

wcfiore:
> Can you redo that plot for a few values of ECC just to check? And maybe explicitly include the terms here so we can see them?

Certaintly, I'm not sure why I omitted the terms themselves. This is my full math, including re-deriving the lower-order terms and the partial derivatives:

[Third_Order_ELL1.pdf](https://github.com/nanograv/PINT/files/11443962/Third_Order_ELL1.pdf)

And here are just the third-order terms by themselves:

$\Delta_\mathrm{R}^{(e^3)} = - \frac{x}{12}\Big(5e_2^3\sin 2\phi + 3e_1^2e_2\sin 2\phi - 6e_1e_2^2\cos 2\phi - 4e_1^3\cos 2\phi - 4e_2^3\sin 4\phi  + 12e_1^2e_2\sin 4\phi + 12e_1e_2^2\cos 4\phi  - 4e_1^3\cos 4\phi\Big)$

Here is another plot with a few combinations of ECC and OM:
ell1test

wcfiore:
> I see that you put in the partial derivatives. Did you test fitting? Does it converge OK?

I have tested fitting and everything seems to be working.

dlakaplan:
Can you fix the black formatting?

dlakaplan:
If you click on the failed test you can see:
```
black: install_deps> python -I -m pip install 'black~=23.0'
black: freeze> python -m pip freeze --all
black: black==23.3.0,click==8.1.3,mypy-extensions==1.0.0,packaging==23.1,pathspec==0.11.1,pip==23.1.2,platformdirs==3.5.0,setuptools==67.7.2,tomli==2.0.1,wheel==0.40.0
black: commands[0]> black --check src tests examples
would reformat /home/runner/work/PINT/PINT/src/pint/models/stand_alone_psr_binaries/ELL1_model.py

Oh no! 💥 💔 💥
1 file would be reformatted, 272 files would be left unchanged.
black: exit 1 (7.38 seconds) /home/runner/work/PINT/PINT> black --check src tests examples pid=1747
  black: FAIL code 1 (10.29=setup[2.92]+cmd[7.38] seconds)
  evaluation failed :( (10.64 seconds)
Error: Process completed with exit code 1.
```
In general you should just run the right version of black over your code, or try `tox -e black`.  You might find that you can automatically make your editor do this.

dlakaplan:
Here is a maple notebook that does the derivation and converts the result to python.  It also does a couple of partial derivatives.  I don't know who else can open this, but probably you can at least look at it within Mathematica.
[ELL1_3rdorder.mw.zip](https://github.com/nanograv/PINT/files/11510187/ELL1_3rdorder.mw.zip)

wcfiore:
I checked my code against the output from David's Maple notebook for all the partial derivatives and did find one math mistake and a typo, which I've corrected. I copied the code for each partial derivative into the attached jupyter notebook, as well as the corresponding Maple output, to check them against each other. As it is now, everything seems to be correct.

[test_ELL1_expressions.ipynb.zip](https://github.com/nanograv/PINT/files/11512166/test_ELL1_expressions.ipynb.zip)

wcfiore:
Here's my version of David's Maple notebook. It produces the partial derivatives, which I copied into the python notebook I shared above. It also shows that the various trig algebra that happens along the way is correct.

[ELL1_3rdorder.mw.zip](https://github.com/nanograv/PINT/files/11512814/ELL1_3rdorder.mw.zip)

wcfiore:
I removed "WIP" from the title as I think this is ready to review. However, I suppose it is also worth taking a look at the example notebooks as discussed last week, which I have not yet done.

dlakaplan:
@wcfiore : if you merge in the master branch I think the CI tests will pass

src/pint/models/binary_ell1.py
```diff
@@ -79,21 +79,25 @@ class BinaryELL1(PulsarBinary):
...
     Notes
     -----
     This includes o(e^2) expression for Roemer delay from Norbert Wex and Weiwei Zhu
-    This is equaiton (1) of Zhu et al (2019) but with a corrected typo:
+    This is equation (1) of Zhu et al (2019) but with a corrected typo:
         In the first line of that equation, ex->e1 and ey->e2
         In the other lines, ex->e2 and ey->e1
     See Email from Norbert and Weiwei to David on 2019-Aug-08
     The dre expression comes from Norbert and Weiwei; the derivatives
     were calculated by hand for PINT
```
abhisrkckl, line 96: Can you write full names here? 

It will be nice to have that email stored somewhere for posterity. (This may not be possible given privacy concerns.)

src/pint/models/binary_ell1.py
```diff
@@ -79,21 +79,25 @@ class BinaryELL1(PulsarBinary):
     ----------
     - Lange et al. (2001), MNRAS, 326 (1), 274–282 [1]_
     - Zhu et al. (2019), MNRAS, 482 (3), 3249-3260 [2]_
+    - Fiore et al. (2023), arXiv:2305.13624 [astro-ph.HE] [3]_
 
     .. [1] https://ui.adsabs.harvard.edu/abs/2019MNRAS.482.3249Z/abstract
     .. [2] https://ui.adsabs.harvard.edu/abs/2001MNRAS.326..274L/abstract
-
+    .. [3] https://arxiv.org/abs/2305.13624
```
abhisrkckl, line 86: Are the O(e^3) terms given in this paper?

abhisrkckl (commented):
Can you add an entry in CHANGELOG-unreleased.md?
Also, can you add your name to `AUTHORS.rst`?

src/pint/models/binary_ell1.py
wcfiore: yes, Equation 4.

src/pint/models/binary_ell1.py
wcfiore: This text was introduced with the o(e^2) terms, not by me...I don't actually know which David is meant here. It could be Kaplan but I think it is Nice? 

It does say "Norbert Wex" and "Weiwei Zhu" on line 90, and then it changes to first names. I could change it to full names all the way through if you think that is best.

dlakaplan: I think the email was to David Nice.

wcfiore: ok, changed that to David Nice and changed "Norbert"/"Weiwei" to NW/WZ after the first mention of their full names. I think this clarifies things.

CHANGELOG-unreleased.md
```diff
@@ -25,7 +25,8 @@ the released changes.
 - method `AllComponents.param_to_unit` to get units for any parameter, and then made function `utils.get_unit`
 - can override/add parameter values when reading models
 - docs now include list of observatories along with google maps links and clock files
+- added second-order Roemer delay terms to ELL1 model
```
abhisrkckl, line 28: Shouldn't this be 3rd order?

wcfiore: yep, my mistake. Fixed.

abhisrkckl:
Is this ready to go?

wcfiore:
Should be all ready!

abhisrkckl:
I have verified these expressions independently using Mathematica. Everything checks out as far as I can see.
I am attaching a PDF of my Mathematica notebook here for posterity. (Github does not accept the actual notebook.)
[Verify_ELL1_O(e^3).pdf](https://github.com/nanograv/PINT/files/11884677/Verify_ELL1_O.e.3.pdf)

PR #1581, opened 2023-05-24
Include list of observatories in docs

dlakaplan:
Will include a list of observatories in the docs, including google maps URLs to check positions and clock files.  See:
https://nanograv-pint--1581.org.readthedocs.build/en/1581/observatory_list.html

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -23,6 +23,7 @@ the released changes.
 - Documentation: Explanation for offsets
 - Example: `phase_offset_example.py`
 - method `AllComponents.param_to_unit` to get units for any parameter, and then made function `utils.get_unit`
+- docs now include list of observatories along with google maps links and clock files
 ### Fixed
 - fixed docstring for `add_param_from_top`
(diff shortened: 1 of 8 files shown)


dlakaplan:
I don't know why that test is failing (not related to my PR).  But if somebody can look at the RTD output and give any feedback that would be great.  I'll fix the error eventually.

dlakaplan:
I've added clock files with links to the global clock repository.  I'm linking to the human-readable pages (not the raw files).  I construct the URLs by:
`https://ipta.github.io/pulsar-clock-corrections/{format}/clock/{filename}.html`
where `format` is `tempo` or `T2runtime` depending on whether `clock_fmt` is `tempo` or `tempo2`.  I can't verify that this is correct with the global clock code, but it seems to work.  @aarchiba : are you able to confirm or fix?

abhisrkckl:
I think this looks good. I'll merge this if it's ready.

abhisrkckl:
I think that is fine.

dlakaplan:
In that case, this is probably OK to merge now.

PR #1591, opened 2023-06-13
Changed default toa errors for event files.  can also set explicitly

dlakaplan:
Default errors are 1 us for all event files.  Can also set as a `Quantity` or `float`

### DIFF
--- a/src/pint/event_toas.py
+++ b/src/pint/event_toas.py
@@ -1,4 +1,21 @@
-"""Generic function to load TOAs from events files."""
+"""Generic functions to load TOAs from events files, along with specific implementations for different missions.
+
+**Instrument-specific Functions**
+
+.. autofunction:: pint.event_toas.load_NuSTAR_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.load_NICER_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.load_RXTE_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.load_IXPE_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.load_Swift_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.load_XMM_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.get_NuSTAR_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.get_NICER_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.get_RXTE_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.get_IXPE_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.get_Swift_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+.. autofunction:: pint.event_toas.get_XMM_TOAs(eventname [, minmjd, maxmjd, errors, ephem, planets])
+
+"""
 import os
 from functools import partial
 
--- a/src/pint/event_toas.py
+++ b/src/pint/event_toas.py
@@ -4,6 +4,7 @@ import os
 import astropy.io.fits as pyfits
 from astropy.time import Time
 from astropy.coordinates import EarthLocation
+from astropy import units as u
 import numpy as np
 from loguru import logger as log
 
@@ -210,6 +211,7 @@ def load_fits_TOAs(
     timeref=None,
     minmjd=-np.inf,
     maxmjd=np.inf,
+    errors=1 * u.us,
 ):
     """
     Read photon event times out of a FITS file as a list of PINT :class:`~pint.toa.TOA` objects.
@@ -239,6 +241,9 @@ def load_fits_TOAs(
         minimum MJD timestamp to return
     maxmjd : float, default "infinity"
         maximum MJD timestamp to return
+    errors : astropy.units.Quantity or float, optional
+        The uncertainty on the TOA; if it's a float it is assumed to be
+        in microseconds
 
     Returns
     -------
@@ -261,6 +266,7 @@ def load_fits_TOAs(
         timeref=timeref,
         minmjd=minmjd,
         maxmjd=maxmjd,
+        errors=errors,
     )
 
     return toas.to_TOA_list()
@@ -279,6 +285,7 @@ def get_fits_TOAs(
     planets=False,
     include_bipm=False,
     include_gps=False,
+    errors=1 * u.us,
 ):
     """
     Read photon event times out of a FITS file as :class:`pint.toa.TOAs` object
@@ -318,6 +325,9 @@ def get_fits_TOAs(
         Use TT(BIPM) instead of TT(TAI)
     include_gps : bool, optional
         Apply GPS to UTC clock corrections
+    errors : astropy.units.Quantity or float, optional
+        The uncertainty on the TOA; if it's a float it is assumed to be
+        in microseconds
 
     Returns
     -------
@@ -367,6 +377,9 @@ def get_fits_TOAs(
     if weights is not None:
         new_kwargs["weights"] = weights
 
+    if not isinstance(errors, u.Quantity):
+        errors = errors * u.microsecond
+
     # mask out times/columns outside of mjd range
     mjds_float = np.asarray([r[0] + r[1] for r in mjds])
     idx = (minmjd < mjds_float) & (mjds_float < maxmjd)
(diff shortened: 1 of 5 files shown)


dlakaplan:
It's sort of weird to have errors associated with individual photons, but I agree they probably need to be > 0 to make basic fitting/chi^2 calculations work.

paulray:
Looks good to me!  Thanks for the quick fix!

paulray:
Oops, the CI doesn't like the repeated `errors` kwarg

dlakaplan:
Yes, I fixed that.  I was actually thinking that an explicit test of this for non-Fermi satellites would also be useful (I hadn't had time/found the right test yet).  For the Fermi case it is tested.

scottransom:
@dlakaplan The time tagging of Fermi photons is only guaranteed to about that level, though, right?  So there really is an error.  @paulray probably knows the exact number.

scottransom:
Fermi LAT webpage says "<10us, relative to spacecraft time"
https://fermi.gsfc.nasa.gov/ssc/data/analysis/documentation/Cicerone/Cicerone_Introduction/LAT_overview.html

paulray:
Smith et al. (2008) says: "Ground tests using cosmic ray muons demonstrated that the LAT measures event times with precision relative to UTC significantly better than a microsecond (Smith et al. 2006). On orbit, satellite
telemetry indicates comparable accuracy. The contribution to the barycentered time resolution from uncertainty in the LAT’s position is negligible."

dlakaplan:
I guess my point was more about hard-coding a single value for all spacecraft, rather than having some uncertainty present.  But this is fine in any case.

src/pint/event_toas.py
```diff
@@ -581,10 +585,12 @@ def load_RXTE_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
...
     :func:`get_RXTE_TOAs`
     """
-    return load_event_TOAs(eventname, "xte", minmjd=minmjd, maxmjd=maxmjd)
+    return load_event_TOAs(
+        eventname, "xte", minmjd=minmjd, maxmjd=maxmjd, errors=errors
+    )
 
 
-def load_NICER_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
+def load_NICER_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf, errors=1 * u.us):
```
paulray, line 593: Please make this `0.1*u.us`

dlakaplan: Do you want that to be across all event files?  Again, I don't know that we should put too much physical meaning here.

src/pint/event_toas.py
```diff
@@ -550,7 +550,7 @@ def get_event_TOAs(
     )
 
 
-def load_RXTE_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
+def load_RXTE_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf, errors=1 * u.us):
```
paulray, line 553: Probably `2.5*u.us` is a better value here.

src/pint/event_toas.py
```diff
@@ -615,10 +624,16 @@ def load_NICER_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
...
+        eventname,
+        "nicer",
+        minmjd=minmjd,
+        maxmjd=maxmjd,
+        errors=errors,
+    )
 
 
-def load_IXPE_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
+def load_IXPE_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf, errors=1 * u.us):
```
paulray, line 636: This one should be `20*u.us`

src/pint/event_toas.py
```diff
@@ -649,10 +667,16 @@ def load_IXPE_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
...
+        eventname,
+        "ixpe",
+        minmjd=minmjd,
+        maxmjd=maxmjd,
+        errors=errors,
+    )
 
 
-def load_XMM_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
+def load_XMM_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf, errors=1 * u.us):
```
paulray, line 679: I'd use `48*u.us` for XMM

src/pint/event_toas.py
```diff
@@ -683,10 +711,16 @@ def load_XMM_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
...
+        eventname,
+        "xmm",
+        minmjd=minmjd,
+        maxmjd=maxmjd,
+        errors=errors,
+    )
 
 
-def load_NuSTAR_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
+def load_NuSTAR_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf, errors=1 * u.us):
```
paulray, line 723: NuSTAR, with Matteo's calibrations, is about `65*u.us`

src/pint/event_toas.py
paulray: Only NICER event files have 100 ns absolute accuracy to UTC.  The other missions are all worse

src/pint/event_toas.py
```diff
@@ -717,10 +754,16 @@ def load_NuSTAR_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
...
+        eventname,
+        "nustar",
+        minmjd=minmjd,
+        maxmjd=maxmjd,
+        errors=errors,
+    )
 
 
-def load_Swift_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf):
+def load_Swift_TOAs(eventname, minmjd=-np.inf, maxmjd=np.inf, errors=1 * u.us):
```
paulray, line 766: Maybe `300*u.us` for Swift XRT data

dlakaplan:
I made all of the uncertainties separate and set by instrument.  It would actually be better to turn the instrument-by-instrument functions into some factory functions.

paulray:
> I made all of the uncertainties separate and set by instrument. It would actually be better to turn the instrument-by-instrument functions into some factory functions.

Looks good!  I might be better to make a factory function, but I don't think it needs to be part of this PR. I'm in favor of merging this once the tests pass.

dlakaplan:
I can probably use `functools` to do that quickly.  Let me try while I'm on another call - I can probably do it before the tests pass

dlakaplan:
It's pretty easy using `functools.partial`.  The only downside is that the signature changes to it's less clear what the arguments are:
```
functools.partial(<function get_event_TOAs at 0x...60d0>, mission='nicer', errors=<Quantity 0.1 us>)
    Read photon event times out of a NICER file as a :class:`pint.toa.TOAs` object
    
    Correctly handles raw event files, or ones processed with axBary to have
    barycentered TOAs. Different conditions may apply to different missions.
    
    The minmjd/maxmjd parameters can be used to avoid instantiation of TOAs
    we don't want, which can otherwise be very slow.
    
    Parameters
    ----------
    eventname : str
        File name of the FITS event list
    minmjd : float, default "-infinity"
        minimum MJD timestamp to return
    maxmjd : float, default "infinity"
        maximum MJD timestamp to return
    errors : astropy.units.Quantity or float, optional
        The uncertainty on the TOA; if it's a float it is assumed to be
        in microseconds
    ephem : str, optional
        The name of the solar system ephemeris to use; defaults to "DE421".
    planets : bool, optional
        Whether to apply Shapiro delays based on planet positions. Note that a
        long-standing TEMPO2 bug in this feature went unnoticed for years.
        Defaults to False.
```
Do you think this is OK?  The docstring can still be correct.  This is easier to update/maintain.

dlakaplan:
Let me know if you think the latest push is OK.  It's easier than before, but you still need to set each function explicitly (more complicated things are possible but seemed harder).  It's a lot easier to read the definitions and make sure they are all consistent.  The downside is that the signature is harder to understand.

paulray:
Yeah, that looks good. It is nice to remove all the redundancy.

dlakaplan:
The other side of this is that the functions created with `partial` don't seem to be in the default RTD:
https://nanograv-pint--1591.org.readthedocs.build/en/1591/_autosummary/pint.event_toas.html

paulray:
Oh, that's not optimal

dlakaplan:
Yes, I'll see if there's a quick fix.

dlakaplan:
Oddly, https://github.com/sphinx-doc/sphinx/pull/5305 says it will add autodocs for partial functions, and I think our sphinx should be recent enough.  But on the other hand I see people in the original issues trying with 1.8 and still having problems.  I don't know if there were divergences between the sphinx 1.8 releases (which we use) and the 2.0/3.0/... releases.  We are pinned to <2, but the current release is 7

dlakaplan:
@paulray : please look at the lated RTD build.  I explicitly added the instrument-specific functions.  It's still not perfect, as they don't collapse like normal.  And now the instruments need to be included in 3 places.  I still feel that something like this is useful, but maybe it's better to go back a couple of pushes and deal with this other stuff later.

paulray:
I think this looks fine. My only suggestion would be to try to get the preferred `get_XXX_TOAs` functions to appear in the docs before the `load_XXX_TOAs` functions. Is that possible?

dlakaplan:
Changed

dlakaplan:
It is possible to define these functions dynamically.  But I haven't found a nice way.  The best I can do is:
```
for mission in ["RXTE", "NICER", "NuSTAR", "XMM", "Swift", "IXPE"]:
    exec(
        f"get_{mission}_TOAs = partial(get_event_TOAs, mission=mission.lower(), errors=_default_uncertainty[mission])"
    )
    exec(f"get_{mission}_TOAs.__doc__ = _get_event_docstring.format(mission)")
```
which seems too messy.

PR #1593, opened 2023-06-16
Allow merging a list of TOAs objects with only one element

rossjjennings:
@dlakaplan did some code cleanup in PR #1511 that inadvertently broke the `merge_TOAs()` function in the case when a list of length 1 is passed (it currently raises a `TypeError` in this case). Trying to merge such a list might seem like an odd thing to want to do, but it is something we ran across in doing IPTA "combination" using PINT Pal for a pulsar with only NANOGrav data, where we had a configuration file that only included a single .tim file.

The simplest solution seemed to be to change PINT to handle this case more sensibly, by just returning the single TOAs object rather than raising an exception, so this is what this PR does.

### DIFF
--- a/src/pint/toa.py
+++ b/src/pint/toa.py
@@ -2692,7 +2692,7 @@ def merge_TOAs(TOAs_list, strict=False):
     """
     # don't duplicate code: just use the existing method
     t = copy.deepcopy(TOAs_list[0])
-    if len(TOAs_list > 1):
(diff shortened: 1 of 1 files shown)


dlakaplan:
I don't understand the IPTA workflow (a check there that there is >1 tim file could also help) but this seems fine.  Could you add an entry to the unreleased CHANGELOG?

PR #1595, opened 2023-06-16
Standardize tests and fix some deprecation warnings

### LINKED ISSUE #1592: Should we standardize tests?
Right now we have two different strategies in use for test. Some seem like straight pytest syntax:

```
class TestModelBuilding:
    def setup_method(self):
        self.parfile = os.path.join(datadir, "J0437-4715.par")

    def test_from_par(self):
        tm = get_model(self.parfile)
        assert len(tm.components) == 6
        assert len(tm.DelayComponent_list) == 4
        assert len(tm.PhaseComponent_list) == 2
```

and others subclass unittest:

```
class TestTOA(unittest.TestCase):
    """Test of TOA class"""

    def setUp(self):
        self.MJD = 57000

    def test_units(self):
        with self.assertRaises(u.UnitConversionError):
            t = TOA(self.MJD * u.m)
```

Would it be desirable to make them all standard pytest, or is there some benefit to the unittest framework?

I found this because there are a ton of deprecation warnings about old nose syntax, which I am fixing in a PR (`setup(self)` should be `setup_method(self)`, see https://docs.pytest.org/en/stable/deprecations.html#support-for-tests-written-for-nose )

paulray:
Fixes #1592

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -12,6 +12,7 @@ the released changes.
 - Applied `sourcery` refactors to the entire codebase
 - Changed threshold for `test_model_derivatives` test to avoid CI failures
 - Unreleased CHANGELOG entries should now be entered in `CHANGELOG-unreleased.md` instead of `CHANGELOG.md`. Updated documentation accordingly.
+- Changed tests to remove `unittest` and use pure pytest format
 ### Added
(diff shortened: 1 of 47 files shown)


dlakaplan:
Looks good, but I don't think I saw a CHANGELOG entry

paulray:
Right. I'll add CHANGELOG entries for this and the other one that needs it.

Issue #1594, opened 2023-06-16
sampler.chain is deprecated


paulray:
Several places in event_optimize and related places, we use `sampler.chain` to get the chain. This is deprecated:
```
DeprecationWarning: 'chain' is deprecated. Use 'get_chain()' instead.
```

I *think* those could all simply be replaced with `sampler.get_chain()` but I haven't checked or tested this.
@devbhakt can you take a look?

devbhakt:
Hi Paul, the pull request I just submitted fixes the deprecation and uses get_chain() instead.

PR #1596, opened 2023-06-18
Sampler.chain deprecated fix

devbhakt:
This pull request addresses the issue #1594 and uses the built in attribute `sampler.get_chains()` to get the samples instead of `sampler.chain`.

### DIFF
--- a/docs/examples/MCMC_walkthrough.broken
+++ b/docs/examples/MCMC_walkthrough.broken
@@ -127,7 +127,8 @@ To make this run relatively fast for demonstration purposes, nsteps was purposef
 
 ```python
 fitter.phaseogram()
-samples = sampler.sampler.chain[:, 10:, :].reshape((-1, fitter.n_fit_params))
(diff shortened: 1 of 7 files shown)


paulray:
Thanks for doing this!  

The changes are a bit more complicated that I would have guessed. Do you think we need any tests to ensure we are getting the chains right?

devbhakt:
I do not think we need any tests to ensure that we are getting the chains right. The main issue with `sampler.get_chain()` is that it returns the samples in the array [nsteps, nwalkers, ndim], but the codes were set up with `sampler.chain` which returns [nwalkers, nsteps, ndim]. Doing a simple transpose as I have it set up ensures that the samples are in the right array configuration as before. I ran a test mcmc run with event optimize 10 times and the difference between the samples returned by `sampler.chain` and the tranposed `sampler.get_chain()` was 0.0 each time.

paulray:
Great, thanks!

PR #1606, opened 2023-07-18
Correction to docstring for pbprime functions

coclar:
While going through the binary orbit classes to implement a new model, I found that the docstrings for the `pbprime()` methods had an incorrect description. These should calculate the orbital period as a function of time, but the docstrings said that they compute the time-derivative of the orbital period (this is actually computed in `pbdot_orbit()`).

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -14,4 +14,5 @@ the released changes.
 - Fixed RTD by specifying theme explicitly.
 - `.value()` now works for pairParameters
 - Setting `model.PARAM1 = model.PARAM2` no longer overrides the name of `PARAM1`
+- Fixed an incorrect docstring in pbprime() functions. 
 ### Removed
--- a/src/pint/models/stand_alone_psr_binaries/binary_orbits.py
+++ b/src/pint/models/stand_alone_psr_binaries/binary_orbits.py
@@ -29,7 +29,7 @@ class Orbit:
         return (orbits - norbits) * 2 * np.pi * u.rad
 
     def pbprime(self):
-        """Derivative of binary period with respect to time."""
+        """Instantaneous binary period as a function of time."""
         raise NotImplementedError
 
     def pbdot_orbit(self):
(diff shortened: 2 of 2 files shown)


dlakaplan:
Thanks for the fix.  Do you think that this is used correctly otherwise?

coclar:
In `binary_orbits.py`, yes - `pbprime()` is always correctly used as the orbital period. 

However, after digging a bit deeper, there are some lines that I am unsure about in `binary_generic.py`:

The `pb()` and `pbdot()` methods are used by the binary models (ELL1model, BTmodel, etc.) obtain `pbprime()` and `pbdot_orbit()` from the relevant `orbits_cls`: 
https://github.com/nanograv/PINT/blob/40dec33c013096096dd335c221393945033a3a66/src/pint/models/stand_alone_psr_binaries/binary_generic.py#L424C1-L425C41

https://github.com/nanograv/PINT/blob/40dec33c013096096dd335c221393945033a3a66/src/pint/models/stand_alone_psr_binaries/binary_generic.py#L447C1-L448C45

but then the `PSR_BINARY` class also has its own `pbprime()` method:
https://github.com/nanograv/PINT/blob/40dec33c013096096dd335c221393945033a3a66/src/pint/models/stand_alone_psr_binaries/binary_generic.py#L681C5-L682C51
I'm really not sure what this is supposed to be doing! 

`self.pb()` retrieves `self.orbits_cls.pbprime()`, which should already account for any PBDOT, but then `self.pbdot() * self.tt0` is subtracted from it, which would presumably "undo" the PBDOT effects to arrive back at the initial orbital period? But even this is incorrect if there are higher-order orbital frequency derivatives (i.e. pbdot_orbit() is not a constant value). 

I'm struggling to follow the flow here though - it's possible that this method is always overwritten by the inheriting classes in `binary_orbits.py` that implement their own `pbprime()`s, and is therefore never used.

dlakaplan:
Sorry for the delay in dealing with this.  Do you think you could rebase it on the latest `master` and put in an entry in `CHANGELOG-unreleased.md`?

coclar:
No worries, done!

Issue #1585, opened 2023-06-07
Option to include AbsPhase in the timing model during creation 


abhisrkckl:
Currently, `AbsPhase` is included in the timing model during creation only if TZRMJD is defined in the par file. This is understandable since the `TimingModel` object does not have access to the `TOAs` object, which is needed for computing the TZR TOA.

The `AbsPhase` component gets added to the timing model silently when a `Residuals` object is created since it has access to a `TOAs` object at that stage. This is not very transparent. I encountered this while investigating why the par file shown in `pintk` magically had a TZR TOA even when the input par file didn't have it.
 
I propose we move the `AbsPhase` creation logic from the `Residuals` class to the `TimingModel` class in the following way:
- `TimingModel` should have an `add_tzr_toa` method that takes a `TOAs` object as input and created an `AbsPhase` component. The `Residuals` class will call this function instead of having this logic in place.
- `ModelBuilder.__call__` and `get_model` functions should have a `toas_for_tzr` option. If this is given, an `AbsPhase` component will be created based on the given `TOAs` object.
- `get_model_and_toas` function should have an `add_tzr` Boolean option. If this is given, the `AbsPhase` component will be created based on the given TOAs object.
- These options will be superseded by the TZR TOA in the par file if present with a warning.

PR #1607, opened 2023-07-21
Make the usage of TZR TOA explicit

abhisrkckl:
Related: #1585 

1. Move the TZR TOA creation from `TimingModel.phase()` to a new function `TimingModel.add_tzr_toa()`. This will throw an error if `AbsPhase` is already present in the model.
2. `Residuals.__init__()`, `Residuals.calc_phase_resids()` and `Residuals.calc_time_resids()` now have an explicit `use_abs_phase` option. The existing behavior is to implicitly use the absolute phase, and create a TZR TOA implicitly if not available in the timing model.
3. `ModelBuilder.__call__()` and `get_model()` have a new option `toas_for_tzr`. If a TOAs object is given, it creates a TZR TOA using that TOAs object if not already present and does nothing if the TZR TOA is already given in the model.
4. `get_model_and_toas` has a new option `add_tzr_to_model`. If True (default), this option will create a TZR TOA in the timing model using the created TOA object if not already present and does nothing otherwise.
5. Logs an info message every time a TZR TOA gets added to the model after its creation.
6. Fix docstrings for `get_model_and_toas()` and `get_toas()`.
7. Fix bug where `DelayComponent_list` is missing.
8. Fix bug where missing PLANET_SHAPIRO is accessed.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -16,4 +16,6 @@ the released changes.
 ### Fixed
 - Deleting JUMP1 from flag tables will not prevent fitting
 - Docstrings for `get_toas()` and `get_model_and_toas()`
+- Set `DelayComponent_list` and `NoiseComponent_list` to empty list if such components are absent
+- Fix invalid access of PLANET_SHAPIRO in models without Astrometry
 ### Removed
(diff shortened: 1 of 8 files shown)


paulray (approved):
Looks good to me.

PR #1610, opened 2023-07-30
Autocorrelation check for event_optimize

devbhakt:
I have an autocorrelation routine for `event_optimize`. The routine runs the sampler and determines convergence of the chains by 2 criteria:

1.  The chain length is a specified ratio times the estimated autocorrelation time for each parameter.  (The default is set to 10 in `event_optimize`)
2. The change in estimated autocorrelation times is less than 1%

The check is conducted every 100 steps and then every 25 steps once 10% convergence is reached.

### DIFF
--- a/src/pint/scripts/event_optimize.py
+++ b/src/pint/scripts/event_optimize.py
@@ -622,7 +622,7 @@ def main(argv=None):
         action="store_true",
     )
     parser.add_argument(
-        "--autocorr",
+        "--no-autocorr",
         help="Turn the autocorrelation check function off",
         default=True,
         action="store_false",
--- a/src/pint/scripts/event_optimize.py
+++ b/src/pint/scripts/event_optimize.py
@@ -247,6 +247,82 @@ def get_fit_keyvals(model, phs=0.0, phserr=0.1):
     return fitkeys, np.asarray(fitvals), np.asarray(fiterrs)
 
 
+def autocorr_check(sampler, pos, nsteps, burnin, csteps=100, crit1=10):
+    """Return the converged sampler and the mean autocorrelation time per 100 steps
+    Parameters
+    ----------
+    Sampler
+        The Emcee Ensemble Sampler
+    pos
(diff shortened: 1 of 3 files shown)


src/pint/scripts/event_optimize.py
```diff
@@ -545,6 +621,12 @@ def main(argv=None):
         default=False,
         action="store_true",
     )
+    parser.add_argument(
+        "--autocorr",
+        help="Turn the autocorrelation check function off",
```
abhisrkckl, line 626: What is the behavior of event_optimize when "--autocorr" is given? If it is to turn off the autocorrelation, I think this option will be better called "--no-autocorr".

devbhakt: I've updated the option now

src/pint/scripts/event_optimize.py
```diff
@@ -247,6 +247,82 @@ def get_fit_keyvals(model, phs=0.0, phserr=0.1):
...
+    The criteria to check for convergence is:
+        1. the chain has to be longer than the specified ratio times the estimated autocorrelation time
+        2. the change in the estimated autocorrelation time is less than 1%
+    """
+    autocorr = []
+    old_tau = np.inf
+    converged1 = False
+    converged2 = False
+    for sample in sampler.sample(pos, iterations=nsteps, progress=True):
+        if converged1 == False:
```
abhisrkckl, line 279: if not converged1

src/pint/scripts/event_optimize.py
```diff
@@ -247,6 +247,82 @@ def get_fit_keyvals(model, phs=0.0, phserr=0.1):
...
+                        log.info(
+                            "10 % convergence reached with a mean estimated integrated step: "
+                            + str(x)
+                        )
+                    else:
+                        continue
+            else:
+                continue
+        else:
+            if converged2 == False:
```
abhisrkckl, line 302: if not converged2

Issue #1059, opened 2021-07-07
Mask parameters have unhelpful names


aarchiba:
A parfile line that looks like
```
JUMP -fe Rcvr_800 -3.26e-6
```
will produce a parameter on the timing model `mo.JUMP17` or something. It is not clear how the number `17` is assigned to this JUMP, or whether this is consistent in the face of par file changes (for example re-ordering lines, or adding or removing other JUMPs). This makes comparing par files difficult.

I suggest constructing the names of mask parameters in some more consistent way. Perhaps `JUMP_fe_Rcvr_800` for this one? Some care must be taken to ensure that the names are unique and also valid Python identifiers; perhaps just substituting all questionable characters with `_` and appending a number if necessary would resolve this.

Such a change would impact user code that relies on naming parameters by number, including all NANOGrav yaml configurations, but that code is almost certainly broken anyway - minor changes can unexpectedly result in renaming JUMPs.

aarchiba:
Is there any objection to simply naming them `NAME_key_value` or `NAME_key_value1_value2`? Implementation will be a little cumbersome (some values are not acceptable python strings) but can be done.

PR #1615, opened 2023-08-04
FDJUMP implementation (Hacky version)

abhisrkckl:
This PR implements the system-dependent FD parameters (FDJUMPs).
(See #1480)

In PINT, `maskParameter`s and `prefixParameter`s have identical naming schemes. This makes it very challenging to implement FDJUMPs, which have properties of both `maskParameter`s and `prefixParameter`s.

So I am doing a stopgap measure here, by renaming "FDJUMPn" --> "FDnJUMP" so that the mask and prefix indices don't interfere with each other. This also means that we would need an upper bound on the number of allowed prefix indices (n). I have set this to 20. This is controlled by the `pint.models.fdjump.fdjump_max_index` variable.

The above choice also leads to an inconsistency with tempo2 due to the parameter name being "FDnJUMP" instead of "FDJUMPn". I have added a preprocessing step to ensure that the tempo2 par files with "FDJUMPn" are read correctly. In addition, tempo2-compatible par files with "FDJUMPn" can be written out using the `format="tempo2"` option.

A proper solution to this will be to overhaul the `maskParameter` naming scheme (e.g. #1059). However, I am not doing this here since we need this quickly for the IPTA EDR3.

I have also put a boolean parameter FDJUMPLOG to control whether the FDjumps act on the log-frequency or the frequency. This is False by default to make it consistent with tempo2. Note that regular FD parameters act on the log-frequency .


abhisrkckl:
@rossjjennings @gooddc Can you try this and see if it does what you need?

gooddc:
The changes don't seem to fully work with timing_model.py 

Trying to set the list of free_params equal to a list including FD1JUMP returns the following error:

> UnknownParameter                          Traceback (most recent call last)
> Cell In[11], line 2
>       1 # Set free params based on list in the config file (want to update JUMP handling differently soon)
> ----> 2 fo.model.free_params = tc.get_free_params(fo)
>       4 # Do the fit
>       5 try:
> 
> File ~/mambaforge/envs/IPTA_Env_dev_version/lib/python3.9/site-packages/pint/models/timing_model.py:566, in TimingModel.free_params(self, params)
>     564 @free_params.setter
>     565 def free_params(self, params):
> --> 566     params_true = {self.match_param_aliases(p) for p in params}
>     567     for p in self.params:
>     568         getattr(self, p).frozen = p not in params_true
> 
> File ~/mambaforge/envs/IPTA_Env_dev_version/lib/python3.9/site-packages/pint/models/timing_model.py:566, in <setcomp>(.0)
>     564 @free_params.setter
>     565 def free_params(self, params):
> --> 566     params_true = {self.match_param_aliases(p) for p in params}
>     567     for p in self.params:
>     568         getattr(self, p).frozen = p not in params_true
> 
> File ~/mambaforge/envs/IPTA_Env_dev_version/lib/python3.9/site-packages/pint/models/timing_model.py:603, in TimingModel.match_param_aliases(self, alias)
>     600         continue
>     601     return pint_par
> --> 603 raise UnknownParameter(f"{alias} is not recognized as a parameter or alias")
> 
> UnknownParameter: FD1JUMP is not recognized as a parameter or alias

abhisrkckl:
@goodc Can you try with "FD1JUMP1" instead of "FD1JUMP"?

Does this usually work with other mask parameters? What happens when you set `model.free_params` to `["JUMP"]` instead of `["JUMP1"]`?

gooddc:
Okay, I didn't realize it needed to be FD1JUMP1 not FD1JUMP. JUMP doesn't work: only JUMP1, JUMP2, etc. For multiple FDJUMPs will the formatting be FD1JUMP1, FD1JUMP2, FD1JUMP3 or FD1JUMP1, FD1JUMP2, FD1JUMP3?

abhisrkckl:
> For multiple FDJUMPs will the formatting be FD1JUMP1, FD1JUMP2, FD1JUMP3 or FD1JUMP1, FD1JUMP2, FD1JUMP3?

For multiple FD1JUMPs it will be.FD1JUMP1, FD1JUMP2, etc.
For multiple FD2JUMPs it will be.FD2JUMP1, FD2JUMP2, etc.
and so on.

In FDxJUMPy, the x is the FD index (power of the radio frequency in the FD delay expression), and y is the mask index.

gooddc:
Ah okay. I'm not sure I fully understand how this compares to the way that tempo2 is doing this. Does that mean that FD1JUMPn is tied to FD1, FD2JUMPn is tied to FD2, etc.?

abhisrkckl:
> Does that mean that FD1JUMPn is tied to FD1, FD2JUMPn is tied to FD2, etc.?

Yes.

There is an extra complication though. In tempo2, FDx delay is proportional to (log(freq/1_GHz))**x, whereas FDJUMPx (FDxJUMP in PINT) is proportional to (freq/1_GHz)**x. You can control this in PINT using the FDJUMPLOG parameter, which is not there in tempo2. It is set to be tempo2-compatible by default.

src/pint/models/model_builder.py
```diff
@@ -64,6 +65,20 @@ def parse_parfile(parfile):
     return parfile_dict
 
 
+def _replace_fdjump_in_parfile_dict(pardict):
```
dlakaplan, line 68: Can you add a docstring to explain what this function does (and why it's needed)?

src/pint/models/model_builder.py
```diff
@@ -330,6 +345,9 @@ def _pintify_parfile(self, parfile, allow_name_mixing=False):
             parfile_dict = parse_parfile(parfile)
         else:
             parfile_dict = parfile
+
+        parfile_dict = _replace_fdjump_in_parfile_dict(parfile_dict)
```
dlakaplan, line 349: Right now this seems to be a very special single case.  Is there a point to having a slightly more general dictionary (re)parsing step where other functions could be added?  Maybe that's too complex for this, but again a comment line would help.

src/pint/models/fdjump.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class FDJump(DelayComponent):
+    """A timing model for system-dependent frequency evolution of pulsar
+    profiles.
+
+    This model expresses the delay as a polynomial function of the
+    observing frequency/logarithm of observing frequency. This is
+    intended to compensate for the delays introduced by frequency-dependent
+    profile structure when a frequency-independent template profile is used.
```
dlakaplan, line 22: Can you explain the indexing scheme and tempo2 conversion here?

src/pint/models/fdjump.py
```diff
@@ -0,0 +1,155 @@
...
+            if self.param_regex.match(mask_par)
+        ]
+
+        for fdj in self.fdjumps:
+            # prevents duplicates from being added to phase_deriv_funcs
+            if fdj in self.deriv_funcs.keys():
+                del self.deriv_funcs[fdj]
+            self.register_deriv_funcs(self.d_delay_d_FDJUMP, fdj)
+
+    def get_fd_index(self, par):
```
dlakaplan, line 71: Can you add a full docstring?

src/pint/models/fdjump.py
```diff
@@ -0,0 +1,155 @@
...
+        """Extract the index from an FDJUMP parameter name."""
+        if m := self.param_regex.match(par):
+            return int(m.groups()[0])
+        else:
+            raise ValueError(
+                f"The given parameter {par} does not correspond to an FDJUMP."
+            )
+
+    def get_freq_y(self, toas):
```
dlakaplan, line 80: docstring?

src/pint/models/model_builder.py
abhisrkckl: added.

src/pint/models/model_builder.py
abhisrkckl: I think we shouldn't delay this for the general case. But I agree that implementing the general case will be useful.

I have added comments to this effect.

src/pint/models/fdjump.py
abhisrkckl: Added.

src/pint/models/fdjump.py
abhisrkckl: added.

src/pint/models/fdjump.py
abhisrkckl: added.

src/pint/models/fdjump.py
```diff
@@ -0,0 +1,207 @@
...
+    This is intended to compensate for the delays introduced by frequency-dependent
+    profile structure when a different profiles are used for different systems.
+
+    The default behavior is to have FDJUMPs as polynomials of the observing
+    frequency (rather than log-frequency). This is different from the convention
+    used for global FD parameters. This choice is made to be compatible with tempo2.
+    This is controlled using the `FDJUMPLOG` parameter. `FDJUMPLOG Y` may not be
+    tempo2-compatible.
+
+    Caveat
```
dlakaplan, line 30: Would this be better as "Note" which I think gets formatted differently by RTD (e.g., https://nanograv-pint--1615.org.readthedocs.build/en/1615/_autosummary/pint.toa_select.TOASelect.html#pint.toa_select.TOASelect)?  

(I think in many cases we have "Notes", but it might be that "Note" is preferred and those should be updated in another PR)

src/pint/models/fdjump.py
```diff
@@ -0,0 +1,207 @@
...
+
+    This model expresses the delay as a polynomial function of the
+    observing frequency/logarithm of observing frequency in the SSB frame.
+    This is intended to compensate for the delays introduced by frequency-dependent
+    profile structure when a different profiles are used for different systems.
+
+    The default behavior is to have FDJUMPs as polynomials of the observing
+    frequency (rather than log-frequency). This is different from the convention
+    used for global FD parameters. This choice is made to be compatible with tempo2.
+    This is controlled using the `FDJUMPLOG` parameter. `FDJUMPLOG Y` may not be
```
dlakaplan, line 27: I think double quotes (``FDJUMPLOG``) would look better here and elsewhere.

src/pint/models/fdjump.py
```diff
@@ -0,0 +1,207 @@
...
+        super().__init__()
+
+        # Matches "FDpJUMPq" where p and q are integers.
+        self.param_regex = re.compile("^FD(\\d+)JUMP(\\d+)")
+
+        self.add_param(
+            boolParameter(
+                name="FDJUMPLOG",
+                value=False,
+                description="Whether to use log-frequency for computing FDJUMPs.",
```
dlakaplan, line 73: maybe "log-frequency (True) or linear-frequency (False)"?  Or otherwise state the alternative?

src/pint/models/fdjump.py
```diff
@@ -0,0 +1,207 @@
...
+            )
+        )
+        for j in range(1, fdjump_max_index + 1):
```
dlakaplan, line 76: So when you add this component, do jumps 1-20 all get added right away?

src/pint/models/fdjump.py
```diff
@@ -0,0 +1,207 @@
...
+
+        for fdj in self.fdjumps:
+            # prevents duplicates from being added to phase_deriv_funcs
+            if fdj in self.deriv_funcs.keys():
+                del self.deriv_funcs[fdj]
+            self.register_deriv_funcs(self.d_delay_d_FDJUMP, fdj)
+
+    def get_fd_index(self, par):
+        """Extract the FD index from an FDJUMP parameter name. In a parameter name
+        "FDpJUMPq", p is the FD/prefix index and q is the mask index.
```
dlakaplan, line 104: Use consistent style here and in the top-level docstring.

src/pint/models/fdjump.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def fdjump_delay(self, toas, acc_delay=None):
+        """Calculate frequency dependent delay.
+
+        If FDJUMPLOG is Y, use the following expression:
+            Z. Arzoumanian, The NANOGrav Nine-year Data Set: Observations, Arrival
+            Time Measurements, and Analysis of 37 Millisecond Pulsars, The
+            Astrophysical Journal, Volume 813, Issue 1, article id. 65, 31 pp.(2015).
+            Eq.(2):
+            FDJUMP_delay = sum_i(c_i * (log(obs_freq/1GHz))^i)
```
dlakaplan, line 159: add better line breaks to make this more clear in RTD (https://nanograv-pint--1615.org.readthedocs.build/en/1615/_autosummary/pint.models.fdjump.FDJump.html#pint.models.fdjump.FDJump.fdjump_delay).  You could also use a proper math rendering (although that is harder to read in code)

tests/test_fdjump.py
```diff
@@ -0,0 +1,104 @@
+from pinttestdata import datadir
```
dlakaplan, line 1: Do you do any explicit tests where you set an FDJUMP and just check whether the residuals for the selected TOAs change while the others do not?  I see tests of overall fits but nothing that specific.

src/pint/models/fdjump.py
dlakaplan: If so, you might explain that too in the docs, and how to keep the unneeded ones from having any effect.

src/pint/models/fdjump.py
```diff
@@ -0,0 +1,207 @@
...
+    The default behavior is to have FDJUMPs as polynomials of the observing
+    frequency (rather than log-frequency). This is different from the convention
+    used for global FD parameters. This choice is made to be compatible with tempo2.
+    This is controlled using the `FDJUMPLOG` parameter. `FDJUMPLOG Y` may not be
+    tempo2-compatible.
+
+    Caveat
+    ------
+    `FDJUMP`s have two indices: the polynomial/FD/prefix index and the system/mask
+    index. i.e., they have properties of both `maskParameter`s such as `JUMP`s and
```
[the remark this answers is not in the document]
abhisrkckl: This doesn't seem to work properly in a note box.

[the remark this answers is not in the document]
dlakaplan: OK.  In that case I think still using double back ticks is better.  Otherwise the formatting gets confused.

src/pint/models/fdjump.py
abhisrkckl: changed the top level docstring to use double quotes

src/pint/models/fdjump.py
abhisrkckl: Changed.

src/pint/models/fdjump.py
abhisrkckl: Changed.

src/pint/models/fdjump.py
abhisrkckl: I have removed the reference as it looks cluttered. That paper does not discuss FDJUMPs anyway.
I have instead mentioned that the log-expression is similar to global FD 

I think proper math rendering will make it harder to read in code. So I am keeping the expressions as they are

abhisrkckl: Added line breaks

src/pint/models/fdjump.py
abhisrkckl: Changed this to a note box.

tests/test_fdjump.py
abhisrkckl: Added a new test (test_residual_change).

PR #1616, opened 2023-08-10
Chromatic delay similar to dispersion delay

abhisrkckl:
This PR defines a chromatic delay (`ChromaticCM`) using CM, CM1, CM2, etc. similar to `DispersionDM`.

The delay is given by the following expression (consistent with tempo2):

   cm * DMconst * (freq / 1 MHz)**(-alpha)

The chromatic index alpha is called TNCHROMIDX and it is a constant in time in this model.

I have also defined an abstract base class `Chromatic` similar to `Dispersion` so that piecewise-constant CM similar to `DispersionDMX` can be implemented in the future.

Partially implements #1661


src/pint/models/chromatic_model.py
```diff
@@ -0,0 +1,383 @@
...
+            MJDParameter(
+                name="CMEPOCH", description="Epoch of CM measurement", time_scale="tdb"
+            )
+        )
+        self.add_param(
+            floatParameter(
+                name="CMIDX",
+                units=u.dimensionless_unscaled,
+                value=4.0,
+                description="Chromatic measure index",
```
dlakaplan, line 238: Should this be "spectral index"?

src/pint/models/chromatic_model.py
```diff
@@ -0,0 +1,383 @@
...
+        self.add_param(
+            MJDParameter(
+                name="CMEPOCH", description="Epoch of CM measurement", time_scale="tdb"
+            )
+        )
+        self.add_param(
+            floatParameter(
+                name="CMIDX",
+                units=u.dimensionless_unscaled,
+                value=4.0,
```
dlakaplan, line 237: Why a default of 4?  Would 2 make more sense?

src/pint/models/chromatic_model.py
abhisrkckl: In PTA literature "spectral" is often used to describe the dependence on the Fourier conjugate frequency of TOAs (i.e, red noise frequencies, GW frequencies, etc). On the other hand, radio frequency dependence is often described as "chromatic". This is why I am using the term "chromatic" here.

src/pint/models/chromatic_model.py
abhisrkckl: That will be DM, which has a separate class associated with it. I am using 4, because that is the value often associated with scattering noise, which is the most important source of non-DM chromatic noise.

src/pint/models/chromatic_model.py
dlakaplan: I was suggesting "Chromatic measure spectral index". But maybe that is too confusing.

dlakaplan:
Should there be a DMX version of this?

abhisrkckl:
I'll do them separately.

abhisrkckl:
OK... I have compared the code carefully with TEMPO2 and it matches as far as I can tell.

Unfortunately, I couldn't actually do the comparison with tempo2 due to what I suspect is a bug in tempo2. Basically, the post-fit values for CM are not displayed on the terminal or written into an output par file. I have raised this issue in the IPTA slack.

For the time being I think this is ready for review.

abhisrkckl:
Also, I have removed the chromatic index fitting code since it was not working consistently. This is probably because there is too much covariance with DM when the chromatic index is allowed to vary.

abhisrkckl:
Example application on InPTA DR1 data of J1939+2134

Before fitting CM & CM1
image
CHI2R                  3.2021938751881196 
TRES                1.3759850757276195468

After fitting CM & CM1
image
CHI2R                  1.5363535810916809 
TRES                0.9524124603021290128

dlakaplan:
Looks good.   I don't see a test with a fit for CM1, though.  Is that possible?

Similarly, would it make sense to do a fit for CM with alpha=-2 and show that it reproduces the DM fit?

abhisrkckl:
> Looks good. I don't see a test with a fit for CM1, though. Is that possible?
>Similarly, would it make sense to do a fit for CM with alpha=-2 and show that it reproduces the DM fit?

Added both.

tests/test_chromatic.py
```diff
@@ -0,0 +1,78 @@
...
+    m1.add_component(ChromaticCM())
+    m1.TNCHROMIDX.value = 2
+    m1.CM.value = dmval
+    m1.CM.frozen = False
+
+    ftr = WLSFitter(t, m1)
+    ftr.fit_toas(maxiter=3)
+
+    assert ftr.resids.chi2_reduced < 1.5
+    assert (ftr.model.CM.value - m0.DM.value) / ftr.model.CM.uncertainty_value < 2
```
dlakaplan, line 78: Shouldn't they agree to much better than 2 sigma?  If you are fitting the same data with essentially the same model...

abhisrkckl: I have reduced the tolerance to 1.1

dlakaplan: I guess I would have expected this to work with a `isclose` type test, but this should be fine.

And similarly can you have a test for the uncertainty (again, should basically reproduce the DM value)?

abhisrkckl:
@dlakaplan I am done unless there is any other suggestion.

tests/test_chromatic.py
abhisrkckl: added

Issue #1295, opened 2022-06-22
The order of parameters in TimingModel.free_params and TimingModel.designmatrix is different


abhisrkckl:
Is there a reason for this discrepancy? Naively one would expect them to be the same.

PR #1617, opened 2023-08-14
Make `TimingModel.params` and `TimingModel.ordered_params` identical.

abhisrkckl:
See #1295

These parameter lists had different orders because Python dicts were unordered until Python 3.7. Since this is no longer the case, `TimingModel.params` and `TimingModel.ordered_params` have no reason to be different anymore.

### DIFF
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -501,19 +501,23 @@ class TimingModel:
 
     @property_exists
     def params(self):
-        """List of all parameter names in this model and all its components (order is arbitrary)."""
-        # FIXME: any reason not to just use params_ordered here?
-        p = self.top_level_params
-        for cp in self.components.values():
-            p = p + cp.params
-        return p
+        """List of all parameter names in this model and all its components.
+        This is the same as `params_ordered`."""
+
+        # Historically, this was different from `params_ordered` because Python
+        # dictionaries were unordered until Python 3.7. Now there is no reason for
+        # them to be different.
+
+        return self.params_ordered
 
     @property_exists
     def params_ordered(self):
         """List of all parameter names in this model and all its components, in a sensible order."""
+
         # Define the order of components in the list
         # Any not included will be printed between the first and last set.
         # FIXME: make order completely canonical (sort components by name?)
+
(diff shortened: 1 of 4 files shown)


dlakaplan:
Would it be better to just remove the `ordered_params` and make all instances `params`?

abhisrkckl:
I think we should keep it for the time being since other people's codes may depend on it. We can deprecate it in the future and then remove it once people get enough time to change their code.

src/pint/models/timing_model.py
```diff
@@ -500,20 +499,30 @@ def __getattr__(self, name):
...
+    def params_ordered(self):
+        """List of all parameter names in this model and all its components.
+        This is the same as `params`."""
+
+        # Historically, this was different from `params` because Python
+        # dictionaries were unordered until Python 3.7. Now there is no reason for
+        # them to be different.
+
+        warn(
+            "`TimingModel.params_ordered` is now deprecated and maybe removed in the future. "
```
dlakaplan, line 511: "maybe" -> "may be"

dlakaplan: also, shouldn't there be some sort of way that the two lines of the warning are connected?  Is this actually tested?

dlakaplan: Is this best done as a generic `warn` or https://docs.python.org/3/library/exceptions.html#DeprecationWarning?

abhisrkckl: > also, shouldn't there be some sort of way that the two lines of the warning are connected? Is this actually tested?

Python treats such strings as a single string with no newline in between. In a script,
```
"foo"
"bar"
```
is the same as `"foobar"`.

abhisrkckl: > Is this best done as a generic warn or

The `warn` function does invoke a `DeprecationWarning`.

dlakaplan: ok, I didn't read down far enough.

dlakaplan:
OK, I think I can merge this when you are done.

abhisrkckl:
I am done.

PR #1618, opened 2023-08-14
Remove `GLSFitter` dependence from `Residuals`

abhisrkckl:
This requires a bit of code duplication, but I think it is worth decoupling these modules. Also, this avoids a bunch of unnecessary fitting steps while computing the chisq.


src/pint/models/timing_model.py
```diff
@@ -62,6 +62,7 @@
 from pint.toa import TOAs
 from pint.utils import (
     PrefixError,
+    normalize_designmatrix,
```
dlakaplan, line 65: Is this actually needed?

src/pint/utils.py
```diff
@@ -2129,3 +2131,20 @@ def get_unit(parname):
...
+    norm = np.sqrt(np.sum(M**2, axis=0))
+
+    bad_params = [params[i] for i in np.where(norm == 0)[0]]
+    if len(bad_params) > 0 and params is not None:
+        warn(
+            f"Parameter degeneracy found in designmatrix! The offending parameters are {bad_params}.",
+            DegeneracyWarning,
+        )
+    norm[norm == 0] = 1
+    M1 = M / norm
```
dlakaplan, line 2148: couldn't this just be:
`return M/norm, norm`?

src/pint/utils.py
```diff
@@ -2129,3 +2131,20 @@ def get_unit(parname):
 
     ac = AllComponents()
     return ac.param_to_unit(parname)
+
+
+def normalize_designmatrix(M, params):
```
dlakaplan, line 2136: This needs a docstring, maybe with a bit of math in it explaining what is happening.

dlakaplan:
I assume there is no big performance hit when using this?  Or maybe it's even faster?

src/pint/utils.py
abhisrkckl: done.

src/pint/models/timing_model.py
abhisrkckl: removed.

abhisrkckl:
It's the same code that was in GLSFitter, sans a few unnecessary lines. There is no reason it should be slower.

dlakaplan:
I agree that there is no reason.  and it probably does less copying of objects.  but I just wanted to verify empirically.

abhisrkckl:
I did the profiling. The old code takes about 4 s whereas the new code takes about 1 s.

dlakaplan:
great! that's what I was hoping.

src/pint/utils.py
abhisrkckl: Added.

src/pint/utils.py
```diff
@@ -2129,3 +2131,36 @@ def get_unit(parname):
...
+    """Normalize each row of the design matrix.
+
+    This is used while computing the GLS chi2 and the GLS fitting step. The
+    normalized and unnormalized design matrices Mn and M are related by
+        M = Mn @ S
```
dlakaplan, line 2141: what is "@"?  Matrix multiplication?

abhisrkckl: Yes. It's a numpy notation.

abhisrkckl:
There is an API change in this that has to be documented.

dlakaplan:
Is this ready to merge?

dlakaplan:
I guess on the PINT call last week we were discussing whether or not this would lead to changes in any example notebooks.  Have you looked to see what needs to be done?

abhisrkckl:
I checked. The change doesn't affect any notebooks.

abhisrkckl:
> Is this ready to merge?

Yes.

dlakaplan:
But the `understanding_fitters` notebook does:
```
# Not sure how to do this properly yet.
# glsfit2 = pint.fitter.GLSFitter(toas=t, model=glsfit.model, residuals=glsfit.resids)
# glsfit2.fit_toas(maxiter=0)
```
presumably this is trying to construct residuals and just doing a fit because the API required it?

abhisrkckl:
I'll check this.

abhisrkckl:
> But the understanding_fitters notebook does:

I don't think those commented out lines are needed. I have removed them.

The API change is only in Residuals and not GLSFitter. The understanding_fitters notebook does not use Residuals directly at all. So I don't think we need any other change in this file.

PR #1619, opened 2023-08-17
Download data file automatically in `high_level_benchmark.py`

abhisrkckl:
This avoids getting wrong profiling results if the large tim file for J0740+6620 is not present in the directory.


dlakaplan:
Looks fine.  I can merge when you are ready.

abhisrkckl:
Please do so.

PR #1620, opened 2023-08-21
Minor doc updates

dlakaplan:
Minor doc updates to explain default NHARMS and missing derivative functions


dlakaplan:
Error message for a missing derivative now says:
`AttributeError: Derivative function for 'RNAMP' is not provided or not registered; parameter 'RNAMP' may not be fittable. "

dlakaplan:
For the ELL1H model, docstrings now explicitly give the default value of NHARMS in PINT vs. TEMPO2

PR #1624, opened 2023-08-22
Setting model.PARAM1 = model.PARAM2 no longer overrides the name of PARAM1

### LINKED ISSUE #1622: Assigning to a Parameter can override its name
For instance, if I have a model `m` and do:
```
m.POSEPOCH = m.PEPOCH
```
then there is still a `POSEPOCH` attribute in the model, but its name is now `PEPOCH`.  This is largely unintended but comes up accidentally.  Better would be to do:
```
m.POSEPOCH.quantity = m.PEPOCH.quantity
```

I have a fix for this but I'm not sure it's a big enough issue to deal with.  It overrides `setattr` and explicitly copies only `quantity`, `uncertainty`, `frozen` when the names don't match.

dlakaplan:
Fixes #1622 .  Includes a test.


src/pint/models/timing_model.py
```diff
@@ -498,6 +498,16 @@ def __getattr__(self, name):
...
+    def __setattr__(self, name, value):
+        """Mostly this just sets ``self.name = value``.  But in the case where they are both :class:`Parameter` instances
+        with different names, this copies the ``quantity``, ``uncertainty``, ``frozen`` attributes only.
+        """
+        if isinstance(value, Parameter) and name != value.name:
+            for p in ["quantity", "uncertainty", "frozen"]:
+                setattr(getattr(self, name), p, getattr(value, p))
+        else:
+            self.__dict__[name] = value
+
```
abhisrkckl, line 510: I think this should not work for incompatible parameters.

For example, `m.F0 = m.F1` should throw an error. But `m.JUMP1=m.JUMP2` should work.
Maybe we can check the units?

dlakaplan: I think the incompatible units will be rejected by the quantity setting.  So I don't think an explicit unit check is required.  But I put a test in to demonstrate this.

Issue #1623, opened 2023-08-22
Readthedocs build is failing.


abhisrkckl:
The RTD builds for two of my PRs (independent of each other) keeps failing. The error message I get is the following. It's the same in both cases.

> Running Sphinx v7.2.2
> 
> Traceback (most recent call last):
>   File "/home/docs/checkouts/readthedocs.org/user_builds/nanograv-pint/envs/1621/lib/python3.11/site-packages/sphinx/config.py", line 358, in eval_config_file
>     exec(code, namespace)  # NoQA: S102
>     ^^^^^^^^^^^^^^^^^^^^^
>   File "/home/docs/checkouts/readthedocs.org/user_builds/nanograv-pint/checkouts/1621/docs/conf.py", line 395, in <module>
>     'html_theme': html_theme,
>                   ^^^^^^^^^^
> NameError: name 'html_theme' is not defined
> 
> The above exception was the direct cause of the following exception:
> 
> Traceback (most recent call last):
>   File "/home/docs/checkouts/readthedocs.org/user_builds/nanograv-pint/envs/1621/lib/python3.11/site-packages/sphinx/cmd/build.py", line 293, in build_main
>     app = Sphinx(args.sourcedir, args.confdir, args.outputdir,
>           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/docs/checkouts/readthedocs.org/user_builds/nanograv-pint/envs/1621/lib/python3.11/site-packages/sphinx/application.py", line 211, in __init__
>     self.config = Config.read(self.confdir, confoverrides or {}, self.tags)
>                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/docs/checkouts/readthedocs.org/user_builds/nanograv-pint/envs/1621/lib/python3.11/site-packages/sphinx/config.py", line 181, in read
>     namespace = eval_config_file(filename, tags)
>                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/docs/checkouts/readthedocs.org/user_builds/nanograv-pint/envs/1621/lib/python3.11/site-packages/sphinx/config.py", line 371, in eval_config_file
>     raise ConfigError(msg % traceback.format_exc()) from exc
> sphinx.errors.ConfigError: There is a programmable error in your configuration file:
> 
[...]

PR #1625, opened 2023-08-23
Fix RTD by specifying theme explicitly

dlakaplan:
#1623


dlakaplan:
https://docs.readthedocs.io/en/latest/faq.html#how-do-i-change-behavior-when-building-with-read-the-docs suggests that the environment variable `READTHEDOCS` should be set when building, but maybe it isn't?

dlakaplan:
In the end I don't think pinning the versions of `sphinx` or `sphinx-rtd-theme` mattered.  What mattered was the change to `docs/conf.py` to specify the theme explicitly even when on RTD.  Not sure why this worked before or what changed, but the results look OK now.

The pinned versions are also probably OK but can be changed back if needed.

PR #1627, opened 2023-08-25
Fix sphinx build by specifying theme explicitly

dlakaplan:
Also pinning some versions, but that doesn't seem to help or hurt.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -11,4 +11,5 @@ the released changes.
 ### Changed
 ### Added
 ### Fixed
+- Fixed RTD by specifying theme explicitly.
 ### Removed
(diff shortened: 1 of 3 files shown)


abhisrkckl:
Is the "if not on_rtd" part required?

dlakaplan:
I left it.  I can comment it out if that is clearer.

abhisrkckl:
Wouldn't it be better not to pin versions if it makes no difference? Readthedocs or sphinx may drop support for old versions in the future.

abhisrkckl:
I think that if statement just looks confusing

dlakaplan:
My worry is that a major version bump (e.g., sphinx 6->7) may break things anyway.  Some people prefer pinning exact versions to general requirements.

PR #1632, opened 2023-08-29
Fix conversion from ICRS -> ECL when pm uncertainties are None

dlakaplan:
This should have been fixed by #1598 , but that only handled ECL -> ICRS and not the inverse.  I had put in tests there but they actually didn't work (you can't assign uncertainties to `None` with the normal methods). So I fixed the conversion and fixed the tests.

### DIFF
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -550,12 +550,16 @@ class AstrometryEquatorial(Astrometry):
             ra=self.RAJ.quantity,
             dec=self.DECJ.quantity,
             obstime=self.POSEPOCH.quantity,
-            pm_ra_cosdec=self.RAJ.uncertainty * np.cos(self.DECJ.quantity) / dt
-            if self.RAJ.uncertainty is not None
-            else 0 * self.RAJ.units / dt,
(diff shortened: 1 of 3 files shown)


src/pint/models/astrometry.py
```diff
@@ -550,8 +550,12 @@ def as_ECL(self, epoch=None, ecl="IERS2010"):
...
             dec=self.DECJ.quantity,
             obstime=self.POSEPOCH.quantity,
-            pm_ra_cosdec=self.RAJ.uncertainty * np.cos(self.DECJ.quantity) / dt,
-            pm_dec=self.DECJ.uncertainty / dt,
+            pm_ra_cosdec=self.RAJ.uncertainty * np.cos(self.DECJ.quantity) / dt
+            if self.RAJ.uncertainty is not None
+            else 0 * self.RAJ.units / dt,
+            pm_dec=self.DECJ.uncertainty / dt
+            if self.DECJ.uncertainty is not None
+            else 0 * self.DECJ.units / dt,
```
abhisrkckl, line 558: Please put the if expressions inside parentheses for readability. I am surprised that the inline if expressions are formatted this way by black.

Issue #1633, opened 2023-08-31
Allow CHI2 and DOF in par files


dlakaplan:
@scottransom wants this

PR #1634, opened 2023-08-31
Include CHI2 in par output

dlakaplan:
#1633

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -12,6 +12,7 @@ the released changes.
 - Moved design matrix normalization code from `pint.fitter` to the new `pint.utils.normalize_designmatrix()` function.
 - Made `Residuals` independent of `GLSFitter` (GLS chi2 is now computed using the new function `Residuals._calc_gls_chi2()`).
 ### Added
+- CHI2 now in par files
 ### Fixed
 - Fixed RTD by specifying theme explicitly.
 - `.value()` now works for pairParameters
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -12,7 +12,7 @@ the released changes.
 - Moved design matrix normalization code from `pint.fitter` to the new `pint.utils.normalize_designmatrix()` function.
 - Made `Residuals` independent of `GLSFitter` (GLS chi2 is now computed using the new function `Residuals._calc_gls_chi2()`).
 ### Added
-- CHI2 now in par files
+- CHI2, CHI2R, TRES, DMRES now in par files
 ### Fixed
 - Fixed RTD by specifying theme explicitly.
 - `.value()` now works for pairParameters
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -13,7 +13,7 @@ the released changes.
 - Moved design matrix normalization code from `pint.fitter` to the new `pint.utils.normalize_designmatrix()` function.
 - Made `Residuals` independent of `GLSFitter` (GLS chi2 is now computed using the new function `Residuals._calc_gls_chi2()`).
 ### Added
-- CHI2, CHI2R, TRES, DMRES now in par files
+- CHI2, CHI2R, TRES, DMRES now in postfit par files
 - Added `WaveX` model as a `DelayComponent` with Fourier amplitudes as fitted parameters
 - `Parameter.as_latex` method for latex representation of a parameter.
 - `pint.output.publish` module and `pintpublish` script for generating publication (LaTeX) output.
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -698,6 +698,7 @@ class Fitter:
             if self.toas.clock_corr_info["include_bipm"]
             else "TT(TAI)"
         )
+        self.model.CHI2.value = self.resids.chi2
 
     def reset_model(self):
         """Reset the current model to the initial model."""
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -415,13 +415,7 @@ class Fitter:
                         pn, prefitpar.str_quantity(prefitpar.value), "", par.units
                     )
                 elif par.frozen:
-                    if (
-                        par.name == "START"
-                        and prefitpar.value is None
-                        or par.name != "START"
-                        and par.name == "FINISH"
-                        and prefitpar.value is None
-                    ):
+                    if par.name in ["START", "FINISH"] and prefitpar.value is None:
                         s += ("{:" + spacingName + "s} {:20s} {:28g} {} \n").format(
                             pn, " ", par.value, par.units
                         )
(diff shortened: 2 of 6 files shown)


dlakaplan:
Are other parameters useful?  It has NTOA, but would NDOF be helpful?  RMS?

abhisrkckl:
> It has NTOA, but would NDOF be helpful? RMS?

I think NDOF in the case of wideband TOAs will be useful.

scottransom:
I just looked at what is currently being output (at least for `pintempo` via a quick check) and it is the full chi^2 with no NDOF information.  Together that is not super useful as you need to compute the reduced chi^2 yourself by first computing NDOF and then dividing.  TEMPO gives the RMS from the residuals in the output (as TRES), which is useful, and always gives chi^2, NDOF, and the reduced chi^2 on STDOUT and in tempo.lis.  So I recommend that we add TRES to the output parfile, as well, and have CHI2 be the reduced chi^2 as that is probably what people are most interested in at a quick glance.

dlakaplan:
For WB data, would some `DMRES` make sense as well?

dlakaplan:
I added `CHI2R`, `TRES`, `DMRES` (wb only).  So they look like (NB):
```
PSR                            1748-2021E
EPHEM                               DE421
CLK                          TT(BIPM2021)
UNITS                                 TDB
START              53478.2858714195382639
FINISH             54187.5873241702319097
TIMEEPH                              FB90
T2CMETHOD                        IAU2000B
DILATEFREQ                              N
DMDATA                                  N
NTOA                                   62
CHI2                   59.574705174097026
CHI2R                  1.0638340209660184
TRES                 21.18210726637331637
...
```
and WB:
```
PSR                            1748-2021E
EPHEM                               DE421
CLK                          TT(BIPM2021)
UNITS                                 TDB
START              54999.9999999976728819
FINISH             57999.9999999851471876
TIMEEPH                              FB90
T2CMETHOD                        IAU2000B
DILATEFREQ                              N
DMDATA                                  Y
NTOA                                   20
CHI2                    28.49620151951231
CHI2R                  0.8381235741033033
TRES                0.8281346247756675277
DMRES               8.596551156465835e-05
...
```
Are the units of `TRES` supposed to be microseconds?

scottransom:
Excellent.  And yes, microseconds (at least in TEMPO!)

dlakaplan:
We have a test which says:
```
def test_CHI2():
    """Should be present after fit in PINT, not in TEMPO/TEMPO2"""
    m, t = get_model_and_toas(
        os.path.join(datadir, "NGC6440E.par"), os.path.join(datadir, "NGC6440E.tim")
    )

    f = fitter.WLSFitter(toas=t, model=m)
    assert "CHI2" in f.model.as_parfile()
    assert "CHI2" not in f.model.as_parfile(format="tempo2")
    assert "CHI2" not in f.model.as_parfile(format="tempo")
```
are those as intended?

dlakaplan:
And should `CHI2R` come in all formats?  What about `TRES`/`DMRES`?

dlakaplan:
This is basically ready, unless we want to change which parameters are included in which output formats.

dlakaplan:
* TEMPO says `CHI2` and `CHI2R` are unrecognized, but doesn't fail
* TEMPO2 did not fail, and maybe didn't complain
* `TRES` should be fine
* TEMPO2 does give a warning for `DMRES` but does not fail
* TEMPO does give a warning for `DMRES` but does not fail

abhisrkckl:
The RTD build was successfully created in the latest CI run. I don't know why it is showing as failed. Shall I merge this?

paulray:
One minor question about philosophy. CHI2 doesn't make sense in a model, except when referred to a set of TOAs. Do we want to add any additional information (perhaps in a comment?) about the TOAs used to compute CHI2. In general this is difficult/impossible since the set of TOAs can be constructed in many ways and edited, but sometimes there is a unique .tim file name that the CHI2 is referring to. Probably we should just ignore this and accept that there is no way to know what TOAs are being referred to, but I thought I'd at least bring it up.

dlakaplan:
I don't really disagree.  But I think that the common workflow is you have 1 set of TOAs but make many many par files with different iterations/strategies.  And the tim file can more easily be sorted just by which is the longest/most recent.  So knowing which parfile has the lowest chi^2 can help when looking at a directory you haven't touched in a while.  I did think if there was some way to associate a TOA string with this (file name(s), etc) but in the end this is what tempo(2) do, so I think we are safe.

scottransom:
Yeah, I agree that it would be nice to know something about the TOAs used.  But we already have START/FINISH and NTOA, so that gives a hint about what was used.  And the main reason I want to see (reduced) CHI2 is simply to see if it is a small number (i.e. it was a good fit/model) or a big number (i.e. it was an experiment that didn't work out and things aren't phase connected).

CHANGELOG-unreleased.md
```diff
@@ -13,6 +13,7 @@ the released changes.
 - Moved design matrix normalization code from `pint.fitter` to the new `pint.utils.normalize_designmatrix()` function.
 - Made `Residuals` independent of `GLSFitter` (GLS chi2 is now computed using the new function `Residuals._calc_gls_chi2()`).
 ### Added
+- CHI2, CHI2R, TRES, DMRES now in par files
```
abhisrkckl, line 16: Is this in all par files or only in pst-fit par files?

src/pint/fitter.py
```diff
@@ -415,24 +415,30 @@ def get_summary(self, nodmx=False):
...
-                        or par.name != "START"
-                        and par.name == "FINISH"
+                    if par.name in ["START", "FINISH"] and prefitpar.value is None:
+                        s += ("{:" + spacingName + "s} {:20s} {:28g} {} \n").format(
+                            pn, " ", par.value, par.units
+                        )
+                    elif par.name in ["START", "FINISH"]:
+                        s += ("{:" + spacingName + "s} {:20g} {:28g} {} \n").format(
+                            pn, prefitpar.value, par.value, par.units
+                        )
```
abhisrkckl, line 425: Can these be converted into fstrings? 

Curly brackets can be escaped in fstrings like this:
```
foo = 1.5
f"{foo}  {{bar}}" == "1.5  {bar}"
```

src/pint/fitter.py
```diff
@@ -415,24 +415,30 @@ def get_summary(self, nodmx=False):
...
+                    elif (
+                        par.name in ["CHI2", "CHI2R", "TRES", "DMRES"]
                         and prefitpar.value is None
                     ):
                         s += ("{:" + spacingName + "s} {:20s} {:28g} {} \n").format(
                             pn, " ", par.value, par.units
```
abhisrkckl, line 431: fstrings?

CHANGELOG-unreleased.md
dlakaplan: only post-fit.  I will clarify

src/pint/fitter.py
dlakaplan: possibly.  I didn't write these and didn't want to mess with the formatting.

abhisrkckl: OK..

PR #1636, opened 2023-08-31
Mixed wideband + narrowband TOAs objects should not be allowed.

abhisrkckl:
Raise an exception in the case of a tim file with mixed narrowband and wideband TOAs.

### DIFF
--- a/tests/test_wideband_dm_data.py
+++ b/tests/test_wideband_dm_data.py
@@ -61,19 +61,6 @@ def wb_model(tmpdir):
 
 
 @pytest.fixture
-def wb_toas(wb_model):
-    toas = get_TOAs(io.StringIO(tim))
-    for i in range(9):
-        r = Residuals(toas, wb_model)
-        if np.all(r.time_resids < 1 * u.ns):
-            break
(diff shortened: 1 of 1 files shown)


dlakaplan:
Is a warning rather than an exception appropriate here?  Can PINT actually do a sensible fit with only a fraction of data being WB?

abhisrkckl:
> Can PINT actually do a sensible fit with only a fraction of data being WB?

I don't think so. AFAIK, Tim Pennucci's papers as well as the NG12.5 paper, where the WB fitting formalism are given, assume that all the TOAs should be wideband. I am not sure if anyone has worked out how to accurately mix narrowband and wideband toas.

dlakaplan:
It looks like this would work for `get_toas` and `get_model_and_toas`.  Can you confirm?

abhisrkckl:
Yes. get_model_and_toas calls get_toas under the hood. So this should work for both.

dlakaplan:
Great.  Then I can probably merge whenever you want.

abhisrkckl:
Yes. This is ready.

PR #1637, opened 2023-08-31
DMWaveX model for Fourier representation of DM noise

abhisrkckl:
This is a clone of the WaveX model for DM variations.

### DIFF
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1366,6 +1366,86 @@ def wavex_setup(model, T_span, freqs=None, n_freqs=None):
     return model.components["WaveX"].get_indices()
 
 
+def dmwavex_setup(model, T_span, freqs=None, n_freqs=None):
+    """
+    Set-up a DMWaveX model based on either an array of user-provided frequencies or the wave number
+    frequency calculation. Sine and Cosine amplitudes are initially set to zero
+
+    User specifies T_span and either freqs or n_freqs. This function assumes that the timing model does not already
+    have any WaveX components. See add_dmwavex_component() or add_dmwavex_components() to add components
+    to an existing DMWaveX model.
+
+    Parameters
+    ----------
+
+    model : pint.models.timing_model.TimingModel
+    freqs : iterable of float or astropy.quantity.Quantity, None
+        User inputed base frequencies
+    n_freqs : int, None
+        Number of wave frequencies to calculate using the equation: freq_n = 2 * pi * n / T_span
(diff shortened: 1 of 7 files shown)


abhisrkckl:
Here are some test results:

Before applying DMWaveX: (reduced chi2 ~ 30)
image

After applying DMWaveX with 3 harmonics (reduced chisq ~ 9)
image

After applying DMWaveX with 9 harmonics (reduced chisq ~ 7.75)
image

DM time series reconstruction (9 harmonics)
image

src/pint/utils.py
```diff
@@ -1366,6 +1366,86 @@ def wavex_setup(model, T_span, freqs=None, n_freqs=None):
     return model.components["WaveX"].get_indices()
 
 
+def dmwavex_setup(model, T_span, freqs=None, n_freqs=None):
+    """
+    Set-up a DMWaveX model based on either an array of user-provided frequencies or the wave number
+    frequency calculation. Sine and Cosine amplitudes are initially set to zero
+
+    User specifies T_span and either freqs or n_freqs. This function assumes that the timing model does not already
+    have any WaveX components. See add_dmwavex_component() or add_dmwavex_components() to add components
```
dlakaplan, line 1375: "any WaveX components": should be "DMWaveX"?

dlakaplan:
Looks fine.  Do we need any utility to extract the DM time series (e.g., to convert to DMX) or is the `dmwavex_dm` method sufficient?

abhisrkckl:
I think `dmwavex_dm` should be enough. Also, I think most users will only need `TimingModel.total_dm` normally.

src/pint/utils.py
abhisrkckl: changed.

src/pint/utils.py
```diff
@@ -1366,6 +1366,86 @@ def wavex_setup(model, T_span, freqs=None, n_freqs=None):
...
+    nyqist_freq = 1.0 / (2.0 * T_span)
+    if freqs is not None:
+        if isinstance(freqs, u.quantity.Quantity):
+            freqs.to(u.d**-1)
+        else:
+            freqs *= u.d**-1
+        if len(freqs) == 1:
+            model.DMWXFREQ_0001.quantity = freqs
+        else:
+            np.array(freqs)
```
dlakaplan, line 1428: should this be assigned to `freqs`?

abhisrkckl: Corrected,

abhisrkckl: the same mistake was in wavex also. I have corrected that as well.

Issue #374, opened 2018-02-22
Add radial velocities?


dlakaplan:
Would it be possible to add support for companion radial velocities into the PINT binary delay model?  This would help eliminate possible sources of error when doing simultaneous optical/timing fits.  Presumably the radial velocity of the pulsar is already present since it is needed to correct TOAs.  All that would be needed is the ability to compute velocities for the other body in the orbit.

Adding support for 3D motion would be even better.

scottransom:
@dlakaplan I don't think that the radial velocity of the pulsar, per se, gets used to correct TOAs.  But the ability to *get* the orbital radial velocity is certainly available.  If you wanted to use that extra information to do a joint fit, it would simply mean tweaking one of the current Fitter classes.  Should be doable quite easily.

dlakaplan:
OK, it may not be computed directly.  I wasn't necessarily thinking of using it in a joint fit: the precision and number of data points are usually so different from optical to timing that it is hard to do it at once.  But just to get the predicted RV of the companion for a given set of Time objects would define the basis function for optical fitting.

I guess the alternative is to just compute orbital phases (which I have done, although can't immediately recall) from within PINT and just fit for the amplitude.   That may ultimately be simpler.

paulray:
This is very closely related to issue #100

aarchiba:
Orbital radial velocities would be valuable. Would correcting them all to topocentric be of use? I guess the optical tools must already correct to barycentric velocities so all that is needed is the orbital part?

Worth noting is that we would get the radial velocity of the pulsar, and not the companion; for normal models the conversion factor may be simple (but for three- and more-body systems, for example those with planets, one would want the model itself to report companion velocities). 

More broadly, are models able to attach model-specific extra information to individual TOAs? If requested perhaps? Reflex motion of pulsar and/or companion, for example?

paulray:
Now that Scott merged model.orbital_phase() with all the anomalies it should be easy to add a couple utilities to compute line of sight velocity and acceleration, along the lines of `los_velocity()` and `los_acceleration()` here:
https://github.com/ewanbarr/pyorbit/blob/master/pyorbit/utils.py

Great little project for a student...

abhisrkckl:
Implemented in #1638

PR #1638, opened 2023-09-01
Radial velocities

dlakaplan:
#374 

Works for DD, DDS, BT, ELL1, ELL1H models.  Returns the RV of the pulsar or companion.

### DIFF
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -833,8 +833,8 @@ class TimingModel:
             bbi.a1() * np.sin(psi) * (1 - bbi.ecc() ** 2) / (1 + bbi.ecc() * np.cos(nu))
         )
 
-    def radial_velocity(self, barytimes):
-        """Return line-of-sight velocity of the pulsar at barycentric MJD times.
+    def pulsar_radial_velocity(self, barytimes):
+        """Return line-of-sight velocity of the pulsar relative to the system barycenter at barycentric MJD times.
 
         Parameters
         ----------
@@ -858,8 +858,7 @@ class TimingModel:
 
         Notes
         -----
-        This is the radial velocity of the pulsar.  For the radial velocity of the companion,
-        this must be multiplied by -1 times the mass of the pulsar divided by the mass of the companion.
+        This is the radial velocity of the pulsar.
 
         See [1]_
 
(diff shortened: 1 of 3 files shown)


abhisrkckl:
>Works for DD and ELL1 models

Does it work for variations of these models like ell1h?

dlakaplan:
> > Works for DD and ELL1 models
> 
> Does it work for variations of these models like ell1h?

I haven't tried it yet - that's why this is still WIP.

dlakaplan:
(also needs tests)

dlakaplan:
I think this is probably ready, unless there are other tests.

src/pint/models/timing_model.py
```diff
@@ -799,6 +799,87 @@ def orbital_phase(self, barytimes, anom="mean", radians=True):
...
+
+        Raises
+        ------
+        ValueError
+            If an astropy Time object is passed with scale!="tdb".
+
+        Returns
+        -------
+        array
+            The line-of-sight position
```
paulray, line 823: Can we explicitly say this position is with respect to the system barycenter (if that is indeed the case)?

src/pint/models/timing_model.py
```diff
@@ -799,6 +799,87 @@ def orbital_phase(self, barytimes, anom="mean", radians=True):
...
+
+        Returns
+        -------
+        array
+            The line-of-sight velocity
+
+        Notes
+        -----
+        This is the radial velocity of the pulsar.  For the radial velocity of the companion,
+        this must be multiplied by -1 times the mass of the pulsar divided by the mass of the companion.
```
paulray, line 862: Why not just add `companion_radial_velocity()` that takes the mass ratio as an argument?

paulray:
This is great! Thanks for adding it!

src/pint/models/timing_model.py
dlakaplan: done

src/pint/models/timing_model.py
dlakaplan: good idea.  done

src/pint/models/timing_model.py
```diff
@@ -799,6 +799,88 @@ def orbital_phase(self, barytimes, anom="mean", radians=True):
...
+        ]
+        bbi = b.binary_instance  # shorthand
+        psi = nu + bbi.omega()
+        return (
+            2
+            * np.pi
+            * bbi.a1()
+            / (bbi.pb() * np.sqrt(1 - bbi.ecc() ** 2))
+            * (np.cos(psi) + bbi.ecc() * np.cos(bbi.omega()))
+        ).cgs
```
abhisrkckl, line 846: Why is this in cgs rather than si?

src/pint/models/timing_model.py
```diff
@@ -799,6 +799,88 @@ def orbital_phase(self, barytimes, anom="mean", radians=True):
...
+
+        Notes
+        -----
+        This is the radial velocity of the pulsar.
+
+        See [1]_
+
+        .. [1] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.24
+        """
+        # this should also updaate the binary instance
```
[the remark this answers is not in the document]
dlakaplan: fixed

src/pint/models/timing_model.py
```diff
@@ -799,6 +799,88 @@ def orbital_phase(self, barytimes, anom="mean", radians=True):
         # return with radian units or return as unitless cycles from 0-1
         return anoms * u.rad if radians else anoms / (2 * np.pi)
 
+    def pulsar_radial_velocity(self, barytimes):
+        """Return line-of-sight velocity of the pulsar relative to the system barycenter at barycentric MJD times.
```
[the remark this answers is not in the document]
dlakaplan: I think "line-of-sight" or "radial". both is redundant

src/pint/models/timing_model.py
dlakaplan: because we are astronomers?  it still is a `Quantity` so a user can convert it as needed, but I wanted to get it out of the mix of light-seconds/day that it started in.

scottransom:
Looks great!  I'll 2nd Paul's thanks for doing this!

Issue #1641, opened 2023-09-08
`Quantity.decompose()` is slower than `Quantity.to()`


abhisrkckl:
I did a profiling run on PINT, and found that the performance issues of PINT arises from a large number of factors. One of the culprits was the frequent use of `Quantity.decompose()` function and `Quantity.to()` function with a string argument.

Here is a simple comparison:

`%timeit (dmu*DMconst).decompose()`
> 65.2 µs ± 798 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

`%timeit (dmu*DMconst).to("Hz")`
> 50.5 µs ± 495 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

`%timeit (dmu*DMconst).to(u.Hz)`
> 46 µs ± 274 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In this case, `decompose()` is slower by 40%.

I think we should change `decompose()` to `to()` wherever possible.

scottransom:
Optimizations with regards to units has come up multiple times in the past.  This is a very good page to take a look at:
https://docs.astropy.org/en/stable/units/index.html#astropy-units-performance
Note especially how useful it is to define composite units once and also to apply them using the `<<` operator rather than with a multiplication.  Those cause multiple orders of magnitude speed improvements.

abhisrkckl:
`%timeit (dmu*DMconst).to_value(u.Hz)`
> 43.6 µs ± 616 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

`%timeit (dmu*DMconst).to_value("Hz")`
> 47.4 µs ± 924 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

`%timeit (dmu*DMconst).to(u.Hz).value`
> 50.1 µs ± 2.71 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

`%timeit (dmu*DMconst).to("Hz").value`
> 52.9 µs ± 971 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

dlakaplan:
So maybe the issue is that new code doesn't obey the optimizations that were already decided on. I don't see recommendations in the coding style doc.

abhisrkckl:
> So maybe the issue is that new code doesn't obey the optimizations that were already decided on.

[...]

PR #1639, opened 2023-09-06
WIP: Fit white noise parameters

abhisrkckl:

Issue #1512, opened 2023-01-22
A way to enforce consistency between different time columns?


dlakaplan:
Is there a way overloading the setter or something so that changes to `toa['mjd']` propagate to `mjd_float`, the `tdb` columns, etc?

PR #1642, opened 2023-09-08
WIP: Performance improvements

abhisrkckl:
1. Change `Quantity.decompose()` to `Quantity.to()` --- see #1641 
2. Optionally cache parameter masks.
3. Minimize Quantity operations in `ScaleToaError.scale_toa_sigma`

Item 1 gives a slight performance advantage (~5%) overall.

Item 2 gives almost 10x improvement in timing models where maskParameters (e.g. JUMPs) are present. The caching is of course turned off by default, and can be enabled by calling `model.lock(toas)`. Mutating the `toas` object after calling `model.lock(toas)` leads to undefined behavior (in practice, some unhelpful exception down the line in the best case scenario, and wrong result in the worst case scenario).

Item 3 gives around 40% improvement in models where ScaleToaError is present.

These performance improvements are important for Bayesian timing & noise analysis (#1407) and frequentist noise estimation (#1639).

### DIFF
--- a/src/pint/models/astrometry.py
+++ b/src/pint/models/astrometry.py
@@ -203,10 +203,10 @@ class Astrometry(DelayComponent):
         rd = self.get_d_delay_quantities(toas)
 
         px_r = np.sqrt(rd["ssb_obs_r"] ** 2 - rd["in_psr_obs"] ** 2)
-        dd_dpx = 0.5 * (px_r**2 / (u.AU * const.c)) * (u.mas / u.radian)
+        dd_dpx = 0.5 * px_r**2 * ((u.mas / u.radian) / (u.AU * const.c))
 
         # We want to return sec / mas
-        return dd_dpx.decompose(u.si.bases) / u.mas
+        return dd_dpx.to(u.s) / u.mas
 
     def d_delay_astrometry_d_POSEPOCH(self, toas, param="", acc_delay=None):
         """Calculate the derivative wrt POSEPOCH"""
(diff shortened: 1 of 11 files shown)


abhisrkckl:
Performance comparison:
```
from pint.logging import setup as setup_log
from time import time

setup_log(level="WARNING")

m, t = get_model_and_toas("J1909-3744.par", "J1909-3744.basic_sim.tim")

res = Residuals(t, m)

repeat = 100
start = time()
for _ in range(repeat):
    res.calc_chi2()
end = time()

print(f"time elapsed without caching = {(end-start)/repeat:.2e} s")


m.lock(t)
res = Residuals(t, m)

repeat = 100
start = time()
for _ in range(repeat):
    res.calc_chi2()
end = time()

print(f"time elapsed with caching = {(end-start)/repeat:.2e} s")
```

> time elapsed without caching = 5.44e-03 s
> time elapsed with caching = 4.69e-04 s

dlakaplan:
Would using `@cache` or `@cached_property` be better than writing your own caching?

abhisrkckl:
> Would using @cache or @cached_property be better than writing your own caching?

That's probably a better idea. I'll try this.

abhisrkckl:
> If you are putting in changes to how units are handled, it would also make sense to add those guidelines to the coding style documents.

I am still experimenting with this and figuring out what's performance-critical and what's not. I am planning to make my changes as small as possible.

abhisrkckl:
> There are some other areas where modifying TOAs can require recalculating various things, such as in https://github.com/nanograv/PINT/issues/1512 . Rather than have an explicit lock() method, would it work to have a TOA timestamp/checksum associated with the TOAs object and then copy that into the model? Then this could be copied into the model and checked.

Wouldn't a checksum be expensive to compute, since it has to process the entire TOA object?

abhisrkckl:
I have decided that it will be better to divide this into multiple PRs focusing on different aspects of performance. So I am closing this.

PR #1644, opened 2023-09-25
Optimize usage of astropy.units

abhisrkckl:
#1641 

This is a superset of #1643. 
So this should be merged only after merging #1643.
I am doing it this way to make sure that the performance comparisons are meaningful.


dlakaplan:
If you want to do this, then I think https://nanograv-pint.readthedocs.io/en/latest/coding-style.html should be updated accordingly so we have a record of what to do.
(note that the coding style currently also allows python 2.7: should be fixed).

abhisrkckl:
Closing this because this is a superset of #1643, and that was found to be inadequate.

PR #1645, opened 2023-09-26
Simulate correlated noise 

abhisrkckl:
Simulate correlated noise realizations as present in the `TimingModel` object using `pint.simulation` and `zima`.

This brings some more of the `libstempo.toasim` functionality to  `pint.simulation`.

This works for *any* type of correlated noise, including EcorrNoise, PLRedNoise and PLDMNoise.

Also fixes bug #1649

### DIFF
--- a/src/pint/utils.py
+++ b/src/pint/utils.py
@@ -1328,8 +1328,9 @@ def wavex_setup(model, T_span, freqs=None, n_freqs=None):
             "Both freqs and n_freqs are specified. Only one or the other should be used"
         )
 
-    if n_freqs <= 0:
+    if n_freqs is not None and n_freqs <= 0:
         raise ValueError("Must use a non-zero number of wave frequencies")
+
     model.add_component(WaveX())
     if isinstance(T_span, u.quantity.Quantity):
         T_span.to(u.d)
@@ -1356,11 +1357,11 @@ def wavex_setup(model, T_span, freqs=None, n_freqs=None):
 
     if n_freqs is not None:
         if n_freqs == 1:
-            wave_freq = 2.0 * np.pi / T_span
+            wave_freq = 1 / T_span
             model.WXFREQ_0001.quantity = wave_freq
         else:
             wave_numbers = np.arange(1, n_freqs + 1)
-            wave_freqs = 2.0 * np.pi * wave_numbers / T_span
+            wave_freqs = wave_numbers / T_span
             model.WXFREQ_0001.quantity = wave_freqs[0]
             model.components["WaveX"].add_wavex_components(wave_freqs[1:])
     return model.components["WaveX"].get_indices()
@@ -1408,8 +1409,9 @@ def dmwavex_setup(model, T_span, freqs=None, n_freqs=None):
             "Both freqs and n_freqs are specified. Only one or the other should be used"
         )
 
-    if n_freqs <= 0:
+    if n_freqs is not None and n_freqs <= 0:
         raise ValueError("Must use a non-zero number of wave frequencies")
+
     model.add_component(DMWaveX())
     if isinstance(T_span, u.quantity.Quantity):
         T_span.to(u.d)
@@ -1436,11 +1438,11 @@ def dmwavex_setup(model, T_span, freqs=None, n_freqs=None):
 
     if n_freqs is not None:
         if n_freqs == 1:
-            wave_freq = 2.0 * np.pi / T_span
+            wave_freq = 1 / T_span
             model.DMWXFREQ_0001.quantity = wave_freq
         else:
             wave_numbers = np.arange(1, n_freqs + 1)
-            wave_freqs = 2.0 * np.pi * wave_numbers / T_span
+            wave_freqs = wave_numbers / T_span
             model.DMWXFREQ_0001.quantity = wave_freqs[0]
             model.components["DMWaveX"].add_dmwavex_components(wave_freqs[1:])
     return model.components["DMWaveX"].get_indices()


abhisrkckl:
Here is a noise realization created using PINT (X axis in MJD, Y axis in s):
image

And here is the ENTERPRISE analysis result (blue lines represent injected values):
image

CHANGELOG-unreleased.md
```diff
@@ -20,6 +20,7 @@ the released changes.
 - Added radial velocity methods for binary models
 - Support for wideband data in `pint.bayesian` (no correlated noise).
 - Added `DMWaveX` model (Fourier representation of DM noise)
+- Simulate correlated noise using `pint.simulate` functions and the `zima` script.
```
dlakaplan, line 23: `pint.simulation`, not `pint.simulate` (I think).

Also, is zima needed?  Or should it be clear that this works within zima too, but can work fine just using `pint.simulation`?

tests/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.ipynb
+chains/*
```
dlakaplan, line 2: are these actually generally needed?

dlakaplan:
Do you actually have any tests which look at the nature of the simulated noise?  Or just that make sure there are simulated TOAs returned?  Fitting the results may not be simple but maybe you can compare simulated TOAs with and without correlated noise to make sure they are different in the right way?  Or some other test?

CHANGELOG-unreleased.md
abhisrkckl: corrected.

abhisrkckl: Corrected.

tests/.gitignore
abhisrkckl: Removed.

tests/test_zima.py
```diff
@@ -92,3 +94,41 @@ def test_zima_fuzzdays(tmp_path):
...
+    ftr.fit_toas()
+    rc = sum(ftr.resids.noise_resids.values())
+    r = ftr.resids.time_resids
+    rw = r - rc
+    sigma = ftr.resids.get_data_error()
+
+    # This should be independent and standard-normal distributed.
+    x = (rw / sigma).to_value("")
+    assert np.isclose(np.std(x), 1, atol=0.2)
+    assert np.isclose(np.mean(x), 0, atol=0.01)
```
abhisrkckl, line 134: > Do you actually have any tests which look at the nature of the simulated noise? Or just that make sure there are simulated TOAs returned? Fitting the results may not be simple but maybe you can compare simulated TOAs with and without correlated noise to make sure they are different in the right way? Or some other test?

@dlakaplan Is this test sufficient? It is checking if the simulated TOAs can be whitened using the original model.

dlakaplan: I think that's reasonable.  I guess I don't know if the different noise realizations would play into it here: do you know that the phase of the red noise will be the same?

dlakaplan: Maybe I don't understand then.  If each realization has a different phase, how can you know that this will properly whiten the data?

abhisrkckl: I carefully went through the `DownhillGLSFitter` code. My understanding is as follows:

When we call `DownhillGLSFitter.fit_toas()`, it effectively fits the correlated noise amplitudes (such as Fourier amplitudes for PLRedNoise) together with the timing model parameters. While the TimingModel object doesn't know about these amplitudes, the `DownhillGLSFitter` object knows about them via the `DownhillGLSFitter.current_state.xhat` property. This is a vector that contains the timing model parameter deviations along with the correlated noise amplitudes. The `noise_resids` property is computed by taking the inner product of these amplitudes with the correlated noise basis matrix. 

So, in effect, the `DownhillGLSFitter` object *does* know about the basis amplitudes, and hence the correlated noise phase.

dlakaplan: So in that case, are you actually sure that you are whitening the data with the input model?  Can you check that what the fitter finds is the same (or close to) what you injected?

abhisrkckl: Here is a comparison between injected and recovered red noise.
image


For posterity, here is the code for obtaining the above plot:
```
from pint.fitter import DownhillGLSFitter
from astropy import time
import numpy as np

parfile = datadir / "B1855+09_NANOGrav_9yv1.gls.par"
m = get_model(parfile)

m.remove_component("ScaleToaError")
m.remove_component("EcorrNoise")
m.remove_component("DispersionDMX")
m.remove_component("PhaseJump")
m.remove_component("FD")
m.PLANET_SHAPIRO.value = False

m.F0.frozen = True
m.F1.frozen = True

t = make_fake_toas_uniform(
	m.START.value,
	m.FINISH.value,
	1000,
	m,
	add_noise=True,
)

U = m.noise_model_designmatrix(t)
b = m.noise_model_basis_weight(t)
rn_inj = (U @ (b**0.5 * np.random.normal(size=len(b)))) << u.s
t.adjust_TOAs(time.TimeDelta(rn_inj))

res = Residuals(t, m)

ftr = DownhillGLSFitter(t, m)
ftr.fit_toas()

rn_fit = sum(ftr.resids.noise_resids.values())

plt.errorbar(t.get_mjds(), res.time_resids, t.get_errors(), label="Pre-fit Residuals", ls="")
plt.plot(t.get_mjds(), rn_inj, label="Injected red noise")
plt.plot(t.get_mjds(), rn_fit, label="Recovered red noise")
plt.legend()
plt.xlabel("MJD")
plt.ylabel("residuals (s)")
plt.show()
```

abhisrkckl: I think the recovered signal is consistent with the injected signal. Note that I am freezing F0 and F1 in the above case so that fitting for them don't distort the recovered red noise. 

In practice, the red noise recovery will be distorted by the F0 & F1 fit like so:
image

dlakaplan: OK, that looks good.  And can you also verify that if you simulate or fit using different noise models you don't get good residuals?

abhisrkckl: Here is what happens when I set a wrong value for PLREDAMP before fitting.
image

For comparison, the fit with correct values for the same dataset: 
image

dlakaplan: OK, I think this all makes sense.

abhisrkckl:
Here is the result of a roundtrip fitting test.
1. Simulate a PLRedNoise
2. Fit using WaveX
3. Compute the PSD from WaveX amplitudes
4. Fit a powerlaw to the measured PSD
5. Compare against the injected PSD

image

PR #1647, opened 2023-10-02
Derivatives of scaled TOA uncertainty wrt white noise parameters

abhisrkckl:
This is useful for speeding up #1639 using a derivative-aware algorithm.


abhisrkckl:
Closing this because this has been merged with #1639.

PR #1648, opened 2023-10-04
Sensibly handle unfittable params, No unfittable params in `plk`

abhisrkckl:
Unfittable parameters are parameters for which no derivatives are defined so that they cannot be fit.
1. Fitter objects should not be created any unfittable parameters are unfrozen. Currently, it fails with an unclear error message while calling `fit_toas()`.
2. `TimingModel.designmatrix()` should fail with an informative error message upon encountering free unfittable parameters. Currently, it fails with an unclear error message.
3. `pintk` interface shouldn't list unfittable parameters. This will help unclutter `pintk`.


dlakaplan:
The errors here might be due to function parameters which are visible but not intrinsically part of the model.  So maybe just a check for parameter type as well.

abhisrkckl:
> The errors here might be due to function parameters which are visible but not intrinsically part of the model. So maybe just a check for parameter type as well.

A subset of the errors were indeed indirectly related to funcParameters. But they were actually due to a bug in the ELL1k model which I have fixed.

abhisrkckl:
Plk interface before excluding unfittable params:
image

After:
image

dlakaplan:
This looks good.  Is it ready?

abhisrkckl:
Yes.

PR #1650, opened 2023-10-12
Fix versioneer

paulray:
Upgraded for compatibility with Python 3.12 (revealed in conda package processing)


paulray:
This passes so I think it can be merged. To really test it, we need to tag a new version and see if it works on conda-forge.

PR #1652, opened 2023-10-13
Update to python 3.12 for testing

dlakaplan:


dlakaplan:
https://docs.python.org/3/whatsnew/3.12.html
```This means that distutils, setuptools, pkg_resources, and easy_install will no longer available by default; to access these run pip install setuptools in the [activated](https://docs.python.org/3/library/venv.html#venv-explanation) virtual environment.```
so have to add these explicitly.

dlakaplan:
@paulray : this required a couple of tweaks because of how `setuptools` is now included.  But I think it works.  Should we change our default tests to use 3.12 or just keep it as a demo?

paulray:
Well, we have oldestdeps on Python 3.8 and the oldest packages we support, so I think it makes sense to have the tests also run on the most recent versions to catch any compatibility issues that crop up. I'd go with 3.12 for our tests.

abhisrkckl:
I'll merge this if ready.

dlakaplan:
I think so.

PR #1655, opened 2023-10-19
Added elongation for `pintk`

dlakaplan:
Screen Shot 2023-10-19 at 10 37 00 AM


abhisrkckl:
What happens in the case of a timing model without an Astrometry component? (e.g. for already barycentered TOAs)

dlakaplan:
I think that is a separate issue.  The buttons for e.g., `orbital phase` and `frequency` are present even for data that don't have a binary and don't have frequency info.  Pressing those leads to non-fatal errors.  Probably that should be fixed separately.

abhisrkckl:
I'll merge this if this is ready.

dlakaplan:
I think it's ready.  I added #1658 to deal with the other cases you asked about

PR #1663, opened 2023-10-23
Disabled axes for pintk when not in timing model or data

dlakaplan:
* Disabled "elongation" axes when no Astrometry component present
* Disabled "orbital phase" axes when no Binary component present
* Disabled "frequency" axes when <=1 frequency value present or when frequencies are not finite
* Also changed default `pintk` fitters to downhill versions (#1665)
* Also changed general `is_binary` property to rely on class inheritance rather than component name

Example on a barycentered dataset (so no Astrometry or Binary component and all event data):
Screen Shot 2023-10-23 at 12 43 27 PM

### DIFF
--- a/src/pint/pintk/plk.py
+++ b/src/pint/pintk/plk.py
@@ -14,6 +14,8 @@ from matplotlib.backends.backend_tkagg import FigureCanvasTkAgg
 
 import pint.pintk.pulsar as pulsar
 import pint.pintk.colormodes as cm
+from pint.models.pulsar_binary import PulsarBinary
+from pint.models.astrometry import Astrometry
 
 import tkinter as tk
 import tkinter.filedialog as tkFileDialog
@@ -529,14 +531,16 @@ class PlkXYChoiceWidget(tk.Frame):
                 if self.master.psr.fitted
                 else self.master.psr.prefit_model
             )
-            if choice == "elongation":
-                if not any(x.startswith("Astrometry") for x in model.components):
-                    self.xbuttons[ii].configure(state="disabled")
-                    self.ybuttons[ii].configure(state="disabled")
-            if choice == "orbital phase":
-                if not any(x.startswith("Binary") for x in model.components):
-                    self.xbuttons[ii].configure(state="disabled")
-                    self.ybuttons[ii].configure(state="disabled")
+            if choice == "elongation" and not any(
+                isinstance(x, Astrometry) for x in model.components
+            ):
+                self.xbuttons[ii].configure(state="disabled")
+                self.ybuttons[ii].configure(state="disabled")
+            elif choice == "orbital phase" and not any(
+                isinstance(x, PulsarBinary) for x in model.components
+            ):
+                self.xbuttons[ii].configure(state="disabled")
+                self.ybuttons[ii].configure(state="disabled")
             if choice == "frequency" and (
                 (len(np.unique(self.master.psr.all_toas["freq"])) <= 1)
                 or np.any(np.isinf(self.master.psr.all_toas["freq"]))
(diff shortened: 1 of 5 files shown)


abhisrkckl:
> Disabled "frequency" axes when <=1 frequency value present

I think this should also be disabled when infinite-frequency TOAs are present.

dlakaplan:
Done.  also fixed a bug regarding initial log level

src/pint/pintk/plk.py
```diff
@@ -524,6 +524,39 @@ def setChoice(self, xid="mjd", yid="pre-fit"):
...
+                    if not self.master.psr.fitted:
+                        self.ybuttons[ii].configure(state="disabled")
+                if self.master.psr.fitted and not any(
+                    x.startswith("Binary")
+                    for x in self.master.psr.prefit_model.components
+                ):
+                    self.ybuttons[ii].configure(state="disabled")
+            if choice == "frequency" and (
+                (len(np.unique(self.master.psr.all_toas["freq"])) <= 1)
+                or np.any(np.isinf(self.master.psr.all_toas["freq"]))
```
abhisrkckl, line 555: Shoudn't this be 
`np.all(np.isfinite(self.master.psr.all_toas["freq"]))`
?

dlakaplan: Isn't that the same thing?  `any(isinf)` vs `all(finite)`?

abhisrkckl: No. any(isinf) == not all(finite)
(de Morgan's identity)

dlakaplan: but here we are excluding data with any infinite frequencies.

abhisrkckl: OK.. My bad. I thought the code was conditionally enabling the buttons instead of disabling them.

src/pint/pintk/plk.py
```diff
@@ -524,31 +524,18 @@ def setChoice(self, xid="mjd", yid="pre-fit"):
...
-                    x.startswith("Astrometry")
-                    for x in self.master.psr.pstfit_model.components
-                ):
                     self.ybuttons[ii].configure(state="disabled")
             if choice == "orbital phase":
-                if not any(
-                    x.startswith("Binary")
-                    for x in self.master.psr.prefit_model.components
-                ):
+                if not any(x.startswith("Binary") for x in model.components):
```
abhisrkckl, line 537: Wouldn't it be more robust to have `isinstance(model.components[x], PulsarBinary)`?

src/pint/pintk/plk.py
```diff
@@ -524,31 +524,18 @@ def setChoice(self, xid="mjd", yid="pre-fit"):
...
+                self.master.psr.postfit_model
+                if self.master.psr.fitted
+                else self.master.psr.prefit_model
+            )
             if choice == "elongation":
-                if not any(
-                    x.startswith("Astrometry")
-                    for x in self.master.psr.prefit_model.components
-                ):
+                if not any(x.startswith("Astrometry") for x in model.components):
```
abhisrkckl, line 533: Wouldn't it be more robust to have isinstance(model.components[x], Astrometry)?

src/pint/pintk/plk.py
dlakaplan: possibly.  it would involve more importing.  I'll see what it looks like

scottransom: Note also that the timing model has a `.is_binary` parameter. That is IMO the proper way to do that check (and where we should be making smarter code for identifying whether a model is a binary or not).

abhisrkckl:
Can you also change the `TimingModel.is_binary` method to use `isinstance` instead of a string-based condition?

dlakaplan:
I did that.  Not sure it's better (it requires a local import to avoid circular import errors).

dlakaplan:
I had to undo those changes because of various failures

abhisrkckl:

dlakaplan:
I think so

PR #1666, opened 2023-10-28
Compute the noise uncertainties only once

abhisrkckl:
The most expensive part of white noise fit is the uncertainty computation using the numerical Hessian. 
This PR removes unnecessary repetitions of this step.


abhisrkckl:
Execution time for `DownhillWLSFitter.fit_toas()` with free white noise parameters:
Before - 
After - 1.2 s

PR #1670, opened 2023-11-02
More realistic simulation of narrowband TOAs of multiple frequencies

abhisrkckl:
`multi_freqs_in_epoch` option in `make_fake_toas_uniform` and `make_fake_toas_fromMJDs`.
`--multifreq` option in `zima`
Moved some tests from `test_zima.py` to `test_fake_toas.py`

Note that the default behavior doesn't change.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -20,6 +20,7 @@ the released changes.
 - Update CI tests for Python 3.12
 - Made `test_grid` routines faster
 - `pintk` uses downhill fitters by default
+- Optionally generate multi-frequency TOAs in an epoch using `make_fake_toas_uniform`
 ### Added
 - CHI2, CHI2R, TRES, DMRES now in postfit par files
 - Added `WaveX` model as a `DelayComponent` with Fourier amplitudes as fitted parameters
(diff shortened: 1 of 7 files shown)


dlakaplan:
Can you give an example of how this works here?

dlakaplan:
And can you also add in the docstrings a more explicit explanation of the two modes?

dlakaplan:
Basically if you add in the way you explained this in the issue, that would be great.

abhisrkckl:
> Can you give an example of how this works here?

I have added a new example script `simulation_example.py` which demonstrates the various capabilities of the `pint.simulate` module.


> And can you also add in the docstrings a more explicit explanation of the two modes?
> Basically if you add in the way you explained this in the issue, that would be great.

I have expanded the docstrings with this explanation.

dlakaplan:
Should your new simulation notebook be included in the RTD examples?  I don't think it is yet.

dlakaplan:
Actually I'm not sure about that.  It isn't in the `tutorials.rst` but I still see it at https://nanograv-pint--1670.org.readthedocs.build/en/1670/examples/simulation_example.html

abhisrkckl:
It's there in `tutorials.rst`

dlakaplan:
OK, good.  I didn't see that before in the changed files for whatever reason.  I can then merge this when you are ready.

abhisrkckl:
It can be merged once the tests pass.

PR #1671, opened 2023-11-03
Make whitened residuals available

abhisrkckl:
This partially overlaps with the stale PR #1000.

whitened_resids = (resids - noise_resids) / scaled_toa_error

If the fitting has been done properly, the whitened residuals should be normally distributed with unit variance.

### DIFF
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -500,6 +500,12 @@ class Residuals:
             )
         return (phase_resids / self.get_PSR_freq(calctype=calctype)).to(u.s)
 
+    def calc_whitened_resids(self):
+        r = self.calc_time_resids()
+        nr = sum(self.noise_resids.values())
+        sigma = self.get_data_error()
+        return (r - nr) / sigma
+
     def _calc_gls_chi2(self, lognorm=False):
         """Compute the chi2 when correlated noise is present in the timing model.
         If the system is not singular, it uses Cholesky factorization to evaluate this.
--- a/src/pint/residuals.py
+++ b/src/pint/residuals.py
@@ -504,7 +504,7 @@ class Residuals:
         r = self.calc_time_resids()
         nr = sum(self.noise_resids.values())
         sigma = self.get_data_error()
-        return (r - nr) / sigma
+        return ((r - nr) / sigma).to(u.dimensionless_unscaled)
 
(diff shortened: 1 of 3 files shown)


dlakaplan:
Does this work for WB data?

src/pint/residuals.py
```diff
@@ -500,6 +500,12 @@ def calc_time_resids(
             )
         return (phase_resids / self.get_PSR_freq(calctype=calctype)).to(u.s)
 
+    def calc_whitened_resids(self):
```
dlakaplan, line 503: Is it better to have a separate method, or `calc_time_resids()` with a flag?

tests/test_residuals.py
```diff
@@ -438,3 +438,46 @@ def test_resid_mean_phase():
     mean, _ = pint.residuals.weighted_mean(full, w)
     assert mean == r.calc_phase_mean()
     assert r.calc_phase_mean() == r2.calc_phase_mean()
+
+
+@pytest.mark.parametrize(
```
dlakaplan, line 443: This is fine I guess, but is it necessary if you aren't actually iterating over multiple inputs?

abhisrkckl:
> Does this work for WB data?

It should. But I haven't tried it yet.

src/pint/residuals.py
abhisrkckl: I think it should be a separate method since whitened residuals are dimensionless, whereas time resids have the dimentions of time.

tests/test_residuals.py
abhisrkckl: The idea here is to make sure it works without and with red noise.

dlakaplan: ok, I had mis-read it.  I now see there are 2 versions.

src/pint/residuals.py
dlakaplan: Then probably add a "See Also" to the other residual methods

abhisrkckl: OK.

abhisrkckl: Updated documentation.

dlakaplan:
Looks good to me.  ready to go?

abhisrkckl:
> Looks good to me. ready to go?

Yes.

PR #1675, opened 2023-11-14
Docstring for `TimingModel.get_deriv_funcs` 

abhisrkckl:


dlakaplan:
Seems fine.  Does it need a CHANGELOG entry?

abhisrkckl:
I don't think so. This is really minor.

PR #1677, opened 2023-11-15
Remove broken example `MCMC_walkthrough.broken`

abhisrkckl:


dlakaplan:
Actually I think this is easy to fix

dlakaplan:
#1679  at least allows the notebook to run.  I don't know that it is returning the expected results.

abhisrkckl:
This is being fixed in #1679. So I am closing this.

PR #1679, opened 2023-11-16
WIP: fix `MCMC_walkthrough.ipynb`

dlakaplan:
This notebook runs, although I'm not sure if the results are sensible/as intended.

### DIFF
--- a/docs/examples/MCMC_walkthrough.py
+++ b/docs/examples/MCMC_walkthrough.py
@@ -1,0 +1,210 @@
+# ---
+# jupyter:
+#   jupytext:
+#     formats: ipynb,py:percent
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.13.8
+#   kernelspec:
+#     display_name: Python 3 (ipykernel)
+#     language: python
+#     name: python3
+# ---
+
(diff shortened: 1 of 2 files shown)


dlakaplan:
https://nanograv-pint--1679.org.readthedocs.build/en/1679/examples/MCMC_walkthrough.html

dlakaplan:
Strangely, when I run it locally I don't get the same errors and I don't get a good phase-o-gram

dlakaplan:
I thought I fixed the second error that appears by the phaseogram, but it's still there in the RTD.

dlakaplan:
So I fixed the syntax errors and now don't get good phase-o-grams.  But at least now I see consistent results.

dlakaplan:
Also, the second example discussed using a phase term in the likelihood (and hard a hard bound on the prior for that).  But I don't see it in the model.  With the phase prior removed it now doesn't give `-inf` for the likelihood.

scottransom:
The absolute phase information for the fit used to be tacked on invisibly at the end of the parameter vectors (that's why we loop up to `:-1` for regular params).  But now absolute phase is handled in PINT.  So maybe that needs to be changed now.

abhisrkckl:

dlakaplan:
Yes, please.

PR #1681, opened 2023-11-17
Set `UNITS` to TDB if it is not given in the par file.

abhisrkckl:
TDB is assumed implicitly if no UNITS is given. This PR makes it explicit.


dlakaplan:
Looks fine.  Ready to merge?

abhisrkckl:
Yes.

PR #1682, opened 2023-11-17
Simulate TOA flags

abhisrkckl:
The infrastructure for doing this is already there in `pint.toa` and is straightforward to use. This PR just uses that in `pint.simulation`.

### DIFF
--- a/src/pint/simulation.py
+++ b/src/pint/simulation.py
@@ -190,6 +190,7 @@ def make_fake_toas_uniform(
     include_bipm=False,
     include_gps=True,
     multi_freqs_in_epoch=False,
+    flags=None,
 ):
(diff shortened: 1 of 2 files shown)


dlakaplan:
This is fine, but is it needed?  You can just modify the TOAs after creation like:
`t['be']='GUPPI'` 
(from your test).

dlakaplan:
Maybe an addition to the docs explaining some of the flag access methods would be more useful.  but I guess if you want to keep this too then it''s fine and I can merge it.

abhisrkckl:
> Maybe an addition to the docs explaining some of the flag access methods would be more useful.

I agree. I can add that in a later PR. 

This one is ready to merge.

Issue #1660, opened 2023-10-23
`PLDMNoise` doesn't work correctly with wideband TOAs


abhisrkckl:
`PLDMNoise` is a GP representation of DM variations. Hence, while computing the wideband residuals, it should be applied to both TOAs *and* DMs. However, we don't have the framework to do this properly for a correlated noise component, and thus `PLDMNoise` is only applied to TOAs. This will lead to wrong results while analyzing wideband data.

abhisrkckl:
@blarsen10

blarsen10:
My limited understanding of the other DM modeling features in PINT is that the `DispersionDM` and `DispersionDMX` classes apply only to ToAs, while the `ScaleDmError` and `DispersionJump` classes apply only to wideband DMs. My intention when adding `PLDMNoise` was that it would behave similarly to `DispersionDM` and `DispersionDMX`, which means applying to ToAs but not wideband DMs. However, if my understanding is wrong and the `DispersionDM` and `DispersionDMX` classes also work with wideband ToAs, then maybe `PLDMNoise` should also do so in a similar way. Is that the case here? Or is there some other reason why this `PLDMNoise` should be made to work with wideband DMs? Apologies that I am not up to speed on the requirements for wideband analyses right now.

abhisrkckl:
I am working with @vhaasteren on doing this properly.

abhisrkckl:
Not fully implemented in [#1868](https://github.com/nanograv/PINT/pull/1868)

PR #1694, opened 2023-12-15
Convert `WaveX` to `PLRedNoise` and `DMWaveX` to `PLDMNoise`

abhisrkckl:
`WaveX` and `PLRedNoise` are both Fourier representations of achromatic red noise. The difference is that `WaveX` is a deterministic model with free Fourier coefficients whereas `PLRedNoise` is a GP model where the Fourier coefficients have been analytically marginalized. The `WaveX` coefficients can be converted to `PLRedNoise` spectral parameters (amplitude and spectral index) by maximizing the following likelihood function:
image
Here, a_j and b_j are the `WaveX` coefficients, epsilon represents their measurement uncertainties, and P(f) is a power law spectrum.

This likelihood function has a different statistical interpretation than the analytically marginalized likelihood used for Bayesian analysis (it acts on timing residuals directly). Hence, its results are not expected to be fully consistent with the Bayesian results. Nevertheless, this conversion gives results fairly close to the Bayesian estimates (mostly within the 3-sigma level in the tests I have done). 

The inverse `PLRedNoise` --> `WaveX` conversion without using the TOAs is not possible since `PLRedNoise` has less information than `WaveX`. 

A similar relationship and conversion also apply to `DMWaveX` and `PLDMNoise`. 
Please note that this does not work properly for wideband data due to bug #1660.

Also includes some sourcery refactors.


abhisrkckl:
Red noise conversion example (simulation):
image

abhisrkckl:
DM noise conversion example (simulation):
image

src/pint/utils.py
```diff
@@ -2764,3 +2774,170 @@ def woodbury_dot(Ndiag, U, Phidiag, x, y):
...
+        return 0.5 * np.sum(
+            (a**2 / (sigma**2 + da**2))
+            + (b**2 / (sigma**2 + db**2))
+            + np.log(sigma**2 + da**2)
+            + np.log(sigma**2 + db**2)
+        )
+
+    gamma_val, log10_A_val = minimize(mlnlike, [4, -13], method="Nelder-Mead").x
+
+    hess = Hessian(mlnlike)
```
kerrm, line 2843: Do there need to be any checks on the numerical results here?  (Positive definitiveness of the Hessian, positivity of the the resulting inverse, order of magnitude sanity check on uncertainty?)

src/pint/utils.py
```diff
@@ -2764,3 +2774,170 @@ def woodbury_dot(Ndiag, U, Phidiag, x, y):
...
+    fs = np.array(
+        [model[f"DMWXFREQ_{idx:04d}"].quantity.to_value(u.Hz) for idx in idxs]
+    )
+    f0 = np.min(fs)
+
+    assert np.allclose(
+        np.diff(np.diff(fs)), 0
+    ), "`pldmnoise_from_dmwavex` requires the DMWaveX frequencies to be uniformly spaced."
+
+    def powl_model(params):
```
kerrm, line 2910: Maybe make these external functions?  They seem common between the two methods and so redundancy could be reduced.  (The same thing applies to the error estimation at the end -- a general method to extra uncertainties on the PL parameters could be used for both functions.

abhisrkckl: I have refactored the repeating code into a separate function.

src/pint/utils.py
abhisrkckl: > order of magnitude sanity check on uncertainty?

This may not be practical, since there may be datasets where the parameters are not well-constrained, with the large uncertainties reflecting that.

abhisrkckl: > positive definitiveness of the Hessian, positivity of the the resulting inverse

I have added assertions for this.

dlakaplan:
This looks good, but it looks like the fit comparisons are only done in the example and not in the tests.  Are additional tests possible?

abhisrkckl:
@dlakaplan I have added quantitative tests.

src/pint/utils.py
```diff
@@ -1050,25 +1046,24 @@ def dmxparse(fitter, save=False):
         if isinstance(save, bool):
             save = "dmxparse.out"
         DMX = "DMX"
-        lines = []
-        lines.append("# Mean %s value = %+.6e \n" % (DMX, DMX_mean))
-        lines.append("# Uncertainty in average %s = %.5e \n" % ("DM", DMX_mean_err))
-        lines.append(
+        lines = [
```
dlakaplan, line 1049: not f-strings?

src/pint/utils.py
```diff
@@ -2764,3 +2741,234 @@ def woodbury_dot(Ndiag, U, Phidiag, x, y):
...
+                (len(model1.components[component].get_indices()) + 1) / T_span,
+                frozen=False,
+            )
+        else:
+            model1.components[component].add_dmwavex_component(
+                (len(model1.components[component].get_indices()) + 1) / T_span,
+                frozen=False,
+            )
+
+    return np.argmin(aics), np.array(aics) - np.min(aics)
```
dlakaplan, line 2974: Any tests if there are NaN's in this?  Or if more than one value is at the minimum (probably indicative of an error)?

src/pint/utils.py
abhisrkckl: changed to fstrings

src/pint/utils.py
```diff
@@ -2764,3 +2741,234 @@ def woodbury_dot(Ndiag, U, Phidiag, x, y):
...
+    Returns
+    -------
+    pint.models.timing_model.TimingModel
+        The timing model with a converted `PLRedNoise` component.
+    """
+    from pint.models.noise_model import PLRedNoise
+
+    mlnlike = _get_wx2pl_lnlike(model, "WaveX", ignore_fyr=ignore_fyr)
+
+    gamma_val, log10_A_val = minimize(mlnlike, [4, -13], method="Nelder-Mead").x
```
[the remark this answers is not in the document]
abhisrkckl: added convergence checks

src/pint/utils.py
abhisrkckl: Added check.

abhisrkckl: > Or if more than one value is at the minimum (probably indicative of an error)

This scenario is not an error, although it is very unlikely. As I understand it, it just means that two Nharm values are equally good. In this case, `np.argmin` will return the earliest occurrence of the minimum, which corresponds to a lower Nharm value. I think this is the correct behavior.

abhisrkckl:
@dlakaplan Can you please merge this if it looks fine?

src/pint/utils.py
dlakaplan: My point is that if they are close but not equal, that's OK.  but if they are exactly equal then they are probably 0 or something like that, so potentially an error.

abhisrkckl: I don't think that will lead to an error. All that we do with AIC values is that we take their minimum (using argmin) to find the best model. We don't do any other calculation with it. As long as it is a finite value (i.e., not nan or inf), it should be fine even if two values happen to be equal.

dlakaplan: But if they are exactly equal, it seems unlikely that this happens with good data values.  But if you don't think that's an issue, then I won't worry.

dlakaplan: wouldn't it also return an array of indices?  I think that would break your test, although maybe it's not used elsewhere

abhisrkckl: `np.argmin` returns the first occurrance only.

For example, `np.argmin([2,4,5,1,2,1])` returns 3.

PR #1701, opened 2023-12-26
Better exceptions for unimplemented binary models (BTX etc)

abhisrkckl:
Improved error messages
1. Suggest changing BTX -> BT and DDFWHE -> DDH
2. State that MSS, BT1P etc are not *yet* implemented (see https://tempo.sourceforge.net/ref_man_sections/binary.txt)
3. Default error message otherwise

Also has some sourcery refactors


tests/test_model.py
```diff
@@ -92,9 +92,6 @@ def test_model():
     # run tempo1 also, if the tempo_utils module is available
     did_tempo1 = False
     with contextlib.suppress(Exception):
-        import tempo_utils
-
-        log.info("Running TEMPO1...")
```
abhisrkckl, line 97: It is not actually running tempo, but reading outputs from a text file.

src/pint/models/timing_model.py
```diff
@@ -3827,6 +3823,26 @@ def search_binary_components(self, system_name):
...
+                raise UnknownBinaryModel(
+                    "`BINARY  DDFWHE` is not supported, but the same model "
+                    "is available as `BINARY  DDH`."
+                )
+            elif system_name in ["MSS", "EH", "H88", "DDT", "BT1P", "BT2P"]:
+                # Binary model list taken from
+                # https://tempo.sourceforge.net/ref_man_sections/binary.txt
+                raise UnknownBinaryModel(
+                    f"`The binary model {system_name} is not yet implemented."
+                )
```
abhisrkckl, line 3844: This is the functional change in this PR.

dlakaplan:
This looks reasonable.  Can merge when you are ready.

abhisrkckl:
Yes.. This is ready to go.

PR #1706, opened 2023-12-28
CI ephemeris test no longer requires access to static NANOGrav site

dlakaplan:
In general the ephemeris downloader is smart enough to look at astropy first and then a list of mirror sites.  But in `check_ephemeris_connection.py` it also had a command to look explicitly at `data.nanograv.org`.  This meant that any disruption to that site would break CI.

This removes that one command.  The remainder of the ephemeris files are still searched using the standard method.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -10,6 +10,7 @@ the released changes.
 ## Unreleased
 ### Changed
 - Moved `get_derived_params` to `timing_model`
+- `check_ephemeris_connection` CI test no longer requires access to static NANOGrav site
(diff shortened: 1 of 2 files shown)


dlakaplan:
Was there a reason that:
```
urllib.request.urlopen("https://data.nanograv.org/static/data/ephem/de405.bsp").read()
```
was used in addition to `pint.solar_system_ephemerides.load_kernel('de405')`?

abhisrkckl:
Looks fine. Shall I merge this?

dlakaplan:
Yes, I think this is ready.

Issue #1691, opened 2023-12-10
Clock corrections applied to TZRMJD even if TZRSITE is BAT


gcerib:
Dear all,

I am in the field of very-high energy pulsars with imaging Cherenkov telescopes such as MAGIC and CTA. Since we typically have to integrate tens of hours of observations (scattered across many months) to detect the pulsed gamma rays and build a TOA, we are forced to apply the barycenter corrections to each single _event_ we record, with most of them being just background cosmic rays, and obtain the pulse profile from already barycentered data. I find convenient for me to produce par files where TZRSITE is set to "@" or "bat", so that I can manipulate the phase=0 definition in a relatively easy way by changing TZRMJD, and adapt it to various different references present in literature.

I think I've found an inconsistency between the behaviour of PINT and tempo2 in the specific case in which TZRSITE is set to "@" and the clock is set to TT(BIPM****). Specifically, PINT applies the clock corrections to TZRMJD, whereas tempo2 doesn't. This causes the absolute phases calculated by tempo2 and PINT to differ by ~(-27us/F0), even when the BATs are identical down to the precision of np.longdouble. I was suggested to report it here by Paul Ray over the facebook pulsar group.

I believe I've identified the origin of the discrepancy in the definition of get_TZR_toa(self, toas) in $PINT/src/pint/models/absolute_phase.py and the respective function refphs_init(pulsar* psr, int npsr) in $TEMPO2/refphs.C .
The tempo2 function contains:
```
21             if (obs->telID[0]=='@'            /* Have barycentric arrival time */
22                     || strcasecmp(obs->telID,"bat")==0)
23             {
24                 obs->clockCorr=0;  /* therefore don't do clock corrections */
25                 obs->delayCorr=0;
26             }
```
whereas the PINT function does not have such a check. The docstring of get_TZR_toa() even mentions that "We are treating the TZRMJD as a special TOA. Note that any observatory clock corrections will be applied to this TOA, as with any other TOA. This does not affect the value of the TZRMJD parameter, however." Indeed, if I add such a check on the value of self.TZRSITE in get_TZR_toa() of absolute_phase.py, the discrepancy in the phases disappears:
```
111         if self.TZRSITE.value != '@' and self.TZRSITE.value.lower() != 'bat':
112             tz = toa.get_TOAs_array(
113                 (self.TZRMJD.quantity.jd1 - 2400000.5, self.TZRMJD.quantity.jd2),
114                 obs=self.TZRSITE.value,
[...]

PR #1707, opened 2024-01-02
Disable TZRMJD corrections if TZRSITE is bat

gcerib:
This is a pull request following the discussion with @abhisrkckl , aimed at resolving #1691.
It introduces a check on TZRSITE in absolute_phase.py, and disables clock corrections on TZRMJD if the former is one of the aliases for the SSB. This is necessary for PINT to consistently reproduce results by tempo2, which implements such a check.

I managed to run tox to test the changes, it somehow took 32 hours on my server. It fails on tests/test_precision.py::test_pulsar_mjd_close_to_mjd_on_leap_second_days . However, so does the version I originally checked out, prior to applying any change to absolute_phase.py . Please find attached the short summary I obtained from both the original (unchanged) version and the one with changes.

[tox_result.txt](https://github.com/nanograv/PINT/files/13814305/tox_result.txt)
[tox_result__orig.txt](https://github.com/nanograv/PINT/files/13814306/tox_result__orig.txt)

Cheers,
GC


abhisrkckl:
Can you add a test to verify that your original issue has been resolved?

src/pint/models/absolute_phase.py
```diff
@@ -108,16 +108,28 @@ def get_TZR_toa(self, toas):
...
-            obs=self.TZRSITE.value,
-            freqs=self.TZRFRQ.quantity,
-            include_bipm=clkc_info["include_bipm"],
-            include_gps=clkc_info["include_gps"],
-            ephem=toas.ephem,
-            planets=toas.planets,
-            tzr=True,
-        )
+        if self.TZRSITE.value != "@" and self.TZRSITE.value.lower() != "bat":
+            tz = toa.get_TOAs_array(
```
abhisrkckl, line 112: The barycenter has more aliases than "@" and bat. It may be better to change this comparison to
`self.TZRSITE.value.lower() not in [get_observatory("@" ).name] + get_observatory("@" ).aliases`

abhisrkckl: Also, please add a similar condition for the geocenter ("geo") as well.

gcerib: Hi. In tempo2's https://bitbucket.org/psrsoft/tempo2/src/master/refphs.C, this is what it does:
```
 21             if (obs->telID[0]=='@'            /* Have barycentric arrival time */
 22                     || strcasecmp(obs->telID,"bat")==0)
 23             {
 24                 obs->clockCorr=0;  /* therefore don't do clock corrections */
 25                 obs->delayCorr=0;
 26             }
 27             else if (strcmp(obs->telID,"STL")==0)
 28             {
 29                 obs->clockCorr=0;  /* don't do clock corrections */
 30                 obs->delayCorr=1;
 31             }
 32             else if (strcmp(obs->telID,"STL_FBAT")==0)
 33             {
 34                 obs->clockCorr=0;  /* don't do clock corrections */
 35                 obs->delayCorr=1;
 36             }
 37             else
 38             {
 39                 obs->clockCorr=1;
 40                 obs->delayCorr=1;
 41             }
```
It doesn't involve geocentric TOAs at all, and indeed, a par file with "COE" as TZRSITE will get the clock corrections in both PINT and tempo2. If we now disable them in PINT, I am afraid we will end up in the mirrored situation in which PINT doesn't apply the clock corrections to geocentric TZRMJD, whereas tempo2 does, producing again absolute phase differences between the two packages.

Regarding the aliases for "@" and "BAT", this can be easily changed. I have little experience with pytest, but I can try and build a test by looking at other ones in the tests folder.

abhisrkckl: OK... Then you can ignore the part about the geocenter.

About STL and STL_FBAT, I don't know what they are. They are not supported by PINT in any case.

abhisrkckl: I dug into this a bit more. The `include_bipm` and `include_gps` attributes of the `BarycenterObs` class are explicitly set to False in https://github.com/nanograv/PINT/blob/master/src/pint/observatory/special_locations.py

So this bug should not have occurred in the first place. 

I think the bug is in the `get_observatory()` function. When I do `print(get_observatory("@")).include_bipm`, I get True. This is because the `get_observatory()` function wrongly sets this from default arguments. I think the correction should be in `get_observatory()`. This special case treatment is not actually needed.

src/pint/models/absolute_phase.py
```diff
@@ -108,16 +108,28 @@ def get_TZR_toa(self, toas):
...
+                obs=self.TZRSITE.value,
+                freqs=self.TZRFRQ.quantity,
+                include_bipm=False,
+                include_gps=False,
+                ephem=toas.ephem,
+                planets=toas.planets,
+                tzr=True,
+            )
         log.debug("Done with TZR_toa")
         self.tz_cache = tz
```
gcerib, line 134: @abhisrkckl  I think in here `self.tz_clck_info` should also reflect the fact that if TZRMJD is barycentered, no corrections have been applied. What do you think?

gcerib: Sorry, it would be line 124/136, it is just outside the commit and cannot comment directly on it.

abhisrkckl: Please see my comment above. I think if we fix the bug in `get_observatory`, this should get fixed automatically.

src/pint/models/absolute_phase.py
gcerib: I see! So the check for the barycenter in absolute_phase.py (and toa.py) is unnecessary, provided it is fixed in get_observatory(), am I correct?

abhisrkckl: I think so.

gcerib: Ok, so what about this replacement in get_observatory()?
```
488    site = Observatory.get(name)
489    site.include_gps = include_gps
490    site.include_bipm = include_bipm
491    site.bipm_version = bipm_version
492    return site
```
with
```
488    site = Observatory.get(name)
489    import pint.observatory.special_locations
490    if not isinstance(site,special_locations.BarycenterObs):
491        site.include_gps = include_gps
492        site.include_bipm = include_bipm
493        site.bipm_version = bipm_version
494    return site
```
But this would ignore the arguments passed to get_observatory completely. Perhaps it would even be worth having it through a warning about it.

I tested this change already on my system, and it solves both problems (with TZRMJD and normal barycentered TOAs).

abhisrkckl: I think it will be better to make this more general. We can change the default argument values for include_bipm and include_gps to None, and (re-)set these values only if they are not None.

gcerib: Thanks. If I set defaults to None and reset them only if they are provided, as you suggest, it doesn't actually work, because in pint.toa get_observatory() gets called with explicit arguments:
```
2206         for obs, grp in self.get_obs_groups():
2207             site = get_observatory(
2208                 obs,
2209                 include_gps=include_gps,
2210                 include_bipm=include_bipm,
2211                 bipm_version=bipm_version,
2212             )
``` 
Any idea?

gcerib:
> Can you please add a test to verify that your original issue has been resolved?

@abhisrkckl  I've added the test you requested, hope it helps!

src/pint/models/absolute_phase.py
abhisrkckl: I meant changing the `get_observatory` function itself to something like this:

```
def get_observatory(
    name, include_gps=True, include_bipm=True, bipm_version=bipm_default
):
    """Convenience function to get observatory object with options.

    This function will simply call the ``Observatory.get`` method but
    will manually modify the global observatory object after the method is called.
    Name-matching is case-insensitive.

    If the observatory is not present in the PINT list, will fallback to astropy.

    Parameters
    ----------
    name : str
        The name of the observatory
    include_gps : bool, optional
        Set False to disable UTC(GPS)->UTC clock correction.
    include_bipm : bool, optional
        Set False to disable TAI TT(BIPM) clock correction.
    bipm_version : str, optional
        Set the version of TT BIPM clock correction files.

    .. note:: This function can and should be expanded if more clock
        file switches/options are added at a public API level.

    """
    site = Observatory.get(name)
    if include_gps is not None:
        site.include_gps = include_gps
    if include_bipm is not None:
        site.include_bipm = include_bipm
        site.bipm_version = bipm_version
    return site
```

gcerib: Hi,
Yes. That is precisely what I did. It **still fails**, even if I set the defaults of get_observatory to None. This is because toa.py:get_TOAs() sets include_bipm and include_gps to True even when it is called without them (and should therefore default to None), if all TOAs provided do not have yet the clock corrections (as is the case):
```
 283     if all("clkcorr" not in f for f in t.table["flags"]):
 284         if include_gps is None:
 285             include_gps = True
 286         if bipm_version is None:
 287             bipm_version = bipm_default
 288         if include_bipm is None:
 289             include_bipm = True
 290         # FIXME: should we permit existing clkcorr flags?
 291         t.apply_clock_corrections(
 292             include_gps=include_gps,
 293             include_bipm=include_bipm,
 294             bipm_version=bipm_version,
 295             limits=limits,
 296         )
```

The conditions in lines 284 and 288 are met by default, i.e. when calling get_TOAs without explicitly setting include_gps and include_bipm, and those parameters are passed over to apply_clock_corrections, which in turn passes them to get_observatory. Therefore, even if a TOA in a tim file is already defined in the barycenter, whenever it is read with get_TOAs it will get the clock corrections, unless the user manually sets get_TOAs(..., include_gps=False, include_bipm=False). This would, however, affect all TOAs in the tim file, not only the barycentered ones!

The behavior with TZRMJD when TZRSITE is BAT follows from the same origin. While building the TZR TOA, get_TZR_toa(self,toas) in absolute_phase.py reads the clock correction information from the provided toas, and calls get_TOAs_array() explicitly, passing the include_gps and include_bipm parameters that were applied on the toas. Since these were set to True when reading them for the first time, it ends up applying the correction also on TZRMJD, even if it is defined in BAT..

At least, this was my attempt to follow the code around. It involves an interplay of get_observatory() [observatory/__init__.py], get_TOAs(), get_TOAs_array(), apply_clock_corrections.py [toa.py], and get_TZR_toa.py [absolute_phase.py].

gcerib: Hi @abhisrkckl ,

Unfortunately I am on a technical shift to the telescopes in La Palma, I don't have much time these days to continue digging on this. Did you have a chance to have a look at the problem?

So far, the only solution that worked consistently is adding a check for the BarycenterObs class in get_observatory() (https://github.com/nanograv/PINT/pull/1707#discussion_r1440816152), and it passes the tests I've implemented. but it also has some drawbacks. Namely, if apply_clock_corrections() is called on some TOAs multiple times, and there is a barycentered TOA in the list, all calls after the first one will fail.

Cheers,
Giovanni

Issue #1708, opened 2024-01-04
`pint.observatory` has spooky action at a distance


abhisrkckl:
The way observatories are read relies heavily on side effects. This makes it very difficult to understand and debug that module unless one already knows what to look for.

aarchiba:
Are you thinking of the registry/alias map? https://github.com/nanograv/PINT/blob/master/src/pint/observatory/__init__.py#L145 That's there to make it easier to define observatories - no need to list them all somewhere central - but there are less confusing ways to do that.

Or the lazy loading of clock files? That is motivated by performance concerns - clock files can be large, and loading them all can be slow, unnecessarily so when most operations require only one, or at most a few. The pattern of storing something in _variable and having a method _load_variable() is not ideal; one should be able to replace it with `@cached_property`.

aarchiba:
Perhaps more of a problem is that various actions - registering the observatories, in particular - are done on import of the relevant modules. This is not nice, but it is not easy to resolve without requiring users to call some setup function before doing anything with observatories. Something could probably be arranged, perhaps by putting load-everything calls in get_observatory and the like.

abhisrkckl:
I appreciate the need for lazy loading of clock files, and this issue is not about that. I did find that part of the code quite opaque, and more comments will likely be very helpful.

My complaint is about the implementation of the registry/alias map mechanism. It is not at all obvious where and how the registry gets populated, and there is probably a way to make this mechanism less reliant on non-obvious side effects.

More serious is the fact that the `get_observatory` function rewrites `Observatory` objects silently, leading to bugs like #1691. (PR #1715 is an attempt at fixing this.)

aarchiba:
The registry is a pattern used elsewhere in PINT. The idea is that the Observatory class is constructed so that when an observatory object is constructed, it registers itself automatically in a central registry. This registration enables file readers to translate names into Observatory instances. 

The pattern is also used for timing models, although in this case the registry is not for instances of TimingModel subclasses but for the subclasses themselves. It exists for the same reason.

[...]

PR #1711, opened 2024-01-10
`get_observatory` shouldn't overwrite `include_bipm` and `include_gps` unless given explicitly 

abhisrkckl:
#1691 is a symptom of this, although the root cause is elsewhere.

#1707 was an attempt at fixing this.

Related issue - #1708


src/pint/observatory/__init__.py
```diff
@@ -143,24 +144,6 @@ class Observatory:
...
-                raise ValueError(
-                    f"Observatory {name.lower()} already present and overwrite=False"
-                )
-            log.warning(f"Observatory '{name.lower()}' already present; overwriting...")
-
-            cls._register(obs, name)
-            return obs
-        cls._register(obs, name)
-        return obs
-
```
abhisrkckl, line 163: Everything here can be done in `__init__()`. Having a non-trivial `__new__()` makes the objects non-copyable. That's why I am moving this code to `__init__()`.

CHANGELOG-unreleased.md
```diff
@@ -40,4 +40,5 @@ the released changes.
 - Consistent naming in `TimingModel.get_params_mapping()`
 - Better exceptions for unsupported/unimplemented binary models (BTX, MSS, etc.)
 - Emit warnings when `WaveX`/`DMWaveX` is used together with other representations of red/DM noise
+- `get_observatory()` no longer overwrites `include_gps` and `include_bipm` of `Observatory` objects unless explicitly stated.
```
paulray, line 43: I think we should add "Fixes issue #1691 (BIPM+GPS clock corrections incorrectly applied to BAT TOAs)"

paulray (approved):
Looks good to me. BAT times should not get BIPM or GPS corrections applied, regardless of whether they are TZRMJD times or TOA times. BAT times are already fully corrected to TDB.

CHANGELOG-unreleased.md
abhisrkckl: Added.

abhisrkckl:
@paulray Can you please merge this?

Issue #1369, opened 2022-08-04
Discrepancy in phase prediction between PINT and tempo2 polycos.


theXYZT:
Hi,

I am trying to figure out what's causing a particular discrepancy in pulsar phase prediction between PINT and generating polycos with tempo2.

Here is a par file I am using, `b1937.par` (it is a modified version of the NanoGrav 12.5 year narrowband par file for B1937+21):

```
PSR              B1937+21
LAMBDA   301.9732445257902  1     0.0000000054618
BETA      42.2967522555839  1     0.0000000058937
PMLAMBDA           -0.0184  1              0.0041
PMBETA             -0.4007  1              0.0060
PX                  0.2819  1              0.0509
POSEPOCH        55599.0000
F0    641.9282322943167856  1  0.0000000000229207
F1     -4.330889418718D-14  1  1.374778598040D-19
PEPOCH        55599.000000
DM                 71.0201
DMEPOCH          58245.375
SOLARN0               0.00
EPHEM               DE436
ECL                 IERS2010
CLK                 TT(BIPM2017)
UNITS               TDB
TIMEEPH             FB90
T2CMETHOD           TEMPO
CORRECT_TROPOSPHERE N
PLANET_SHAPIRO      N
DILATEFREQ          N
NTOA                 17024
TRES                  1.82
TZRMJD  55616.41780189249631
TZRFRQ            1742.942
TZRSITE                  GB
MODE                     1
```

When I use `tempo2` to generate polycos, it generates the following relevant entry:

```
B1937+21    7-May-18   90000.00   58245.37500000000            71.020167 -0.706 -6.330
 146774936445.058162  641.928232294317   ao   90   12   327.000                
 -3.00147350705225993e-07  2.72493109523728450e+00 -1.19447953427931179e-04
 -5.36771151182867317e-09  1.79612760415288616e-10  4.84839643606338133e-13
  2.66425801966225905e-15 -5.05089187190312671e-16 -2.43473904299295782e-19
  2.19028473393043386e-19 -9.16295733411736226e-23 -3.35126294498144066e-23
```

Using PINT, I execute the following code:

```python
import pint.models
import pint.toa

m = pint.models.get_model(folder / "b1937.par")

ts = pint.toa.get_TOAs_list([pint.toa.TOA((58245, 0.375), 0, freq=327 * u.MHz, obs="ao")],
                            ephem=m.EPHEM.value, bipm_version="BIPM2017")

ph = m.phase(ts, abs_phase=False)
[...]

Issue #595, opened 2020-01-19
Polycos with Doppler shifts 0


emmacarli:
Hello all,
I am generating polycos for a custom observatory. They are produced, but the Doppler shift due to Earth motion is 0 in each entry. This is the main effect I would like to correct for, to fold long observations with other software.

I have input my custom observatory in /PINT/src/pint/observatory/observatories.py as follows:
`TopoObs('acre',
	aliases=['acreroad'],
	itrf_xyz =[3573741.1, -269156.74, 5258407.3],
)`

This is the Python script I used:
```
#%% Import packages
from pint import polycos, observatory, models
import astropy
import sys

#%% Start a log
log_handle = open('PINT_polycos_generator.log', 'w')

sys.stdout = log_handle
sys.stderr = log_handle

#%%Generate model
B032954_model = models.model_builder.get_model('B0329+54.par')
#this parameter file's data was obtained from ATNF

#%% Generate polycos
pl = polycos.Polycos() #initiate an empty polycos instance
pl.generate_polycos(model = B032954_model, mjdStart = 57969.1, mjdEnd = 57969.4, obs = 'acre', segLength = 5, ncoeff = 12, obsFreq = 407.5, method = 'TEMPO') #segment length in minutes, observing frequency in MHz, method TEMPO2 unavailable

#%% Write to TEMPO style polyco table (PRESTO uses these for folding)
pl.write_polyco_file(format='tempo', filename='PINT_polycos.dat')

```
This is an example of a set of polycos from PINT_polycos.dat : 

> J0332+543404-Aug-17 022630.00 57969.101736111109656 0.0                  0.0   0.0
1390110980.94504384 1.399541538720    acre 5      12   407.5                0.0  
-2.25647100421503604e-10 6.05368936507820563e-03  -1.82265915169433213e-08 
-5.16412032720972806e-10 -1.29915760566270875e-10 3.98289910480010468e-10  
6.39002894891700953e-11  -1.61703415153186620e-10 -1.19489570516592519e-11 
2.82082280282038799e-11  7.22100001451911620e-13  -1.75994574970452220e-12 

I believe pulsar number, date, UTC, MJD, 0 DM (dedispersed data) are correct; although the first two are not separated. Then there is **0 doppler shift** and 0 fit residual (not sure if the latter is normal, either). Then reference phase, frequency, observatory, time span, number of coefficients, observing frequency and binary phase (single pulsar) seem all correct as well.

I can provide the contents of the log if needed.

Thank you!
[...]

PR #1713, opened 2024-01-15
WIP: basic support for polyco doppler and rms

dlakaplan:
Previously the Doppler and RMS fields for polycos were set to 0 explicitly (#595, maybe #1369).

I have added computation of the Doppler shift and RMS.  However, I am unsure how to test them since we don't have comparisons against external polycos (and as per #1369 it is not clear if they would succeed).

The Doppler shifts here seem very similar to those in #595.  The RMS values are ~1 dex lower - this may be due to differences in how the polycos are computed, or an error in units.

For reference, when I use the same routine that #595 did I now get:
```
In [7]: polycos[['doppler','logrms']]
Out[7]: 
<Table length=21>
     doppler              logrms      
    float64[1]           float64      
------------------ -------------------
0.8277697583935151  -10.05876484179034
0.8214061155480782 -10.187670625757141
0.8252718627984125  -10.06929480188947
0.8333408761473038  -10.06110378949105
0.8333122527054521 -10.262116883531544
               ...                 ...
0.8192315787385138 -10.041679764534035
0.8191044627908792  -9.993124663597873
0.8270383391991994 -10.054704235766907
0.8307247245349052  -10.12410055651501
0.8241657455674353 -10.016446104786814
0.8166066521214685 -10.011765829681192
```

### DIFF
--- a/src/pint/polycos.py
+++ b/src/pint/polycos.py
@@ -793,13 +793,17 @@ class Polycos:
             rdcPhase = rdcPhase.int - (dt * model.F0.value * 60.0) + rdcPhase.frac
             dtd = dt.astype(float)  # Truncate to double
             rdcPhased = rdcPhase.astype(float)
-            coeffs = np.polyfit(dtd, rdcPhased, ncoeff - 1)[::-1]
+            rawcoeffs = np.polyfit(dtd, rdcPhased, ncoeff - 1)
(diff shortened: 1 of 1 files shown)


dlakaplan:
@mcknightryan, @emmacarli: let me know what you think.

PR #1715, opened 2024-01-24
Don't overwrite include_gps and include_bipm for ssb

abhisrkckl:
It turned out that #1711 didn't fully solve the original issue #1691.
This PR attempts to solve it. 

Includes tests provided by @gcerib to check the original issue.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -40,5 +40,6 @@ the released changes.
 - Consistent naming in `TimingModel.get_params_mapping()`
 - Better exceptions for unsupported/unimplemented binary models (BTX, MSS, etc.)
 - Emit warnings when `WaveX`/`DMWaveX` is used together with other representations of red/DM noise
-- `get_observatory()` no longer overwrites `include_gps` and `include_bipm` of `Observatory` objects unless explicitly stated (BIPM and GPS clock corrections no longer incorrectly applied to BAT TOAs).
+- `get_observatory()` no longer overwrites `include_gps` and `include_bipm` of `Observatory` objects unless explicitly stated 
+- BIPM and GPS clock corrections are no longer incorrectly applied to BAT TOAs
 ### Removed
--- a/src/pint/observatory/__init__.py
+++ b/src/pint/observatory/__init__.py
@@ -476,8 +476,12 @@ def get_observatory(
         file switches/options are added at a public API level.
 
     """
-    if include_bipm is not None or include_gps is not None:
-        site = deepcopy(Observatory.get(name))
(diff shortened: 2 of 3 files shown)


aarchiba:
This fails a number of tests and needs fixing.

More seriously, it is not clear whether you can safely apply `deepcopy` to an Observatory instance without triggering the Observatory registration machinery. At a minimum some tests to check the registry after doing this would be valuable. This might explain some of the mysterious failures.

paulray:
This test is failing, but it isn't super clear to me if it is a bug in the test or in the load_Fermi_TOAs() code.
The difference between the two methods of loading Fermi TOAs is 27 µs so one is getting BIPM corrections and the other is not.
```
_______________________________ test_toas_fermi ________________________________

    def test_toas_fermi():
        eventfileraw = datadir / "J0030+0451_w323_ft1weights.fits"
        ft2file = datadir / "lat_spacecraft_weekly_w323_p202_v001.fits"
        get_satellite_observatory("Fermi", ft2file, overwrite=True)
    
        tl = load_Fermi_TOAs(eventfileraw, weightcolumn="PSRJ0030+0451")
        ts = toa.get_TOAs_list(
            tl, include_gps=False, include_bipm=False, planets=False, ephem="DE405"
        )
    
        t = time.Time([toa.mjd for toa in tl], scale="tt")
        flags = [toa.flags for toa in tl]
        ts2 = toa.get_TOAs_array(
            t,
            "fermi",
            include_gps=False,
            include_bipm=False,
            planets=False,
            ephem="DE405",
            flags=flags,
        )
>       assert np.all(ts["mjd"] == ts2["mjd"])
E       AssertionError: assert False
E        +  where False = <function all at 0x7ff69f8e8330>(<Column name=...2.959435096476 == <Column name=...2.959435096156
E        +    where <function all at 0x7ff69f8e8330> = np.all
E           Full diff:
E             <Column name='mjd' dtype='object' length=27>
E           -  56876.47375201528
E           ?                 ^^
E           +   56876.4737520156
E           ? +                ^
E           -  56877.00298295838...
E           
E           ...Full output truncated (92 lines hidden), use '-vv' to show)

/home/runner/work/PINT/PINT/tests/test_toa_create.py:246: AssertionError
----------------------------- Captured stderr call -----------------------------
WARNING  (pint.observatory              ): Observatory 'fermi' already present; overwriting...
WARNING  (pint.toa                      ): No ephemeris provided to TOAs object or compute_TDBs. Using DE421
=============================== warnings summary ===============================
```

abhisrkckl:
Closing this because #1763 is probably a better solution for this.

PR #1717, opened 2024-02-05
Allow simulated TOAs to maintain a non-zero mean

dlakaplan:
Normally when creating fake TOAs, the residuals are set to zero with `subtract_mean=True`.  This means that when comparing a set of TOAs (real or fake) against another set of fake TOAs, they will not connect coherently.  

This PR allows the option of setting `subtract_mean=False` to the various fake TOA creation routines to preserve coherence.

### DIFF
--- a/src/pint/simulation.py
+++ b/src/pint/simulation.py
@@ -1,8 +1,9 @@
 """Functions related to simulating TOAs and models
 """
+from __future__ import annotations
 from collections import OrderedDict
 from copy import deepcopy
-from typing import Optional, Union, List, Dict
+from typing import Optional, List
 import pathlib
 
 import astropy.units as u
--- a/src/pint/simulation.py
+++ b/src/pint/simulation.py
@@ -22,7 +22,7 @@ __all__ = [
 ]
 
 
-def zero_residuals(ts, model, maxiter=10, tolerance=None):
+def zero_residuals(ts, model, subtract_mean=True, maxiter=10, tolerance=None):
     """Use a model to adjust a TOAs object, setting residuals to 0 iteratively.
 
     Parameters
@@ -42,7 +42,9 @@ def zero_residuals(ts, model, maxiter=10, tolerance=None):
     if tolerance is None:
         tolerance = 1 * u.ns if pint.utils.check_longdouble_precision() else 5 * u.us
     for i in range(maxiter):
-        r = pint.residuals.Residuals(ts, model, track_mode="use_pulse_numbers")
+        r = pint.residuals.Residuals(
+            ts, model, subtract_mean=subtract_mean, track_mode="use_pulse_numbers"
+        )
         resids = r.calc_time_resids(calctype="taylor")
         if maxresid is not None and (np.abs(resids).max() > maxresid):
             log.warning(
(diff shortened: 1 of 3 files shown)


src/pint/simulation.py
```diff
@@ -22,7 +22,7 @@
 ]
 
 
-def zero_residuals(ts, model, maxiter=10, tolerance=None):
+def zero_residuals(ts, model, subtract_mean=True, maxiter=10, tolerance=None):
```
aarchiba, line 25: Since these are not keyword-only arguments, someone might have been using these positionally and their code will break. Perhaps it's better to put the new argument at the end? Or if breaking their code is okay, make these arguments keyword-only?

tests/test_fake_toas.py
```diff
@@ -236,6 +236,60 @@ def test_fake_fromMJDs(MJDs):
...
+        model=model,
+        error=1 * u.us,
+        add_noise=True,
+    )
+    toas2 = pint.simulation.make_fake_toas_fromMJDs(
+        t2,
+        model=model,
+        error=1 * u.us,
+        add_noise=True,
+    )
```
aarchiba, line 267: This test enforces that the default for `subtract_mean` is True. Is that desired?

src/pint/simulation.py
dlakaplan: I used the order since more people would change `subtract_mean` compared to the others, although that isn't a big deal.  But I made them keyword-only as you suggested.

src/pint/simulation.py
```diff
@@ -1,7 +1,10 @@
 """Functions related to simulating TOAs and models
 """
+from __future__ import annotations
```
aarchiba, line 3: This is only relevant if there are forward references, as far as I can tell, and I couldn't spot any. Is this actually needed?

dlakaplan:
I think using `|` to join types is only in 3.10 and later.  I guess I could just use `Union` if using `future` is bad.

aarchiba:
> I think using `|` to join types is only in 3.10 and later. I guess I could just use `Union` if using `future` is bad.

I don't think either of those is an effect of using `from __future__ import annotations`? That is supposed to be specific to PEP 563: https://docs.python.org/3/library/__future__.html#module-contents

dlakaplan:
I haven't dug deeply, but when I didn't have that it failed the python 3.8 tests, and from some googling the import seemed to fix it.  The CI tests did succeed.

aarchiba:
The import lets you get away with using 3.8-incompatible syntax under python 3.8 by deferring evaluation of the annotations. A more conventional approach would be to use Union instead of |.

dlakaplan:
> The import lets you get away with using 3.8-incompatible syntax under python 3.8 by deferring evaluation of the annotations. A more conventional approach would be to use Union instead of |.

Yes, that's what I had mentioned before: if doing it this way is objectionable I'll use `Union`

As a related point, should we define some top-level types (say for file-like objects that might be `Union[str, Path, ...]` or things that we can pass for time-like objects which are commonly used (`Union[float, Quantity, Time]`)?  Or do those only at the module level?

aarchiba:
What is the impact of `from __future__ import annotations` on mal-formed type specifications? It ensures the python interpreter does not see them, but how do IDEs and `mypy` handle these? I assume the tools don't mind newer-than-3.8 syntax but to they correctly highlight mangled type declarations?

If we want to put in type annotations more broadly I do think that declaring certain standard types in a central location is a good idea. Perhaps a module `pint.types`, to avoid import-order headaches? Although this won't work for types that include PINT-defined classes like TOAs. I guess those could perhaps be defined immediately after the relevant classes?

But declaring standard types globally will not just save typing; it should encourage people writing or modifying PINT functions to ensure they accept standard (combinations of) types rather than whatever is convenient.

aarchiba:
The notebooks test failure appears to be an import error (`pint.fitter.Fitter` when `Fitter` is what was imported), but not one introduced by this PR. I'm not sure why it's surfacing here and not sooner?

Issue #1709, opened 2024-01-04
We should have type annotations


abhisrkckl:
Type hinting was introduced in Python 3.5. Now that we no longer support older python versions (<3.8), there is no reason why we cannot have type annotations everywhere. It makes life so much easier while using an IDE with code highlighting.

scottransom:
I just found another really amazing usecase with vscode that is enabled by type hinting.

In the example `fit_NGC6440E.py` we have the following:

```
...
from pint.models import get_model_and_toas
...
# Read the timing model and the TOAs
m, t = get_model_and_toas(parfile, timfile)
```

If we are editing code below that, the editor does not know what `m` and `t` are.

However, if I simply put a wrapper around `get_model_and toas()` like so:
```
import pint.models.timing_model
def foo(par: str, tim: str) -> tuple[pint.models.timing_model.TimingModel, pint.toa.TOAs]:
    return get_model_and_toas(par, tim)

m, t = foo(parfile, timfile)
```
Then anywhere later in the code, vscode knows that `m` and `t` are `TimingModel` and `TOAs` class instances, respectively, and you get tab completion (and autocompletion) on all of the methods and attributes of those classes.

That's really cool.  I definitely think we should add type hints to many of our most common methods and functions for this reason alone.

scottransom:
Interestingly, you can get the same behavior without adding type hints to the definition of `get_model_and_toas()` but my pre-hinting the variables that you are going to use in `fit_NGC6440E.py` itself:

```
import pint.toa
import pint.models.timing_model

m: pint.models.timing_model.TimingModel
t: pint.toa.TOAs
m, t = get_model_and_toas("1748-2446as.par", "Ter5as_forDMs.tim")
```
That works just fine, and the editor knows what `m` and `t` are and allows autocompletion etc.

scottransom:
It looks like there are a couple projects that do code analysis to guess type hinting information and then automatically modify your code.  So this could be a first step if we wanted to do this with much of or most of PINT:
https://github.com/Instagram/MonkeyType

aarchiba:
[...]

PR #1718, opened 2024-02-10
Use cached_property and types

aarchiba:
Addresses part of #1708 and also demonstrates gradual typing towards #1709. Unsure whether this resolves the first.

### DIFF
--- a/src/pint/observatory/__init__.py
+++ b/src/pint/observatory/__init__.py
@@ -21,14 +21,17 @@ done, but if you are using a different subset of PINT these imports may be
 necessary.
 """
 
-from copy import deepcopy
 import os
 import textwrap
 from collections import defaultdict
+from collections.abc import Callable
+from copy import deepcopy
 from io import StringIO
 from pathlib import Path
+from typing import Optional, Union
 
 import astropy.coordinates
+import astropy.time
 import astropy.units as u
 import numpy as np
 from astropy.coordinates import EarthLocation
@@ -97,7 +100,7 @@ def _load_gps_clock():
         )
 
 
-def _load_bipm_clock(bipm_version):
+def _load_bipm_clock(bipm_version: str):
     bipm_version = bipm_version.lower()
     if bipm_version not in _bipm_clock_versions:
         try:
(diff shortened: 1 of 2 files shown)


aarchiba:
I am hoping that the type annotations improve the readthedocs build.

src/pint/observatory/__init__.py
```diff
@@ -428,17 +468,17 @@ def get_TDBs(self, t, method="default", ephem=None, options=None):
         else:
             raise ValueError(f"Unknown method '{method}'.")
 
-    def _get_TDB_default(self, t, ephem):
+    def _get_TDB_default(self, t: astropy.time.Time, ephem):
         return t.tdb
 
-    def _get_TDB_ephem(self, t, ephem):
+    def _get_TDB_ephem(self, t: astropy.time.Time, ephem):
```
dlakaplan, line 474: Same for this instance.

dlakaplan: and later

src/pint/observatory/topo_obs.py
```diff
@@ -277,11 +306,11 @@ def get_dict(self):
         return {self.name: output}
 
     def get_json(self):
```
dlakaplan, line 308: return type?

src/pint/observatory/topo_obs.py
```diff
@@ -277,11 +306,11 @@ def get_dict(self):
...
-        """Return as a JSON string"""
+        """Return as a JSON string."""
         return json.dumps(self.get_dict())
 
-    def separation(self, other, method="cartesian"):
-        """Return separation between two TopoObs objects
+    def separation(self, other: "TopoObs", method: str = "cartesian"):
```
dlakaplan, line 312: return type?

src/pint/observatory/topo_obs.py
```diff
@@ -365,19 +393,18 @@ def clock_corrections(self, t, limits="warn"):
...
         available clock corrections specific to the telescope.
         """
         t = super().last_clock_correction_mjd()
-        self._load_clock_corrections()
         for clock in self._clock:
             t = min(t, clock.last_correction_mjd())
         return t
 
-    def _get_TDB_ephem(self, t, ephem):
+    def _get_TDB_ephem(self, t: Time, ephem) -> Time:
```
dlakaplan, line 407: type of `ephem`?

src/pint/observatory/topo_obs.py
```diff
@@ -389,8 +416,8 @@ def _get_TDB_ephem(self, t, ephem):
         # Topocenter to Geocenter
         # Since earth velocity is not going to change a lot in 3ms. The
         # differences between TT and TDB can be ignored.
-        earth_pv = objPosVel_wrt_SSB("earth", t.tdb, ephem)
-        obs_geocenter_pv = gcrs_posvel_from_itrf(
+        earth_pv: PosVel = objPosVel_wrt_SSB("earth", t.tdb, ephem)
```
dlakaplan, line 419: What is the advantage to typing of this sort?  Is this to prevent any changes to the API?

src/pint/observatory/__init__.py
```diff
@@ -428,17 +468,17 @@ def get_TDBs(self, t, method="default", ephem=None, options=None):
         else:
             raise ValueError(f"Unknown method '{method}'.")
 
-    def _get_TDB_default(self, t, ephem):
+    def _get_TDB_default(self, t: astropy.time.Time, ephem):
```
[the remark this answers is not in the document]
aarchiba: I didn't add it because I wasn't sure what the type was. String I think?

src/pint/observatory/topo_obs.py
aarchiba: This tells the type checker to signal a problem if the function does not actually return a PosVel, or if the surrounding code uses the value in a way incompatible with a PosVel. More, it informs the reader what type this has, in case they don't know off the top of their head what type to expect. In this case it seemed useful because I wasn't ready to go digging about and annotate those two functions.

src/pint/observatory/topo_obs.py
aarchiba: I was trying to apply incremental typing just to those functions I touched. But if we're going all in on type annotations then I can be more aggressive.

src/pint/observatory/__init__.py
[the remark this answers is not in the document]
dlakaplan: Yes, I think it's a string

src/pint/observatory/topo_obs.py
dlakaplan: ok, sure.  either way.

src/pint/observatory/topo_obs.py
dlakaplan: ok, that makes sense.

PR #1720, opened 2024-02-17
Improve clock error reporting

aarchiba:
When PINT tries to read a clock file with out-of-order entries, it will raise an exception. (Such files have never worked correctly on any of TEMPO, TEMPO2, or PINT.) This PR makes sure that includes the first two out-of-order MJDs to aid in diagnosing the problem.

Helped with both https://github.com/ipta/pulsar-clock-corrections/issues/18 and https://github.com/ipta/pulsar-clock-corrections/issues/19

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,11 @@
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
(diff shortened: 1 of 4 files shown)


dlakaplan:
Looks good to me.  Is this ready for merging?

aarchiba:
Yes, it's ready to go.

PR #1721, opened 2024-02-18
Remove ERFA warnings

aarchiba:
This PR sets `pytest` so that ERFA warnings become exceptions, then adjusts things so that the test suite passes. For the most part, this amounts to ignoring/suppressing ERFA warnings that are generated when the test suite (usually because of hypothesis) supplies very unusual dates to Time functions. There is at least one instance where loading the Parkes clock file requires handling MJD 0, which meant emitting an ERFA warning every time you worked with Parkes data. This has been fixed; "dubious year" warnings are no longer emitted when loading clock files.

### DIFF
--- a/src/pint/observatory/clock_file.py
+++ b/src/pint/observatory/clock_file.py
@@ -523,16 +523,22 @@ def read_tempo2_clock_file(
         mjd = mjd[1:]
         clk = clk[1:]
         comments = comments[1:]
-    return ClockFile(
-        mjd,
-        clk * u.s,
-        filename=filename,
-        comments=comments,
-        leading_comment=leading_comment,
-        header=header,
-        friendly_name=friendly_name,
-        valid_beyond_ends=valid_beyond_ends,
-    )
+    with warnings.catch_warnings():
+        # Some clock files have dubious years in them
+        # Most are removed by automatically ignoring MJD 0, or with "bogus_last_correction"
+        # But Parkes incudes a non-zero correction for MJD 0 so it isn't removed
+        # In any case, the user doesn't need a warning about strange years in clock files
+        warnings.filterwarnings("ignore", r".*dubious year", erfa.ErfaWarning)
+        return ClockFile(
+            mjd,
+            clk * u.s,
+            filename=filename,
+            comments=comments,
+            leading_comment=leading_comment,
+            header=header,
+            friendly_name=friendly_name,
+            valid_beyond_ends=valid_beyond_ends,
+        )
 
 
 ClockFile._formats["tempo2"] = read_tempo2_clock_file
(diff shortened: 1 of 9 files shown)


aarchiba:
If this is helpful, a wider warning-suppression effort could be carried out - not too painful, just requires understanding which warnings come from test cases and users should see them, and which warnings come from normal situations and users shouldn't.

aarchiba:
I also replaced some random tests (that break parallel testing and won't produce reproducible failures) with hypothesis (which is intended for the purpose).

aarchiba:
The oldestdeps failure looks like we are triggering an internal EFRA bug that has been fixed with the upgrade from 2.0.0.3 to 2.0.1. I'm not quite sure what to do about it.

kerrm:
These changes look positive to me!  Two quick things:

1. I'm pretty sure I could produce BIPM corrections using the same code that produced the ones in the tempo2 repository, and thus remove the bogus last entries.  Is that a better solution than handling the <=2019 ones separately?
2. Typo: "            warnings.filterwarnings("ignore", r".***dubuious** year", erfa.ErfaWarning)

src/pint/observatory/clock_file.py
```diff
@@ -864,7 +872,11 @@ def evaluate(self, t, limits="warn"):
         corrections : astropy.units.Quantity
             The corrections in units of microseconds.
         """
-        if np.any(t > self.clock_file.time[-1]):
+        with warnings.catch_warnings():
+            # FIXME: could these warnings be useful?
+            warnings.filterwarnings("ignore", r".*dubuious year", erfa.ErfaWarning)
```
kerrm, line 877: typo!

aarchiba:
> These changes look positive to me! Two quick things:
> 
>     1. I'm pretty sure I could produce BIPM corrections using the same code that produced the ones in the tempo2 repository, and thus remove the bogus last entries.  Is that a better solution than handling the <=2019 ones separately?

The global-clock-corrections repository already generates the post-2019 ones from BIPM files. But I vaguely remember there were, at least potentially, complexities in older versions (a slightly different format of the BIPM files maybe?). Or maybe I just didn't want to mess with the versions everyone was already using.

It's not ideal to have the special case in PINT, but PINT already has information about which observatory clock files come with bogus last corrections built in to it (though in a config file). And the BIPM stuff seems global/generic enough. I considered just dropping the last correction on all BIPM files - they all include a thousand-day forecast, so dropping the last of those days isn't doing much harm.

src/pint/observatory/clock_file.py
aarchiba: Since the warning suppression wasn't working because of the typo, and the test suite passed anyway, clearly it wasn't needed here. (I feel better not suppressing the warning in library code unless I can clearly justify the need to.)

scottransom:
Looks good!  Thanks a bunch!

PR #1722, opened 2024-02-19
Allow very old Princeton format TOAs

scottransom:
A very minor change that checks to see if a Princeton-format TOA uses the 39126 day subtracted MJD format as described here: https://tempo.sourceforge.net/ref_man_sections/toa.txt
And then makes that work.

This is needed to process some archival TOAs.


scottransom:
Haha.  yeah.  As soon as I made the PR, I knew that this request would be made...  ;-)

PR #1725, opened 2024-02-25
WIP: mypy checking as part of CI

aarchiba:
If we do want to apply static typing to PINT (#1709 ), a possible goal is for the static type checker `mypy` to successfully check the entire code base. 

`mypy` is designed for gradual typing, and in fact there is a [guide to applying mypy to an existing codebase](https://mypy.readthedocs.io/en/stable/existing_code.html). A key first step is to [enable mypy as part of CI](https://mypy.readthedocs.io/en/stable/existing_code.html#run-mypy-consistently-and-prevent-regressions). Of course, this requires an appropriate `mypy` setup; in particular it requires ignoring a lot of errors. As long as these are ignored, `mypy` isn't doing a particularly strict job, but this is a starting point for gradual typing.

Next steps:

1. Get `mypy` working on the whole codebase and running in CI, but with sufficiently permissive settings that it passes. *This PR does this.*
2. Select a small subset of the codebase (ideally something that is widely used) and enable stricter `mypy` checking, fixing that part of the codebase so it passes. *This PR does this for `pint.utils` and a few other tiny parts.*
3. Add type annotations to the subset where `mypy` is checking, making sure `mypy` continues to pass.
4. Add a new piece of code to the subset `mypy` checks and apply the above.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -35,6 +35,9 @@ jobs:
           - os: ubuntu-latest
             python: '3.12'
             tox_env: 'notebooks'
+          - os: ubuntu-latest
+            python: '3.12'
+            tox_env: 'mypy'
 #          - os: ubuntu-latest
 #            python: '3.8'
 #            tox_env: 'docs'
--- a/.gitignore
+++ b/.gitignore
@@ -139,3 +139,6 @@ tests/datafile/par_*.par
 tests/datafile/fake_toas.tim
 tests/datafile/*.converted.par
 tests/datafile/_test_pintempo.out
+
+# mypy
+.mypy_cache
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -41,3 +41,7 @@ loguru
 gprof2dot
 py-cpuinfo
 pytest-xdist
+mypy==1.8.0
+GitPython
+types-setuptools
+types-tqdm
--- a/setup.cfg
+++ b/setup.cfg
@@ -129,3 +129,67 @@ skip_glob =
     src/pint/extern/*
 include_trailing_comma = True
 combine_as_imports = True
+
+[mypy]
+warn_unused_configs = True
+files = src/pint
+
+[mypy-pint.templates.*]
+ignore_errors = True
+
+[mypy-pint.derived_quantities]
+ignore_errors = True
+
+[mypy-pint.observatory.*]
+ignore_errors = True
+
+[mypy-pint.models.stand_alone_psr_binaries.*]
+ignore_errors = True
+
+[mypy-pint.models.timing_model]
+ignore_errors = True
+
+[mypy-pint.models.pulsar_binary]
+ignore_errors = True
+
+[mypy-pint.fitter]
+ignore_errors = True
+
+[mypy-pint.polycos]
+ignore_errors = True
+
+[mypy-pint.output.publish]
+ignore_errors = True
+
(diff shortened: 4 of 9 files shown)


aarchiba:
Python 3.12 is more strict with string syntax; previous versions were happy with something like `"\Delta"` because `\D` wasn't a recognised escape sequence; they would pass it through without complaining. Python3.12 emits a SyntaxWarning in this case, because future versions of Python will consider that an error. These SyntaxWarnings have been getting lost in our test suite's blizzard of warnings, but `mypy` emits them. I probably should have switched to python3.11 for running `mypy` but I figured we needed to fix those anyway.

aarchiba:
It's probably instructive to look at the changes I had to make to get `mypy` to pass:

- A few type annotations were actually wrong
- The try import except ImportError doesn't work very well (but we didn't need it)
- There is some weirdness with the Fitter's covariance matrix attribute because it doesn't exist unless the value has been computed
- There was a case where `e` had been used both as `except ... as e` and as a normal variable in the same scope
- Some global variables needed types
- `models/__init__.py` needed to have `__all__` (which it should have had anyway)
- mypy is confused by `@thing.setter` so I had to disable it for that line
- mypy doesn't know about the `:P` f-string format introduced by the uncertainties package so I had to disable mypy for that line
- I had to put an `if TYPE_CHECKING:` to cope with some of our circular importing
- mypy did not like me passing `**kwargs` into a function that only accepted specific keyword arguments so I had to disable mypy for that line

So, not too difficult a battle, but there are some things that are okay that mypy doesn't like.

aarchiba:
Adding typed, documented instance variable declarations helps the docs (adds a Variables section): https://nanograv-pint--1725.org.readthedocs.build/en/1725/_autosummary/pint.fitter.Fitter.html#pint-fitter-fitter

aarchiba:
`toa.py` passes without changes, but this demonstrates gradual typing: none of the functions have type annotations so they pass automatically.

aarchiba:
I'd actually say this is ready to go: if we want to start working on static typing for pint, I'd say apply #1718 and then this one; we can then start adding annotations and removing `ignore_errors` and mypy will keep us from making a mess.

aarchiba:
Curses. Astropy's type annotations pre-6.0 clash with mypy's, and oldestdeps doesn't use astropy 6.0. I'll have to back out the changes to `derived_quantities`.

aarchiba:
Didn't have to back them out entirely, just separate the actual units into `quantity_input` and the types (`u.Quantity`) into the function signature. With astropy 6.0 (5.0 maybe?) we could include the units in the form `u.Quantity[u.m/u.s]` on both input and output of the function, and at least in principle a static type checker could detect that we were putting a length in where a speed was needed. I don't know that any actual type checker can check this, but it would collect all the type-like information in one place tidily. And `quantity_input` (in sufficiently advanced astropy) can understand this. Plus, I believe, Sphinx can use this, so we should be able to drop the type annotations in the docstrings and they'll get automatically filled in in the rendered docs (might need to set something to make this happen).

src/pint/utils.py
```diff
@@ -1197,22 +1225,21 @@ def merge_dmx(model, index1, index2, value="mean", frozen=True):
             )
     if value.lower() == "first":
         dmx = getattr(model, f"DMX_{index1:04d}").quantity
-    elif value.lower == "second":
+    elif value.lower() == "second":
```
aarchiba, line 1228: Bug found!

PR #1726, opened 2024-03-09
Binary derivatives units and exact

DotanGazith:
Fixed the references in ELL1 model (references 1&2 were flipped)
Fixed the units of binary delay derivatives with respect to "angle" parameters (omega\E). Specifically in BT, DD models.
Added an exact derivative class for BT model. I kept it in a separate class to not clash with performance decisions that might drive the approximate derivative implemented.

### DIFF
--- a/src/pint/models/binary_ell1.py
+++ b/src/pint/models/binary_ell1.py
@@ -1,4 +1,5 @@
 """Approximate binary model for small eccentricity."""
+
 import astropy.units as u
 import numpy as np
 from astropy.time import Time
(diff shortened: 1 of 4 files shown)


DotanGazith:
Added a fix for the units that caused 3 of the errors.
The black formatting goes back and forth, I format it and the automatic pre-commit un-formats it, not sure what its actual preferences are.

dlakaplan:
Hi, @DotanGazith .  It looks like the black test is failing just on a single file.  I don't know exactly how you have things set up locally but this should be easy to fix in general.

dlakaplan:
More broadly, do you want to expose your exact BT model to the user?  I think you could do that (make a wrapper class in `/models/binary_bt.py`).  You could also add appropriate tests with the same structure as tests that involve BT models, such as `tests/test_B1953.py`).  Finally, it might be good to add some tests to explicitly compare the normal BT against this version (at the moment there shouldn't be any variations since you just copied things, but in case one or the other changes).  But some of those additions are up to you.

Issue #1727, opened 2024-03-13
Failures in Numpy 2.0: np.compat was removed


matteobachetti:
It's always a good idea to add routine tests on development versions of key dependencies. Lately, it has become clear that the next release of Numpy will break a lot of things, and Astropy and many other projects are trying to keep up with all the bugs that these API changes are introducing.
In stingray and HENDRICS, some advanced pulsar functionality depends on PINT, so we end up testing some functionality of PINT with the development version of Numpy. We are getting this error: 
```
../../.tox/py312-test-devdeps/lib/python3.12/site-packages/pint/models/stand_alone_psr_binaries/binary_orbits.py:28: in orbit_phase
    norbits = np.array(np.floor(orbits), dtype=np.compat.long)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

attr = 'compat'

    def __getattr__(attr):
        # Warn for expired attributes
        import warnings

    (...)
    
        if attr in __expired_attributes__:
>           raise AttributeError(
                f"`np.{attr}` was removed in the NumPy 2.0 release. "
                f"{__expired_attributes__[attr]}"
            )
E           AttributeError: `np.compat` was removed in the NumPy 2.0 release. There's no replacement, as Python 2 is no longer supported.. Did you mean: 'concat'?

../../.tox/py312-test-devdeps/lib/python3.12/site-packages/numpy/__init__.py:397: AttributeError
```

Example: https://github.com/StingraySoftware/HENDRICS/actions/runs/8271392969/job/22631073816?pr=161

abhisrkckl:
I think #1728 fixes this. Can this be closed?

matteobachetti:
I think so! Thanks!

PR #1728, opened 2024-03-14
Removed np.compat.long (deprecated)

dlakaplan:
Addresses #1727.  I think that in this case `np.compat.long` is just an alias for `int` (that's what I see when I look at it on my local system).


abhisrkckl:
This looks fine to me. Shall I merge this?

dlakaplan:
Please do

PR #1731, opened 2024-04-01
FDJUMPDM : System-dependent DM offset

abhisrkckl:
This PR implements the system-dependent DM offset for narrowband TOAs. Such offsets can arise in large datasets where the template profiles used for different observing systems may not have been dedispersed with the same fiducial DM. 

This is different from DMJump. DMJump applies to wideband TOAs and only provides a DM offset without a corresponding delay. 

FDJumpDM should only be used for narrowband TOAs, and its effect on wideband timing is yet unexplored.

The name FDJUMPDM was introduced by the IPTA DR3 data combination team partly because the name DMJUMP was already taken. This parameter has recently been added to tempo2.

Also adds an explanation for FDJUMPs and FDJUMPDMs in the documentation.

### DIFF
--- a/tests/test_fdjump.py
+++ b/tests/test_fdjump.py
@@ -130,7 +130,7 @@ def test_fdjumpdm_offset(model_and_toas):
     model, toas = model_and_toas
 
     model2 = deepcopy(model)
-    model2.FDJUMPDM.value = 0
+    model2.FDJUMPDM1.value = 0
 
     mask = model.FDJUMPDM1.select_toa_mask(toas)
     not_mask = np.setdiff1d(np.arange(len(toas)), mask)
--- a/tests/test_fdjump.py
+++ b/tests/test_fdjump.py
@@ -7,6 +7,7 @@ from pint.fitter import DownhillWLSFitter
 import pytest
 from io import StringIO
 import numpy as np
+from copy import deepcopy
 
 
 @pytest.fixture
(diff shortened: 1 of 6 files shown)


tests/test_fdjump.py
```diff
@@ -17,6 +17,8 @@ def model_and_toas():
         "FD1JUMP -sys GM_GWB_1460_100_b1 0.02 1\n"
         "FD2JUMP -sys GM_GWB_500_100_b1 0.001 1\n"
         "FD2JUMP -sys GM_GWB_1460_100_b1 0.002 1\n"
+        "FDJUMPDM -sys GM_GWB_500_100_b1 0.00002 1\n"
+        "FDJUMPDM -sys GM_GWB_1460_100_b1 0.00001 1\n"
```
dlakaplan, line 21: Do you think it would be worth including lines like these in the docstring or in `explanation.rst`, just to give a concrete example?

dlakaplan:
This looks pretty simple, but it might be helpful if there were a way to have a warning if it is used with WB data.  I realize that I don't think there is an easy way to that: is there?  Might it be better as a longer-term project to identify if a component has restrictions on the type of data it can be used with?

abhisrkckl:
> This looks pretty simple, but it might be helpful if there were a way to have a warning if it is used with WB data. I realize that I don't think there is an easy way to that: is there? Might it be better as a longer-term project to identify if a component has restrictions on the type of data it can be used with?

We can check with the `DMDATA` parameter? But I don't think that's a sure thing. It doesn't really enforce anything except for getting printed out into the par file.

Basically, we need both the TimingModel and TOAs objects together to do this type of validation.

One option is to check when a `Residuals` object is created. If a narrowband TOAs object is used with a par file with `DMJUMP` in it, the constructor of the `Residuals` class can throw a warning. Similar thing for `FDJUMPDM`.

Another option is to add additional checks in the `get_model_and_toas` function before returning the created objects.

dlakaplan:
I think those example lines will be helpful - thanks!

dlakaplan:
What about a test just verifying that the DM values are indeed properly offset?  Something simpler than the fits?

dlakaplan:
I think this looks good.  OK to merge?

abhisrkckl:
Yes. please merge this.

PR #1735, opened 2024-04-17
Changed error bars to scaled for pintk

### LINKED ISSUE #1734: PINTk error bars do not show scaled uncertainties?
As far as I can tell, the error bars in pintk are just `toas.get_errors()`, rather than `resids.get_data_error()` that would incorporate EFAC/EQUAD etc.  It seems like we should show the latter, at least optionally.

@scottransom ?

dlakaplan:
Closes #1734


abhisrkckl:

dlakaplan:
Yes, thanks.

PR #1736, opened 2024-04-17
Updated docs for ELL1H

dlakaplan:
Made it clear that when STIGMA is used, NHARMS is ignored.  In those cases NHARMS will remain unset unless it is explicitly used.

Made it clear that when NHARMS < 7 is specified, it will pick 7.

Both in the docstring and added warnings.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -36,6 +36,7 @@ the released changes.
 - `freeze_params` option in `wavex_setup` and `dmwavex_setup`
 - `plrednoise_from_wavex`, `pldmnoise_from_dmwavex`, and `find_optimal_nharms` functions
 - fake TOAs can be created with `subtract_mean=False`, to maintain phase coherence between different data sets
+- Better explanation of ELL1H behavior when H3/H4/STIGMA supplied and when NHARMS is used
 ### Fixed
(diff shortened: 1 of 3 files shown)


dlakaplan:
Updated docstring:
```
When `H3` only is supplied, `NHARMS` is ignored, and the approximate version is used (Eqn. 19) appropriate for medium inclinations.

    When `H3` and `H4` are supplied, `NHARMS` is taken to be `max(7,NHARMS)`, and the approximate version is used (Eqn. 19) appropriate for medium inclinations.
    Note that the default value in `pint` for `NHARMS` is 7, while in `tempo2` it is 4.

    When `H3` and `STIGMA` are supplied, `NHARMS` is ignored since the exact version is used (Eqn. 29) appropriate for very high inclinations.
```

dlakaplan:
Updated warnings:
```
WARNING  (pint.models.binary_ell1       ): Requested NHARMS=12 will be ignored, since will use exact parameterization with STIGMA specified
```

and
```
WARNING  (pint.models.binary_ell1       ): Requested NHARMS=5, but setting it to 7 since H4 is also specified
```

abhisrkckl:
This looks good. I'll merge this if ready.

dlakaplan:
I think #1738 first will make it easier.

PR #1738, opened 2024-04-18
Fixed common CI failure in `tests/test_observatory.py::test_json_observatory_input_latlon`

dlakaplan:
I think this failure was just that sometimes it was adding to a database in a different order.  This should fix that.

### DIFF
--- a/tests/test_observatory.py
+++ b/tests/test_observatory.py
@@ -297,7 +297,10 @@ def test_json_observatory_input_latlon(sandbox):
     for p in gbt_orig.__dict__:
         if p not in ["location", "_clock"]:
             # everything else should be identical
-            assert getattr(gbt_orig, p) == getattr(gbt_reload, p)
+            # but check for either order
+            assert (getattr(gbt_orig, p) == getattr(gbt_reload, p)) or (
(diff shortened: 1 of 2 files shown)


dlakaplan:
@scottransom : I think this might fix the bug you are seeing in #1732 and elsewhere.

tests/test_observatory.py
```diff
@@ -297,7 +297,10 @@ def test_json_observatory_input_latlon(sandbox):
     for p in gbt_orig.__dict__:
         if p not in ["location", "_clock"]:
             # everything else should be identical
-            assert getattr(gbt_orig, p) == getattr(gbt_reload, p)
+            # but check for either order
+            assert (getattr(gbt_orig, p) == getattr(gbt_reload, p)) or (
+                getattr(gbt_orig, p) == getattr(gbt_reload, p)[::-1]
+            )
```
abhisrkckl, line 303: Will a set comparison be more robust?
like `set(getattr(gbt_orig, p)) == set(getattr(gbt_reload, p))`

dlakaplan: yes, I realized that later.

PR #1740, opened 2024-04-22
WIP: Better algorithm for TCB to TDB conversion

abhisrkckl:
Implements a more general algorithm that covers more parameters.


abhisrkckl:
I think I have a better approach. So closing this...

PR #1741, opened 2024-04-23
Bayesian information criterion

abhisrkckl:
A function for computing BIC similar to AIC

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -39,6 +39,7 @@ the released changes.
 - Better explanation of ELL1H behavior when H3/H4/STIGMA supplied and when NHARMS is used
 - FDJumpDM component for System-dependent DM offsets
 - Documentation: Explanation for FDJUMP and FDJUMPDM
+- `bayesian_information_criterion()` function 
 ### Fixed
 - `MCMC_walkthrough` notebook now runs
 - Fixed runtime data README 
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -9,6 +9,7 @@ the released changes.
 
 ## Unreleased
 ### Changed
+- Better documentation for `akaike_information_criterion()`
 ### Added
 - `bayesian_information_criterion()` function 
 ### Fixed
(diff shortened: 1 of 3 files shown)


dlakaplan:
Can you not add in the WB version?  Is it that hard to do (I assume we would just add the chi^2 for TOAs to the chi^2 for DM)?

Also, while not necessary to do it for all of the functions, could you add type annotations for this and the AIC function?

dlakaplan:
It also might be nice with AIC and BIC to have a note illustrating how they could be used (i.e., compute for two models, how to compare) and add references?

abhisrkckl:
More detailed documentation added.

abhisrkckl:
> Can you not add in the WB version? 

I'll do that as a separate PR.

The problem is that the `WidebandTOAResiduals.calc_chi2` function calls `WidebandTOAFitter`, and I don't fully understand what it does under the hood. Depending on what exactly it does, my understanding of how to implement the normalization term in the log-likelihood may be wrong. I want to get a better hang of what's going on there before attempting to add AIC & BIC for WB data.

dlakaplan:
Docs are good.  I still think it would be nice to incrementally add type hinting, but if you don't want to do that I can merge this.

abhisrkckl:
I tried adding type hints. But importing TimingModel and TOAs into pint.util leads to circular imports.

dlakaplan:
OK, I see.  I _think_ you can give type hints as strings, and they will only be evaluated later.  But I am not 100% sure (https://mypy.readthedocs.io/en/stable/runtime_troubles.html#runtime-troubles). I feel like I did this recently but I can't remember where.

dlakaplan:
So yes, look at:
https://github.com/nanograv/PINT/blob/master/src/pint/models/astrometry.py#L641
I think there I put in a type hint for a return value as a string since the class was only defined later in that file.

abhisrkckl:
Thanks... I have added type hints.

dlakaplan:
Looks great.  I can merge it if you are set.

abhisrkckl:
Yes. Please do.

PR #1743, opened 2024-04-24
Make Astrometry faster & avoid long double operations where possible

abhisrkckl:
Avoid creating `SkyCoord` objects during Astrometry computations. This was partly implemented in #1646 for models with proper motion. This PR does the same thing for cases without proper motion. This has a significant performance advantage.

Also, avoid long double operations where possible. This is done by including a new column `tdbfloat` in the TOAs table. This is the same thing as `tdbld`, but reduced to float64 precision.

Includes some refactorings suggested by sourcery.

### DIFF
--- a/src/pint/models/pulsar_binary.py
+++ b/src/pint/models/pulsar_binary.py
@@ -309,11 +309,11 @@ class PulsarBinary(DelayComponent):
                 method_name = f"{p.lower()}_func"
                 try:
                     par_method = getattr(self.binary_instance, method_name)
-                except AttributeError:
+                except AttributeError as e:
                     raise MissingParameter(
                         self.binary_model_name,
                         f"{p} is required for '{self.binary_model_name}'.",
-                    )
+                    ) from e
                 par_method()
 
     # With new parameter class set up, do we need this?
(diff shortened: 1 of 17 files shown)


abhisrkckl:
**Profiling test**
(using the NGC6440E example dataset.)

Before
----------
image

After
------
image

i.e., a 45% reduction in execution time.

dlakaplan:
Have you verified that going to float instead of long double doesn't change precision?

src/pint/models/pulsar_binary.py
```diff
@@ -377,7 +376,7 @@ def update_binary_object(self, toas, acc_delay=None):
                 # it's already in ICRS
                 updates["obs_pos"] = tbl["ssb_obs_pos"].quantity
                 updates["psr_pos"] = self._parent.ssb_to_psb_xyz_ICRS(
-                    epoch=tbl["tdbld"].astype(np.float64)
+                    epoch=tbl["tdbld"]
```
dlakaplan, line 379: Should this be the ld version, or the float version?

src/pint/models/pulsar_binary.py
```diff
@@ -390,7 +389,7 @@ def update_binary_object(self, toas, acc_delay=None):
                     PulsarEcliptic(ecl=self._parent.ECL.value)
                 ).cartesian.xyz.transpose()
                 updates["psr_pos"] = self._parent.ssb_to_psb_xyz_ECL(
-                    epoch=tbl["tdbld"].astype(np.float64), ecl=self._parent.ECL.value
+                    epoch=tbl["tdbld"], ecl=self._parent.ECL.value
```
dlakaplan, line 392: Should this be the ld version, or the float version?

abhisrkckl: changed.

src/pint/models/pulsar_binary.py
abhisrkckl: changed.

abhisrkckl:
Closing this because the effects of the two separate changes are difficult to test together. 
I'll make separate PRs for each.

Issue #1742, opened 2024-04-24
macos-latest fails due to precision issue


abhisrkckl:
#1695 and #1741 are affected at least.

The error I'm seeing is 

> ImportError while loading conftest '/Users/runner/work/PINT/PINT/conftest.py'.
> conftest.py:49: in <module>
>     pint.utils.require_longdouble_precision()
> .tox/py312-test/lib/python3.12/site-packages/pint/utils.py:156: in require_longdouble_precision
>     raise PINTPrecisionError(
> E   pint.utils.PINTPrecisionError: PINT needs higher precision floating point than you have available. PINT uses the numpy longdouble type to represent modified Julian days, and this machine does not have sufficient numerical precision to represent sub-microsecond times with np.longdouble. On an M1 Mac you will need to use a Rosetta environment, or on a Windows machine you will need to us a different Python interpreter. Some PINT operations can work with reduced precision, but you have requested one that cannot.

This has nothing to do with the PR contents.

abhisrkckl:
Perhaps this is useful: https://github.com/grahampugh/Rosetta-2-install

dlakaplan:
#1745 implements one of the possibilities above, and seems to avoid the error.  Not sure if that's the right answer but we could do it temporarily if needed.

dlakaplan:
I'm going to mark this as tabled: I don't think we have a general solution, but we have something that will allow CI to proceed.

PR #1745, opened 2024-04-24
Try macos 12 to see if it is still intel

dlakaplan:
Change `macOS-latest` to `macOS-12` to see if that will force use of an intel processor (and hence have the precision needed).

Deals with #1742, although not optimally.


abhisrkckl:
I think this is a good enough solution for the time being while we figure out a better solution.

abhisrkckl:
Can you add a changelog entry?
I'll merge this since this issue is affecting other PRs.

dlakaplan:
I think you can merge when ready

PR #1747, opened 2024-04-25
Allow "CLK UNCORR" in par files 

abhisrkckl:
Required to correctly deal with IPTA MDC1 dataset.


dlakaplan:
How about a test to make sure that this is actually applied?

abhisrkckl:
> How about a test to make sure that this is actually applied?

Test added.

abhisrkckl:
Yes.

PR #1748, opened 2024-04-29
Make `Astrometry` & `SolarSystemShapiro` faster

abhisrkckl:
Avoid creating `SkyCoord` objects during `Astrometry` computations. This was partly implemented in #1646 for models with proper motion. This PR does the same thing for cases without proper motion. This has a significant performance advantage.
The tests introduced by #1646 also cover these changes.

In `SolarSystemShapiro`, avoid doing expensive and unnecessary slices of the `TOAs` table.

Spun off from #1743, to make testing easier.


abhisrkckl:
Performance
-----------------

Here is the profiling code:
```

m, t = get_model_and_toas(examplefile("NGC6440E.par.good"), examplefile("NGC6440E.tim"))
res = Residuals(t, m)

%timeit res.calc_phase_resids()
```

Before: 13.3 ms ± 45 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

After: 6.01 ms ± 19.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

dlakaplan:
I think this looks good.  Is any explicit new testing required, or is the fact that all of the usual tests passed here sufficient (I suspect the latter since you haven't yet changed the precision)?

Any changes wrt skipping the SkyCoord creation that need tests?

abhisrkckl:
I think the existing tests are sufficient.

dlakaplan:
So I can merge?

abhisrkckl:
Yes.

PR #1750, opened 2024-04-29
Allow pintk to read par files with `UNITS TCB` and `BINARY T2` + Exclude function parameters in pintk

abhisrkckl:


dlakaplan:
Looks straightforward.  Can merge when ready.

abhisrkckl:
Yeah, This is ready.

PR #1752, opened 2024-04-30
WIP: adding type hinting to almost all of utils.py

dlakaplan:


dlakaplan:
Note that I moved the type definitions to `pint.types` since there were a number of circular import issues.

src/pint/utils.py
```diff
@@ -2762,7 +2980,9 @@ def sherman_morrison_dot(Ndiag, v, w, x, y):
     return result, logdet_C
 
 
-def woodbury_dot(Ndiag, U, Phidiag, x, y):
+def woodbury_dot(
+    Ndiag: np.ndarray, U: np.ndarray, Phidiag: float, x: np.ndarray, y: np.ndarray
```
abhisrkckl, line 2984: Phidiag is also an array.

abhisrkckl:
Aside from the small correction above, I think this is ready to go.

src/pint/utils.py
dlakaplan: ok, thanks.  I had been following the docstring, but now both are fixed.

PR #1753, opened 2024-04-30
Fix split swx

LuluAgazie:
Fixing naming convention bugs in `pint.utils.split_swx` and added a test for it to 'test_utils.py'


dlakaplan:
Could you add an entry to `CHANGELOG-unreleased.md`?

PR #1755, opened 2024-05-03
Type hints in `pint.toa` and `get_model` & `get_model_and_toas` functions

abhisrkckl:
#1709


dlakaplan:
I don't know if we want to follow #1756 and define custom types.  Like `file_like` for where you have `Union[str, IO]`?

abhisrkckl:
Yes I'll use those definitions once that pr is merged.

src/pint/toa.py
```diff
@@ -567,15 +569,15 @@ def _parse_TOA_line(line, fmt="Unknown"):
...
-    alias_translation=None,
+    toatime: time.Time,
+    toaerr: u.Quantity,
+    freq: u.Quantity,
+    obs: Observatory,
+    dm: u.Quantity = 0.0 * pint.dmu,
+    name: str = "unknown",
+    flags: dict = {},
+    format: str = "Princeton",
+    alias_translation: dict = None,
```
dlakaplan, line 580: I thought that if the default was `None`, it needed an `Optional` tag since `None` is not a `dict`?
(don't know if that is present elsewhere too).

src/pint/toa.py
```diff
@@ -1456,7 +1468,7 @@ def __getitem__(self, index):
             # FIXME: what to do if length zero? How to ensure it's a string array even then?
             return np.array(r)
 
-    def __setitem__(self, index, value):
+    def __setitem__(self, index: toas_index_like, value):
```
dlakaplan, line 1471: type for `value`?

src/pint/toa.py
```diff
@@ -1703,11 +1715,13 @@ def get_pulse_numbers(self):
...
             return None
 
-    def get_flags(self):
+    def get_flags(self) -> table.Column:
         """Return a numpy array of the TOA flags."""
         return self.table["flags"]
 
-    def get_flag_value(self, flag, fill_value=None, as_type=None):
+    def get_flag_value(
+        self, flag: str, fill_value=None, as_type=None
```
dlakaplan, line 1723: types for `fill_value` and `as_type`?

src/pint/toa.py
abhisrkckl: Added Optional

dlakaplan:
I think there are still a few functions that need a return type of `None`

src/pint/toa.py
abhisrkckl: added Any

src/pint/toa.py
abhisrkckl: added

abhisrkckl:
Yes.

abhisrkckl:
#1709

PR #1756, opened 2024-05-03
Custom type definitions

dlakaplan:
In `__init__.py` defining:
```
# Something that is a Quantity or can behave like one (with units assumed)
quantity_like = Union[float, np.ndarray, u.Quantity]
# Something that is a Time or can behave like one
time_like = Union[float, np.ndarray, u.Quantity, time.Time]
file_like = Union[str, Path, IO]
```
that can be used elsewhere (instead of exhaustively repeating those).

### DIFF
--- a/src/pint/__init__.py
+++ b/src/pint/__init__.py
@@ -18,6 +18,9 @@ import astropy.units as u
 import numpy as np
 import pkg_resources
 from astropy.units import si
+from pathlib import Path
+
+from typing import Union, IO
 
 from pint import logging
 from pint.extern._version import get_versions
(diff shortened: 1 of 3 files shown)


dlakaplan:
#1709

dlakaplan:
Are there other types to add? Or is this sufficient to get started?  And should those type definitions by in the top-level `__init__.py` or elsewhere?

abhisrkckl:
I think these are good enough to start with. We can add more as needed.

dlakaplan:
OK, probably ready to merge then.  This will simplify a few other in-process PRs.

PR #1761, opened 2024-05-07
Fix #1759 (coordinate comparison identity)

### LINKED ISSUE #1759: tests/test_Galactic.py::TestGalactic::test_proper_motion_identity failing
This test evaluates a position at the POSEPOCH and compares against the baseline position.  The threshold is set to 1e-11 arcsec.  Normally this passes but I've been having intermittent failures.  I don't know why it's intermittent (it should be deterministic) and only on some architectures.  I also don't know if this is due to #1748. 

Any ideas @abhisrkckl ?

dlakaplan:
After astropy update to 6.1.0, the comparison between coordinates at the same epoch `tests/test_Galactic.py::TestGalactic::test_proper_motion_identity` fails. This is because:

`A.separation(B) = 1.3320388147503583e-14`
but
`B.separation(A) = 3.975693351829395e-16`
I think this happens because the RA comparison is done in different units in those cases.  In one it's done in degrees, so the error is ~1e-16 deg (machine precision).  The other is done in hourangle (~1e-16 h).  The latter is then converted to degrees which makes the comparison fail.

Not sure why this is different now but it's not a situation that comes up a lot.  This should just get around it.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -20,4 +20,5 @@ the released changes.
 - test for `pint.utils.split_swx()` 
 ### Fixed
 - `pint.utils.split_swx()` to use updated `SolarWindDispersionX()` parameter naming convention 
+- Fix #1759 by changing order of comparison
 ### Removed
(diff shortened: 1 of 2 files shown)


dlakaplan:
@abhisrkckl : please look at this.  If OK (and if it passes) then it can be merged and this latest issue can be closed.

PR #1763, opened 2024-05-13
Moved include_gps and include_bipm out of Observatory objects

### LINKED ISSUE #1691: Clock corrections applied to TZRMJD even if TZRSITE is BAT

paulray:
This is an attempt to clean up handling of `include_gps` and `include_bipm`

Rather than having those as attributes of an `Observatory` object, they are passed as arguments to the `site.clock_correction()` method.  So instead of this:
```
site = get_observatory("ao", include_bipm=True)
clock_corr = site.clock_corrections(time)
```
the flow now looks like this:
```
site = get_observatory("ao")
clock_corr = site.clock_corrections(
        time, include_gps=True, include_bipm=True, bipm_version="BIPM2015"
    )
```
This has no effect on most users, since `get_TOAs()` still accepts the same arguments, it just passes `include_bipm` to the clock correction method rather than the observatory constructor.

In addition, the `include_gps` and `include_bipm` arguments now are interpreted as the user wanting UTC times to get the UTC->GPS correction added and TT times to get the TT(BIPM) corrections included.  Even if they are True, those corrections will *NOT* get applied to a time that is already in TDB.  This behavior should fix #1691.

The one significant thing on the user side that I've seen so far is that a bunch of observatories (mainly TeV and GW) in observatories.json had `include_bipm=False` included in their definition (making it their default). This is now lost. I argue this is a good thing since whether or not BIPM corrections are included is not really a feature of the observatory. So, users will now have to specify `include_bipm=False` if that is the behavior they want when loading TOAs from those observatories.


paulray:
Fixes #1766

paulray:
Also made a fix to the `clock_correction()` method of `TopoObs`, where if `self._clock` is False/None and `self.clock_files` is True, it would replace the clock corrections with 0.0. I believe this was wrong and it should just add 0.0 to the prior corrections (the BIPM and GPS corrections), so I changed `corr =` to `corr +=` in that line.

abhisrkckl:
How does this interact with #1747?

Are clock GPS and BIPM corrections disabled when "CLOCK UNCORR" is given in the par file?

dlakaplan:
> The one significant thing on the user side that I've seen so far is that a bunch of observatories (mainly TeV and GW) in observatories.json had `include_bipm=False` included in their definition (making it their default). This is now lost. I argue this is a good thing since whether or not BIPM corrections are included is not really a feature of the observatory. So, users will now have to specify `include_bipm=False` if that is the behavior they want when loading TOAs from those observatories.

Will most users know how/whether to do that?  Should the various wrappers to load spacecraft TOAs have that set explicitly now?

paulray:
> > The one significant thing on the user side that I've seen so far is that a bunch of observatories (mainly TeV and GW) in observatories.json had `include_bipm=False` included in their definition (making it their default). This is now lost. I argue this is a good thing since whether or not BIPM corrections are included is not really a feature of the observatory. So, users will now have to specify `include_bipm=False` if that is the behavior they want when loading TOAs from those observatories.
> 
> Will most users know how/whether to do that? Should the various wrappers to load spacecraft TOAs have that set explicitly now?

Hmm. Most people will hopefully use `get_model_and_toas()` which will make the right choice based on the setting of CLOCK in your model.  Otherwise, it is up to the user. If they call `get_TOAs()` with no kwargs, they will get the defaults for `include_bipm` and `include_gps`, which are True.  Maybe there should not be defaults for those?  I just don't see how it makes sense for those settings to be part of the observatory. How does the fact that I'm using LIGO or HAWC imply that I don't want a BIPM timescale? I think we just need to educate people by emailing the PINT list and putting info in the release notes.

dlakaplan:
> > > The one significant thing on the user side that I've seen so far is that a bunch of observatories (mainly TeV and GW) in observatories.json had `include_bipm=False` included in their definition (making it their default). This is now lost. I argue this is a good thing since whether or not BIPM corrections are included is not really a feature of the observatory. So, users will now have to specify `include_bipm=False` if that is the behavior they want when loading TOAs from those observatories.
> > 
> > 
> > Will most users know how/whether to do that? Should the various wrappers to load spacecraft TOAs have that set explicitly now?
> 
> Hmm. Most people will hopefully use `get_model_and_toas()` which will make the right choice based on the setting of CLOCK in your model. Otherwise, it is up to the user. If they call `get_TOAs()` with no kwargs, they will get the defaults for `include_bipm` and `include_gps`, which are True. Maybe there should not be defaults for those? I just don't see how it makes sense for those settings to be part of the observatory. How does the fact that I'm using LIGO or HAWC imply that I don't want a BIPM timescale? I think we just need to educate people by emailing the PINT list and putting info in the release notes.

If we remove the defaults that would make it more explicit that users need to set this.  Maybe that is a good plan.  But need to make sure the upstream uses like `get_fits_TOAs` also do the same.

paulray:
The problem with removing the defaults is that everyone who has code that calls `get_TOAs()` without those kwargs will then get an exception. This might be aggravating, but on the other hand it forces people to think about what they really want.

I also looked at what happens when you call `get_TOAs()` with a `model=` argument. It can set `use_bipm`, `bipm_version` and `planets`, plus it will set `use_gps` to False *only* in the case that `CLK UNCORR` is in the model. 

So really BIPM is associated with the model and it makes sense to me that either its setting should be inherited from the model, or it should force the user to specify it explicitly.  On the other hand, I'm now re-thinking the GPS correction.  That really is a function of the observatory clock.  If the observatory clock corrections put you on UTC(GPS), as is the most common case, then you do want to apply the UTC(GPS) to UTC corrections.  Since this is site dependent, probably I should **not** have taken it out of the Observatory class.  The difference is that you should NOT use different settings for BIPM between different sites if you have multi-observatory datasets, but you SHOULD use different settings for GPS for each observatory if some use GPS clocks and some don't.

Do people agree that I should revert my changes to `use_gps`, but not `use_bipm`?

paulray:
In thinking about `use_gps` more, I changed its name to `apply_gps2utc` since that better characterizes what it does. It really is a feature of the observatory clock, and just part of the necessary clock corrections to go from raw observatory times to UTC (along with the observatory clock file like pks2gps.clk).  So, I think it should not be a kwarg to so many functions like `get_TOAs()`.  If you really want to use an observatory and NOT apply the correction from GPS to UTC that the observatory normally thinks should be done, you could instantiate the site yourself as `site = get_observatory('gbt',apply_gps2utc=False)`.  However, this does NOT overwrite the registry so if you later call `get_observatory('gbt')` you will get the normal version with `apply_gps2utc` False.

I'd like to hear opinions but I don't think we need to make it so easy to turn off the gps2utc corrections in normal code.  That setting should be part of observatories.json and that's basically it. Thoughts?

paulray:
Ok, question to consider. How easy should it be to not apply the UTC(GPS)->UTC corrections?
Is it sufficient to just need to modify `observatories.json` to turn it off for any observatory where you don't want it?

Or, if you want to be able to do it programmatically, what is acceptable? It is very painful to make sure that every call to `get_observatory()` has the kwarg to disable it.  But right now when you call `get_observatory('gbt',apply_gps2utc=False)` it makes a deepcopy of observatory object and then modifies that. This does NOT affect the registry and thus that setting will not "stick" so if you call `get_observatory()` again without the kwarg, the setting will disappear.  It seems to me OK to have a programmer call `get_observatory('gbt',apply_gps2utc=False,overwrite=True)` once and have that update the registry. Then future calls with no kwargs will get the observatory object with `apply_gps2utc=False`.  But this should be a pretty rare use case since normally you would want this correction for any observatory where it is correct to apply it.

paulray:
Ah, the deepcopy was added in #1711. I am going to revert that behavior in this PR.

CHANGELOG-unreleased.md
```diff
@@ -16,6 +16,10 @@ the released changes.
 - Type hinting for most of the `pint.utils` module
 - `funcParameter`s are no longer listed in the `pintk` interface.
 - Updated location of CCERA
+- Removed `include_bipm`, and `bipm_version` from the Observatory class. Now they are passed as arguments to `site.clock_correction()`
+- Renamed `include_gps` to `apply_gps2utc` in the Observatory class
+- Removed `apply_gps2utc` from `TOAs.clock_corr_info` since it can be different for different observatories. It is not a global setting.
+- The following observatories no longer have a default of include_bipm=False: magic, lst, virgo, lho, llo, geo600, kagra, hess, hawc
```
abhisrkckl, line 22: `include_bipm=False` in backticks

src/pint/data/runtime/observatories.json
```diff
@@ -1520,8 +1530,6 @@
         "origin": "An Imaging Atmospheric Cherenkov Telescope.\nThese coordinates are provided (from geodetic coordinates : lat = 23°16'18''S, lon = 16°30'00''E at 1800m asl) by the Maxime Regeard and Arache Djannati-Atai on behalf of the H.E.S.S. Collabotation, 2023 January 18."
     },
     "hawc": {
-        "include_bipm": false,
-        "include_gps": true,
```
abhisrkckl, line 1524: Isn't it better to set apply_gps2utc as false explicitly?

I am assuming that's the default behaviour.

src/pint/observatory/__init__.py
```diff
@@ -59,6 +59,7 @@
 
 # The default BIPM to use if not explicitly specified
 # FIXME: this should be auto-detected by checking the index file to see what's available
+# ALSO: Why is the a feature of the Observatory class?  Seems like the wrong place for it.
```
abhisrkckl, line 62: May be we can move this to the `pint` namespace or `pint.toa` namespace?

src/pint/observatory/__init__.py
```diff
@@ -272,12 +295,19 @@ def get(cls, name):
...
-            return cls._registry[cls._alias_map[name]]
+            site = cls._registry[name]
+        elif name in cls._alias_map:
+            site = cls._registry[cls._alias_map[name]]
+        else:
+            site = None
+        if site is not None:
+            if overwrite and apply_gps2utc is not None:
+                # This will modify the Observatory object in the registry, so it will "stick" until overwritten
+                site.apply_gps2utc = apply_gps2utc
```
abhisrkckl, line 309: Can we have a warning here telling the user to be extra careful?

src/pint/observatory/__init__.py
```diff
@@ -340,7 +371,13 @@ def timescale(self):
         astropy.time.Time()."""
         raise NotImplementedError
 
-    def clock_corrections(self, t, limits="warn"):
+    def clock_corrections(
+        self,
+        t,
```
abhisrkckl, line 376: What is the type of t? Please add a type hint or mention it in the docstring.

src/pint/observatory/__init__.py
```diff
@@ -469,30 +504,24 @@ def get_observatory(
...
-            site.bipm_version = bipm_version
-
-        return site
+    if apply_gps2utc is None:
+        site = Observatory.get(name)
+    else:
+        # If we are overriding the gps2utc setting then  site and update apply_gps2utc
+        # Otherwise, will just use default setting for this site.
+        log.debug(f"Observatory.get(apply_gps2utc={apply_gps2utc})")
+        site = Observatory.get(name, apply_gps2utc=apply_gps2utc, overwrite=True)
```
abhisrkckl, line 522: A warning here?

src/pint/observatory/__init__.py
```diff
@@ -771,7 +800,9 @@ def update_clock_files(bipm_versions=None):
     # FIXME: what to do about GPS and BIPM files?
 
     if bipm_versions is not None:
-        o = get_observatory("arecibo")
+        o = get_observatory(
+            "arecibo"
+        )  # Why arecibo? Would be better to use an observatory with no clock file of its own, right?
```
abhisrkckl, line 805: I agree.

src/pint/observatory/special_locations.py
```diff
@@ -86,20 +72,19 @@ def __init__(
...
     """Observatory-derived class for the solar system barycenter.
 
-    Time scale is assumed to be tdb."""
+    Time scale is assumed to be TDB, so no GPS or BIPM corrections should be applied."""
 
     def __init__(
         self,
         name,
         aliases=None,
+        apply_gps2utc=False,
```
abhisrkckl, line 81: Should this even be an option?

abhisrkckl: Is there any scenario where an SSB TOA will have a GPS correction?

src/pint/observatory/topo_obs.py
```diff
@@ -335,7 +313,13 @@ def _load_clock_corrections(self):
                 )
             )
 
-    def clock_corrections(self, t, limits="warn"):
+    def clock_corrections(
+        self,
+        t,
```
abhisrkckl, line 318: type of t?

CHANGELOG-unreleased.md
paulray: Done

src/pint/data/runtime/observatories.json
paulray: The default for `include_bipm` is true (I didn't change that).
However, I agree that having it set explicitly for each observatory is good.
I've done that for all the observatories where it was obvious (e.g. the clock file was XXX2gps.clk).
The correct choice for other observatories will need to come from people with specific knowledge of how that observatory does time stamps.

src/pint/observatory/special_locations.py
paulray: Good point. Removed.

src/pint/observatory/topo_obs.py
paulray: it is an astropy `Time`. It says in the docstring.

src/pint/observatory/__init__.py
paulray: I looked at this and in the end decided it was better where it was. The lowest level clock correction calculator is in the Observatory class. So I left it and deleted that comment.

What do you think about making the default auto-detect the most recent BIPM available?  I could add that.

src/pint/observatory/__init__.py
paulray: Added.

src/pint/observatory/__init__.py
paulray: Updated doctoring

src/pint/observatory/__init__.py
paulray: I added one in `Observatory.get` so no need for one here.

src/pint/observatory/__init__.py
abhisrkckl: > What do you think about making the default auto-detect the most recent BIPM available?

That makes sense.

paulray:
This is ready to merge. @dlakaplan or @scottransom do you want to give it a look and merge? I addressed @abhisrkckl's comments.

abhisrkckl:
I'll merge this if it is ready to go.

paulray:
> I'll merge this if it is ready to go.

It would be great to merge soon. I just had to fix several conflicts caused by adding type hints.

abhisrkckl:
Are you going to push more commits? If not, I'll merge this once the tests pass.

Issue #1577, opened 2023-05-18
Add CITATION.cff file


abhisrkckl:
https://citation-file-format.github.io/

abhisrkckl:
Who all should be listed as authors in the CITATION.cff file?

abhisrkckl:
Added in #1767

PR #1764, opened 2024-05-14
Update README with paper info

abhisrkckl:
This is a purely documentation update.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -20,6 +20,7 @@ the released changes.
 - `pintk` now reads and automatically converts TCB par files and par files with `BINARY T2`.
 - Test for `pint.utils.split_swx()`
 - Custom type definitions for type hints
+- PINT Paper BibTeX in the README file
 ### Fixed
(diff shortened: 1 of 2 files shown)


dlakaplan:
Looks good, but do you want to wait until you hear back about paper 2?

abhisrkckl:
OK That's fine. I'll remove the part about the second paper. That can be added later

abhisrkckl:
Removed reference to paper 2

dlakaplan:
Wouldn't it be better to do this as a citation.cff file (like #1577) instead?

abhisrkckl:
I think that's less visible than this. 

Also, the cff file needs a lot of information and is a lot of work (I tried before and gave up).

dlakaplan:
I just tried in #1767 and it looks OK.  We can differentiate between the full citation (all contributors) and the preferred journal citation.  Which isn't to say that we can't do both.

Personally I think including the full bibtex doesn't look great.  How about just a link to https://ui.adsabs.harvard.edu/abs/2021ApJ...911...45L/exportcitation?

abhisrkckl:
Closing this because cff file was added

PR #1765, opened 2024-05-14
WIP: Better algorithm for TCB -> TDB conversion

abhisrkckl:


dlakaplan:
So for people adding new parameters, are there instructions on how to determine the relevant factor?

abhisrkckl:
I'll add that.

abhisrkckl:
HOWTO entry added.

dlakaplan:
The explanation you have is very thorough, but it seems like it could still connect a little more directly to the factors used in the code.

docs/tcb2tdb-factors.rst
```diff
@@ -0,0 +1,72 @@
...
+EQUADs. Since we are not converting TOA variances, it doesn't make sense to convert TOA covariances
+either. Hence, ECORRs and red and DM noise parameters are not converted. This means that 
+the noise parameters must ALWAYS be re-estimated after a TCB <-> TDB conversion.
+
+Another exception to this are FD parameters and FD jumps, which involve polynomials of logarithms.
+Such functions transform in a non-linear fashion during TCB <-> TDB conversion, and we have chosen
+not to apply the conversion to such parameters. They too must be re-estimated after a TCB <-> TDB 
+conversion.
+
+In such cases, the parameter can be excluded from TCB <-> TDB conversion by passing `convert_tcb2tdb=True`
```
dlakaplan, line 71: Not `False`?

abhisrkckl: oops!

dlakaplan:
Are you going to update this as discussed to change the default behavior?  Or will that be a future PR?

abhisrkckl:
I'll update it.

abhisrkckl:
> Are you going to update this as discussed to change the default behavior? Or will that be a future PR?

Done.

src/pint/models/tcb_conversion.py
```diff
@@ -19,7 +26,7 @@
 IFTE_K = 1 + IFTE_KM1
 
 
-def scale_parameter(model, param, n, backwards):
+def scale_parameter(model: TimingModel, param: str, n: int, backwards: bool):
```
dlakaplan, line 29: should this have `-> None` for the return?

dlakaplan:
It looks like you put the conversion factor into a lot of unrelated tests.  Does that actually make a difference?  If the tests don't do TCB/TDB conversion, is there any step that would notice if the conversion factor is present or not?

abhisrkckl:
Since we made the conversion factor mandatory, I had to include them in tests whenever Parameter object was getting created. Otherwise the yest will fail.

dlakaplan:
OK, I guess I thought that the factor need not be mandatory but that the conversion step would fail without it.

src/pint/models/tcb_conversion.py
```diff
@@ -76,42 +83,35 @@ def transform_mjd_parameter(model, param, backwards):
...
+            and isinstance(par.param_comp, MJDParameter)
+        )
 
         par.value = (par.value - tref) * factor + tref
         if par.uncertainty_value is not None:
             par.uncertainty_value *= factor
 
 
-def convert_tcb_tdb(model, backwards=False):
+def convert_tcb_tdb(model: TimingModel, backwards: bool = False):
```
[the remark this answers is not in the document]
abhisrkckl: added

src/pint/models/tcb_conversion.py
```diff
@@ -52,14 +59,14 @@ def scale_parameter(model, param, n, backwards):
...
-        par = getattr(model, param)
+    if (param in model) and model[param].quantity is not None:
+        par = model[param]
         par.value *= factor
         if par.uncertainty_value is not None:
             par.uncertainty_value *= factor
 
 
-def transform_mjd_parameter(model, param, backwards):
+def transform_mjd_parameter(model: TimingModel, param: str, backwards: bool):
```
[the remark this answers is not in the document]
abhisrkckl: added

src/pint/models/tcb_conversion.py
abhisrkckl: added

abhisrkckl:
> OK, I guess I thought that the factor need not be mandatory but that the conversion step would fail without it.

I think that will be more confusing. If we are going to throw an error for missing scaling factors, I'd prefer it to be thrown when the parameter is added, rather than at a later stage. This will also force the devs to think more carefully about the parameters, IMO.

abhisrkckl:
OK... Please merge this. I got all the necessary feedback last week.

PR #1767, opened 2024-05-15
Try citation.cff

dlakaplan:
This can be updated.  The paper citation and overall author list need not be the same.


dlakaplan:
Can also add the zenodo DOI based on fixed versions if/when we do that.

dlakaplan:
Should probably update the full author list if people think this is worth doing.

dlakaplan:
Although, this may not work for > 1 paper so it may not be workable.

abhisrkckl:
> Although, this may not work for > 1 paper so it may not be workable.

Unfortunately, that doesn't work. According to the authors of cff, this is a feature.
https://github.com/citation-file-format/citation-file-format/issues/339

abhisrkckl:
> Should probably update the full author list if people think this is worth doing.

I think this is super helpful.

abhisrkckl:
> Can also add the zenodo DOI based on fixed versions if/when we do that.

ApJ people asked me to upload the version corresponding to the paper to zenodo. So we'll have a new zenodo DOI soon.

dlakaplan:
I updated the author list to the full list of contributors, but kept the paper as-is.  We'll see what we do when we have a second paper.

dlakaplan:
> > Can also add the zenodo DOI based on fixed versions if/when we do that.
> 
> ApJ people asked me to upload the version corresponding to the paper to zenodo. So we'll have a new zenodo DOI soon.

Yes, exactly.

citation.cff
```diff
@@ -0,0 +1,135 @@
...
+  given-names: "Rutger"
+- family-names: "van Kerkwijk"
+  given-names: "Marten"
+- family-names: "Zimmerman"
+  given-names: "Josef"
+title: "PINT: A Modern Software Package for Pulsar Timing"
+version: 1.0
+doi: 10.3847/1538-4357/abe62f
+date-released: 2024-04-26
+url: "https://github.com/github-linguist/linguist"
```
abhisrkckl, line 92: Please update this.

PR #1775, opened 2024-05-30
Whitened residuals in `pintk`

abhisrkckl:
image


dlakaplan:
This looks good.  My only question is if it's worth reordering the axis options. So that e.g., `WB DM res` and `white-res` are at the top near `pre-fit` and `post-fit`, and `WB DM err` is near `TOA error`.  I don't think this matters too much either way but it could make things more clear for some users.

dlakaplan:
Also needs CHANGELOG

abhisrkckl:
Reordered the axes.
image

dlakaplan:
great.  I can merge if you are done (and I can merge the other PR first if you'd prefer).

abhisrkckl:
I'm done. This can be merged first.

Issue #1661, opened 2023-10-23
Chromatic delay models


abhisrkckl:
PINT should have variable-index chromatic delay models

1. Taylor series representation (like DispersionDM)
2. Piecewise representation (like DMX)
3. Fourier series representation (like DMWaveX)
4. Fourier GP representation (like PLDMNoise)
5. Chromatic events (e.g. for J1713+0747)

abhisrkckl:
The chromatic exponential dip parameters for J1713+0747 look like this:

```
EXPEP_1        54764.272428904194001       
EXPEP_2        57500.886666587524999       
EXPPH_1        1.6641670367524487e-06      
EXPPH_2        2.1964077367938582999e-06   
EXPTAU_1       112.00425959054773          
EXPTAU_2       26.57155879530331           
EXPINDEX_1     -1.9148109887274356         
EXPINDEX_2     -1.5060816448307765999
```

abhisrkckl:
`TNChromIdx` is already part of `ChromaticCM`.

abhisrkckl:
The tempo2 exponential dip model is `dip = A * (f / 1 MHz)**gamma * exp(-(t-t0) / tau) * heaviside(t-t0)`. This implies that A has units of time.

abhisrkckl:
Since we need to fit for t0 also, we can't use the Heaviside theta function; it is not differentiable at 0. I am trying using the logistic function which tends to the Heaviside theta function in the appropriate limit.

PR #1777, opened 2024-06-03
Piecewise-constant chromatic variations (CMX)

abhisrkckl:
This depends on #1616. So That PR needs to be merged first. --- DONE.

Partially implements #1661.


kerrm:
This looks like a lot of work!
Overall it looks fine to me.
I think the big thing is to add to the test an explicit verification that the chromatic calculation is correct.  You've got some TOAs at different freqs, so verify that
(1) the delay is correct given the amplitude and chromatic index
(2) that the delay does *not* affect TOAs outside of any ranges
(3) that the lookup of ranges is correct, so do (1) for multiple events

mtlam:
I took a look and agree with @kerrm, the base functionality looks great. After any potential additional testing components might be added beyond the recent commits, I think it would be great to get this functionality in.

abhisrkckl:
> I think the big thing is to add to the test an explicit verification that the chromatic calculation is correct. You've got some TOAs at different freqs, so verify that
> (1) the delay is correct given the amplitude and chromatic index
> (2) that the delay does not affect TOAs outside of any ranges
> (3) that the lookup of ranges is correct, so do (1) for multiple events

Added tests for this

abhisrkckl:
I think this can be merged.

PR #1784, opened 2024-06-17
Minimal fixes for numpy 2.0

dlakaplan:
Trying minimal fixes to assure numpy 2.0 compatibility.  A lot of these are related to how float128 objects are treated, and which functions accept them.

### DIFF
--- a/src/pint/scripts/event_optimize.py
+++ b/src/pint/scripts/event_optimize.py
@@ -151,7 +151,7 @@ def profile_likelihood(phs, *otherargs):
     Likelihood is calculated as per eqn 2 in Pletsch & Clark 2015.
     """
     xvals, phases, template, weights = otherargs
-    phss = phases.astype(np.float64) + phs
+    phss = phases.astype(np.float64) + phs.astype(np.float64)
     phss %= 1
     probs = np.interp(phss, xvals, template, right=template[0])
     if weights is None:
--- a/tests/test_model_derivatives.py
+++ b/tests/test_model_derivatives.py
@@ -163,19 +163,27 @@ def test_derivative_equals_numerical(parfile, param):
         h3 = model.H3.value
         stepgen = numdifftools.MaxStepGenerator(abs(h3) / 2)
     elif param == "FB0":
-        stepgen = numdifftools.MaxStepGenerator(np.abs(model.FB0.value) * 1e-2)
+        stepgen = numdifftools.MaxStepGenerator(
+            np.abs(model.FB0.value.astype(np.float64)) * 1e-2
+        )
     elif param == "FB1":
-        stepgen = numdifftools.MaxStepGenerator(np.abs(model.FB1.value) * 1e3)
+        stepgen = numdifftools.MaxStepGenerator(
+            np.abs(model.FB1.value.astype(np.float64)) * 1e3
+        )
     elif param == "FB2":
-        stepgen = numdifftools.MaxStepGenerator(np.abs(model.FB2.value) * 1e5)
+        stepgen = numdifftools.MaxStepGenerator(
+            np.abs(model.FB2.value.astype(np.float64)) * 1e5
+        )
     elif param == "FB3":
-        stepgen = numdifftools.MaxStepGenerator(np.abs(model.FB3.value) * 1e7)
+        stepgen = numdifftools.MaxStepGenerator(
+            np.abs(model.FB3.value.astype(np.float64)) * 1e7
+        )
     else:
         stepgen = None
     df = numdifftools.Derivative(f, step=stepgen)
 
     a = model.d_phase_d_param(toas, delay=None, param=param).to_value(1 / units)
-    b = df(getattr(model, param).value)
+    b = df(getattr(model, param).value.astype(np.float64))
     if param.startswith("FB"):
         assert np.amax(np.abs(a - b)) / np.amax(np.abs(a) + np.abs(b)) < 1.5e-6
     else:
--- a/tests/test_model_derivatives.py
+++ b/tests/test_model_derivatives.py
@@ -154,7 +154,7 @@ def test_derivative_equals_numerical(parfile, param):
                 dphase = m.phase(toas, abs_phase=False) - phase
             except ValueError:
                 return np.nan * np.zeros_like(phase.frac)
-        return dphase.int + dphase.frac
+        return np.float64(dphase.int + dphase.frac)
 
     if param == "ECC":
         e = model.ECC.value
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -43,4 +43,5 @@ the released changes.
 - Moved the test in `test_pmtransform_units.py` into a function.
 - Fixed bug in residual calculation when adding or removing phase wraps
 - Fix #1766 by correcting logic and more clearly naming argument (clkcorr->undo_clkcorr)
+- Minimal fixes to allow usage of numpy 2.0
 ### Removed
--- a/src/pint/fitter.py
+++ b/src/pint/fitter.py
@@ -490,9 +490,11 @@ class Fitter:
         """
 
         return self.model.get_derived_params(
-            rms=self.resids.toa.rms_weighted()
-            if self.is_wideband
-            else self.resids.rms_weighted(),
+            rms=(
+                self.resids.toa.rms_weighted()
+                if self.is_wideband
+                else self.resids.rms_weighted()
+            ),
             ntoas=self.toas.ntoas,
             returndict=returndict,
         )
@@ -1242,7 +1244,7 @@ class DownhillFitter(Fitter):
             for fp, x in zip(free_noise_params, xs):
                 getattr(res.model, fp).value = x
 
-            return -res.lnlikelihood()
+            return -res.lnlikelihood().astype(np.float64)
 
         if not res.model.has_correlated_errors:
 
--- a/src/pint/models/model_builder.py
+++ b/src/pint/models/model_builder.py
@@ -1081,14 +1081,14 @@ def convert_binary_params_dict(
             log.debug("Converting KIN to/from IAU <--> DT96")
             entries = parfile_dict["KIN"][0].split()
             new_value = _convert_kin(float(entries[0]) * u.deg).value
-            parfile_dict["KIN"] = [" ".join([repr(new_value)] + entries[1:])]
+            parfile_dict["KIN"] = [" ".join([str(new_value)] + entries[1:])]
 
         # Convert KOM if requested
         if convert_komkin and "KOM" in parfile_dict:
             log.debug("Converting KOM to/from IAU <--> DT96")
             entries = parfile_dict["KOM"][0].split()
             new_value = _convert_kom(float(entries[0]) * u.deg).value
-            parfile_dict["KOM"] = [" ".join([repr(new_value)] + entries[1:])]
+            parfile_dict["KOM"] = [" ".join([str(new_value)] + entries[1:])]
 
         # Drop SINI if requested
         if drop_ddk_sini and binary_model_guesses[0] == "DDK":
--- a/src/pint/models/parameter.py
+++ b/src/pint/models/parameter.py
@@ -103,7 +103,7 @@ def _return_frequency_asquantity(f):
     astropy.units.Quantity
     """
 
-    return u.Quantity(f, u.MHz, copy=False)
+    return f << u.MHz
 
 
 class Parameter:
--- a/src/pint/plot_utils.py
+++ b/src/pint/plot_utils.py
@@ -168,10 +168,10 @@ def phaseogram_binned(
             ax2.scatter(phss, mjds, s=size, color=colarray)
             ax2.scatter(phss + 1.0, mjds, s=size, color=colarray)
     else:
-        profile = np.zeros(bins, dtype=np.float_)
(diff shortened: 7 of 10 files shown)


dlakaplan:
I'll try to fix the obvious things here.  Some issues may be related to astropy.  And I also see an issue around longdoubles.  I see behavior like:
```
In [1]: np.longdouble('41498.0')
Out[1]: np.longdouble('41498.0')
```
with numpy 2.0, while previously it returned just the numerical value.  The value it returns is useable, but it confuses the float/string conversion tasks.  It's further confused by the attempt to change fortran values to normal exponentials, since the word `longdouble` gets changed to `longeouble`.  Not sure what the fix is (the fix for the second part shouldn't be hard, but it doesn't work even without that).

dlakaplan:
Many errors are like:
```
tests/test_jump.py::test_jump_params_to_flags - ValueError: Unable to avoid copy while creating an array as requested.
```
That seems to be traced back to:
```
/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/models/parameter.py:106: in _return_frequency_asquantity
    return u.Quantity(f, u.MHz, copy=False)
```
which itself relies on astropy.  I don't know why `copy=False` is specified here, and not anywhere else.  I'll see if it's any easy fix otherwise.

dlakaplan:
Looks like 64-bit floats are also returned looking like strings, which cases errors like:
```
FAILED tests/test_t2binary2pint.py::test_t2binary2pint - ValueError: could not convert string to float: 'np.float64(-2.8244442419999984)'
```

dlakaplan:
And some were caused by the numerical derivative tests not accepting 128-bit floats either.  Trying to cast those to 64.

dlakaplan:
@abhisrkckl : I think there are a number of failures in the `test_noisefit` module, which seems to call `numdifftools.extrapolation.convolve1d` which calls `scipy.ndimage.correlate1d` in some of the matrix math.  And that can't support 128-bit.  Is this a calculation that would be safe to cast to 64-bit?  Do you think that numdifftools or scipy will be updated instead (I don't know if this is a known issue for them, or if you can file one).

dlakaplan:
Seems like changes to the `repr` function in np 2.0 are responsible for the errors:
```
FAILED tests/test_precision.py::test_longdouble_str_roundtrip_is_exact - ValueError: invalid literal for long double: np.longeouble('41498.0')
FAILED tests/test_precision.py::test_longdouble2str_same_as_str_and_repr - assert "np.longdouble('41498.0')" == '41498.0'
```

dlakaplan:
The failure in the notebook build is in:
```
examples/noise-fitting-example.ipynb
```
and it's also related to the noise fitting/convolve1d/correlate1d issue above.

dlakaplan:
This passes all tests now.  I think that all of the changes I have made are minimal and do not affect functionality (some are changes to the precision used in tests, others just change `repr` -> `str`).  But @abhisrkckl it would be good to check.

tests/test_model_derivatives.py
```diff
@@ -154,7 +154,7 @@ def f(value):
                 dphase = m.phase(toas, abs_phase=False) - phase
             except ValueError:
                 return np.nan * np.zeros_like(phase.frac)
-        return dphase.int + dphase.frac
```
abhisrkckl, line 157: Is this really OK? I'd think a test of phase would need more precision?

tests/test_model_derivatives.py
```diff
@@ -163,19 +163,27 @@ def f(value):
         h3 = model.H3.value
         stepgen = numdifftools.MaxStepGenerator(abs(h3) / 2)
     elif param == "FB0":
-        stepgen = numdifftools.MaxStepGenerator(np.abs(model.FB0.value) * 1e-2)
```
abhisrkckl, line 166: Why aren't FB parameters `float64`s to begin with?

abhisrkckl: I don't think even the rotational frequency needs more precision.

abhisrkckl:
I checked the released and dev versions of `scipy` and `astropy`. They seem to allow numpy v2.0.0 in their config files. I think they may have worked with the numpy devs to make sure that nothing breaks. So I think we are OK as long as our own tests pass.

abhisrkckl:
If we merge this we'll have to do a new release ASAP. I am pretty sure the pip and conda versions are broken right now since we don't explicitly disallow numpy>=2.

abhisrkckl:
When I do a grep I see that there are a number of `repr` calls in `polycos.py`. Do you think those should be corrected?

tests/test_model_derivatives.py
dlakaplan: Agree.  Maybe that is more general bookkeeping that could be done elsewhere (reduce precision where not needed).  FB0 is set to `longdouble=True` on creation.  I could change that here, or we can just deal with that later.

scottransom: Rotational frequency is one of the only timing parameters that *does* need to be long double (40 years of timing a MSP with 100ns timing precision gives a nominal absolute frequency precision of 100ns/(40yrs) ~ 1e-16, and there are 3 more sig figs before the decimal place!).  However, the FB terms can have numerical stability issues (since we are doing polynomial calculations with terms that can have amplitudes of like 1e-100) and so I think that is why we went with long double with those.  Numerical derivatives make the numerical stability issues even worse.  So I think we should be very careful with this...

dlakaplan: Careful in the sense that we shouldn't change the precision of the parameter, or even the precision of the derivative test?  If the former then we can just proceed.  If the latter then we need to understand how to do the tests better.

abhisrkckl: The Nyquist frequency for 40 years is about 0.3 nHz. If we have a 700 Hz pulsar, shouldn't the space needed to represent F0 be log2(700 Hz / 0.3 nHz) ~ 41 bits?

abhisrkckl: Ok... If we consider the numerical stability of Horner method, it makes sense to me that we'll need long double precision.

dlakaplan:
> When I do a grep I see that there are a number of `repr` calls in `polycos.py`. Do you think those should be corrected?

I think those are OK.  They are only in printing the individual entries.  Like:

```
Middle Point mjd : <Quantity 55000. d>
Time Span in mjd : <Quantity 0.20833333 d>
Time Start in mjd : <Quantity 54999.89583333 d>
Time Stop in mjd : <Quantity 55000.10416667 d>
Reference Phase : Phase(int=<Quantity [8.93793914e+10]>, frac=<Quantity [0.361084]>)
Reference Freq in Hz : np.longdouble('186.494081566982')
Number of Coefficients : 12
Coefficients : array([-1.61789223e-08,  3.19012578e-01, -1.16058620e-04, -5.12350025e-08,
        6.16132626e-12,  4.91933105e-14, -3.45712955e-18, -6.24302027e-21,
        2.10181635e-23, -5.34503450e-25, -4.58956216e-28,  6.75413195e-30],
      dtype=float128)
```
a lot of those are quantities.    There are still tests for polyco functionality which pass.  So maybe we can leave this alone.

abhisrkckl:
Ok...

abhisrkckl:
OK... This looks ready to go. Shall I merge this?

dlakaplan:
I think so, unless @paulray or @scottransom have any concerns.

tests/test_model_derivatives.py
scottransom: Oops.  I forgot a factor of F0.  For 100ns timing precision and a 40-year baseline, if you have a 1.5ms pulsar, then the nominal absolute freq error is roughly deltaPhi / deltaT ~ (100ns/1.5ms) / 40 yrs ~ 5e-14.  So that is 13 or 14 sig figs.  Then there are 3 more sig figs for the freq before the decimal point.  Double precision floats can give you ~15 sig figs (or digits of precision).  And we are needing that precision right now -- if you look at the recent data releases from all the PTAs, the F0 errors for the best pulsars are better than 1e-13, meaning 16 full digits of precision.

As for your question about bits, the number of bits needed in the mantissa must therefore be more than log2(1e16) = 53.15, and IEEE doubles have 52 bits in the mantissa.

(I think it is really important that pulsar students learn stuff like this -- we need to keep teaching them.  :-)

tests/test_model_derivatives.py
```diff
@@ -154,7 +154,7 @@ def f(value):
                 dphase = m.phase(toas, abs_phase=False) - phase
             except ValueError:
                 return np.nan * np.zeros_like(phase.frac)
-        return dphase.int + dphase.frac
+        return np.float64(dphase.int + dphase.frac)
```
scottransom, line 157: Hmmm.  That sum was long double before, right?  This type of conversion worries me a bit.  Because the int in this case can be tens of billions or maybe even hundreds of billions if people aren't careful.  Given 15 digits of decimal precision in a float64, that means that we only have 4-5 digits of precision for the fractional phase.  Is that accurate enough?

tests/test_model_derivatives.py
```diff
@@ -163,19 +163,27 @@ def f(value):
         h3 = model.H3.value
         stepgen = numdifftools.MaxStepGenerator(abs(h3) / 2)
     elif param == "FB0":
-        stepgen = numdifftools.MaxStepGenerator(np.abs(model.FB0.value) * 1e-2)
+        stepgen = numdifftools.MaxStepGenerator(
+            np.abs(model.FB0.value.astype(np.float64)) * 1e-2
```
scottransom, line 167: I missed the rational for all of these explicit casts to float64.  What was the reason?  As I mentioned, I think the FB terms (of which there are often many, like 10+ for redbacks) are highly numerically unstable.  Do we need to do this for some reason?

tests/test_model_derivatives.py
abhisrkckl: Ok... I was thinking the precision was related to the total number of bits, not the bits allotted for the mantissa. Thanks for the explanation.

tests/test_model_derivatives.py
scottransom: Interestingly, I think I may have just run into an issue with TOA simulation that is exactly related to this.  I need to simulate good TOAs for a very fast MSP over 20+ years and here is what I got when I plotted the raw simulated TOAs.  Note the quantization.  I think that is likely the result of the a float64 representing phase being "saturated".  I will try to investigate (eventually).
sim_prefit

tests/test_model_derivatives.py
dlakaplan: The numerical derivative routines in `numdifftools` rely on `numdifftools.extrapolation.convolve1d` which calls `scipy.ndimage.correlate1d`.  Those routines fail explicitly with an error about not recognizing `float128`.  The last release of `numdifftools` was 2015, so I don't think it will be updated.  

The casting is not in the actual calculation, but only in the tests which compare our analytical derivatives to the numerical values.  So I don't think this will affect our results.

tests/test_model_derivatives.py
dlakaplan: Yes, it was long double before, but this is just a test, not the actual code that is used.

src/pint/scripts/event_optimize.py
```diff
@@ -151,7 +151,7 @@ def profile_likelihood(phs, *otherargs):
     Likelihood is calculated as per eqn 2 in Pletsch & Clark 2015.
     """
     xvals, phases, template, weights = otherargs
-    phss = phases.astype(np.float64) + phs
+    phss = phases.astype(np.float64) + phs.astype(np.float64)
```
dlakaplan, line 154: Another instance where data are cast explicitly to 64-bit.  Here it seemed like `phases` was already 64-bit, and then `phs` was added to that, so there would be no/minimal loss in precision.

dlakaplan:
@scottransom : I just noted above the two instances where casting to 64-bit affects the actual code and not just tests.  I think casting the likelihood is very unlikely to cause any issues.  The phases in `event_optimize` might be more complicated, but it was already partly 64-bit.

src/pint/scripts/event_optimize.py
dlakaplan: @kerrm : do you think this is OK?  Probably it was silently casting without telling us, so maybe we just have to do it explicitly now.

kerrm:
Hi David -- yes, that's fine!

src/pint/scripts/event_optimize.py
matteobachetti: As a general approach, `astype(float)` might be considered instead of using the Numpy type. Python uses 64-bit floats anyway in most architectures, and `float` is sure to be supported over time (while, e.g., Numpy  has recently deprecated and eliminated `np.float`)

dlakaplan: Thanks, @matteobachetti .  I think it might be better to cast to an explicit precision here rather than leave it architecture dependent.  That way if the architecture doesn't support it, it will fail explicitly rather than just work with altered precision.  But maybe I'm thinking about this the wrong way.

abhisrkckl:
Can this be merged? Are there any unresolved concerns?

dlakaplan:
> Can this be merged? Are there any unresolved concerns?

I thought @devbhakt was going to check over the `event_optimize` code and give us his opinion.  But it looks like @kerrm thinks it's OK, so if you don't want to wait it's probably OK to merge.

abhisrkckl:
Ok here goes nothing. YOLO!

PR #1790, opened 2024-06-20
Handle BIPM and wideband DMs correctly in `make_fake_toas_fromtim` 

abhisrkckl:


src/pint/simulation.py
```diff
@@ -514,25 +514,14 @@ def make_fake_toas_fromtim(
...
-        if hasattr(model, "PLANET_SHAPIRO") and model.PLANET_SHAPIRO.value is not None
-        else False
-    )
-
-    input_ts = pint.toa.get_TOAs(timfile, ephem=ephem, planets=planets)
+    input_ts = pint.toa.get_TOAs(timfile, model=model)
 
     if input_ts.is_wideband():
         dm_errors = input_ts.get_dm_errors()
         ts = update_fake_dms(model, input_ts, dm_errors, add_noise)
```
dlakaplan, line 521: You are creating `ts` here if it's Wideband, but if it's not Wideband then I don't know if that variable is used.  which makes the command below suspect.

abhisrkckl: Corrected.

abhisrkckl:
This is ready to go.

src/pint/simulation.py
```diff
@@ -514,25 +514,14 @@ def make_fake_toas_fromtim(
...
-        model.EPHEM.value
-        if hasattr(model, "EPHEM") and model.EPHEM.value is not None
-        else None
-    )
-    planets = (
-        model.PLANET_SHAPIRO.value
-        if hasattr(model, "PLANET_SHAPIRO") and model.PLANET_SHAPIRO.value is not None
-        else False
-    )
+    ts = pint.toa.get_TOAs(timfile, model=model)
```
dlakaplan, line 517: So setting the model attribute will propagate BIPM, planets, etc, right?

abhisrkckl: It should.

PR #1793, opened 2024-06-26
Updates for scipy 1.14

dlakaplan:
Note that I couldn't base this off the version with the numpy 2.0 updates applied, since they are not available to update my fork (probably because the tests are failing).  So hopefully this merge doesn't get complicated.


dlakaplan:
And this is failing in oldestdeps.  I was worried about that (I couldn't find when the name change happened).  I'll add a conditional  import

dlakaplan:
Looks like the name change happened in https://github.com/scipy/scipy/releases/tag/v1.6.0.  So as long as our oldestdeps includes releases prior to that, we need to support it.  And currently we go back to 1.4.1.

PR #1704, opened 2023-12-27
Consistent naming in `get_params_mapping`

abhisrkckl:

PR #1794, opened 2024-06-26
Fix removal of top-level parameter

dlakaplan:
Previously there was a bug such that if you tried to remove a parameter that was part of the top-level `TimingModel` it would fail.  This is a small fix and a test.


dlakaplan:
Such removals may not always make sense, but you might remove something only to replace it later and this should facilitate that.

abhisrkckl:
This was introduced by #1704. This used to work before that, I think.

dlakaplan:
OK.  So hopefully this can be merged to fix this.

PR #1795, opened 2024-06-26
Allow setting of parameter quantity/value implicitly

dlakaplan:
Often I will accidentally do things like:
```
model.F0 = 10*u.Hz
```
or otherwise set the parameter to a value when I mean to set `parameter.quantity` or `parameter.value`.  This can be annoying.  However, nothing prevents you from doing this even if it makes no sense (I don't think it's caught by the validate step, but later attempts to fit the data end up with weird exceptions).

My thoughts to fix this were to either catch this and through an explicit exception or to try to interpret the intent.  This attempts to do the latter, although if that is too messy I can change it to the former.

So now with this PR if you do:
```
model.F0 = 10*u.Hz
```
you get:
```
WARNING  (pint.models.timing_model      ): Setting 'F0.quantity' to '10 Hz' although 'quantity' not specified
```
and similarly if you do:
```
model.PLANET_SHAPIRO = False
```
you get:
```
WARNING  (pint.models.timing_model      ): Setting 'PLANET_SHAPIRO.value' to 'True' although 'value' not specified
```
In both cases the quantity/value is correctly set.

### DIFF
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -545,17 +545,6 @@ class TimingModel:
         """Mostly this just sets ``self.name = value``.  But in the case where they are both :class:`Parameter` instances
         with different names, this copies the ``quantity``, ``uncertainty``, ``frozen`` attributes only.
         """
-        # print(name, value)
-        # try:
-        #     print(getattr(self, name).__class__)
-        # except:
-        #     pass
-        # try:
-        #     print(f"{name} in params: {name in self.params}")
-        # except:
-        #     print(f"Cannot tell if {name} is in params")
-        # if hasattr(self, "top_level_params"):
-        #     print(f"{self.top_level_params}")
         if isinstance(value, (Parameter, prefixParameter)) and name != value.name:
             for p in ["quantity", "uncertainty", "frozen"]:
                 setattr(getattr(self, name), p, getattr(value, p))
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -555,7 +555,7 @@ class TimingModel:
             getattr(self, name).quantity = value
         elif isinstance(value, (float, str, bool, int)) and name != "name":
             log.warning(
-                f"Setting '{name}.value' to '{value}' although 'value' not specified"
+                f"Setting '{name}.value' to '{value}' (assumed units '{getattr(self,name).units}') although 'value' not specified"
             )
             getattr(self, name).value = value
         else:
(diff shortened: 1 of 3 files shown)


src/pint/models/timing_model.py
```diff
@@ -545,9 +545,30 @@ def __setattr__(self, name, value):
...
+        #     print(f"{self.top_level_params}")
         if isinstance(value, (Parameter, prefixParameter)) and name != value.name:
             for p in ["quantity", "uncertainty", "frozen"]:
                 setattr(getattr(self, name), p, getattr(value, p))
+        elif isinstance(value, u.Quantity):
+            log.warning(
+                f"Setting '{name}.quantity' to '{value}' although 'quantity' not specified"
+            )
+            getattr(self, name).quantity = value
+        elif isinstance(value, (float, str, bool, int)) and name != "name":
```
abhisrkckl, line 567: This can also be a Time object.

abhisrkckl:
Does it make sense to have the same feature for Component class?

abhisrkckl:
I have also done this accidentally and wasted too long debugging. I think this is a good feature to have.

src/pint/models/timing_model.py
abhisrkckl: I think a bit more validation is needed here. Right now this will allow one to set a string as the value of a floatParameter, if I understand correctly. I think that is a potential foot gun.

dlakaplan: Time objects added

dlakaplan: I think your validation comment is handled by the `@setter` methods already.  e.g., `m.F0.value = 'abc'` is prevented already, and this works here too

dlakaplan: similarly, a quantity with the wrong units will be prevented.

dlakaplan:
The tests now demonstrate when this does/does not work

dlakaplan:
> Does it make sense to have the same feature for Component class?

I don't think I've ever made that error.  Can you share an example of what you mean?

src/pint/models/timing_model.py
abhisrkckl: Ok...

abhisrkckl:
I thought it would be nice to have the same parameter access API for the timing model and components, just like it works now.

abhisrkckl:

dlakaplan:
I didn't know if we want to discuss it a bit more.  Like throw an exception vs. try to do something intelligent.

abhisrkckl:
Ok... Let us wait for comments from others.

scottransom:
I've definitely done this multiple times, myself.  So I like this addition.  Has anyone thought of any potential downsides?

dlakaplan:
OK, now when setting `.value` implicitly the warning message lists the units:
```
Setting 'F0.value' to 'abc' (assumed units 'Hz') although 'value' not specified
```

abhisrkckl:
Can you please add a CHANGELOG entry?

PR #1797, opened 2024-06-30
Powerlaw chromatic noise model (`PLChromNoise`)

abhisrkckl:
See #1661

### DIFF
--- a/src/pint/models/noise_model.py
+++ b/src/pint/models/noise_model.py
@@ -24,7 +24,7 @@ class NoiseComponent(Component):
         # need to think about the design now. If we do not define the list
         # here and calling the same name from other component, it will get
         # it from the component that hosts it. It has the risk to duplicate
-        # the list elements.
+        # the list elements.What is the origin of these defaults?
         self.dm_covariance_matrix_funcs_component = []
         self.basis_funcs = []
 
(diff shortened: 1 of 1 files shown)


src/pint/models/__init__.py
```diff
@@ -51,13 +57,3 @@
...
-
-# Define a standard basic model
-StandardTimingModel = TimingModel(
-    "StandardTimingModel",
-    [AstrometryEquatorial(), Spindown(), DispersionDM(), SolarSystemShapiro()],
-)
-# BTTimingModel = generate_timing_model("BTTimingModel",
-#         (Astrometry, Spindown, Dispersion, SolarSystemShapiro, BT))
-# DDTimingModel = generate_timing_model("DDTimingModel",
-#         (Astrometry, Spindown, Dispersion, SolarSystemShapiro, DD))
```
abhisrkckl, line 63: This looks unnecessary. None of the tests depend on it, at least. It looks like vestigial lines of an early design approach that changed afterward.

src/pint/utils.py
```diff
@@ -3026,7 +3026,7 @@ def woodbury_dot(
 
     logdet_N = np.sum(np.log(Ndiag))
     logdet_Phi = np.sum(np.log(Phidiag))
-    _, logdet_Sigma = np.linalg.slogdet(Sigma)
+    _, logdet_Sigma = np.linalg.slogdet(Sigma.astype(float))
```
dlakaplan, line 3029: Are there no cases where long double precision is needed for this?

src/pint/models/noise_model.py
```diff
@@ -556,6 +556,127 @@ def pl_dm_cov_matrix(self, toas):
...
+    This model should be used in combination with the ChromaticCM model.
+
+    Parameters supported:
+
+    .. paramtable::
+        :class: pint.models.noise_model.PLChromNoise
+
+    Note
+    ----
+    Ref: Lentati et al. 2014
```
dlakaplan, line 583: Can you give the full reference?

tests/test_noise_models.py
```diff
@@ -25,22 +25,33 @@
...
 
 
 @pytest.fixture()
 def model_and_toas():
     parfile = examplefile("B1855+09_NANOGrav_9yv1.gls.par")
     timfile = examplefile("B1855+09_NANOGrav_9yv1.tim")
     model, toas = get_model_and_toas(parfile, timfile)
-    model = add_DM_noise_to_model(model)
+    add_DM_noise_to_model(model)
+    add_chrom_noise_to_model(model)
```
dlakaplan, line 54: you are adding both types to the same model?  Is that what you want to do?  You don't want to add to two separate models and compare?

src/pint/utils.py
abhisrkckl: I don't think so. For comparison, ENTERPRISE uses only double precision for the same computation.

src/pint/models/noise_model.py
abhisrkckl: added

tests/test_noise_models.py
abhisrkckl: Yes. The tests in `test_noise_models.py` check things like the shape of weight and basis matrices etc. It does not do any "science" tests. So I think it's safe to piggyback like this. 

Things like simulation and fitting tests are in `test_plchromnoise.py`.

src/pint/utils.py
abhisrkckl: More importantly, the `np.longdouble` type is unsupported in this function.

dlakaplan:
OK, I think this looks good and I can merge if you are ready.

abhisrkckl:
Yes, please merge

PR #1798, opened 2024-07-02
Run tests in parallel

abhisrkckl:
Github runners now provide multiple cores. 
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Our tests can be sped up significantly if we run them in parallel.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -77,7 +77,7 @@ jobs:
         python -c "import pip; print(f'pip {pip.__version__}')"
         python -c "import tox; print(f'tox {tox.__version__}')"
     - name: Run tests
-      run: tox -p auto -e ${{ matrix.tox_env }}
+      run: tox -e ${{ matrix.tox_env }}
     - name: Upload coverage to codecov
       if: "endsWith(matrix.tox_env, '-cov')"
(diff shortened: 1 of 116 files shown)


abhisrkckl:
The test failures seem to be due to `astropy` cache issues. #1374 may be a solution.

abhisrkckl:
Here is a summary of what I am doing:
1. Parallelize tests using `pytest-xdist`. I am using 6 processes here. This is greater than the number of CPU cores available (4), but that's OK because many of our tests are IO-limited rather than CPU-limited. I am getting ~4-5x improvement in test execution time.
2. Re-run failed tests using `pytest-rerunfailures`. This handles intermittent failures that occur due to random chance.
3. In the macos test, `test_toa_selection.py` keeps failing in the parallel mode (but is OK when run serially). So I am running that test serially and everything else parallelly for macos.

dlakaplan:
This is good.  I remember we had tried this a while ago but for some reason didn't implement it in CI.  But I don't see any reason not to.  Even a factor of ~2 change helps.

Is this ready to merge?

abhisrkckl:
Yes

PR #1800, opened 2024-07-02
Fix strange maxMJD default in event_optimize

paulray:
This just updates the --maxMJD default in event_optimize to be the stop time of the input event file if no other value is specified. Previously it was 57250, a seemingly random time in 2015.


paulray:
@devbhakt is this change OK?

devbhakt:
I think the change works with fits files but it breaks if you're using a pickle file for the events and don't specify a maxMJD

abhisrkckl:
Please add a CHANGELOG

PR #1801, opened 2024-07-09
Type hints for derived quantities

dlakaplan:
Using the older syntax for combining type hints and quantity input since we still support old astropy (#1709)

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -39,6 +39,7 @@ the released changes.
 - `pint.models.chromatic_model.ChromaticCM` for a Taylor series representation of the variable-index chromatic delay.
 - Whitened residuals (`white-res`) as a plotting axis in `pintk`
 - `TOAs.get_Tspan()` method
+- Type hints in `pint.derived_quantities`
 ### Fixed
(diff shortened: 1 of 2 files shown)


dlakaplan:
I had to change the docstrings to raw docstrings (`r""" ... """`) to avoid stricter checking for escape sequences in python 3.12.

dlakaplan:
This is broadly failing the oldestdeps.  I haven't tracked it down, but it might be something we table until we update the oldest astropy.

dlakaplan:
So I think the issue with the older astropy is a conflict between the quantities decorator and typehints decorator for the return types.  e.g.,

```
@u.quantity_input(x=u.m)
def v(x: u.Quantity) -> u.Quantity
    return x/(10*u.s)
```
seems to fail with the older astropy since it interprets the return type for a return unit.  Without the `@u.quantity_input` this works fine, and in newer astropy it works.  But not with the old ones.  Instead we need to do:
```
@u.quantity_input(x=u.m)
def v(x: u.Quantity) -> u.m/u.s
    return x/(10*u.s)
```
and specify the units for the return rather than the type/class.  Probably this is OK and combining the unit checking with type hints is still a net positive, but we might want to check back on this when we update oldestdeps to see if we can make it more uniform.

Issue #1809, opened 2024-07-24
model1.par = model2.par has unexpected results


dlakaplan:
From @scottransom :
```
from io import StringIO
import pint.models as pm
import pint.fitter as pf

par1 = """
RAJ      17:48:04.53480592  0          0.00005548
DECJ     -24:46:34.6586741  0           0.0279874
F0              118.538254  1  0.0000000000054826
F1            8.485977D-15  0  9.543894501911D-20
PEPOCH        53000.000000
DM              237.062679  0            0.005850
SOLARN0              10.00
EPHEM               DE421
CLK                 UNCORR                          
UNITS               TDB
"""

par2 = """
RAJ      17:48:04.53480592  0          0.00005548
DECJ     -24:46:34.6586741  0           0.0279874
F0              666.666666  1  0.0000000000054826
F1            8.485977D-15  0  9.543894501911D-20
PEPOCH        53000.000000
DM              237.062679  0            0.005850
SOLARN0              10.00
EPHEM               DE421
CLK                 UNCORR                          
UNITS               TDB
"""

m1 = pm.get_model(StringIO(par1))
m2 = pm.get_model(StringIO(par2))
print(m1)
# The following works if you use this instead:
# m1.F0.value = m2.F0.value
m1.F0 = m2.F0
# Following should(?) be different, but it isn't
print(m1)

# These show that the F0s are both consistent
print(m1.F0)
print(m2.F0)
assert(m1.F0==m2.F0)
```

dlakaplan:
When the above is run, it looks like the assignment into `m1` is not part of the correct component.  so rather than copying by reference or value it's doing a third even weirder thing

dlakaplan:
`model1.par1 = model1.par2` is explicitly allowed, and copies `.quantity`, `.uncertainty`, and `.frozen` attributes.  But it needs to have the two parameters have different names in order to work

dlakaplan:
The rationale for that exclusion is that it would otherwise override the name of `par1`.  That usage was for things like:
```
m.POSEPOCH = m.PEPOCH
```

scottransom:
Could you still get that same behavior above (i.e. copying if the names are different) by simply checking that the names are different first, and if they are the same then give a reference?  As mentioned in the PINT Slack channel, I think assigning by reference should be the default since that is what Python does by default.

dlakaplan:
[...]

PR #1810, opened 2024-07-25
Finds the correct component when assigning parameters

dlakaplan:
Addresses #1809 

When doing:
```
m1.F0 = m2.F0
```
it will correctly find `F0` in the `Spindown` component and do the assignment there.  The assignment will be via reference, so future changes to `m2.F0` will change `m1.F0`.

The tests also check that this works OK for top-level parameters.

### DIFF
--- a/src/pint/models/timing_model.py
+++ b/src/pint/models/timing_model.py
@@ -542,12 +542,31 @@ class TimingModel:
         )
 
     def __setattr__(self, name, value):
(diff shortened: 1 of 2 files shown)


abhisrkckl:

dlakaplan:
I think it's fine, but would also welcome input from @scottransom

scottransom:
Looks good to me, too!  Thanks!  I'll merge it.

PR #1817, opened 2024-08-07
Type hints in `residuals.py` & bump `black` to 24.x

abhisrkckl:


dlakaplan:
Do we need to bump `black` now?  I thought the annual updates happened in Jan, so maybe we are just very late?

abhisrkckl:
We were using `black` 23 in `tox.ini` and black 24 in the pre-commit file. It was working till now for some reason.
But `black` 24.8 dropped last week, breaking the `black` CI test for this PR. Bumping the black version was the only way to make it work. 

I think we should be bumping the `black` version anyway, and this is as good a time as any, in my opinion.

abhisrkckl:
#1709

dlakaplan:
OK, that makes sense.

dlakaplan:
I can merge this when you are set.

abhisrkckl:
This is ready to go.

PR #1818, opened 2024-08-08
Delete `download_data.sh`

abhisrkckl:


abhisrkckl:
Based on email from Emma van Wateren

dlakaplan:
Is the local `de432s.bsp` still needed?

abhisrkckl:
I will test removing that.

abhisrkckl:
I have removed `de432s.bsp`. It seems to not cause any problems.

PR #1802, opened 2024-07-11
Fourier series representation of chromatic noise (`CMWaveX`)

abhisrkckl:
See #1661.

Like `WaveX` & `DMWaveX`

#1797 needs to be merged first. I have merged that PR into this one to implement some validation steps that depend on both `CMWaveX` and `PLChromNoise`.

PR #1824, opened 2024-08-12
WIP: Brute force method to find the optimal number of harmonics for `WaveX`, `DMWaveX`, and `CMWaveX`

abhisrkckl:


abhisrkckl:
Here is an example for achromatic red noise and DM noise
image

requirements.txt
```diff
@@ -10,3 +10,4 @@ uncertainties
 loguru
 nestle>=0.2.0
```
dlakaplan, line 11: should we discuss having a new requirement more broadly?  If this is only needed for a subset of tasks, should it be optional?

src/pint/models/binary_ddk.py
```diff
@@ -49,14 +49,19 @@ class BinaryDDK(BinaryDD):
     of the system from Earth, the finite size of the system, and the
     interaction of these with the proper motion.
 
-    From Kopeikin (1995) this includes :math:`\Delta_{\pi M}` (Equation 17), the mixed annual-orbital parallax term, which changes :math:`a_1` and :math:`\omega`
-    (:meth:`~pint.models.stand_alone_psr_binaries.DDK_model.DDKmodel.delta_a1_parallax` and :meth:`~pint.models.stand_alone_psr_binaries.DDK_model.DDKmodel.delta_omega_parallax`).
+    From Kopeikin (1995) this includes :math:`\\Delta_{\\pi M}` (Equation 17),
```
dlakaplan, line 52: aren't these changes already done in #1802?

abhisrkckl: I have merged that PR into this one since `CMWaveX` is needed for this PR. That's why it is appearing here also. It's probably best to review this PR after merging #1802. That way, the diff will only show the lines changed by this PR.

requirements.txt
dlakaplan: I think this was actually adding `joblib`?

dlakaplan: In particular, I wonder about how using `joblib` compares to the use of `concurrent.futures`?  I see some discussion online.  It seems like we might want to stick with one library for that functionality.

abhisrkckl: I will try `concurrent.futures`.

Issue #1826, opened 2024-08-15
Solar Wind geometry not implemented error 


mtlam:
Having this issue on my local machine and the notebook server. I'm taking a NANOGrav par file and trying to add in SWX parameters. Each is of the form:
```
SWXDM_0001   0.001   1
SWXP_0001   2
SWXR1_0001   <float>
SWXR2_0001   <float>
```
which should allow fitting some DM over that time range. But even with one of the above blocks, I get the following error:
`NotImplementedError: Solar Wind geometry not implemented for power-law index p <= 1`
I get this regardless of what value I put into the `SWXP` parameters.

I tried to diagnose with a test installation of PINT but running into other compatibility issues that prevent me from easily going forward. The source code itself looks correct from what I can tell but still having the problem.

mtlam:
@dlakaplan did some digging and found that on loading a test model that `SWXP_0001` is shown in `m.free_params`. I tested and if I change the second line in the block above to `SWXP_0001   2   0` to explicitly include the frozen flag, then the code does work.

mtlam:
Resolved in #1827.

PR #1827, opened 2024-08-15
Made SWXP frozen by default, even for first segment

dlakaplan:
#1826

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -24,6 +24,7 @@ the released changes.
 - Documentation: Fixed empty descriptions in the timing model components table
 - BIC implementation
 - `event_optimize`: Fixed a bug that was causing the results.txt file to be written without the median values. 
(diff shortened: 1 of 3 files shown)


dlakaplan:
This may still not be perfect.  For the first segment, `SWXDM_0001` will still be unfrozen if the fit flag is not specified. i.e., 
for:
```
SWXDM_0001   0.0 
SWXP_0001   2
SWXR1_0001   53216.008946369795
SWXR2_0001   53505.07804959211
SWXDM_0002   0.0 
SWXP_0002   2
SWXR1_0002   53516.008946369795
SWXR2_0002   53605.07804959211
```
SWXDM_0001 is not frozen, while SWXDM_0002 is.  So there is a difference in how these are treated.

dlakaplan:
OK, now I think it behaves correctly.

mtlam:
Looks good to me, thanks, @dlakaplan!

abhisrkckl:
This looks good. Shall I merge it?

dlakaplan:
Yes, I think this is set.

Issue #1830, opened 2024-08-20
Inability to use local ephemeris files


dlakaplan:
`solar_sytem_ephemerides.load_kernel(path=...)` should be able to take a local path to use a local ephemeris (note that I think there is a bug in that routine, but what I say below applies even when that bug is fixed).   This is actually also available in `solar_sytem_ephemerides.objPosVel_wrt_SSB()`.  But when that function is called in the TOA creation, no `path` argument is allowed.  So I don't think that one can actually use a local file for the ephemeris.

I'm not sure what the right solution is.  One could try to specify additional arguments to where `objPosVel_wrt_SSB` is called, although that happens in many places and not all with direct calls. Or we could make the loading of the kernel respect previous calls within the same session, so that if it was called once externally (e.g., `pint.solar_system_ephemerides.load_kernel("de118",path=...)`) then that would persist.  Or something else?

I'm not sure of the right solution here, or how general a problem this is.  Presumably it could occur in HPC environments where individual nodes cannot access external networks.  @JPGlaser : do you see this?  Or do I understand the usage incorrectly?

PR #1831, opened 2024-08-21
Can now use local ephemeris file

dlakaplan:
I think there was an intent to allow this a while ago, but the mechanism appeared broken and not uniformly implemented (#1830).  This PR should solve that:

* A small fix to ``solar_system_ephemeris._load_kernel_local()`` to actually properly load the files
* A persistent dictionary so that once loaded (locally), PINT will not attempt to re-load the same file
* Tests for this
* A small addition to ``explanation.rst`` explaining how to use a local file

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -34,6 +34,9 @@ jobs:
             tox_env: 'py312-test-cov'
           - os: ubuntu-latest
             python: '3.12'
+            tox_env: 'shorttest'
+          - os: ubuntu-latest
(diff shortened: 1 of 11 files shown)


dlakaplan:
See https://nanograv-pint--1831.org.readthedocs.build/en/1831/explanation.html#ephemerides

abhisrkckl:

dlakaplan:
I think so.  Note that I was unable to get the dictionary working properly for downloaded files, but since those are handled by astropy I don't think it's needed.  So I think this is ready.

PR #1838, opened 2024-09-05
Making bad SINI raise exception to help fit

dlakaplan:
If the fitter encounters a SINI value outside of 0->1, rather than having some cryptic warning messages about bad values in the log and then fail with an error about NaN's in the array, it should be able to avoid taking steps to those places.

Overall the fitters know how to try a step and change that step if it ends in a bad place.  I think all that was missing was to raise an exception for a bad SINI value.  

Also moved `InvalidModelParameters` to `pint.models.parameter` to make it easier to import from within a model (it's a subclass of `ValueError`, so I think either will work).

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -53,9 +53,11 @@ jobs:
 #          - os: ubuntu-latest
 #            python: '3.8'
 #            tox_env: 'codestyle'
-          - os: macos-12
-            python: '3.12'
-            tox_env: 'singletest'
+# uncomment this if needed to run a single test quickly
+# can specify OS and python versions
+          # - os: macos-12
+          #   python: '3.12'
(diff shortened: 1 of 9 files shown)


dlakaplan:
I added a test.  It may look straightforward but it fails without the changes from this PR.

dlakaplan:
I'm not sure if this is a great practice, but I spend a lot of timing waiting for a whole CI process to finish just to see the results of a single test.  So I made a new CI entry (which can be commented out in `ci_test.yaml`) to run that single test quickly, where the test is specified in `tox.ini`.  This way I can get feedback in ~5min and not ~50.  It's not part of this CI so I could do it as a separate PR if needed.  In any case I would comment out the new test before merging.

dlakaplan:
Original motivation:
Screen Shot 2024-09-06 at 9 37 42 AM

PR #1840, opened 2024-09-06
DRAFT: NoiseComponent Free Specs

jeremy-baier:
This MR adds the ability to input free spectral model for red noise, dm noise, and chromatic noise.
It is essentially a copy of PLRedNoise, PLDMNoise, and PLChromNoise, replacing the PL with FreeSpec.


jeremy-baier:
Note: need to check the units on these...

PR #1843, opened 2024-09-10
ORBWAVEs: Fourier-series model for variable orbital periods

coclar:
This pull requests adds the "ORBWAVEs" model, which allows variations in the orbital phase to be expressed as a Fourier series, analogous to the "WAVE" model for timing noise. 

This is crucial for timing redbacks over long time periods, where the Taylor series FBX model can become numerically unstable.


src/pint/models/stand_alone_psr_binaries/binary_orbits.py
```diff
@@ -237,3 +237,280 @@ def d_pbprime_d_par(self, par):
...
+        self,
+        parent,
+        orbit_params=[
+            "PB",
+            "TASC",
+            "ORBWAVE_OM",
+            "ORBWAVE_EPOCH",
+            "ORBWAVEC0",
+            "ORBWAVES0",
+        ],
```
abhisrkckl, line 255: Wouldn't it be better to have PBDOT also here to capture a linear secular variation of PB? That way ORBWAVE_OM can be the same as 1/Tspan like we have in red noise.

dlakaplan:
Hi, Colin.  This looks good but a couple of quick comments:
* Having `ORBWAVE_OM` makes the parameter look like `OM` (longitude of periastron).  Is there another name you can use?  `ORBWAVE_FREQ` or something like that (and change to Hz instead of rad/s)?
* Could you add a description of how this model works in the docstring?

src/pint/models/stand_alone_psr_binaries/binary_orbits.py
coclar: Indeed, @kerrm has added `OrbitWavesFBX` in that file for that purpose. I'll add a test case for that model too.

coclar:
Thanks David, I'll add a docstring. 

I chose the name `ORBWAVE_OM` to be analogous with `WAVE_OM`, although the units are different between the two (rad/s vs. 1/d).

dlakaplan:
> Thanks David, I'll add a docstring.
> 
> I chose the name `ORBWAVE_OM` to be analogous with `WAVE_OM`, although the units are different between the two (rad/s vs. 1/d).

OK, as long as the docstring is sufficiently clear, this may be OK.

PR #1845, opened 2024-09-27
Move exceptions to new module

dlakaplan:
All of the custom warnings and exceptions will be in `pint.exceptions` now.  This means that are easier to find/import, and also easier to use (since you can see which are already defined).

(the one exception is in `noise_model.py`, since some of the exceptions there require access to other functions defined in that module)


dlakaplan:
Again, I'm not sure this is a good idea, but if we want it it's here.

src/pint/exceptions.py
```diff
@@ -0,0 +1,162 @@
...
+class ConvergenceFailure(ValueError):
+    pass
+
+
+class MaxiterReached(ConvergenceFailure):
+    pass
+
+
+class StepProblem(ConvergenceFailure):
+    pass
```
abhisrkckl, line 24: Can you please add docstrings explaining what each of these exceptions mean?

abhisrkckl:
I think this is a good idea, especially for avoiding circular imports and for improving readability.

abhisrkckl:
Please also add a CHANGELOG entry.

src/pint/exceptions.py
dlakaplan: I don't think I necessarily can, since I didn't write them.  I could probably infer some from usage but that would take a while.

Issue #1846, opened 2024-10-07
Deprecation message while installing PINT using `pip install -e`


abhisrkckl:
DEPRECATION: Legacy editable install of pint-pulsar==1.0.1+233.gc406b12d from file:///home/abhimanyu/Work/PINT (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457

abhisrkckl:
We may have to update our package setup.

dlakaplan:
But the discussion says:

> TL;DR: the execution of setup.py develop when pip does an editable install. The -e / --editable pip install option is not deprecated.
and I (at least) do the latter.  Is this really an issue?

dlakaplan:
https://pypi.org/project/ini2toml/ might be able to translate things for us

PR #1848, opened 2024-10-08
Convert to pyproject install

dlakaplan:
Generated `pyproject.toml` using https://pypi.org/project/ini2toml/.  Removed `setup.cfg`.  Kept `setup.py` as it's needed for versioning.

For me this worked OK with `pip install -e .` Not getting the warning messages (#1846).  Still should test other build methods. It would help if others can try this too.


dlakaplan:
Looks like install during the GitHub CI tests worked...

dlakaplan:
Just did a fresh install:
```
conda create -n pintinstalltest python=3.12
conda activate pintinstalltest   
pip install .  
```
and it seemed to work too

dlakaplan:
@paulray : are you able to see if this can still be packaged for pypi and conda (is there any way to do that without releasing a new version)?

paulray:
In our Makefile, when I do a `make release`, it runs the following:
```
dist: clean ## builds source and wheel package
        python setup.py sdist
        python setup.py bdist_wheel
        ls -l dist
```
@dlakaplan is that still the right thing to run?

paulray:
I get a few deprecation and other warnings when I build the source and wheel packages. See attached log. Are these ignorable, or should we fix them?
[distlog.txt](https://github.com/user-attachments/files/17440016/distlog.txt)

PR #1850, opened 2024-10-17
Phases calculation using the Designmatrix feature

devbhakt:
This adds in a calculation that using the designmatrix to calculate the phases of each photon during each step of the MCMC run in `event_optimize`. This leads to a dramatic speed up in the MCMC run. 

Here are some test runs I have done with real gamma-ray data. The samples line up well with the exception of the internal `PHASE` parameter that is written into the fitter within `event_optimize`. This parameter is an internal parameter that serves as analogous to absolute phase and does not effect the timing model.
J1231_compare_samps_triangle
J2028_compare_samps_triangle

### DIFF
--- a/src/pint/scripts/event_optimize.py
+++ b/src/pint/scripts/event_optimize.py
@@ -414,6 +414,16 @@ class emcee_fitter(Fitter):
             self.model, phs, phserr
         )
         self.n_fit_params = len(self.fitvals)
+        self.M, _, _ = self.model.designmatrix(self.toas)
+        self.M = self.M.transpose() * -self.model.F0.value
+        self.phases = self.get_event_phases()
+        self.calc_phase = False
+
+    def calc_phase_matrix(self, theta):
+        d_phs = np.zeros(len(self.toas))
+        for i in range(len(theta) - 1):
+            d_phs += self.M[i + 1] * (self.fitvals[i] - theta[i])
+        return (self.phases - d_phs) % 1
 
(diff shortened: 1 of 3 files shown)


src/pint/scripts/event_optimize.py
```diff
@@ -686,6 +699,13 @@ def main(argv=None):
         action="store_true",
         dest="noautocorr",
     )
+    parser.add_argument(
+        "--calc_phase",
+        help="Calculates the phase at each MCMC step using the designmatrix",
+        default=False,
+        action="store_true",
+        dest="calc_phase",
+    )
```
abhisrkckl, line 708: I suggest changing this option to "--linearize_model" or something like that. I don't think "--calc_phase" accurately reflects what it is doing.

src/pint/scripts/event_optimize.py
```diff
@@ -862,6 +882,9 @@ def main(argv=None):
     # This way, one walker should always be in a good position
     pos[0] = ftr.fitvals
 
+    # How phase will be calculated at each step (either with the designmatrix or )
+    ftr.calc_phase = True if args.calc_phase else False
```
abhisrkckl, line 886: This can just be `ftr.calc_phase = args.calc_phase`

abhisrkckl:
Please add a CHANGELOG entry

abhisrkckl:
Shall I merge this? The changes look OK to me.

PR #1854, opened 2024-11-04
add power law solar wind gp as a noise component

jeremy-baier:


abhisrkckl:
Does this model treat NE_SW as a Fourier Gaussian process?

abhisrkckl:
And is this the same model Sai Chaitanya presented in the IPTA meeting?

jeremy-baier:
> Does this model treat NE_SW as a Fourier Gaussian process?
It models solar wind delay using a Fourier basis which is like a DM Fourier basis but multiplied by a geometric factor which encodes information about the pulsar line of sight with respect to the sun.

> And is this the same model Sai Chaitanya presented in the IPTA meeting?

Yes, this is the power-law model which Sai presented at the 2024 IPTA meeting and is described in [this paper](https://arxiv.org/abs/2409.09838). I have been testing this model with the priors and Fourier basis described in the paper.

jeremy-baier:
There is not any physical justification for a power law specifically so I am hoping to code in a free spec here as well at some point.

src/pint/models/noise_model.py
```diff
@@ -556,6 +561,125 @@ def pl_dm_cov_matrix(self, toas):
...
+
+        See the documentation for pl_sw_basis_weight_pair function for details."""
+
+        tbl = toas.table
+        t = (tbl["tdbld"].quantity * u.day).to(u.s).value
+        freqs = self._parent.barycentric_radio_freq(toas).to(u.MHz)
+        nf = self.get_pl_vals()[2]
+        # get the acrhomatic Fourier design matrix
+        Fmat = create_fourier_design_matrix(t, nf)
+        # then get the solar wind chromatic part to multiply by
```
abhisrkckl, line 642: I see that this Fmat  only includes the sin and cos components with fundamental frequency 1/Tspan. This will not fit the data well if there is a linear or quadratic variation in NE_SW. I suggest adding derivatives of NE_SW in `SolarWindDispersion` to take care of this.

src/pint/models/noise_model.py
```diff
@@ -556,6 +561,125 @@ def pl_dm_cov_matrix(self, toas):
...
+        )
+        self.add_param(
+            floatParameter(
+                name="TNSWC",
+                units="",
+                aliases=[],
+                description="Number of SW DM noise frequencies.",
+                convert_tcb2tdb=False,
+            )
+        )
```
abhisrkckl, line 621: I presume this only supports the inverse-squared radial variation of the solar wind. This corresponds to `SWP=2` and `SWM=1`. 

If this is correct, please add a validation step somewhere (I suggest in `get_noise_basis` or `get_noise_weights`) to ensure that this condition is met. If not, this component will be inconsistent with `SolarWindDispersion`.

src/pint/models/noise_model.py
abhisrkckl: See #1859

jeremy-baier: thanks for coding this in!!
i will try to get all of the tests passing on this PR so that we can get these both merged at the next git-flowers meeting

src/pint/models/noise_model.py
jeremy-baier: hi @abhisrkckl ,
i am trying to follow up with all of my open PRs since my term just ended. 😄 
i don’t think this necessarily requires inverse-square solar wind variations, but i agree that we need some way to ensure the same SWP used in the noise run is being plugged into the timing model. i’m not entirely sure how to do this, but i will keep looking into it.

src/pint/models/noise_model.py
jeremy-baier: i have thought about this more as well. it is great to have these polynomial fits in PINT, however, there are 2 reasons that i think they will not always be required with this model. 
1) the spectra will not always be very “red”. infact, i think Sai found that they are often blue. so I think the low frequency power is less important. 
2) DM1 and DM2 will probably absorb much of this since the solar wind is very covariant w DM at low frequencies. i think Caterina shows this in one of her papers.

abhisrkckl: Ok... In any case I have implemented the NE_SW derivatives in another PR which is already merged. The user can use them if needed, but are not necessary.

src/pint/models/noise_model.py
abhisrkckl: One possibility is to check the SWM and SWP parameters and use them to determine how this mode will behave. Or if all of the SWM and SWP values are supported you can throw an error if the condition is not met. The main point is to avoid inconsistency between models.

jeremy-baier:
ok. provided that the tests pass, I think this PR is ready for review perhaps sometime early next week.

src/pint/models/timing_model.py
```diff
@@ -486,18 +486,24 @@ def num_components_of_type(type):
             num_components_of_type(PulsarBinary) <= 1
         ), "Model can have at most one PulsarBinary component."
 
-        from pint.models.solar_wind_dispersion import SolarWindDispersionBase
-
-        assert (
-            num_components_of_type(SolarWindDispersionBase) <= 1
```
jeremy-baier, line 492: common practice is to perturb a fixed, deterministic SW model with a SWGP. for this reason, i have deleted this check, but instead added a warning.

src/pint/models/noise_model.py
```diff
@@ -606,6 +610,179 @@ def pl_dm_cov_matrix(self, toas: TOAs) -> np.ndarray:
...
+        t, f = self.get_time_frequencies(toas)
+        Fmat = create_fourier_design_matrix(t, f)
+        # get solar wind geometry from pint.models.solar_wind_dispersion.SolarWindDispersion
+        solar_wind_geometry = self._parent.solar_wind_geometry(toas)
+        # dispersion constant taken from Handbook of Pulsar Astronomy, eq. 4.5
+        # the cm**-3 at the end reflect the units of the GP perturbations which are solar electron number density
+        # since this is the SW DM value if n_earth = 1 cm^-3. the GP will scale it.
+        dt_DM = (
+            solar_wind_geometry
+            * 4.148808e3
```
abhisrkckl, line 744: What is this constant?

src/pint/models/timing_model.py
abhisrkckl: Makes sense.

src/pint/models/noise_model.py
abhisrkckl: I meant use the already defined constant (`pint.DMconst`) instead of hardcoding the value.

src/pint/models/noise_model.py
```diff
@@ -606,6 +610,179 @@ def pl_dm_cov_matrix(self, toas: TOAs) -> np.ndarray:
...
+        """Return a Fourier design matrix for SW DM noise.
+
+        See the documentation for pl_sw_basis_weight_pair function for details."""
+
+        freqs = self._parent.barycentric_radio_freq(toas).to(u.MHz)
+        # get the achromatic Fourier design matrix
+        t, f = self.get_time_frequencies(toas)
+        Fmat = create_fourier_design_matrix(t, f)
+        # get solar wind geometry from pint.models.solar_wind_dispersion.SolarWindDispersion
+        solar_wind_geometry = self._parent.solar_wind_geometry(toas)
```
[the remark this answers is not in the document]
jeremy-baier: good point !
i think i have just added what you are asking.

jeremy-baier:
seems like everything is now good-to-go as far as i can see.
i would prefer this gets merged whenever you get a chance rather than wait.
thank you very much for your help with this, @abhisrkckl !!

tests/test_noise_models.py
```diff
@@ -34,6 +34,17 @@ def add_DM_noise_to_model(model):
     model.validate()
 
 
+def add_SW_noise_to_model(model):
+    all_components = Component.component_types
+    model.add_component(all_components["PLSWNoise"](), validate=False)
+    model["TNSWAMP"].quantity = -12
+    model["TNSWGAM"].quantity = -2.0  # blue spectrum
```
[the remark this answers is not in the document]
jeremy-baier: Sai found that the spectrum of solar wind GP is sometimes blue since there can be more high frequency power. 
I believe the priors from his paper on gamma are [-5, 5] or so.
We can change the test value to be a red spectrum if you would prefer though.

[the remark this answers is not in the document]
abhisrkckl: No that's fine

PR #1858, opened 2024-11-06
Provide physical formulas for magnetic field calculations.

kerrm:
The previous formulas for the magnetic field at the polar cap and light cylinder relied on a single scale.  I replaced them with physical versions involving the neutron star moment of inertia, radius, etc., and using an explicit conversion units conversion appropriate for the cgs system.  I updated the tests to agree with the new (exact) implementation.  A few other lines were munged by `black`.


abhisrkckl:
Shall I merge this?

PR #1859, opened 2024-11-11
Time derivatives of NE_SW (NE_SW1, ...)

abhisrkckl:
Complements #1854


dlakaplan:
Maybe clarify that you are putting in time derivatives here (and not partial derivatives needed for fitting)?

kerrm:
Two quick comments:
(1) perhaps add a short circuit for non time dependent cases to avoid any performance regression
(2) the test case should include a test of the correctness, e.g. that NE_SW1 = 1 produces a value of 5 a year 0 and 10 at year 5

abhisrkckl:
> Two quick comments:
> (1) perhaps add a short circuit for non time dependent cases to avoid any performance regression
> (2) the test case should include a test of the correctness, e.g. that NE_SW1 = 1 produces a value of 5 a year 0 and 10 at year 5

Added both.

abhisrkckl:

PR #1861, opened 2024-11-14
Fix WAVE_OM units

kerrm:
The WAVE model was implemented some time ago, before the dust had settled on how to handle angular units.  Consequently, the default unit for WAVE_OM was d^-1.  However, it is an angular frequency, so the correct unit is u/d.  This PR corrects the unit and some downstream code.  I also added some tests.  Admittedly the tests are still pretty sparse, but TAOSMHITD.


abhisrkckl:
Please add a changelog entry

abhisrkckl:
Shall I merge this?

PR #1862, opened 2024-11-15
Fixes for using EQUAD/EFAC

dlakaplan:
Two minor fixes:
* `add_param` returns the name of the parameter.  Using when adding `maskParameter`s and you want to track them programmatically.
* Fixed the logic for TOA selection masks.  Previously if it returned an actual selection which was `array([0])` it would treat that as not matching any TOAs.


dlakaplan:
@abhisrkckl : only the MacOS test is failing

abhisrkckl:
This looks OK to me.

PR #1863, opened 2024-11-19
Add an nbin option to photonphase

matteobachetti:
Allow photonphase to plot a pulse profile with a custom number of bins, while maintaining the current value as default

E.g. 256 bins
256 bins

versus default
100 bins


paulray:
Sure, looks good!

abhisrkckl:
Please add a CHANGELOG.

matteobachetti:
Done!

PR #1865, opened 2024-11-27
Add MacOS CI workflow using mamba

dlakaplan:
This implements the full CI test on MacOS using mamba to ensure an x86 processor.  It also removes the old MacOS tests from the testing matrix.  Note that the new tests are not done through `tox`, because `tox` makes its own virtual environment, so they are just called directly.  Hopefully this is sufficient.  Otherwise more development (likely using `tox-conda`) would be needed.  

There was additionally one test that failed repeatedly.  As far as I could tell this was just due to random chance, although I am not sure.  But I fixed it by putting in a fixed random seed.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -96,7 +96,7 @@ jobs:
 #          path: .tox/docs_out/
   test-mamba:
     runs-on: ubuntu-latest
-        steps:
+    steps:
         - uses: actions/checkout@v4
         - name: Set-Up Python Env
           uses: mamba-org/setup-micromamba@v1
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -130,5 +130,5 @@ jobs:
             python -c "import tox; print(f'tox {tox.__version__}')"
         - name: Run tests
           shell: bash -el {0}
-          run: bash -el {0} -c 'tox -e py313-test'
+          run: tox -e py313-test
           
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -116,6 +116,7 @@ jobs:
             python -m pip install --upgrade pip
             python -m pip install tox
         - name: Print Python, pip, and tox versions
+          shell: bash -el {0}
           run: |
             python -c "import sys; print(f'Python {sys.version}')"
             python -c "import pip; print(f'pip {pip.__version__}')"
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -95,7 +95,7 @@ jobs:
 #          name: documentation
 #          path: .tox/docs_out/
   test-mamba:
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
     steps:
         - uses: actions/checkout@v4
         - name: Set-Up Python Env
@@ -106,6 +106,7 @@ jobs:
             cache-environment: true
             cache-downloads: true
             create-args: >-
+              --platform osx-64
               -c conda-forge
               python=3.11
               astropy
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -130,5 +130,5 @@ jobs:
             python -c "import tox; print(f'tox {tox.__version__}')"
         - name: Run tests
           shell: bash -el {0}
-          run: tox -e py313-test
+          run: pytest -v --pyargs tests
           
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -94,3 +94,33 @@ jobs:
 #      with:
 #          name: documentation
 #          path: .tox/docs_out/
+  test-mamba:
+    runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+            - name: Set-Up Python Env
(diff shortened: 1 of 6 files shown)


dlakaplan:
@JPGlaser : I am trying to add a MacOS environment using micromamba like the one you did in https://github.com/ipta/pulsar-env/blob/patch/py-3.11/.github/workflows/test_CondaEnv.yml
To start I am using ubuntu, but will try to transition to x86 macOS.

JPGlaser:
> OK, I think I got this working under `macos-latest`. It's not integrate with the matrix of other tests, but I can probably do that if people (@abhisrkckl , @paulray , @JPGlaser ) think that this is a useful direction.

Looks good to me. It's much easier to control these kinds of tests and package deployments with the micromamba action.

~ Joe G.

abhisrkckl:
I think the macos test using mamba is too much of a special case to belong in the matrix. I think it will be more readable if it's separate.

dlakaplan:
@JPGlaser : I thought I was running the precision tests here before, but I guess I wasn't.  When I try to run the whole test suite it fails on the `require_longdouble_precision` check in `conftest.py` --- not sure why it wasn't called before.  I thought I was using the x86 environment with `--platform osx-64` but maybe not?  Can you look this over and see if there are any obvious issues?  Have you checked the precision explicitly like this elsewhere?

dlakaplan:
Yes, somehow within the `tox` environment I see:
```
024-12-02 17:46:03.808 | DEBUG    | pint.utils:check_longdouble_precision:166 - os posix.uname_result(sysname='Darwin', nodename='Mac-1733160850129.local', release='23.6.0', version='Darwin Kernel Version 23.6.0: Thu Sep 12 23:34:00 PDT 2024; root:xnu-10063.141.1.701.1~1/RELEASE_ARM64_VMAPPLE', machine='arm64')
2024-12-02 17:46:03.832 | DEBUG    | pint.utils:check_longdouble_precision:167 - processor i386
2024-12-02 17:46:03.832 | DEBUG    | pint.utils:check_longdouble_precision:168 - Python 3.13.0 (main, Oct  7 2024, 05:02:14) [Clang 16.0.0 (clang-1600.0.26.4)]
2024-12-02 17:46:03.832 | DEBUG    | pint.utils:check_longdouble_precision:169 - eps: 2.220446049250313e-16
2024-12-02 17:46:03.832 | DEBUG    | pint.utils:check_longdouble_precision:170 - longdouble eps: 2.220446049250313e-16
2024-12-02 17:46:04.021 | DEBUG    | pint.utils:check_longdouble_precision:171 - Info: # Created: 2024-12-02T17:46:03.834905
# PINT_version: 0+untagged.1.g3293987
# User: runner
# Host: Mac-1733160850129.local
# OS: macOS-14.7.1-arm64-i386-64bit-Mach-O
# Python: 3.13.0 (main, Oct  7 2024, 05:02:14) [Clang 16.0.0 (clang-1600.0.26.4)]
```
when it should be python 3.11.  So it's not actually loading within the mamba environment.

dlakaplan:
For some reason the CI has failed several times on:
https://github.com/nanograv/PINT/blob/master/tests/test_phase_offset.py#L38
in the new MacOS tests.  When I run only that test in CI it passes.  When I run it locally it (usually) passes: about 1/100 fails.  Not sure of why it is not passing here.  Will see if adding an explicit random seed makes a difference.

.github/workflows/ci_test.yml
```diff
@@ -94,3 +94,47 @@ jobs:
 #      with:
 #          name: documentation
 #          path: .tox/docs_out/
+  ci-tests-macos:
+    runs-on: macos-latest
```
abhisrkckl, line 98: Maybe rename this to macos-latest-py3.11 or something like that so that the name is similar to the other tests.

.github/workflows/ci_test.yml
```diff
@@ -94,3 +94,47 @@ jobs:
...
+          uses: mamba-org/setup-micromamba@v1
+          with: 
+            init-shell: bash
+            environment-name: pint
+            cache-environment: true
+            cache-downloads: true
+            create-args: >-
+              --platform osx-64
+              -c conda-forge
+              python=3.11
```
abhisrkckl, line 111: Is there a reason why this is not Python 3.13?

abhisrkckl:
Also please update changelog.

.github/workflows/ci_test.yml
dlakaplan: no, just trying to distinguish from the tox version.  Now it's 3.13.

.github/workflows/ci_test.yml
dlakaplan: done.

dlakaplan:
I'm not sure what the original requirement was for, but we had included `typed_ast` in `requirements_dev`.  The build failed for me with macos-latest and python 3.13.  But I also see:
https://github.com/python/typed_ast/issues/179
saying that it was primarily for python 2 and <=3.8, so I'm trying to see if we can remove it entirely, or potentially replace with `ast` if needed.

abhisrkckl:

dlakaplan:
I think this is ready, but would appreciate a look-through from @JPGlaser

JPGlaser:
Yep, everything looks good and things seem to be operating correctly on the tests. I'm good with it getting merged!

PR #1866, opened 2024-12-04
Rerun intermittent failures in CI

abhisrkckl:
This was initially part of #1798 which was merged and then reverted.

### DIFF
--- a/.github/workflows/ci_test.yml
+++ b/.github/workflows/ci_test.yml
@@ -136,5 +136,5 @@ jobs:
             python -m pip install --force-reinstall --no-deps .
         - name: Run tests
           shell: bash -el {0}
-          run: pytest -v --pyargs tests
+          run: pytest -v --pyargs tests --reruns 5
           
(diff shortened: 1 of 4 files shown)


abhisrkckl:
The macos failure is because the runner is deprecated.

dlakaplan:
I was also dealing with a situation like this, and in that case I just put in a fixed random seed so I know that it should pass if things work.  but I don't know if that's a good idea in general?

abhisrkckl:
I think this is a more general solution. This also lets us know that we have a problem if the test fails even after 5 reruns.

dlakaplan:
Since the new MacOS tests are no longer part of `tox.ini`, can you add the appropriate arguments to the git CI file?

abhisrkckl:
> Since the new MacOS tests are no longer part of tox.ini, can you add the appropriate arguments to the git CI file?

Added.

dlakaplan:
Looks good - should I merge?

PR #1867, opened 2024-12-04
Glitch tweak

kerrm:
I noticed a bunch of nearly-duplicate code in the Glitch derivatives, so I pushed that into the "deriv_prep" function.  Otherwise, I switched to f-strings, cleaned up the math and syntax in a few places, and added some missing tests for sanity checks on models.

(E.g., the code attempts to test for a missing GLEP, but there was no test coverage for that.  The code only checked for whether or not a GLEP was present, but it always is!  It's value is set to None, though, so the code now checks for that and the test covers it.

### DIFF
--- a/src/pint/models/glitch.py
+++ b/src/pint/models/glitch.py
@@ -218,7 +218,6 @@ class Glitch(PhaseComponent):
             else:
                 decayterm = u.Quantity(0)
 
-            log.debug(f"Glitch phase for glitch {idx}: {dphs} {dphs.unit}")
             phs[affected] += (
                 dphs
                 + dt[affected]
--- a/src/pint/models/glitch.py
+++ b/src/pint/models/glitch.py
@@ -177,10 +177,7 @@ class Glitch(PhaseComponent):
             glf0d = getattr(self, glf0dnm)
             idx = glf0d.index
             if glf0d.value != 0.0 and getattr(self, "GLTD_%d" % idx).value == 0.0:
-                msg = (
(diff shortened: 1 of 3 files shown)


abhisrkckl:
Please add a CHANGELOG entry.

src/pint/models/glitch.py
```diff
@@ -145,43 +145,41 @@ def setup(self):
...
-                if not hasattr(self, param + "%d" % idx):
+                if not hasattr(self, param + str(idx)):
                     param0 = getattr(self, f"{param}1")
                     self.add_param(param0.new_param(idx))
-                    getattr(self, param + "%d" % idx).value = 0.0
+                    getattr(self, param + str(idx)).value = 0.0
                 self.register_deriv_funcs(
-                    getattr(self, f"d_phase_d_{param[:-1]}"), param + "%d" % idx
+                    getattr(self, f"d_phase_d_{param[:-1]}"), param + str(idx)
                 )
```
abhisrkckl, line 154: It will be better to use fstring instead of the + operator here since that is the style we had decided to prefer.

kerrm:
> Please add a CHANGELOG entry.

I'm never sure where to draw the line but... is it worth noting relatively minor code cleanup in the CHANGELOG?  It's not something, imo, that users of new versions really need to be aware of.

abhisrkckl:
I think it's good for posterity, in case someone wonders in the future when/how/why some change was made.

Issue #1776, opened 2024-05-30
Tests for Gaussianity and whiteness


abhisrkckl:
We have the whitened residuals available via `Residuals.calc_whitened_resids()`.

It will be useful to test the Gaussianity and whiteness of the whitened residuals to make sure that the timing model takes care of everything.

Example: https://arxiv.org/abs/2010.06109
They use the Anderson-Darling test for Gaussianity and the Ljung-Box test for whiteness.

PR #1868, opened 2024-12-05
Simulate correlated DM noise in wideband TOAs

### LINKED ISSUE #1660: `PLDMNoise` doesn't work correctly with wideband TOAs

abhisrkckl:
Partially fixes #1660. Fitting a model to wideband TOAs in the presence of correlated DM noise is still not implemented.

### DIFF
--- a/src/pint/models/noise_model.py
+++ b/src/pint/models/noise_model.py
@@ -13,6 +13,9 @@ from pint.models.timing_model import Component
 
 
 class NoiseComponent(Component):
+
+    introduces_dm_errors = False
+
     def __init__(
         self,
     ):
@@ -232,6 +235,7 @@ class ScaleDmError(NoiseComponent):
     category = "scale_dm_error"
 
     introduces_correlated_errors = False
+    introduces_dm_errors = True
 
     def __init__(
         self,
@@ -469,6 +473,7 @@ class PLDMNoise(NoiseComponent):
     category = "pl_DM_noise"
 
     introduces_correlated_errors = True
+    introduces_dm_errors = True
     is_time_correlated = True
 
     def __init__(
--- a/tests/test_fake_toas.py
+++ b/tests/test_fake_toas.py
@@ -12,6 +12,7 @@ import pint.config
 from pint.fitter import GLSFitter, DownhillGLSFitter
 from pinttestdata import datadir
 import pytest
+from statsmodels.stats.diagnostic import acorr_ljungbox
 
 
 def roundtrip(toas, model):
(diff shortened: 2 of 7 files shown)


tests/test_fake_toas.py
```diff
@@ -511,3 +511,57 @@ def test_simulate_uniform_multifreq(multifreq):
...
+        model=m,
+        add_noise=True,
+        wideband=True,
+        add_correlated_noise=True,
+    )
+
+    assert t.is_wideband()
+
+    # There is correlated noise in DMs
+    assert (
```
dlakaplan, line 563: So this is some sort of a chi^2 test?  Is there no better way to check for correlated noise?

src/pint/models/noise_model.py
```diff
@@ -232,6 +235,7 @@ class ScaleDmError(NoiseComponent):
     category = "scale_dm_error"
 
     introduces_correlated_errors = False
+    introduces_dm_errors = True
```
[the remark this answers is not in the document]
abhisrkckl: Currently there is no way to distinguish between achromatic noise models and noise models that represent DM noise (such as PLDMNoise) that affects the DM part of the wideband likelihood. I added this attribute to distinguish such noise components. The timing model components that introduce a DM correction can already be distinguished by checking if they are derived from the `Dispersion` class.

tests/test_fake_toas.py
abhisrkckl: Yes. A better test will be something like [Ljung-Box test](https://en.wikipedia.org/wiki/Ljung%E2%80%93Box_test). Unfortunately, it is not available in scipy. We will need to add statsmodels as a dependency to use that test.

dlakaplan: It looks like that test would not be very hard to implement, maybe include in PINT utils.  Although it may not be needed just for this, it could be more broadly useful.

abhisrkckl: OK... I have added the Ljung-Box test. For this I added statsmodels as a dependency for tests. It is not a dependency for the package itself since it is not used anywhere in the main code. #1776 may change that though.

dlakaplan:
Looks good.  I wonder if this test should be used elsewhere for correlated noise checks.  How useful is it to test for the presence of correlated noise in the data?  Should we have a utility function for that?  Not needed for this PR - just asking.

Should I merge?

abhisrkckl:
Yes please merge this.

abhisrkckl:
I think it is useful to check for  correlations to make sure that the noise
model is correctly describing the data. Is the noise model is correct the
whitened residuals will be Gaussian and white.

PR #1871, opened 2024-12-10
Fixed some docstrings: removed newlines, other reformatting

### LINKED ISSUE #1870: Timing Models doc page not rendering some of the component docstrings fully
For this [page](https://nanograv-pint.readthedocs.io/en/latest/timingmodels.html). See for example:

> `PLChromNoise` - Model of a radio frequency-dependent noise with a power-law spectrum and
> `PLDMNoise` - Model of DM variations as radio frequency-dependent noise with a

It looks like perhaps it is printing the first line of the docstring only, see for example for `BinaryBTPiecewise` as well.

dlakaplan:
Closes #1870


abhisrkckl:
This looks simple enough. Shall I merge this?

dlakaplan:
Yes, sure.  Could probably do a better pass through the docs at some point but this is still a positive change.

Issue #1873, opened 2024-12-12
Add a `meta` attribute to the TimingModel


JPGlaser:
Hey All,

Currently, `pint_pal` is pinned at version 1.0.1 because we have a few places where we store metadata about the `TimingModel` object as direct attributes. This is no longer valid in version `1.1`. As discussed on the 12/12 telecon, a quick fix would be to make a new `meta` attribute which is a dictionary. This would allow users to add metadata to the `TimingModel` without directly affecting the model itself in possibly weird ways.

See Tests found here:
https://github.com/nanograv/pint_pal/actions/runs/12300670688/job/34329402278?pr=94

~ Joe G.

paulray:
Yeah the CHANGELOG for 1.0.2 has this, which must be the cause:

- Doing model.par = something will try to assign to par.quantity or par.value but will give warning

abhisrkckl:
Added in #1874

PR #1874, opened 2024-12-12
Added metadata to model, and included it in printout

dlakaplan:
```
m=get_model("tests/datafile/NGC6440E.par")
print(m.as_parfile())
```
now returns:
```
# Created: 2024-12-12T13:41:07.865053
# PINT_version: 0.7+6065.g92675b80
# User: David Kaplan (kaplan)
# Host: plock
# Python: 3.9.7 (default, Sep 16 2021, 08:50:36) 
# [Clang 10.0.0 ]
# Format: pint
# read_time: 2024-12-12T13:41:04.663636
# allow_tcb: False
# convert_tcb: False
# allow_T2: False
# original_name: tests/datafile/NGC6440E.par
PSR                            1748-2021E
...
```
So there is now a metadata dictionary, which is initially populated with arguments that were used when reading in the par file along with the name of the file and the time it was read.  Other parameters can be added, and will be included in the info string on printout.

Addresses #1873

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -22,6 +22,7 @@ the released changes.
 - `add_param` returns the name of the parameter (useful for numbered parameters)
 - Rerun intermittent failures in CI
 - micromamba CI environment for testing macOS-latest, without tox
+- models now have metadata dictionary
 ### Fixed
 - Changed WAVE_OM units from 1/d to rad/d.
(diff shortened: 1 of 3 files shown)


JPGlaser:
This looks great and should work perfectly for our cases.

~ Joe G.

dlakaplan:
OK, great.  Maybe @abhisrkckl or someone else can review & merge.

PR #1877, opened 2025-01-08
More correct low-rank expansion of powerlaw noise

vhaasteren:
Has a dependency on #1896 

The existing low-rank expansions for red noise and related processes are based on the Lentati et al. (2014) expansion which uses a diagonal prior for frequency modes at 1/T. This is only approximately correct when fitting for quadratic spindown. For DM variations, which use the same model, contemporary analyses now consequently also include a DM2 term to "make the analysis work". DM2 is not necessarily physical when doing a fully unconstrained fit, so it is more correct to let this term be modeled by the stochastic process.

A low-rank expansion with frequencies below 1/T can improve the expansion while keeping the prior matrix diagonal. This is explored in van Haasteren & Vallisneri (2014): https://doi.org/10.1093/mnras/stu2157

This PR implements that model. Note that newer and more accurate expansions are in development that have a nondiagonal prior matrix. The design idea is that the models in this PR will also be implemented as part of the Classes in this PR: they represent the same model, namely a correct powerlaw. The Lentati expansion is a physically different model with spectral leakage explicitly not present, so that is different.

We have coordinated with the tempo2 devs, and the additional parameters are:

- TNDMFLOG, TNDMFLOG_FACTOR
- TNCHROMFLOG, TNCHROMFLOG_FACTOR
- TNREDFLOG, TNREDFLOG_FACTOR

The FLOG parameters dictate how many log-frequency-spaced extra bins we need below 1/T. The _FACTOR parameters dictate by what fraction the frequency needs to shrink every frequency bin. So setting both to 2 will lead to two new frequency bins: 1/(2T) and 1/(4T)

### DIFF
--- a/tests/test_noise_models.py
+++ b/tests/test_noise_models.py
@@ -26,7 +26,7 @@ correlated_noise_component_labels = [
 def add_DM_noise_to_model(model):
     all_components = Component.component_types
     model.add_component(all_components["PLDMNoise"](), validate=False)
-    model["TNDMAMP"].quantity = 1e-13
+    model["TNDMAMP"].quantity = -13
     model["TNDMGAM"].quantity = 1.2
     model["TNDMC"].value = 30
     model["TNDMFLOG"].value = 4
@@ -37,7 +37,7 @@ def add_DM_noise_to_model(model):
 def add_chrom_noise_to_model(model):
     all_components = Component.component_types
     model.add_component(all_components["PLChromNoise"](), validate=False)
-    model["TNCHROMAMP"].quantity = 1e-14
+    model["TNCHROMAMP"].quantity = -14
     model["TNCHROMGAM"].quantity = 1.2
     model["TNCHROMC"].value = 30
     model["TNCHROMFLOG"].value = 4
--- a/tests/test_noise_models.py
+++ b/tests/test_noise_models.py
@@ -29,6 +29,8 @@ def add_DM_noise_to_model(model):
     model["TNDMAMP"].quantity = 1e-13
     model["TNDMGAM"].quantity = 1.2
     model["TNDMC"].value = 30
+    model["TNDMFLOG"].value = 4
+    model["TNDMFLOG_FACTOR"].value = 2
     model.validate()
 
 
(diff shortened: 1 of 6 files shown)


vhaasteren:
I will merge in changes from #1762 , as those changes overlap with this one. I will get rid of the low-frequency cut-off, and instead use the corner frequency as in #1762 by @abhisrkckl

abhisrkckl:
Hi... can you please merge #1896 again? I just figured out that my fix was not general enough. I have updated the PR to address this.

tests/test_noise_models.py
```diff
@@ -29,6 +29,8 @@ def add_DM_noise_to_model(model):
     model["TNDMAMP"].quantity = 1e-13
```
abhisrkckl, line 29: TNDMAMP is in log10. So isn't this number too small?

tests/test_noise_models.py
```diff
@@ -38,6 +40,8 @@ def add_chrom_noise_to_model(model):
     model["TNCHROMAMP"].quantity = 1e-14
```
abhisrkckl, line 40: Isn't this too small?

vhaasteren: Yes, you are right. So, I check git blame...   and it seems you committed that line @abhisrkckl , haha :p

vhaasteren: Given that 0 is much larger than -14, this value is actually super large

abhisrkckl: 😭

vhaasteren:
@abhisrkckl , in `test_pldmnoise.py` and `test_plchromnoise.py` those parameters are added and used already. Even the PLRedNoise parameters are added there as a check. As far as I can judge, that should be sufficient, no?

vhaasteren:
Hold off on this until we verify with Mike

vhaasteren:
Mike and I agreed to do this the same in PINT and Tempo2. The tempo2 PR is here: https://bitbucket.org/psrsoft/tempo2/pull-requests/93

So they are consistent. I think it's good to go like this @abhisrkckl

PR #1881, opened 2025-01-22
Sourcery and type hints

abhisrkckl:
This is mostly sourcery refactors and type hints. The only non-trivial part is that I removed the definition of `cached_property` in `pint.fitter` since it is no longer needed (it was added to support Python<=3.7).


dlakaplan:
What did you use to sort the imports?  Should we document that somewhere for future consistency?

abhisrkckl:
Just the latest version of isort. The output of that is inconsistent with
black, so I had to blacken the code before committing.

PR #1883, opened 2025-02-06
Store correlated noise amplitudes instead of the residuals

abhisrkckl:
Residuals can be computed from the amplitudes easily, but vice versa is hard.


scottransom (approved):
Looks good!  Thanks!

PR #1889, opened 2025-02-20
Remove broken solar system ephemeris mirror URL

abhisrkckl:


dlakaplan:
This looks fine.  I had hoped to actually understand this issue but even so we can probably just do this.

PR #1891, opened 2025-02-20
Remove `CompositeMCMCFitter`

abhisrkckl:
This fitter was supposed to analyze joint radio and high-energy datasets. However, this is currently broken and cannot be instantiated. It also does not have any test coverage and is not documented. Based on a slack discussion it seems that nobody is using this currently. So I am removing this.


abhisrkckl:
Just pasting the slack discussion for posterity.

image
image

Issue #1893, opened 2025-03-07
Odd results from PINTk due to no mean subtraction


dlakaplan:
In PINTk, the postfit chi^2 info that is printed comes from residuals that are recalculated based on the recent fit:

```
if self.fitted:
            self.postfit_resids = Residuals(
                self.all_toas,
                self.postfit_model,
                subtract_mean=False,
                track_mode=track_mode,
            )
            if (
                self.selected_toas.ntoas > 0
                and self.selected_toas.ntoas != self.all_toas.ntoas
            ):
                self.selected_postfit_resids = Residuals(
                    self.selected_toas,
                    self.postfit_model,
                    subtract_mean=False,
                    track_mode=track_mode,
                )
            else:
                self.selected_postfit_resids = self.postfit_resids
```
In these, `subtract_mean=False`.  I am not sure of the reason for that.  And I can't say I noticed it before but it's giving me weird results now.  If I look at the actual fit (`self.fitter.resids.calc_chi2()`) I get 233 for a recent source I was working on. But the printout from `self.selected_postfit_resids` is very different, 1400491.4986230780929.  I confirmed that this is all due to the mean subtraction.  Maybe the initial model was weird here and so it ended up with a huge mean phase.  Not sure why this hasn't been seen otherwise.

Is there a reason to keep the mean phase unsubtracted here?  I think that in the actual fit it is subtracted.  But in the residual output display it definitely isn't:

<img width="638" alt="Image" src="https://github.com/user-attachments/assets/b9a4b7c2-f6a8-4440-8715-93170595ebeb" />

dlakaplan:
Note that the chi^2 values affect things like the F-test results.

scottransom:
I think I'm responsible for putting the `subtract_mean=False` in there.  It was causing big problems for phase connection when you don't know the correct phase wraps of TOAs.  What does the mean residual mean when some of your TOAs aren't on the correct wrap and so the "residual" is basically just a random number?  Maybe the right thing to do sould be to add a botton to enable/disable this?  I agree that when you have phase connection and want to look at stats, what is there right now is definitely problematic.  But it is also hugely problematic the other way when you are phase connecting!

dlakaplan:
Yes, a button is probably a good idea.  I can look at that

PR #1894, opened 2025-03-07
Added checkbox for optional subtraction of mean in pintk

dlakaplan:
#1893 
Previous behavior (no subtraction of mean):
Screen Shot 2025-03-07 at 1 50 17 PM
with that, post-fit chi^2=1400491.5.

New behavior:
Screen Shot 2025-03-07 at 1 50 35 PM
with that, post-fit chi^2=233.55795.

Note that the actual fit results are the same.


abhisrkckl:
What is the default state of this checkbox? I think having it checked by default will be the expected behavior for most users.

dlakaplan:
Currently it starts unchecked, since that was the previous behavior.  Easy to change, though.

abhisrkckl:
Shall I merge this?

Issue #1900, opened 2025-04-14
`astropy.observatory.sites` is missing NenuFAR (as well as additional LOFAR stations)


blarsen10:
I'm passing this issue along from the IPTA DR3 combination -- the main issue is that NenuFAR site info appears to be missing from `astropy.coordinates.sites` ([link here](https://github.com/astropy/astropy-data/blob/gh-pages/coordinates/sites.json)). The site info should be accessible on the [Tempo2 bitbucket here](https://bitbucket.org/psrsoft/tempo2/src/master/T2runtime/observatory/observatories.dat). I believe T2 and astropy use different coordinate conventions for the sites and that they should be converted from x, y, and z to longitude, latitude, and height.

Also, Joris pointed out to me that while "LOFAR" is already included in astropy, it doesn't have any of the international LOFAR station's positions (which may differ by several thousands of km offset). This seems like it could be important in case the individual stations ever need to be used by themselves, though I don't know if this is required for the DR3 combination. The extra sites can also be found on the T2 bitbucket page (Joris can share more about which sites are "nice to have" vs "totally unnecessary" to add to astropy).

dlakaplan:
Hi, @blarsen10 .  `astropy.coordinates.sites` is just a backup for PINT itself.  If a site is used extensively for pulsar timing we can include the coordinates directly in https://github.com/nanograv/PINT/blob/master/src/pint/data/runtime/observatories.json or https://nanograv-pint.readthedocs.io/en/latest/observatory_list.html.  For instance, that has a number of individual LOFAR stations:
```   "fullname": "The Dutch Low-Frequency Array (LOFAR)",
        "fullname": "The Low-Frequency Array (LOFAR) low-band antenna (LBA) at Effelsberg Observatory",
        "fullname": "The Low-Frequency Array (LOFAR) low-band antenna (LBA) at Effelsberg Observatory?",
        "fullname": "The Low-Frequency Array (LOFAR) high-band antenna (HBA) at Effelsberg Observatory",
        "fullname": "The Low-Frequency Array (LOFAR) high-band antenna (HBA) at Effelsberg Observatory",
        "fullname": "The Low-Frequency Array (LOFAR) low-band antenna (LBA) at Unterweilenbach, Germany",
        "fullname": "The Low-Frequency Array (LOFAR) low-band antenna (LBA) at Unterweilenbach, Germany",
[...]

PR #1901, opened 2025-04-18
Added NENUFAR

dlakaplan:
Based on tempo2 bitbucket position


dlakaplan:
Addresses part of #1900 
See
https://nanograv-pint--1901.org.readthedocs.build/en/1901/observatory_list.html

dlakaplan:
There is a distinction between the core, the Dutch stations, and the international stations (https://science.astron.nl/telescopes/lofar/lofar-system-overview/technical-specification/lofar-array-configuration/).  I don't know how they are generally used for pulsar observations.  Probably a question for a LOFAR user/expert.

PR #1904, opened 2025-05-11
Remove some duplicated code in photon phasoegrams.

kerrm:
The previous implementations had a lot of shared code.  I've now moved most of the code into a single method, and the other (binned) signature calls the fundamental unbinned version.  I added some docstrings too.  The new code should be functionally and API equivalent to the old code.

### DIFF
--- a/src/pint/plot_utils.py
+++ b/src/pint/plot_utils.py
@@ -91,11 +91,13 @@ def phaseogram(
 
     if not binned:
         if weights is None:
-            for ph in [phss, phss+1.0]:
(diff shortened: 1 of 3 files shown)


kerrm:
I don't really use this code.  However, I've added a sanity check test.  NB that it's plotting code, so while I viewed the files and verified they looked OK, the test will just save them to a named temporary file.

devbhakt:
Hi Abhimanyu and Matthew, I took a look at the code and I think it's good to merge.

PR #1910, opened 2025-06-02
Fix wideband DM noise simulations + Utility functions for constructing wideband residuals, scaled uncertainties, and design matrix

abhisrkckl:
#1887 became too unwieldy. So I am splitting it into multiple PRs.

This has two components.
1. The wideband DM noise simulations were incorrect: they were injecting separate realizations of the DMGP signal into the wideband TOAs and DMs. This PR does the right thing.
2. Utility functions for implementing "correct" wideband fitting like design matrix etc.


src/pint/models/timing_model.py
```diff
@@ -1742,12 +1759,58 @@ def scaled_dm_uncertainty(self, toas: TOAs) -> u.Quantity:
             result += nf(toas)
         return result
 
+    def scaled_wideband_uncertainty(self, toas: TOAs) -> np.ndarray:
```
dlakaplan, line 1762: Is there a method to split this into the appropriate pieces and return them as Quantities?  I assume so, but maybe that should be in the docstring?

src/pint/residuals.py
```diff
@@ -1287,3 +1288,14 @@ def dof(self) -> int:
     def reduced_chi2(self) -> float:
         """Return the weighted reduced chi-squared."""
         return self.chi2 / self.dof
+
+    def calc_wideband_resids(self) -> np.ndarray:
```
dlakaplan, line 1292: Again maybe have the docstring here point to the individual functions to make them easier to find.

abhisrkckl: added.

abhisrkckl:
> Mostly this looks fine, but I didn't go through the matrix math in any detail. Are there tests that will show that those calculations are correct?

Do you mean the calculations for the wideband simulation?

tests/test_sini.py
```diff
@@ -63,7 +63,7 @@ def test_sini_fit():
     """
     m = get_model(io.StringIO(s))
     t = pint.simulation.make_fake_toas_uniform(
-        55731, 59051, 100, m, obs="axis", error=1 * u.us, add_noise=True
+        55731, 59051, 1000, m, obs="ao", error=1 * u.us, add_noise=True
```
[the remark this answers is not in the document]
dlakaplan: OK.  Agree that telescope name doesn't matter.  I'll assume this doesn't have much impact on run time?

abhisrkckl:
> OK, great. And will a test now check that the simulated noise residuals work properly?

Added test.

tests/test_timing_model.py
```diff
@@ -42,9 +42,6 @@ def timfile_nojumps():
     return get_TOAs(os.path.join(datadir, "NGC6440E.tim"))
 
 
-len_timfile_nojumps = len(get_TOAs(os.path.join(datadir, "NGC6440E.tim")))
```
abhisrkckl, line 45: This was not used for anything.

abhisrkckl (commented):
For fixing the CI failure

abhisrkckl:
No I just removed one unused line from that file which was creating problem.

dlakaplan:
> No I just removed one unused line from that file which was creating problem.

Oh, OK.  Should I merge this then?

PR #1913, opened 2025-06-06
Make tests run in parallel

abhisrkckl:
The important changes are in `ci_tests.yml` and `tox.ini`. The rest are cleanup and formatting.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -15,6 +15,7 @@ the released changes.
 - TCB <-> TDB conversion for power-law GP noise parameters.
 - Type hints in `pint.fitter`
 - PLSWNoise: a Fourier basis stochastic solar wind model. See Hazboun et al. 2022 for details.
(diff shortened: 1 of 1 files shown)


abhisrkckl:
This was tried earlier in #1798, but had to be reverted because the macOS test kept failing. Since then, the macOS test was rewritten to work on M1 chips. So this should work, hopefully. I am not seeing any of the cryptic errors I was seeing with #1798 here.

dlakaplan:
That all looks pretty simple.  Any rough estimate of the speedup?

abhisrkckl:
All times are in minutes. This includes the overhead of installing packages. Overall, I see ~2x improvement. 

|Test                             | Before | After |
|-----------------------------|-----------|---------|
|macos-latest-py313 | 88        | 46     | 
|oldestdeps                | 28        | 14     |
|py313-test-cov          | 58       | 30     |

dlakaplan:
that's great.  I can merge this when you are ready, or deal with #1910 first.

PR #1925, opened 2025-07-09
Chromatic exponential dip

abhisrkckl:


abhisrkckl:
image

abhisrkckl:
image

abhisrkckl:
Par file example:
```
RAJ            05:00:00                     1
DECJ           15:00:00                     1
POSEPOCH       55000
F0             100                          1
F1             -1e-15                       1
PEPOCH         55000
**EXPDIPEP_1     54764.272428904194001        1      
EXPDIPAMP_1    1.6641670367524487e-06       1
EXPDIPTAU_1    112.00425959054773           1
EXPDIPIDX_1    -1.9148109887274356          1
EXPDIPEPS      0.01** 
TZRMJD         55000
TZRSITE        pks
TZRFRQ         1400
EPHEM          DE440
CLOCK          TT(BIPM2019)
UNITS          TDB
```

abhisrkckl:
@jeremy-baier Please take a look.

src/pint/mcmc_fitter.py
```diff
@@ -19,7 +19,6 @@
     "MCMCFitter",
     "MCMCFitterAnalyticTemplate",
     "MCMCFitterBinnedTemplate",
-    "CompositeMCMCFitter",
```
abhisrkckl, line 22: This was removed in a previous PR, but I missed this line.

jeremy-baier:
could epsilon be hardcoded to some small number to avoid user error? or perhaps a warning is appropriate if epislon is above some value.

abhisrkckl:
> to clarify, if the user sets epsilon to a very small value, this function converges to the heavy-side implementation in Tempo2 and enterprise extensions? i have been playing around with this and it seems like the parameters match up very well.

> could epsilon be hardcoded to some small number to avoid user error? or perhaps a warning is appropriate if epislon is above some value.

As epsilon approaches 0, this approaches the tempo2 version. In practice, this means something much less than the decay timescale. I have checked that 0.01 day works well, and I have included this as the default value. If the user doesn't provide EXPDIPEPS in the par file, this default will be used.

jeremy-baier:
yea, this looks great.
why specifically is it important to be differentiable in time rather than just differentiable w.r.t. the timing model parameters?

abhisrkckl:
I meant differentiable in T. Since it appears as (t-T), differentiable in time == differentiable in T.

dlakaplan:
What does the tempo2 par file look like?  Are the shared parameters the same?

abhisrkckl:
I have added the documentation and set up the parameter aliases such that tempo2 par files are now readable.

src/pint/models/expdip.py
```diff
@@ -0,0 +1,286 @@
...
+import numpy as np
+import astropy.units as u
+from astropy.time import Time
+from pint.models.parameter import floatParameter, prefixParameter
+from pint.models.timing_model import DelayComponent
+from pint.toa import TOAs
+
+
+class SimpleExponentialDip(DelayComponent):
+    """Simple chromatic exponential dip model for events like the profile changes
```
dlakaplan, line 13: Can you actually sketch out the form here?  Not necessarily all of the math, although you could include it for RTD rendering.  But the basic form?  Otherwise things like the index don't make sense.

src/pint/models/expdip.py
```diff
@@ -0,0 +1,286 @@
...
+            if prefix_par.startswith("EXPDIPEP_"):
+                self.register_deriv_funcs(self.d_delay_d_T, prefix_par)
+            elif prefix_par.startswith("EXPDIPAMP_"):
+                self.register_deriv_funcs(self.d_delay_d_A, prefix_par)
+            elif prefix_par.startswith("EXPDIPTAU_"):
+                self.register_deriv_funcs(self.d_delay_d_tau, prefix_par)
+            elif prefix_par.startswith("EXPDIPIDX_"):
+                self.register_deriv_funcs(self.d_delay_d_gamma, prefix_par)
+
+    def get_ffac(self, toas: TOAs) -> np.ndarray:
```
dlakaplan, line 192: Is there ever a time when you might want to change this?  Maybe for LOFAR data or something where 1400 MHz is far outside the band?

src/pint/models/expdip.py
```diff
@@ -0,0 +1,286 @@
...
+
+class SimpleExponentialDip(DelayComponent):
+    """Simple chromatic exponential dip model for events like the profile changes
+    in J1713+0747.
+
+    The dip is modeled as a logistic function multiplied by an exponential decay.
+    This is different from the tempo2 implementation where the exponential decay
+    is multiplied by a Heaviside step function. We cannot fit for the event epoch
+    while using the latter because it is not differentiable at the event epoch.
+    This implementation approaches the tempo2 version in the limit EXPDIPEPS -> 0.
```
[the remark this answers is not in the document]
abhisrkckl: added.

PR #1930, opened 2025-07-18
Calculate integer phases when using polycos in photonphase

matteobachetti:
When using polycos in photonphase, the code crashed because the iphss variable was not defined at line 310. This should fix everything without adding any computational overhead (eval_phase was internally just calculating eval_abs_phase and taking the .frac attribute)


abhisrkckl:
This looks good. Can you add a changelog entry?

PR #1933, opened 2025-08-11
Remove broken script `event_optimize_multiple`

abhisrkckl:


abhisrkckl:
We had discussed this in an earlier email. It was implemented by someone
for some specific purpose, but has been unused since, and lated changes
broke it because it had no test coverage.


Abhimanyu Susobhanan

PR #1935, opened 2025-08-11
Make whitened residuals available pre-fit + Whitened residuals for wideband

abhisrkckl:

PR #1938, opened 2025-08-13
KS test to check if the whitened residuals are unit-normal distributed

abhisrkckl:
Depends on #1935

I am using Kolmogorov-Smirnov test instead of Anderson-Darling test since the former can test if the distribution is *unit*-normal, whereas the latter only tests if the distribution is normal.

I am not implementing the Ljung-Box test here because the `acorr_ljungbox` function in statsmodels assumes that the data is evenly sampled.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -9,37 +9,7 @@ the released changes.
 
 ## Unreleased
 ### Changed
-- Simulation functions no longer subtract the residual mean by default.
-- `cmwavex_setup` now takes the chromatic index as an optional argument.
 ### Added
-- `WidebandTOAResiduals.calc_wideband_resids()` and `TimingModel.scaled_wideband_uncertainty()` methods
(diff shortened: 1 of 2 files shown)


abhisrkckl:
Oh I didn't know that. We can put that in later.

dlakaplan:
> > Depends on #1935
> > I am using Kolmogorov-Smirnov test instead of Anderson-Darling test since the former can test if the distribution is _unit_-normal, whereas the latter only tests if the distribution is normal.
> 
> I haven't looked into this in great detail, but my impression was that the A-D test can look for unit normal as well. e.g., Case 3 at [https://en.wikipedia.org/wiki/Anderson–Darling_test#Test_for_normality](https://en.wikipedia.org/wiki/Anderson%E2%80%93Darling_test#Test_for_normality). but maybe this is not done by the `scipy` function. So probably fine to leave for now and could be added in later?

Actually that should be Case 0 I think.

abhisrkckl:
Yes

PR #1941, opened 2025-08-13
Anderson-Darling test

dlakaplan:
This implements the Anderson-Darling test for data compared to a normal distribution with fixed mean/variance (which https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.anderson.html does not do).  

Based on https://en.wikipedia.org/wiki/Anderson–Darling_test#Test_for_normality, with the p-values computed from https://www.jstatsoft.org/article/view/v009i02


abhisrkckl:

dlakaplan:
I'm trying to improve the docstrings a bit.

dlakaplan:
OK, I think the updated docstrings help (I also added a little more to those for the K-S test to give the output order).  If you think they are OK then it's fine to merge.

PR #1968, opened 2026-02-25
Correctly incorporate correlated noise in `WidebandDownhillFitter` + Plot whitened DM residuals in pintk

abhisrkckl:
This PR implements the correct handling of DMGP in `WidebandDownhillFitter` based on https://arxiv.org/abs/2505.05274

The major changes are as follows.
1. `WidebandDownhillFitter` now uses the wideband design matrix computed by `TimingModel.wideband_designmatrix` instead of constructing it inside the fitter.
2. `WidebandDownhillFitter` now uses the wideband Ndiag computed by `TimingModel.scaled_wideband_uncertainty` instead of constructing it inside the fitter.
3. `WidebandDownhillFitter` now computes the wideband residuals using `WidebandTOAResiduals.calc_wideband_resids` instead of doing it inside the fitter.
4. GLS fitting now incorporates DMGP correctly. Earlier, only DMX was supported.
5. `WidebandDownhillFitter` and `WidebandState` now have a structure more like other downhill fitter and state classes.

The GLS fitting expression is
image
where y, N, and M are the wideband residuals (2ntoa), wideband Ndiag (2ntoa x 2ntoa), and the wideband design matrix (2ntoa x nparam). Phi is the same as in narrowband timing (it contains the prior variances).

Also implements a new option to plot whitened DM residuals in pintk.


dlakaplan:
Hi, @abhisrkckl .  I don't know the correlated-noise code well enough to really review this.  Maybe @rossjjennings or @demorest or somebody else can look more closely?

dlakaplan:
It's hard to parse all of this since it's many changes to different pieces.  I see that a number of `@cached_properties` have been removed in favor of calling other routines.  Are those calculation results still cached elsewhere?

abhisrkckl:
The `@cached_property`s were saving some of the intermediate objects used for computing the chi2. These objects are not computed in the same way or are not computed at all in the new method. So there is no point in having these properties. Note that these cached properties are not present in the other `Fitter`s. This PR makes the `WidebandDownhillFitter` code more similar to the other `Fitter`s. 

I don't think any functionality of performance is lost in removing the cached properties.

scottransom:
I just looked this over.  I didn't check the details of the math, but it seems like that is mostly in the tests. So I'm happy with this given that it seems to also do quite a bit of clean-up.

PR #1975, opened 2026-03-27
Removing `astropy.constants.G` from various calculations

dlakaplan:
To improve consistency, removing `astropy.constants.G` from various derived quantities in favor of `pint.GMsun` or `pint.Tsun`.  The differences are small, ~1e-10.  And issues are still possible (e.g., if a mass is input in kg instead of Msun, it will still need to be converted).  But this is better (I think).


dlakaplan:
Note that there is still extensive use of `c.G` in the DDGR model.  That should probably also be fixed.

dlakaplan:
> Note that there is still extensive use of `c.G` in the DDGR model. That should probably also be fixed.

Now this has been fixed too.

dlakaplan:
As a related note, our value of G*Msun:

```
 Following are from here:
# http://ssd.jpl.nasa.gov/?constants (grabbed on 30 Dec 2013)
GMsun = 1.32712440018e20 * u.m**3 / u.s**2
```

has a newer value:
```
1.32712440041279419e20
```
 from the website cited above.  Fractional difference is ~1e-10.  But we generally don't claim absolute precision at this level (for things like masses) so I'm not sure it's worth changing.

src/pint/models/stand_alone_psr_binaries/DDGR_model.py
```diff
@@ -534,13 +551,13 @@ def d_DR_d_MTOT(self):
         return (
             -self.DR / self.MTOT
             - self.DR * self.d_arr_d_MTOT() / self.arr
-            + 6 * (c.G / c.c**2) / self.arr
+            + 6 * (pint.GMsun / c.c**2) / self.arr / (1 * u.Msun)
```
abhisrkckl, line 554: This doesn't seem necessary. I'd guess GMsun/Msun is no better measured than G.

src/pint/derived_quantities.py
```diff
@@ -351,8 +350,7 @@ def mass_funct(pb: u.Quantity, x: u.Quantity) -> u.Msun:
 
     .. [1] Lorimer & Kramer, 2008, "The Handbook of Pulsar Astronomy", Eqn. 8.34 (RHS)
     """
-    fm = 4.0 * np.pi**2 * x**3 / (const.G * pb**2)
-    return fm.to(u.solMass)
+    return (4 * np.pi**2 * x**3 / pint.GMsun / pb**2).decompose() * u.Msun
```
abhisrkckl, line 353: Same as the above comment. I don't think using GMsun offers an advantage here.

src/pint/models/stand_alone_psr_binaries/DDGR_model.py
dlakaplan: I agree, but I did this mostly for consistency.  If you think this whole PR is pointless we can just ditch it.  Otherwise, this seemed marginally better than the alternative.

src/pint/derived_quantities.py
dlakaplan: I think this one makes more sense.  Using GMsun in SI units allows the conversion of the mass function to Solar units.  Then casting it as Msun means that further calculations that stay in solar units will not lose accuracy.

abhisrkckl:
I will merge this if ready.

dlakaplan:
I think it's ready if you don't have further issues (I agree some of it is not optimal, but not sure what else to do)

Issue #1974, opened 2026-03-27
Is the `orbital` module used?


dlakaplan:
I don't think I see imports of `orbital` or `orbital.kepler`.  Are those deprecated?

PR #1976, opened 2026-03-27
Removed orbital module

dlakaplan:
#1974 : I don't know if this module is used.  I suspect not.  So delete to remove confusion.


dlakaplan:
So the tests that fail are in `tests/test_kepler.py`.  but again I don't see that used elsewhere, so not clear if those tests should also be removed.

dlakaplan:
@scottransom or @paulray : is this actually needed?

scottransom:
I don't think it is needed.  I just looked at the history and I think it was intended as being the core of a pure new orbit model that Anne Archibald was writing. But that obviously never happened. It is probably fine to get rid of it, although it seems like nice code!

dlakaplan:
> I don't think it is needed. I just looked at the history and I think it was intended as being the core of a pure new orbit model that Anne Archibald was writing. But that obviously never happened. It is probably fine to get rid of it, although it seems like nice code!

I have no particular animus, but it can be a little confusing.  Could just add some "DEPRECATED" notes to the files instead.

abhisrkckl:
Yes, I think this is OK to remove. The code will be available in the git history if needed in the future.

dlakaplan:
OK, in that case I think you can merge this.

PR #1977, opened 2026-03-28
Use include_bipm argument when loading Fermi TOAs with weights

coclar:
Fixes a bug where the `include_bipm` argument was hard-coded to `False` when loading Fermi TOAs with weights.


dlakaplan:
This looks fine to merge, but could you add an entry to `CHANGELOG-unreleased`?

coclar:
Thanks @dlakaplan, I've added the note to the changelog.

dlakaplan:
Actually, should `fermiphase` also deal with this argument?

https://github.com/nanograv/PINT/blob/master/src/pint/scripts/fermiphase.py#L122:
```
# For Fermi, we are not including GPS or TT(BIPM) corrections
    ts = get_Fermi_TOAs(
        args.eventfile,
        maxmjd=maxmjd,
        minmjd=minmjd,
        weightcolumn=args.weightcol,
        targetcoord=tc,
        planets=args.planets,
        ephem=args.ephem,
    )
```
so maybe add it to the command-line options and then pass to the function?  Or is that not needed?

coclar:
I agree, and added the relevant option to fermiphase.py. 

But I realised that the better way to do this is to let `get_Fermi_TOAs` respect the options for EPHEM, CLOCK and PLANET_SHAPIRO from a .par file, as is normally done when loading TOAs. I've added this option to get_Fermi_TOAs now, and added the necessary command-line arguments to fermiphase.

dlakaplan:
Looks good.  Can merge if you are done.

paulray:
Yeah, this looks OK to me. Thanks!

coclar:
Yes, ready to merge from my side!

PR #1986, opened 2026-05-05
Fix pintk bug: random models with jumps fails

dlakaplan:
When I was trying to use random models in PINTk with a model that had jumps, it failed in an obscure way involving a table merge.  Somehow the TOA tables got an additional "f" column added as part of the conversion from the JUMP model which just reflected the "-f" flag.  But then the "f" columns had a type mismatch between the real + random TOAs, and that broke things.

I still don't really understand why the column gets added.  But this seems to deal with it in a way that hopefully doesn't break anything else.  If somebody understands the origin of this column let me know.  It's also possible to just restrict this fix to only the PINTk code, so it would have minimal other interactions.


dlakaplan:
@scottransom @abhisrkckl @paulray : I think this one is minor otherwise

PR #1987, opened 2026-05-14
Fix random model bug related to how flag columns are created

dlakaplan:
When flag columns are created to track JUMP selection, if none of the TOAs match that flag then they are all `None`, which means the column type is `object`, not `str`.  This led to errors in `pintk` when using random models since the table merge failed.  Now the default is `""` so the column is still `str`.  This fixes the `pintk` error.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -11,4 +11,5 @@ the released changes.
 ### Changed
 ### Added
 ### Fixed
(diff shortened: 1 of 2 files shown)


abhisrkckl:
This looks good. I'll merge this if it's OK.

dlakaplan:
> This looks good. I'll merge this if it's OK.

Yes, thanks.

Issue #1983, opened 2026-04-29
Add toggle to cache tropospheric delay calculations


jeremy-baier:
profiling has shown that this calculation in particular can be a bottle neck when fitting MSP residuals and partial or total caching of these calculations could make a significant impact on timing model fits with many toas.

dlakaplan:
This may relate to https://github.com/nanograv/PINT/pull/1892 (which may be stalled)

dlakaplan:
Do you have more specifics about what the bottlenecks are here?  I can imagine some things regarding AltAz calculations, but having some specific info would help.

dlakaplan:
Completed via #1988

PR #1988, opened 2026-05-18
Speed up tropospheric delay calculation

dlakaplan:
#1983 

To start, I moved the TOA altitude calculation to the `TOAs` object (and added it to the table). That way it will be calculated once and no other caching is needed.

I verified this on the test NGC6440E dataset.  A single call to:
`model.troposphere_delay(t)` 
went from ~7 ms to ~1.5ms (using `timeit`) after doing the initial calculation.  So this is already considerably faster.  The initial call will take longer.

I also added type annotations to the module.


dlakaplan:
But when I run on a more complex multi-observatory data-set, it now takes twice as long.  So clearly more work is needed.

dlakaplan:
And now I store the delay as a property (can't use decorators since things like numpy arrays are not hashable).  This reduces the time to 2us (vs. 70ms for a 2000-TOA file).  But this is fragile since it depends on a combination of the model and TOA set, so a change to the TOAs would break it.  Not sure if the TOA hashing will fix this.

dlakaplan:
One possible fix is to move all of this to the TOA table - this really doesn't depend on the model

dlakaplan:
OK, so the troposphere tests perturbed TOAs by 0.5 d to get something invalid.  I think this wouldn't happen in real life: any realistic perturbation would lead to negligible changes in the altitude/tropo delay, so it's OK to treat the delays as fixed.  But to make sure that this can be dealt with I added `model.recompute_troposphere_delay(toas)` which will force a recomputation.  Still not sure that this is ideal from a data model perspective, but it is a lot faster.  @jeremy-baier: don't know if it's possible to re-profile this part of the code to see how much a difference it makes in production (i.e., many fit loops etc).

dlakaplan:
I tried this with the same J0437 data-set that Daniel Oliver used.  Using the standard PINT:
```
Fit without troposphere: 11.969788789749146
Fit with troposphere: 10.224065065383911
```
but with this PR:
```
Fit without troposphere: 11.885865926742554
Fit with troposphere: 6.761759996414185
```
To be honest, I don't understand why the fit without troposphere corrections is slow.  I understand the change with the troposphere correction, but not why the initial versions are slow.

dlakaplan:
> I tried this with the same J0437 data-set that Daniel Oliver used. Using the standard PINT:
> 
> ```
> Fit without troposphere: 11.969788789749146
> Fit with troposphere: 10.224065065383911
> ```
> 
> but with this PR:
> 
> ```
> Fit without troposphere: 11.885865926742554
> Fit with troposphere: 6.761759996414185
> ```
> 
> To be honest, I don't understand why the fit without troposphere corrections is slow. I understand the change with the troposphere correction, but not why the initial versions are slow.

This is despite virtually identical chi^2 for the two fits, so I don't think it's a big change in how the fit converges.

dlakaplan:
OK, maybe it is to do with fit convergence.  If I start with the output from a previous fit then the fit without troposphere takes only 6s.

dlakaplan:
And now a much fairer comparison.  I used the same model (0437 but without DMX/RN) to simulate TOAs with and without TROPOSPHERE.  I then fit using the appropriate model.  So the convergence would be similar in both cases.  After doing 5 trials I find:

Current PINT:
```
No troposphere: 11.113752603530884
With troposphere: 22.852966833114625
```
This PR:
```
No troposphere: 12.177623414993286
With troposphere: 12.160136222839355
```
So there is no difference in the "No troposphere" results, but a big difference on the results with troposphere.  Based on this I think this PR does accomplish the goal. Whether it respects the data model or not is still up for debate.

dlakaplan:
@abhisrkckl or @scottransom : can you look at this?

jeremy-baier:
> And now a much fairer comparison. I used the same model (0437 but without DMX/RN) to simulate TOAs with and without TROPOSPHERE. I then fit using the appropriate model. So the convergence would be similar in both cases. After doing 5 trials I find:
> 
> Current PINT:
> 
> ```
> No troposphere: 11.113752603530884
> With troposphere: 22.852966833114625
> ```
> 
> This PR:
> 
> ```
> No troposphere: 12.177623414993286
> With troposphere: 12.160136222839355
> ```
> 
> So there is no difference in the "No troposphere" results, but a big difference on the results with troposphere. Based on this I think this PR does accomplish the goal. Whether it respects the data model or not is still up for debate.

Thanks for working on this !
To clarify, when you say "Whether it respects the data model or not is still up for debate.”, you are basically saying that you are not sure about the regimes of validity for the caching of this delay? Am I interpreting that sentence correctly?

dlakaplan:
> Thanks for working on this ! To clarify, when you say "Whether it respects the data model or not is still up for debate.”, you are basically saying that you are not sure about the regimes of validity for the caching of this delay? Am I interpreting that sentence correctly?

Yes.  In general we keep the TOAs and the model separate.  But the best way to keep the calculations tied to the actual TOAs was to store them in the TOAs, not in the model (as Claude recommended).   I think the risk of some corrupted calculation is low, but non-zero (if you change the TOAs without clearing the cache...).  I may be able to further put in some cache key tied to the TOAs but doing it in a way that will allow for slicing without recomputing may be harder.

dlakaplan:
I added control via `$PINT_DISABLE_CACHE` and `$PINT_DISABLE_TROPOSPHERE_CACHE`

dlakaplan:
And checking with `cProfile`:
with caching:
```
  ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   59.717   59.717 time_troposim.py:1(<module>)
       50    0.001    0.000    1.632    0.033 troposphere_delay.py:147(troposphere_delay)
        3    0.000    0.000    0.039    0.013 troposphere_delay.py:226(delay_model)
        3    0.000    0.000    0.034    0.011 troposphere_delay.py:279(mapping_function)
       18    0.010    0.001    0.011    0.001 troposphere_delay.py:264(_coefficient_func)
        9    0.003    0.000    0.011    0.001 troposphere_delay.py:95(_herring_map)
       18    0.000    0.000    0.006    0.000 troposphere_delay.py:353(_interp)
        3    0.005    0.002    0.006    0.002 troposphere_delay.py:371(_get_year_fraction_fast)
        3    0.000    0.000    0.004    0.001 troposphere_delay.py:296(<listcomp>)
        3    0.000    0.000    0.004    0.001 troposphere_delay.py:302(<listcomp>)
        3    0.000    0.000    0.004    0.001 troposphere_delay.py:308(<listcomp>)
        3    0.000    0.000    0.002    0.001 troposphere_delay.py:329(wet_map)
        3    0.000    0.000    0.001    0.000 troposphere_delay.py:250(zenith_delay)
        6    0.000    0.000    0.001    0.000 troposphere_delay.py:268(_find_latitude_index)
        3    0.000    0.000    0.000    0.000 troposphere_delay.py:242(pressure_from_altitude)
        3    0.000    0.000    0.000    0.000 troposphere_delay.py:188(_validate_altitudes)
        1    0.000    0.000    0.000    0.000 troposphere_delay.py:1(<module>)
        4    0.000    0.000    0.000    0.000 troposphere_delay.py:110(__init__)
        1    0.000    0.000    0.000    0.000 troposphere_delay.py:25(TroposphereDelay)
        3    0.000    0.000    0.000    0.000 troposphere_delay.py:257(wet_zenith_delay)
```
and without:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000  137.026  137.026 time_troposim.py:1(<module>)
       63    0.038    0.001   72.386    1.149 troposphere_delay.py:147(troposphere_delay)
       55    0.001    0.000    1.517    0.028 troposphere_delay.py:226(delay_model)
       55    0.001    0.000    1.419    0.026 troposphere_delay.py:279(mapping_function)
      330    0.455    0.001    0.479    0.001 troposphere_delay.py:264(_coefficient_func)
      165    0.143    0.001    0.423    0.003 troposphere_delay.py:95(_herring_map)
      330    0.001    0.000    0.255    0.001 troposphere_delay.py:353(_interp)
       55    0.235    0.004    0.246    0.004 troposphere_delay.py:371(_get_year_fraction_fast)
       55    0.000    0.000    0.160    0.003 troposphere_delay.py:296(<listcomp>)
       55    0.000    0.000    0.159    0.003 troposphere_delay.py:302(<listcomp>)
       55    0.000    0.000    0.159    0.003 troposphere_delay.py:308(<listcomp>)
       55    0.000    0.000    0.043    0.001 troposphere_delay.py:329(wet_map)
       55    0.001    0.000    0.021    0.000 troposphere_delay.py:250(zenith_delay)
      110    0.001    0.000    0.012    0.000 troposphere_delay.py:268(_find_latitude_index)
       55    0.001    0.000    0.009    0.000 troposphere_delay.py:242(pressure_from_altitude)
       55    0.000    0.000    0.004    0.000 troposphere_delay.py:188(_validate_altitudes)
        1    0.000    0.000    0.000    0.000 troposphere_delay.py:1(<module>)
        4    0.000    0.000    0.000    0.000 troposphere_delay.py:110(__init__)
        1    0.000    0.000    0.000    0.000 troposphere_delay.py:25(TroposphereDelay)
       55    0.000    0.000    0.000    0.000 troposphere_delay.py:257(wet_zenith_delay)
```
so the time for the delay calculation went from 72s to 1.6s.  Note that the data-sets in this case were not identical, so we wouldn't expect identical numbers of calls, but I still think this demonstrates the utility.

scottransom:
I'm all for these changes. There has never been perfect separation between TOAs and model in PINT, simply because we can't do it!  For instance, we get specific clock corrections (i.e. BIPM year) mixed into the TOA table, and we also get solar system ephemeris mixed in (i.e. in planet locations if using planetary Shapiro delay). I'm fully in favor of having these calculations treated in a similar way and stored in the TOA table. I'm happy to merge once @dlakaplan tweaks the CHANGELOG merge conflict that we just got since I just merged another PR. Thanks for doing this!

abhisrkckl:
This looks fine to me. I will merge this if ready.

dlakaplan:
> This looks fine to me. I will merge this if ready.

I think it's largely OK.  But I have had second thoughts about how much/if the calculations should have either metadata in the TOAs table to allow for checking what the source position was and/or an explicit cache key to invalidate the data automatically.  I don't actually think this would come up in normal uses, but I still worry.  I added the RA,Dec of the source to the TOAs metadata: it still isn't used to invalidate anything but it can now be tracked.

scottransom:
I don't really think we should worry about hand-holding our users too much with CORRECT_TROPOSPHERE. It is not enabled by default and is not a part of fitting, and it really only makes a tiny effect on any resulting fit. That is *not* the case, in general, for caching of other things. So I'm personally fine without that extra meta-data. But if it really bugs you @dlakaplan, and you have a good idea how how to make it better without too much complexity, then feel free to go for it! Otherwise, I'm happy if we merge it.

dlakaplan:
I just added a minimal amount of meta-data without any explicit checks.  I think this is OK (no extra overhead, but you can check your work).

scottransom:
That seems good to me. I'm happy merge if you are satisfied, @dlakaplan

dlakaplan:
Sure - I just merged in the most recent updates, so you can merge as soon as the tests pass

PR #1990, opened 2026-05-19
Make `find_prefix_bytime()` work for SWX

abhisrkckl:
This function was not working earlier for SWX, since the SWX parameter is called "SWXDM_" and not "SWX_". This breaks the pattern of "DMX_" and "CMX_". I have added a special case to handle SWX here. 

The actual issue was in the `get_prefix_timeranges()` function.

This is required for including SolarWindX in Vela.jl.


dlakaplan:
Looks good to me.  I can merge if ready.

abhisrkckl:
Please merge this.

Issue #1989, opened 2026-05-18
Require finite TZRFRQ for tempo?


dlakaplan:
Apparently `TZRFRQ inf` breaks tempo polycos.  Not sure if it works for tempo2.  @demorest says this should be `TZRFRQ 0` for tempo.  Don't know if this should be changed only for that format or more generally.

demorest:
To be clear, tempo does allow infinite TZRFRQ, but the way this needs to be represented in the par file is with `TZRFRQ 0`.

abhisrkckl:
I think infinite TZRFRQ should be allowed. Some InPTA par files have `TZRFRQ inf`.

abhisrkckl:
Fixed in #1991

PR #1991, opened 2026-05-19
Fix tempo format TZRFRQ==inf

dlakaplan:
#1989 

When `tempo` format parfiles are written, if `TZRFRQ==inf`, will write `TZRFRQ=0` instead.


dlakaplan:
@demorest : does this solve your issue?

demorest:
That looks like it should, not sure if/when I'll get a chance to test.

dlakaplan:
> That looks like it should, not sure if/when I'll get a chance to test.

Do you think it would do any harm?  There's no rush either way.

abhisrkckl:
This looks OK. Shall I merge this?

dlakaplan:
> This looks OK. Shall I merge this?

Yes, please do.  I don't think this will do any harm.

Issue #1984, opened 2026-04-29
Add ability to cache coordinate transformations within fit iterations


jeremy-baier:
Profiling indicates that astrometric coordinate transformations are being called multiple times within the same iteration of a fit while the astrometry parameters are not changing. The result should be the same for all of those calls but they can sometimes be costly. Caching the result of the transformation until the astrometry parameters change could produce significant speed ups.

dlakaplan:
This may relate to https://github.com/nanograv/PINT/pull/1892 (which may be stalled)

dlakaplan:
Like in #1983, do you have any more specifics on what part of this to look at?  specific lines or functions?

dlakaplan:
I think this can be closed because of #1992?

PR #1992, opened 2026-05-19
Cache SSB astrometry calculations

dlakaplan:
Claude-recommended speedup: #1984

On J0437 data, tried 5 iterations.  chi^2 was within 1e-3 with both versions.  Average time went from 19.3s to 10.9s.

Have `model.clear_ssb_cache()` to clear the cache, and also can control with either:
`$PINT_DISABLE_CACHE=1` or `$PINT_DISABLE_ASTROMETRY_CACHE=1`


jeremy-baier:
amazing ! 
is the `clear_cache` method inherited by the timing model class ? or does the `clear_cache` method just exist as a method of the astrometry delay in particular ?

dlakaplan:
> amazing ! is the `clear_cache` method inherited by the timing model class ? or does the `clear_cache` method just exist as a method of the astrometry delay in particular ?

All component level methods are inherited by the timing model.  But that brings up a good point which is that there could be other caches elsewhere that we want to deal with. So I changed the name to `model.clear_ssb_cache()` to make it explicit.

dlakaplan:
So this mostly works.  But oddly not 100%.  I added env variables to aid testing. And I find that when I test with/without the cache I usually get the results I expect (without cache is slow, and chi^2 changes very slightly).  But not always. I don't yet understand why.

dlakaplan:
So I am now more confused.  Using the master branch or this PR, with or without caching I see bimodal results: sometimes slower, sometimes faster, but when faster I always see a slightly different chi^2 (at the 1e-3 level).  I could understand variation in wall-clock time just due to CPU load.  But not variation in chi^2.  That should be deterministic?

dlakaplan:
OK, I think I understand more.  For some reason, different fits use different numbers of iterations/calls.  That leads to pretty different clock times as well as slightly different chi^2.  Still not sure why that is, but now I understand the origin of the variation that I see.  

Nonetheless, I do see improved performance with this PR in terms of time/call for the ssb methods.  With caching enabled:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       43    0.000    0.000    0.697    0.016 astrometry.py:131(ssb_to_psb_xyz_ICRS)
      151    0.001    0.000    0.025    0.000 astrometry.py:1114(ssb_to_psb_xyz_ECL)
```
or
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       74    0.001    0.000    1.145    0.015 astrometry.py:131(ssb_to_psb_xyz_ICRS)
      272    0.002    0.000    0.037    0.000 astrometry.py:1114(ssb_to_psb_xyz_ECL)
```

and without:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       43    0.000    0.000    1.649    0.038 astrometry.py:131(ssb_to_psb_xyz_ICRS)
      151    0.003    0.000    0.422    0.003 astrometry.py:1114(ssb_to_psb_xyz_ECL)
```
and
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       74    0.001    0.000    2.902    0.039 astrometry.py:131(ssb_to_psb_xyz_ICRS)
      272    0.005    0.000    0.795    0.003 astrometry.py:1114(ssb_to_psb_xyz_ECL)
```

abhisrkckl:
This looks OK. Shall I merge this?

dlakaplan:
> This looks OK. Shall I merge this?

I think this is OK, and it's something we can turn off, but it should be checked closely.

dlakaplan:
Look at:
https://nanograv-pint--1992.org.readthedocs.build/en/1992/explanation.html#caching
my goal is to have a section there for every explicitly cached calculation with instructions on how to clear/turn off.

abhisrkckl:
Is this ready to go?

dlakaplan:
> Is this ready to go?

I would like somebody (you, @scottransom , ...) to check it a little, just to make sure I didn't do anything foolish.  Maybe you have already looked.

docs/explanation.rst
```diff
@@ -826,6 +826,25 @@ well as one might hope in practice and must be considered experimental.
...
+1. Pulsar-to-SSB delay calculations through :meth:`~pint.models.astrometry.Astrometry.ssb_to_psb_xyz_ECL` and 
+:meth:`~pint.models.astrometry.Astrometry.ssb_to_psb_xyz_ICRS`.  The results are cached for a given time
+(usually a set of TOAs), and they are stored separately for ECL and ICRS coordinates.  The cache can be explicitly cleared by 
+using :meth:`~pint.models.astrometry.Astrometry.clear_ssb_cache`. This caching can be turned off by setting 
+``$PINT_DISABLE_ASTROMETRY_CACHE`` to 1.
+
+2. Troposphere delays (if ``CORRECT_TROPOSPHERE`` is ``True``) through :meth:`~pint.models.troposphere_delay.TroposphereDelay.troposphere_delay`, 
+where the results are stored in the TOA table (columns ``alt`` and ``tropo_delay``).  This calculation can be explicitly reset by using 
+:meth:`~pint.models.troposphere_delay.TroposphereDelay.recompute_troposphere_delay`. This caching can be turned off by setting 
+``$PINT_DISABLE_TROPOSPHERE_CACHE`` to 1.
```
abhisrkckl, line 844: $PINT_DISABLE_TROPOSPHERE_CACHE --> PINT_DISABLE_TROPOSPHERE_CACHE

src/pint/models/astrometry.py
```diff
@@ -38,6 +39,17 @@
...
 
+def _epoch_fingerprint(epoch: Optional[time_like]) -> Tuple[Tuple, bytes]:
+    """Return a view of the epoch that is suitable for use as a cache key"""
+    if isinstance(epoch, Time):
+        return (epoch.mjd.shape, epoch.mjd.tobytes())
+    elif isinstance(epoch, u.Quantity):
+        return (epoch.value.shape, epoch.value.tobytes())
+    elif isinstance(epoch, np.ndarray):
+        return (epoch.shape, epoch.tobytes())
+    return None
```
abhisrkckl: Rather than returning None I think it will be more useful to raise a helpful error or warning here.

src/pint/models/astrometry.py
```diff
@@ -117,7 +147,21 @@ def ssb_to_psb_xyz_ICRS(self, epoch: Optional[time_like] = None) -> u.Quantity:
 
         # this is somewhat slow, since it repeatedly created different SkyCoord Objects
         # but for consistency only change the method in the subclasses below
-        return self.coords_as_ICRS(epoch=epoch).cartesian.xyz.transpose()
+        key = (_epoch_fingerprint(epoch), self._astrometric_params_tuple())
```
abhisrkckl: This key computation looks a bit expensive. I think this should be done only if needed. i.e.., only if `self._ssb_cache_icrs is not None and self.use_cache`.

src/pint/models/astrometry.py
```diff
@@ -138,8 +182,20 @@ def ssb_to_psb_xyz_ECL(
...
             (len(epoch), 3) array of unit vectors
         """
-        # TODO: would it be better for this to return a 6-vector (pos, vel)?
-        return self.coords_as_ECL(epoch=epoch, ecl=ecl).cartesian.xyz.transpose()
+        key = (_epoch_fingerprint(epoch), self._astrometric_params_tuple())
+        if (
+            key == self._ssb_cache_key_ecl
+            and self._ssb_cache_ecl is not None
+            and self.use_cache
+        ):
```
abhisrkckl: Same as above, the `key` should be computed only if needed.

src/pint/models/astrometry.py
```diff
@@ -545,12 +621,26 @@ def ssb_to_psb_xyz_ICRS(self, epoch: Optional[time_like] = None) -> u.Quantity:
...
         # return self.coords_as_ICRS(epoch=epoch).cartesian.xyz.transpose()
 
+        key = (_epoch_fingerprint(epoch), self._astrometric_params_tuple())
+        if (
+            key == self._ssb_cache_key_icrs
+            and self._ssb_cache_icrs is not None
+            and self.use_cache
+        ):
+            log.debug(f"Using cached ssb_to_psr data")
+            return self._ssb_cache_icrs.copy()
```
abhisrkckl: Same as above, the `key` should be computed only if needed.

src/pint/models/astrometry.py
```diff
@@ -888,6 +987,22 @@ def __init__(self):
...
+            float(self.PMELONG.value),
+            float(self.PMELAT.value),
+            float(self.PX.value),
+            (
+                float(self.POSEPOCH.value)
+                if isinstance(self.POSEPOCH, u.Quantity)
+                else None
+            ),
+            str(self.ECL.value),
+        )
```
abhisrkckl, line 1033: Also include the VLBIAX, VLBIAY, and VLBIAZ parameters here. Note that they may be None.

src/pint/models/astrometry.py
```diff
@@ -1034,9 +1149,23 @@ def ssb_to_psb_xyz_ECL(
...
+        ``$PINT_DISABLE_CACHE=1`` or ``$PINT_DISABLE_ASTROMETRY_CACHE=1``
         """
         # TODO: would it be better for this to return a 6-vector (pos, vel)?
 
+        key = (_epoch_fingerprint(epoch), self._astrometric_params_tuple())
+        if (
+            key == self._ssb_cache_key_ecl
+            and self._ssb_cache_ecl is not None
+            and self.use_cache
+        ):
```
abhisrkckl: Same as above, the `key` should be computed only if needed.

abhisrkckl: Also, this pattern occurs many times. I think it can be refactored into a function, maybe in `Astrometry`.

abhisrkckl:
Also, 
1. Please make sure that methods like `adjust_TOAs`, `apply_clock_corrections`, `compute_TDBs`, `compute_posvels` etc invalidate the cache.
2. What happens to the astrometry cache when two `TOAs` objects are merged?

docs/explanation.rst
dlakaplan: Do we have a standard for including ENVARS in docstrings?

src/pint/models/astrometry.py
dlakaplan: OK, good point

src/pint/models/astrometry.py
dlakaplan: I don't think it's actually expensive, but easy enough to change.

dlakaplan: Note that it needs to be computed even if `self._ssb_cache_icrs` is None so that it can be used to store the cache info later.  So only compute if `self.use_cache` I think.

src/pint/models/astrometry.py
dlakaplan: OK

src/pint/models/astrometry.py
dlakaplan: done - good idea

dlakaplan:
* I changed the name of the property to `use_ssb_cache` in case we add others later
* I made computation of the key a method, and it will only happen if `use_ssb_cache` is True
* I added more tests

> 1. Please make sure that methods like `adjust_TOAs`, `apply_clock_corrections`, `compute_TDBs`, `compute_posvels` etc invalidate the cache.

I don't know that making them invalidate the cache directly is possible, since those methods only apply to TOAs, not models.  But for the methods that change TOAs the cache will invalidate and the values will be recomputed.  I added some tests for those (not sure it should apply for `compute_posvels`: only for those that change `tdbld`)

> 2. What happens to the astrometry cache when two `TOAs` objects are merged?

Works as expected.  The cache is invalidated since the times change, and is recomputed.

docs/explanation.rst
abhisrkckl: I don't think so.

dlakaplan: But you think I should remove the `$`?  I left it in so it was more clear that it was an ENVAR.

abhisrkckl: I don't have a preference either way, only that it should be consistent. Below, PINT_DISABLE_CACHE is written without $.

dlakaplan: OK, thanks.  will fix.

abhisrkckl:
Shall I merge this once the tests pass?

dlakaplan:
> Shall I merge this once the tests pass?

Your concerns over the various operations are all OK?  If so then OK to merge I think.

PR #1995, opened 2026-05-22
Use truncnorm for bounded Gaussian prior

DotanGazith:
Using `rv_continuous` results in an accumulation of the distribution at the `upper_bound` when sampling.
If not used elsewhere, can remove `pint.models.priors.GaussianRV_gen` and import of `rv_continuous`.


abhisrkckl:
Please add an entry in `CHANGELOG-unreleased.md` and a test.

PR #1996, opened 2026-05-22
Fixed how some gridutils functions treat the logger

dlakaplan:
Some of the functions for doing gridded fits did not respect the global logger settings.  This is better now, although I am still seeing some weird behavior from the clock module.


dlakaplan:
Also added a few annoying messages to "onlyonce" list

abhisrkckl:
This looks fine. I will merge this if ready.

dlakaplan:
I think it is now (or will be after the checks finish).  Still don't fully understand all of the subtleties here, but it is a lot better than before.

PR #1999, opened 2026-05-25
Change diff/unc calculation for pintk

dlakaplan:
Previously `Diff/Unc` used the pre-fit uncertainties.  Now it uses the post-fit

#1963


dlakaplan:
@paulray or @scottransom : if you want to look at this quickly, it addresses one of Paul's notes.

PR #2007, opened 2026-06-10
Fixed binary convert for ELL1H

dlakaplan:
Explicitly run `model.setup()` for `convert_binary()`, so that some things like the ELL1H fit parameters are set.

Also use `NHARMS=7` by default when using H4.  And when `NHARMS=3`, do not use H4.


abhisrkckl:
Is this ready to go?

dlakaplan:
> Is this ready to go?

Yes, please.

PR #2008, opened 2026-06-11
Changed pint.ls to astropy.u.lsec

dlakaplan:
Astropy added a light-second unit years ago.  I think rather than defining our own (which used astropy's definition of c) we should use their version throughout. 

This might break some downstream software, so if needed we could put back the definition of `pint.ls = u.lsec`, although we do use the `pint_units` as keys for string parsing which includes `ls`.


abhisrkckl:
This looks good. I will merge this if ready.

dlakaplan:
> This looks good. I will merge this if ready.

Yes, once the latest merge finishes.

PR #2011, opened 2026-06-30
Support hierarchical triple systems with dual binary components (`BINARY` + `BINARY2`)

vhaasteren:
### Summary
- Adds first-class support for hierarchical triple timing models by allowing two binary delay components in a single PINT model:
  - inner orbit via `BINARY` with standard parameter names (`PB`, `A1`, `T0`, ...)
  - outer orbit via new `BINARY2` with suffixed parameters (`PB_2`, `A1_2`, `T0_2`, ...)
- Introduces outer-orbit wrapper components (`BinaryDD2`, `BinaryBT2`) in a new `pulsar_system_outer` category, ordered before `pulsar_system` so outer delay is accumulated first.
- Preserves physical coupling automatically: the inner binary is evaluated at `t_bary - (accumulated delays)`, so the outer light-travel delay propagates into the inner orbit evaluation epoch.
- Updates model building/validation/parfile I/O to recognize `BINARY2`, enforce one inner + one outer binary maximum, and round-trip `_2` parameters correctly.
- Refactors binary base-component parameter access to be suffix-aware (`_bp()` / `_hasbp()`), enabling fitting and derivative support for outer-orbit parameters without changing standalone binary calculators.
- Keeps single-binary behavior unchanged and backward compatible.

### Scope / Notes
- Outer-orbit support currently includes DD and BT wrappers.
- Outer suffixed orbits intentionally do not support `FBn`/`ORBWAVE` parameterizations.
- This implements coupled delay nesting (not full N-body three-body integration).

### Tests
- Added dedicated triple-system tests covering:
  - component selection and ordering (`outer` before `inner`)
  - parameter namespace isolation (`PB` vs `PB_2`, etc.)
  - `BINARY2` + `_2` parfile read/write round-trip
  - finite residuals and non-zero outer contribution
  - propagation of outer delay into inner evaluation epoch
  - coupled solution differs from naive independent sum
  - finite/non-zero derivatives for outer parameters (fitability)
- Confirmed relevant existing binary/regression tests remain green, including single-binary workflows.

### Example
```par
BINARY   DD
PB       ...
A1       ...
T0       ...
...

BINARY2  BT
PB_2     ...
A1_2     ...
T0_2     ...
...
```

Note: co-developed by @coclar and Claude Code.


abhisrkckl:
Hi... can you merge master into this PR? I will look at this after that.

docs/explanation.rst
```diff
@@ -594,6 +594,9 @@ components.
 
    - If the ``BINARY`` line is present in the parameter file, its value
      determines which binary model to use; if not, no binary model is used.
+     An optional ``BINARY2`` line selects an outer-orbit component for a
+     hierarchical triple (parameters are suffixed with ``_2``, e.g. ``PB_2``;
+     outer wrappers include ``BinaryDD2``, ``BinaryBT2``, and ``BinaryELL12``).
```
abhisrkckl, line 599: Does this incorporate the interactions between the two companion objects? Please state explicitly.

src/pint/models/pulsar_binary.py
```diff
@@ -258,6 +279,69 @@ def __init__(self):
...
+
+        Applies :attr:`param_suffix` so that base-class code written in terms of
+        canonical names (e.g. ``PB``) resolves to the suffixed parameter (e.g.
+        ``PB_2``) for an outer-orbit component. For a normal binary
+        (``param_suffix == ""``) this is just ``getattr(self, name)``.
+        """
+        suffixed = name + self.param_suffix
+        if self.param_suffix and hasattr(self, suffixed):
+            return getattr(self, suffixed)
+        return getattr(self, name)
```
abhisrkckl, line 337: How does this deal with FB parameters (FB0, FB1, ...)? 
Can you add a test for this if this is handled?

docs/explanation.rst
vhaasteren: Yes, at the light-travel / evaluation-epoch level. But this is not a dynamical three-body model, and interactions are not fully incorporated. I'll add a note to the explanation.rst. Maybe @coclar can verify

src/pint/models/pulsar_binary.py
vhaasteren: Outer Roemer delay is chained into the inner epoch. Companions do not interact dynamically per above.

I will look into adding some tests

abhisrkckl: What I meant is, can we use FB parameters for the outer orbit (BINARY2)?

vhaasteren: Ok, talked to Colin. I added `test_inner_orbit_fb_parameterization` to `tests/test_triple_binary.py`, which checks that the inner orbit of a hierarchical triple works with the orbital-frequency parameterization.

It takes the existing DD triple par, replaces the inner `PB` with `FB0 = 1/PB` (and in a second case also `FB1`), and asserts that:
- the model loads, with `FB0`/`FB1` owned by the inner `BinaryDD`;
- the outer `BinaryDD2` carries no `FBn` parameters and keeps `PB_2`;
- the delay is finite and matches the `PB`/`PBDOT` triple (with `PBDOT = -FB1/FB0²`) to numerical noise, about 1e-13 s on delays of order 100 s;
- the `FB0`/`FB1` delay derivatives are finite and nonzero, so they remain fittable inside the triple.

Note that the outer orbit is deliberately `PB`-only: `FBn` and `ORBWAVE` are stripped from the suffixed component, since a hierarchical triple where the outer orbit needs orbital-frequency derivatives is something we consider out-of-scope.

vhaasteren: > What I meant is, can we use FB parameters for the outer orbit (BINARY2)?

No, that's out of scope

abhisrkckl: Ok... please state that in the documentation.

vhaasteren: Ah, right! It's mentioned in `BinaryDD2`, `BinaryBT2`, and `BinaryELL12`. I am adding it now to `explanation.rst`

Thanks!

abhisrkckl:
Looks good! I will merge this if you are done.

vhaasteren:
I think I'm completely done. @coclar anything to add?

coclar:
Nothing else from me!

PR #2012, opened 2026-07-01
Make VLBI frame rotation work correctly when proper motion is present

abhisrkckl:


src/pint/models/astrometry.py
```diff
@@ -709,7 +709,7 @@ def ssb_to_psb_xyz_ICRS(self, epoch: Optional[time_like] = None) -> u.Quantity:
         # Reference: Madison+ 2023, The Astrophysical Journal 777 104 (Equations 9, 10)
         Omega = self.vlbi_coord_rotation()
         Khat = self.xyz_from_radec(ra, dec)
-        result = Omega @ Khat if Omega is not None else Khat
+        result = (Omega @ Khat.T).T if Omega is not None else Khat
```
dlakaplan, line 712: I confess I don't understand this?

dlakaplan:
I don't really understand the difference here, but it looks minor. Probably if you merge in the latest changes I made to fix the Mac CI it will pass?

src/pint/models/astrometry.py
abhisrkckl: When proper motion is absent, `Khat` will be a vector whereas if proper motion is present it will be a 2D array containing the value for each TOA. This change makes sure that the expression handles both cases properly.

dlakaplan:
Should I merge this?

abhisrkckl:
Yes

PR #2015, opened 2026-07-07
Fix Mac CI failures

dlakaplan:
These were maybe due to astropy 8 and other interface changes?


dlakaplan:
This also passes `oldestdeps`

paulray:
Looks good to me

PR #2018, opened 2026-07-15
Fix log-separated frequency computation

abhisrkckl:
There were two issues.
1. The `have_logmode` criterion had `logmode > 0` instead of `logmode >= 0`. This resulted in log-spaced frequencies not being added correctly.
2. `np.log10` was called on astropy quantities. This gives a `UnitTypeError`.


vhaasteren (approved):
I'm surprised about that units issue. I thought that was checked

Anyway, looks good 👍

abhisrkckl:
@vhaasteren It seems the spurious warning came back when I fixed the other issues. I have now fixed the warning also, and improved the test.

abhisrkckl:
@vhaasteren  Can you please merge this if it's OK?

Issue #2016, opened 2026-07-07
Matrix operations with float128 are deprecated


dlakaplan:
If I run `tests/test_residuals.py::test_whitened_res` I see:
```
DeprecationWarning: Calling svd with arguments of dtype=float128 (a.dtype.char = 'g') is deprecated in SciPy 1.18.0 and will be removed in SciPy 1.20.0. Please cast array inputs to one of np.float{32,64} or np.complex{64,128} manually.
    U, s, Vt = scipy.linalg.svd(mtcm, full_matrices=False)
...
DeprecationWarning: Calling linalg.cholesky with arguments of dtype=float128 (a.dtype.char = 'g') is deprecated in SciPy 1.18.0 and will be removed in SciPy 1.20.0. Please cast array inputs to one of np.float{32,64} or np.complex{64,128} manually.
    c = _cholesky(a, lower=lower, overwrite_a=overwrite_a, clean=False,
...
DeprecationWarning: Calling _cho_solve with arguments of dtype=float128 (a.dtype.char = 'g') is deprecated in SciPy 1.18.0 and will be removed in SciPy 1.20.0. Please cast array inputs to one of np.float{32,64} or np.complex{64,128} manually.
    return _cho_solve(c, b, lower, overwrite_b=overwrite_b, check_finite=check_finite)
```
Don't know if this also occurs elsewhere, but it seems like it would be worth checking if casting to 64bit will affect the accuracy of these operations.

abhisrkckl:
None of the timing model computations require matrix operations as far as I understand. All the matrix operations are done on residuals, so I think we can safely cast them to float64. This may give some small speed improvement too.

jeremy-baier:
i think i am already addressing this ( at least in part ) in #1966

PR #2019, opened 2026-07-15
Getting rid of deprecation warnings by casting matrices to 64bit for linear algebra

dlakaplan:
#2016 

This probably isn't a great fix - it does the casting inside various subroutines, so it may be repeated, rather than where the various matrices are created.  But this meant I didn't have to comprehensively find those instances.  If this isn't a good approach I can scrap this PR.


scottransom:
This might be a good use of Claude.  Maybe I (and Claude!) can take a look at this today or tomorrow. There is probably a more elegant solution.

abhisrkckl:
I don't think this is the best solution. The type casting should be done at the source, IMO.

Issue #784, opened 2020-10-01
Can not read INPOP ephemeris 


luojing1211:
PINT currently can not read the ephemeris other than the JPL DExxx format. Our load kernel code assumes the file name would end with ".bsp", which is not true for French INPOP ephemeris. This should be a small fix, but needs more careful design.

abhisrkckl:
TODO: Check this with the detection WG.

abhisrkckl:
I have tried this with the latest latest dev version of PINT. When I "EPHEM INPOP21a", I get a HTTPError code 404 because "inpop21a.bsp" was not found in any of the websites PINT knows about.

In the INPOP web page, they say that they provide INPOP ephemerides in the JPL format (see https://www.imcce.fr/recherche/equipes/asd/inpop/download21a).
The ephemerides are provided in both big endian and little endian formats.

If we are lucky, the solution to this may be as simple as renaming the files to *.bsp and uploading them to data.nanograv.org/* .

dlakaplan:
I don't think that will work. I downloaded the JPL-formatted little-endian version.  I put it in a local directory and appended `file:///...` to the `ephemeris_mirrors` list in `solar_system_ephemerides.py`.  It found the file but says:
```
...
  File "/Users/kaplan/pythonpackages/PINT_dlakaplan/src/pint/solar_system_ephemerides.py", line 53, in _load_kernel_link
    astropy.coordinates.solar_system_ephemeris.set(
  File "/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.9/site-packages/astropy/utils/state.py", line 68, in set
    value = cls.validate(value)
  File "/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.9/site-packages/astropy/coordinates/solar_system.py", line 120, in validate
    cls.get_kernel(value)
  File "/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.9/site-packages/astropy/coordinates/solar_system.py", line 131, in get_kernel
    kernel = _get_kernel(value)
  File "/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.9/site-packages/astropy/coordinates/solar_system.py", line 176, in _get_kernel
    return SPK.open(value)
  File "/Users/kaplan/opt/anaconda3/envs/pintdev/lib/python3.9/site-packages/jplephem/spk.py", line 47, in open
    return cls(DAF(open(path, 'rb')))
[...]

PR #2021, opened 2026-07-24
Incorporate inpop ephemeris

dlakaplan:
#784 

This uses an implementation within PINT to read/process inpop data (adapted from https://github.com/marcelhesselberth/Inpop/tree/main). I avoided the external package directly since it has some peculiarities: it is not vectorized, and it has strict handling of the file names.  So we have to loop over the input data and do something sort of kludgy to use the astropy caching (we cache the file and then set up a tempfile symlink with the right extension, which goes away).

This works in terms of the code, although I am not fully convinced that it is correct.  In particular I see position differences of up to 100km for the Earth-SSB, or ~1us for a test pulsar in residual differences.   
inpop_diff
ephem_difference_de440
(I repeated this for different JPL versions and they are all similar).

So is this just the expected level of ephemeris difference?  Or is there an error in the time handling or units or ...?

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -20,6 +20,7 @@ the released changes.
 - `ssb_to_psb_xyz_ECL` and `ssb_to_psb_xyz_ICRS` are now cached
 - ELL1H with H3+STIGMA: add opt-in ``ell1h_shapiro="absorbed"`` to select Freire & Wex Eq. (28) (Tempo2 ELL1H/T2 mode 1). Default remains Eq. (29) ``"full"`` (`get_model` / `get_model_and_toas` / `ModelBuilder`).
 ### Fixed
+- Remove spurious ``/ Tsun`` factor from analytic DDH ``∂delay/∂STIGMA`` (design matrix / GLS for free ``STIGMA`` was wrong by ``1/Tsun`` since the Maple rewrite in PINT ≥ 1.0).
 - Align ``d_delayS3p_H3_STIGMA_exact_d_STIGMA`` with Eq. (28): ``cos(2*Phi)``.
 - Prefer DD over BT when guessing the binary model for Tempo2 `T2` par files (`allow_T2`), matching Tempo2's `allTerms=1` behavior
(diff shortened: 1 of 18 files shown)


dlakaplan:
Right now some of the CI is failing for reasons unrelated to the actual PR:
https://github.com/marcelhesselberth/Inpop/issues/3

dlakaplan:
Here is an update where I refit the model (RA, Dec, F0, F1):
inpop_diff
It goes from chi^2=97/94 (original) to 2706/94 (INPOP) to 88/94 (INPOP refit).  The rms of the difference is 30ns.  Parameter differences:
```
PARAMETER                                    Model 1                           Model 2   Diff_Sigma1   Diff_Sigma2
-------------------     ----------------------------   -------------------------------   -----------   -----------
RAJ                      17h48m52.74997507s +/-   3e-05    17h48m52.74996363s +/-   3e-05         -0.38         -0.38
DECJ                    -20d21m29.00002704s +/- 0.00011   -20d21m28.99977036s +/- 0.00011          2.29          2.29
F0                                61.48547655399987(33)             61.48547655400124(33)          4.06          4.06 !
F1                                 -1.1809990(22)×10⁻¹⁵              -1.1810004(22)×10⁻¹⁵         -0.62         -0.62
```
So this is plausible, but I would have expected more of a change in the position and less in the spin params (e.g., based on https://arxiv.org/pdf/2304.14677) but I haven't done a deep dive yet.

dlakaplan:
The separation between the fitted positions is 240 uas in RA, 280 uas in Dec.  Those are plausible given Figure 4 in the Liu paper above.

dlakaplan:
Repeating that for a grid of input positions:
inpop_diffgrid
the arrows are the position offsets in uas, and the symbol size is the change in F0 (color is sign).  So this looks relatively smooth/well-behaved, likely a rotation/offset as discussed elsewhere.

dlakaplan:
Note that I have a new standalone INPOP implementation that will avoid the CI errors above (and lead to some speed improvements).  But I am waiting to see which is better to use.

dlakaplan:
@scottransom : would appreciate any feedback here.

PR #2028, opened 2026-07-30
fix(model_builder): prefer DD over BT when guessing T2 binary model

### LINKED ISSUE #2027: Binary model guess: T2 -> DD, never BT
Currently, the `allow_T2` option dynamically guesses the binary model that the `T2` model uses from its parameters. It then uses the simplest model with the same model parameters as in the parfile, and then uses that. There is a priority ordering:

```python
_binary_model_priority = [
    "Isolated",
    "BT",
    "BT_piecewise",
    "ELL1",
    "ELL1H",
    "ELL1k",
    "DD",
    "DDK",
    "DDGR",
    "DDS",
    "DDH",
]
```

The problem with that is that BT is not the same as DD without post-Keplerian parameters. So T2 should always simplify to DD, not to BT. My proposed ordering would be

```python
_binary_model_priority = [
    "Isolated",
    "ELL1",
    "ELL1H",
    "ELL1k",
    "DD",
    "DDK",
    "DDGR",
    "DDS",
    "DDH,
    "BT",
    "BT_piecewise",
]
```

@dlakaplan , this is probably something you'd be best to ask?

vhaasteren:
## Summary
- Reorder `_binary_model_priority` so Keplerian Tempo2 `BINARY T2` pars (`T0`/`ECC`/`OM`/…) map to **DD** instead of **BT**.
- Keep DD-family extensions (`DDK`/`DDGR`/`DDS`/`DDH`) *after* bare DD so plain Keplerian T2 does not mis-map to those models.
- Add a regression case: Keplerian T2 → `DD` in `test_guess_binary_model`.

This fixes #2027 

## Motivation
Tempo2’s T2 model with default `allTerms=1` is effectively the Damour–Deruelle (DD) delay, not the BT truncation. Under `allow_T2`, both BT and DD are often allowed for the same Keplerian parameter set; the old priority put BT first, so PINT silently chose the wrong model (~ns-level residual mismatch vs Tempo2 T2).

Existing ELL1/ELL1H/DDK guesses are unchanged: those parameter sets still exclude DD/BT as appropriate.

## Test plan
- [x] `pytest tests/test_process_parfile.py::test_guess_binary_model tests/test_t2binary2pint.py -v`
- [x] Confirm a Keplerian T2 par with `allow_T2=True` loads as `BINARY DD`
- [x] Confirm DDK (KOM/KIN) and ELL1H (EPS/H3) still guess correctly

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -19,6 +19,7 @@ the released changes.
 - Plot whitened DM residuals in pintk.
 - `ssb_to_psb_xyz_ECL` and `ssb_to_psb_xyz_ICRS` are now cached
 ### Fixed
+- Prefer DD over BT when guessing the binary model for Tempo2 `T2` par files (`allow_T2`), matching Tempo2's `allTerms=1` behavior
 - `WidebandTOAFitter` raises a warning if the model has correlated errors (It used to give wrong results before).
 - Fixed bug where "include_bipm" flag was being ignored when loading Fermi TOAs with weights, now defaults to using EPHEM, CLOCK and PLANET_SHAPIRO from the timing model
 - When flags are created based off jumps uses strings instead of None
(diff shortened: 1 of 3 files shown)


abhisrkckl:
Please add a CHANGELOG entry. Other than that this looks good to me.

vhaasteren:
> Please add a CHANGELOG entry. Other than that this looks good to me.

Done

PR #2029, opened 2026-07-31
feat(ELL1H): opt-in ell1h_shapiro=absorbed for Freire & Wex Eq. 28

vhaasteren:
## Summary
- Add opt-in `ell1h_shapiro={"full","absorbed"}` on `get_model`, `get_model_and_toas`, and `ModelBuilder` (placed immediately before `**kwargs`).
- Default `"full"` keeps today’s Freire & Wex Eq. (29) behavior for `H3`+`STIGMA`.
- `"absorbed"` selects Eq. (28) (`delayS3p_H3_STIGMA_exact`), matching Tempo2 ELL1H / T2 mode 1.
- Align `d_delayS3p_H3_STIGMA_exact_d_STIGMA` with Eq. (28): `cos(2*Phi)`.
- Document both exact modes on `BinaryELL1H` and the standalone `ELL1Hmodel`.

## Motivation
With identical `(A1, EPS1, H3, STIG)`, PINT’s default Eq. (29) and Tempo2’s Eq. (28) differ by a Roemer-absorbable gauge term (harmonics 1–2). Cross-engine residual work needs an explicit, non-breaking way to evaluate the Tempo2-compatible form.

## Test plan
- [x] `pytest tests/test_ell1h_shapiro_mode.py -v` (13 passed)
- [x] Black 24 on touched files
- [x] CI matrix (Python 3.9/3.13, docs, notebooks, macOS)
- [x] Confirm default `get_model(par)` still uses Eq. (29)
- [x] Confirm `get_model(par, ell1h_shapiro="absorbed")` uses Eq. (28) and survives `setup()`

### DIFF
--- a/src/pint/models/model_builder.py
+++ b/src/pint/models/model_builder.py
@@ -121,6 +121,7 @@ class ModelBuilder:
         allow_T2=False,
         force_binary_model=None,
         toas_for_tzr=None,
+        ell1h_shapiro="full",
         **kwargs,
     ):
         """Callable object for making a timing model from .par file.
@@ -155,6 +156,11 @@ class ModelBuilder:
         toas_for_tzr : TOAs or None, optional
             If this is not None, a TZR TOA (AbsPhase) will be created using the
             given TOAs object.
+
+        ell1h_shapiro : {"full", "absorbed"}, optional
+            Freire & Wex convention for ELL1H with H3+STIGMA. "full" (default)
+            uses Eq. (29). "absorbed" uses Eq. (28), matching Tempo2 ELL1H/T2
+            mode 1. Ignored for ELL1H models that are not on the H3+STIGMA path.
 
         kwargs : dict
             Any additional parameter/value pairs that will add to or override those in the parfile.
@@ -206,6 +212,13 @@ class ModelBuilder:
         # Make timing model
         cps = [self.all_components.components[c] for c in selected]
         tm = TimingModel(components=cps)
+        if ell1h_shapiro not in ("full", "absorbed"):
+            raise ValueError(
+                "ell1h_shapiro must be 'full' or 'absorbed', " f"got {ell1h_shapiro!r}"
+            )
+        if "BinaryELL1H" in tm.components:
+            tm.components["BinaryELL1H"].ell1h_shapiro = ell1h_shapiro
+
         self._setup_model(
             tm,
             pint_param_dict,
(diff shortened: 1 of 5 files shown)


src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
```diff
@@ -332,7 +343,7 @@ def d_delayS3p_H3_STIGMA_exact_d_STIGMA(self, H3, stigma, end_harm=None):
                 -3 * np.log(lognum)
                 + 2 * stigma * (stigma - np.sin(Phi)) / lognum
                 - 4 * stigma * np.sin(Phi)
-                + stigma**2 * np.cos(Phi)
+                + stigma**2 * np.cos(2 * Phi)
```
abhisrkckl, line 346: Was this a bug?

src/pint/models/model_builder.py
```diff
@@ -255,6 +268,7 @@ def __call__(
         tm.meta["allow_tcb"] = allow_tcb_
         tm.meta["convert_tcb"] = convert_tcb
         tm.meta["allow_T2"] = allow_T2
+        tm.meta["ell1h_shapiro"] = ell1h_shapiro
```
abhisrkckl, line 271: Please make sure that when PINT writes out a par file, this setting is included in the par file header as a comment.

src/pint/models/stand_alone_psr_binaries/ELL1H_model.py
vhaasteren: Yes, that was a bug. Note that the above code as dead code, as PINT did not use the exact formula. The opt-in I am providing activates it. I need that functionality to get consistency with tempo2, which uses the exact formula.

The bug comes from Eq. (28) of Freire and Wex (2010): https://academic.oup.com/mnras/article/409/1/199/1023583

You can find a cos(2 * Phi) there, not a cos (Phi)

src/pint/models/model_builder.py
vhaasteren: Yes, this happens automatically. Those comments are built from the `tm.meta` dict keys. It will say something like:

```
# ell1h_shapiro: absorbed
```

abhisrkckl:
I think this is good to go. I will merge this if you are done.

vhaasteren:
Thanks! Yes, confirmed, this is done

PR #2033, opened 2026-08-05
fix(FDJUMPDM): use Tempo2-consistent positive DM offset sign (BREAKING CHANGE)

vhaasteren:
## Summary

Flip the `FDJUMPDM` sign convention so that a **positive** `FDJUMPDM` contributes a **positive** DM (and thus a positive dispersion delay) on the selected TOAs.

This matches Tempo2 and is the more natural definition: `+FDJUMPDM` on a mask should act like `+DM` on those TOAs.

### Motivation

PINT previously applied the opposite sign:

```python
jdm[mask] += -fdjumpdm_par.value
```

so a positive fitted `FDJUMPDM` *decreased* DM/delay. That is opposite to Tempo2, where `FDJUMPDM` (implemented as `fdjumpIdx == -2`) enters the residual/phase calculation with the same sense as a positive DM contribution (`formResiduals.C` / `t2FitFunc_fdjump`).

That mismatch made fitted `FDJUMPDM` values non-portable between engines: a solution from Tempo2/libstempo would need a global sign flip to be interpreted correctly in PINT (and vice versa). There is no strong reason for PINT to keep the inverted convention; Tempo2’s definition is both prior art and more logical (“positive parameter → positive DM on the mask”).

This change aligns PINT with Tempo2 rather than documenting or papering over the difference.

### Changes

- `src/pint/models/dispersion_model.py`
  - `fdjump_dm`: `jdm[mask] += fdjumpdm_par.value` (was `+= -…`)
  - `d_dm_d_fdjumpdm`: derivative mask is `+1` (was `-1`)
  - Docstring/class docs state the Tempo2-consistent sign convention explicitly
- `tests/test_fdjump.py`: expectation updated so a positive `FDJUMPDM` increases total DM on the mask
- `docs/explanation.rst`: document the convention
- `CHANGELOG-unreleased.md`: note under Fixed

### Breaking change / migration

**Yes — intentional.** Existing par files (or posteriors) that were fitted with PINT’s old inverted convention need:

```text
FDJUMPDM → -FDJUMPDM
```

when reloading under this fix (or when comparing to Tempo2 fits that already used the positive convention). Values produced by Tempo2/libstempo should now load and fit without a manual sign flip.

If you have hybrid workflows that previously negated PINT vs Tempo2 `FDJUMPDM` by hand, remove that workaround after this lands.

### Test plan

- [x] Unit test `test_fdjumpdm_offset` updated for the new sign
- [x] `pytest tests/test_fdjump.py -v`
- [x] Optional: load a Tempo2-fitted par with non-zero `FDJUMPDM` and confirm PINT’s `total_dm` / residuals match Tempo2 on the selected TOAs without flipping the parameter

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -19,6 +19,7 @@ the released changes.
 - Plot whitened DM residuals in pintk.
 - `ssb_to_psb_xyz_ECL` and `ssb_to_psb_xyz_ICRS` are now cached
 ### Fixed
+- FDJUMPDM sign convention: a positive FDJUMPDM now adds a positive DM (and delay) on selected TOAs, matching Tempo2.
 - `WidebandTOAFitter` raises a warning if the model has correlated errors (It used to give wrong results before).
 - Fixed bug where "include_bipm" flag was being ignored when loading Fermi TOAs with weights, now defaults to using EPHEM, CLOCK and PLANET_SHAPIRO from the timing model
 - When flags are created based off jumps uses strings instead of None
--- a/tests/test_fdjump.py
+++ b/tests/test_fdjump.py
@@ -141,4 +141,5 @@ def test_fdjumpdm_offset(model_and_toas):
     dm2 = model2.total_dm(toas)
 
     assert np.allclose((dm1 - dm2)[not_mask], 0)
-    assert np.allclose((dm1 - dm2)[mask], -model.FDJUMPDM1.quantity)
+    # Positive FDJUMPDM adds positive DM (Tempo2-consistent convention).
+    assert np.allclose((dm1 - dm2)[mask], model.FDJUMPDM1.quantity)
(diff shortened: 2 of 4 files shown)


CHANGELOG-unreleased.md
```diff
@@ -11,4 +11,5 @@ the released changes.
 ### Changed
 ### Added
 ### Fixed
+- FDJUMPDM sign convention: a positive FDJUMPDM now adds a positive DM (and delay) on selected TOAs, matching Tempo2.
```
abhisrkckl, line 16: Please put this in "Changed". This is not a bug.

abhisrkckl:
I don't think anyone is using FDJUMPDMs for anything other than IPTA DR3. I will ask the PINT data combiners' opinion on this just in case.

tests/test_fdjump.py
```diff
@@ -141,4 +141,5 @@ def test_fdjumpdm_offset(model_and_toas):
     dm2 = model2.total_dm(toas)
 
     assert np.allclose((dm1 - dm2)[not_mask], 0)
-    assert np.allclose((dm1 - dm2)[mask], -model.FDJUMPDM1.quantity)
+    # Positive FDJUMPDM adds positive DM (Tempo2-consistent convention).
+    assert np.allclose((dm1 - dm2)[mask], model.FDJUMPDM1.quantity)
```
abhisrkckl, line 145: Please also check the derivative.

PR #2034, opened 2026-08-06
fix(observatory): use Tempo2 clock chains for Jodrell Mark II sites

vhaasteren:
## Summary
- Wire `jbmk2` / `jbmk2roach` / `jbmk2dfb` to the same Tempo2-format clock paths as `jbafb` / `jbroach` / `jbdfb`.
- These sites were imported from `observatories.dat` without `clock_file` / `clock_fmt`, so PINT fell back to an empty TEMPO clock path and applied only a near-constant ~27 µs correction.
- Tempo2 instead routes them through zero-offset aliases (`jbmk2roach2jbroach.clk`, etc.) onto the Lovell backend chains; PINT now uses those equivalent chains directly (alias files are identity and are not in PINT's global clock index).

## Impact
On EPTA DR2 `J1713+0747` (288 `jbmk2roach` TOAs), PINT−Tempo2 residual RMS falls from ~656 ns to ~3 ns after this change. The remaining budget is consistent with other EPTA sites.

## Test plan
- [x] `pytest tests/test_observatory.py::test_jb_mkii_sites_use_tempo2_clock_chains tests/test_observatory.py::test_jbmk2roach_matches_jbroach_clock_corrections tests/test_observatory.py::test_jbmk2dfb_matches_jbdfb_clock_corrections tests/test_observatory.py::test_can_compute_corrections tests/test_observatory.py::test_last_mjd`
- [x] Confirm `jbmk2roach` clock corrections match `jbroach` over MJD 56000–58000
- [x] Spot-check a dataset with `jbmk2roach` TOAs for PINT/Tempo2 residual parity

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -20,6 +20,7 @@ the released changes.
 - `ssb_to_psb_xyz_ECL` and `ssb_to_psb_xyz_ICRS` are now cached
 - ELL1H with H3+STIGMA: add opt-in ``ell1h_shapiro="absorbed"`` to select Freire & Wex Eq. (28) (Tempo2 ELL1H/T2 mode 1). Default remains Eq. (29) ``"full"`` (`get_model` / `get_model_and_toas` / `ModelBuilder`).
 ### Fixed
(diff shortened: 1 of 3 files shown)


dlakaplan:
This looks simple - I can merge when the notebook test has passed (and probably even without that).

PR #2036, opened 2026-08-07
fix(astrometry): propagate sky uncertainties by covariance rotation

vhaasteren:
## Summary
`AstrometryEquatorial.as_ECL` and `AstrometryEcliptic.as_ICRS` (and the ECL→ECL obliquity path) propagate position/PM uncertainties with the "fake proper motion" trick. That treated `(σ_lon, σ_lat)` as a **signed vector**. `as_ICRS` wrapped the result in `np.abs`; `as_ECL` did not. On high-`|ELAT|` pulsars an ECL→ICRS→ECL round trip can therefore assign a **negative** ELONG uncertainty and raise:

`ValueError: Uncertainties cannot be negative but -6.68e-08 deg was supplied`

(seen on PPTA DR3 J0711-6830 / J1017-7156). Even with `abs` on both sides, a vector rotation is the wrong marginal σ after a non-trivial frame change.

This PR replaces the signed-vector path with a **diagonal covariance rotation**: transform each uncertainty axis separately and combine in quadrature (still assuming zero correlation, as before). Uncertainties are always non-negative, and an IERS2010↔IERS2003 obliquity round trip recovers the input σ to numerical precision.

## Test plan
- [x] `pytest tests/test_modelconversions.py -v`
- [x] New regression: high-`|ELAT|` model survives `as_ICRS().as_ECL()` with non-negative σ near the input
- [x] New regression: ECL obliquity round trip preserves diagonal uncertainties
- [ ] Confirm CI green

## Notes
A full ECL→ICRS→ECL identity still cannot recover σ exactly under a diagonal-only assumption (the intermediate frame acquires correlation that we drop). That is unchanged in spirit from the previous helpers; the obliquity-only path MetaPulsar uses is exact.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -20,6 +20,7 @@ the released changes.
 - `ssb_to_psb_xyz_ECL` and `ssb_to_psb_xyz_ICRS` are now cached
 - ELL1H with H3+STIGMA: add opt-in ``ell1h_shapiro="absorbed"`` to select Freire & Wex Eq. (28) (Tempo2 ELL1H/T2 mode 1). Default remains Eq. (29) ``"full"`` (`get_model` / `get_model_and_toas` / `ModelBuilder`).
 ### Fixed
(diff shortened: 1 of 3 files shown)


dlakaplan:
Without looking at every line this looks good in principle, and the tests you've added should address accuracy questions.    However, it might be worth rebasing this since I think the CHANGELOG file looks out of date (there were releases on 10 Aug and 24 Aug).

Issue #2037, opened 2026-08-12
ChromaticNoise class reference frequency hardcoded to 1MHz


jeremy-baier:
This may lead to inconsistencies between different PTA definitions ( tempo2 references to 1400MHz ) and ChromGP is referenced to 1400MHz. This impacts the quadratic filter of chromatic noise ( CM, CM1, CM2 ) as well as CMX and other CM* models. It does not affect PLChromaticNoise ( ChromGP ) which is a noise class.
A simple fix could be to make reference frequency a parameter in the par file.

vhaasteren:
Don't forget to cross-check with tempo2

dlakaplan:
I think it's 1 GHz (not MHz)?

dlakaplan:
I don't have any primary knowledge, but this was the comment in slack:

> As some of you may have noticed, I've been looking at differences between PINT and tempo2. I never looked at Chromatic Noise before until now. It looks like Chromatic Noise has been implemented differently in the two packages as well. PINT uses a 1GHz reference frequency, whereas tempo2 uses 1.4GHz. That makes the units different and index-dependent

jeremy-baier:
> I think it's 1 GHz (not MHz)?

That’s correct. I will update the title accordingly.

jeremy-baier:
See #2038

jeremy-baier:
Okay I checked again and it is actually 1MHz. FD parameters on the other hand are referenced to 1GHz

PR #2038, opened 2026-08-19
Add reference frequency as a parameter for CM, PLDMNoise, PLChromNoise

jeremy-baier:
See #2037

### DIFF
--- a/src/pint/models/chromatic_model.py
+++ b/src/pint/models/chromatic_model.py
@@ -27,12 +27,17 @@ class Chromatic(DelayComponent):
 
         self.alpha_deriv_funcs = {}
 
-    def chromatic_time_delay(self, cm, alpha, freq):
+    def chromatic_time_delay(self, cm, alpha, freq, fref=None):
         """Return the chromatic time delay for a set of frequencies.
 
-        delay_chrom = cm * DMconst * (freq / 1 MHz)**alpha
+        delay_chrom = cm * DMconst * (freq / CM_FREF)**(-alpha)
+
+        If *fref* is None, the reference frequency is taken from the CM_FREF
+        parameter of the parent model.
         """
-        cmdelay = cm * DMconst * (freq / u.MHz) ** (-alpha)
+        if fref is None:
+            fref = self._parent["CM_FREF"].quantity
+        cmdelay = cm * DMconst * (freq / fref) ** (-alpha)
         return cmdelay.to(u.s)
 
     def chromatic_type_delay(self, toas):
@@ -44,7 +49,8 @@ class Chromatic(DelayComponent):
 
         cm = self.cm_value(toas)
         alpha = self._parent["TNCHROMIDX"].quantity
-        return self.chromatic_time_delay(cm, alpha, bfreq)
+        fref = self._parent["CM_FREF"].quantity
+        return self.chromatic_time_delay(cm, alpha, bfreq, fref)
 
     def cm_value(self, toas):
         """Compute modeled CM value at given TOAs.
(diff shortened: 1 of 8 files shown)


jeremy-baier:
One of the failing tests is this:
`FAILED tests/test_chromatic.py::test_chromatic_cm_fit - ValueError: Unable to…`

I am guessing that it is trying to fit the chromatic reference frequency? I thought it was defaulted to not fit, but let me dig into the test.

The other looks unrelated so I will try rerunnign it to see if it passed.

src/pint/models/chromatic_model.py
```diff
@@ -27,12 +27,17 @@ def __init__(self):
 
         self.alpha_deriv_funcs = {}
 
-    def chromatic_time_delay(self, cm, alpha, freq):
+    def chromatic_time_delay(self, cm, alpha, freq, fref=None):
```
dlakaplan, line 30: Is it possible to add type hints to this (e.g., float vs Quantity)?  That would help users give the right inputs.

src/pint/models/chromatic_model.py
```diff
@@ -27,12 +27,17 @@ def __init__(self):
...
-        delay_chrom = cm * DMconst * (freq / 1 MHz)**alpha
+        delay_chrom = cm * DMconst * (freq / CM_FREF)**(-alpha)
+
+        If *fref* is None, the reference frequency is taken from the CM_FREF
+        parameter of the parent model.
         """
-        cmdelay = cm * DMconst * (freq / u.MHz) ** (-alpha)
+        if fref is None:
+            fref = self._parent["CM_FREF"].quantity
+        cmdelay = cm * DMconst * (freq / fref) ** (-alpha)
```
dlakaplan, line 40: You could include an explicit `.decompose()` here to make sure `freq` and `fref` have the same units, although when you cast to `s` that will likely happen anyway.

src/pint/models/chromatic_model.py
dlakaplan: And/or add `Parameters` etc in the docstring

src/pint/models/chromatic_model.py
```diff
@@ -44,7 +49,8 @@ def chromatic_type_delay(self, toas):
 
         cm = self.cm_value(toas)
         alpha = self._parent["TNCHROMIDX"].quantity
-        return self.chromatic_time_delay(cm, alpha, bfreq)
+        fref = self._parent["CM_FREF"].quantity
```
dlakaplan, line 52: Will this always exist and be not-None?

src/pint/models/chromatic_model.py
```diff
@@ -88,11 +94,12 @@ def d_delay_d_cmparam(self, toas, param_name, acc_delay=None):
         param_unit = getattr(self, param_name).units
         d_cm_d_cmparam = np.zeros(toas.ntoas) * cmu / param_unit
         alpha = self._parent["TNCHROMIDX"].quantity
+        fref = self._parent["CM_FREF"].quantity
```
dlakaplan, line 97: Same as above

src/pint/models/chromatic_model.py
```diff
@@ -175,6 +188,17 @@ def __init__(self):
...
             )
         )
+        self.add_param(
+            floatParameter(
+                name="CM_FREF",
+                units=u.MHz,
+                value=1400.0,
+                frozen=True,
+                description="Reference frequency of the chromatic measure",
+                long_double=True,
```
dlakaplan, line 198: I don't think this needs to be a `long_double`

src/pint/models/noise_model.py
```diff
@@ -573,6 +573,16 @@ def __init__(
                 tcb2tdb_scale_factor=1,
             )
         )
+        self.add_param(
+            floatParameter(
+                name="DM_FREF",
```
dlakaplan, line 605: This name sounds like something else (like for referencing the dedispersion).  Is it already in use elsewhere?  If not, can you change to `DMNOISE_FREF` or similar?

PR #2040, opened 2026-09-04
feat: updated accuracy of TCB-->TDB conversion

vhaasteren:
### PR description

This PR updates PINT’s TCB<-->TDB converter to support auditable, no-refit conversion at approximately nanosecond accuracy for the deterministic components that PINT explicitly supports.

My motivation is automatic pulsar-data combination in MetaPulsar. Combining releases from different PTAs sometimes requires converting TCB and TDB par files across timing engines. Refitting during ingestion is undesirable: it changes the scientific product and makes automated combinations harder to reproduce. For supported models, conversion should instead preserve the same physical timing solution.

The old converter used the legacy IFTE common-origin mapping. In particular, its epoch conversion omitted the approximately 65.5 μs TDB offset and used a slightly different rate from the IAU 2006 definition. Its DM scaling also did not match PINT’s actual fixed-frequency forward model.

This implementation follows several requirements we set before writing the code:

- Match PINT’s existing TDB forward model rather than defining an independent convention.
- Use Astropy/ERFA for IAU 2006 coordinate-time conversion.
- Keep radio frequency undilated, consistent with PINT’s `DILATEFREQ N` behavior.
- Avoid duplicated parameter lists where component-owned metadata can express the conversion.
- Convert everything currently supported, while leaving unsupported terms unchanged and warning clearly.
- Distinguish “conversion completed” from “conversion satisfies the tested no-refit contract.”
- Keep PX and UTC/data-span selectors explicitly invariant.
- Require discriminating delay or phase tests before certifying a component.

The resulting converter:

- converts coordinate epochs through Astropy’s TCB/TDB scales, including `TDB0`;
- applies order-aware scaling to DM Taylor coefficients;
- converts FD and FDJUMP coefficients under PINT’s fixed-frequency convention;
- records converted, invariant, unsupported, and unaudited terms in a `TCBTDBConversionReport`;
- remains permissive when it encounters unsupported model terms;
- reports `accepted=True` only when the model lies within the tested no-refit surface;
- preserves that acceptance assessment on repeated/no-op conversion calls.

The documentation and CLI now state the convention and acceptance contract explicitly. The changelog notes that converted epochs and DM values differ from legacy-IFTE output.

MetaPulsar also ingests native tempo2 par files, some of which cannot initially be parsed by PINT. Therefore this PR is not intended to replace MetaPulsar’s tempo2 conversion path. What we actually need longer-term is a corresponding tempo2 PR: one that provides the same auditable, permissive-conversion/strict-acceptance structure, but derives its scaling rules from tempo2’s own forward model and `DILATEFREQ` semantics. I plan to work on that or coordinate with Mike separately.

Thanks to @tcromartie for bringing it to my attention again yesterday during the DWG; it had been on my radar for months. I enjoyed the discussions!

Verification performed:

- Full PINT suite before review fixes: 2,502 passed, 20 skipped, 18 xfailed.
- Updated TCB/TDB suite: 21 passed.
- Parameter and model-construction regressions: 201 passed, 2 skipped, 2 xfailed.
- Formatting, diff checks, and lint checks pass.

### DIFF
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -13,7 +13,9 @@ the released changes.
   keeps radio frequency undilated, converts FD/FDJUMP and order-aware DM
   coefficients, leaves PX and UTC/data-span selectors explicitly invariant,
   and reports unsupported active deterministic terms. Converted epochs and DM
-  values therefore differ from the previous legacy-IFTE conversion.
+  values therefore differ from the previous legacy-IFTE conversion. The
+  compatibility alias `IFTE_K` now denotes the IAU/ERFA rate; the obsolete
+  `IFTE_MJD0` and `IFTE_KM1` module constants have been removed.
 ### Added
 - Time-domain solar wind GP noise components: ridge, squared-exponential, Matérn, and quasi-periodic kernels
 - Documentation page explaining the time-domain solar wind noise model, its interpolation basis, and how it differs from the Fourier-basis noise models
--- a/CHANGELOG-unreleased.md
+++ b/CHANGELOG-unreleased.md
@@ -21,4 +21,7 @@ the released changes.
 - Documentation page explaining the time-domain solar wind noise model, its interpolation basis, and how it differs from the Fourier-basis noise models
 - `TOAs.get_tdb_seconds()`, returning the TDB times of the TOAs in seconds with a selectable dtype
 ### Fixed
+- TCB/TDB parameter scaling uses ``x + x (K^n-1)`` with ``K^n-1`` formed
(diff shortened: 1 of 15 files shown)


vhaasteren:
Extra comment:

`IFTE_K` is still exported as an alias of `TCB_TDB_K`. Nothing in the repo imports it, the `IFTE` epoch origin (`IFTE_MJD0`) is already gone, and the alias is no longer the Irwin & Fukushima rate. I am inclined to remove `IFTE_K` and leave `TCB_TDB_K` / `TCB_TDB_F` as the public constants. Does anyone rely on `from pint.models.tcb_conversion import IFTE_K`?

